PR #17, opened 2020-02-06
Feature/gfsv16b paranetcdf

yangfanglin:
While I am still testing this branch with parallel netCDF enabled, I'd like to give you the opportunity to review the changes  before the testing is completed.   The test was set up to run one and an half cycle.  Changes have already been made to the external repos, including GSI, UPP, GLDAS and UFS_UTIL, to compile with the new parallel netcdf libs and associated esmf lib.   Code managers are yet to make new tags/branches available.  I am still checking if there is any other external apps need to be updated.


yangfanglin:
The changes are rather significant, including
1) running forecasts with parallel netCDF read and write, with different options for different model resolutions and on different platforms.
2) adding all options currently used by v16rt2 into the workflow
3) removing those "*.iau" configure files
4) restructuring config.fv3 and config.resources so that the computation resources can be optimally set  for different computers based on the max cores per node setting.  Removed many redundant and sometimes conflicting parameters. 

Supposedly, out of the box the branch can reproduce v16rt2.

modulefiles/module_base.hera
```diff
@@ -7,22 +7,27 @@ module load intel/18.0.5.274
 module load impi/2018.0.4
 module load wgrib2/2.0.8
 module load hpss/hpss
-module load pnetcdf/1.10.0
-module load netcdf/4.7.0
 module load nco/4.7.0
-module load hdf5/1.10.5
 module load gempak/7.4.2
 
+#module load pnetcdf/1.10.0
```
CoryMartin-NOAA: Do we want to keep the commented out modules? I guess it's okay for now

parm/config/config.eupd
```diff
@@ -11,6 +11,25 @@ echo "BEGIN: config.eupd"
...
-export NAM_ENKF="analpertwtnh=0.9,analpertwtsh=0.9,analpertwttr=0.9"
+# Use NAM_ENKF below for serial EnKF
+##export NAM_ENKF="analpertwtnh=0.9,analpertwtsh=0.9,analpertwttr=0.9"
+
+# LETKF specific settings with model space localization
+export modelspace_vloc=".true."  # model space localization
+export letkf_flag=".true."       # use LETKF instead of serial filter
+export getkf=".true."            # Gain form of LETKF (needed for model-space localization)
+export denkf=".true."            # EnKF approximation (beneficial since less spread removed by analysis)
+export nobsl_max=10000           # max number of obs in each LETKF volume (uses closest nobsl_max). can
```
CoryMartin-NOAA, line 22: are we decreasing this because of runtime concerns?

parm/config/config.gldas
```diff
@@ -12,9 +12,7 @@ export GLDASSH=$HOMEgfs/scripts/exgdas_gldas.sh.ecf
 export gldas_spinup_hours=72
 
 
-if [ $machine = "WCOSS_C" -o $machine = "WCOSS_DELL_P3" ]; then
-    export CPCGAUGE=/gpfs/dell2/emc/modeling/noscrub/Fanglin.Yang/stat/cpcUniGauge
```
CoryMartin-NOAA, line 16: Is this supposed to be removed for Cray/Dell?

parm/config/config.resources
```diff
@@ -40,87 +40,52 @@ if [ $step = "prep" -o $step = "prepbufr" ]; then
 
 elif [ $step = "anal" ]; then
 
-    export wtime_anal="02:30:00"
+    export wtime_anal="06:00:00"
```
CoryMartin-NOAA: Why has this been bumped up to 6 hours? Won't this cause the job to queue longer than necessary?

parm/config/config.resources
```diff
@@ -196,61 +160,40 @@ elif [ $step = "eupd" ]; then
 
 elif [ $step = "ecen" ]; then
 
-    export wtime_ecen="00:30:00"
+    export wtime_ecen="03:00:00"
```
CoryMartin-NOAA: Again, I guess the general comment for a lot of these changes, is why are the wall time requests so much larger, this one is 6x as long. Is that actually what is needed now?

parm/config/config.anal
```diff
@@ -93,9 +104,17 @@ if [ $REALTIME = "NO" ]; then
     fi
 fi
 
-if [ $LEVS = "128" ]; then
-   export GRIDOPTS="nlayers(63)=1,nlayers(64)=1,"
-   export SETUP="gpstop=55,nsig_ext=56,$SETUP"
+# TEST - use experimental dumps
```
KateFriedman-NOAA: Will need to remember to remove pointers to x-dumps or wrap in a date if-block so the experimental versions are used in the future for dates before these files went into production.

parm/config/config.fcst
```diff
@@ -138,6 +138,31 @@ export ishuffle=0
...
+        if [ $RESTILE -ge 768 ]; then
+            export OUTPUT_FILETYPES=" 'netcdf_parallel' 'netcdf_parallel' "
+            export ichunk3d=$((4*RESTILE)) 
+            export jchunk3d=$((2*RESTILE))
+            export kchunk3d=1
+        else
+            export OUTPUT_FILETYPES=" 'netcdf_parallel' 'netcdf' "
+        fi
+    fi
+    if [[ "$machine" == "HERA" ]]; then
```
KateFriedman-NOAA, line 156: Why is Hera different here? Mostly just curious.

parm/config/config.gldas
```diff
@@ -12,9 +12,7 @@ export GLDASSH=$HOMEgfs/scripts/exgdas_gldas.sh.ecf
...
-elif [ $machine = "HERA" ]; then
+if [ $machine = "HERA" ]; then
     export CPCGAUGE=/scratch1/NCEPDEV/global/Fanglin.Yang/stat/cpcUniGauge
 fi
```
KateFriedman-NOAA, line 18: The CPC gauge files are in the Hera GDA just like in the WCOSS GDA so this isn't necessary anymore.

parm/config/config.resources
KateFriedman-NOAA: I have the same question as Cory.

parm/config/config.resources
```diff
@@ -40,87 +40,52 @@ if [ $step = "prep" -o $step = "prepbufr" ]; then
...
+    if [[ "$machine" == "WCOSS_C" ]]; then export memory_gldas="3072M"; fi
 
 elif [ $step = "fcst" ]; then
 
-    if [[ "$machine" == "WCOSS_DELL" ]]; then
-      export npe_node_max=24
-    fi
     export wtime_fcst="01:00:00"
-    export wtime_fcst_gfs="06:00:00"
+    export wtime_fcst_gfs="10:00:00"
```
KateFriedman-NOAA: Hera's max walltime is 8hrs so this won't work on Hera. Perhaps we need to introduce a wtime_max variable for each machine and use it here?

parm/config/config.resources
```diff
@@ -159,7 +124,7 @@ elif [ $step = "eobs" -o $step = "eomg" ]; then
     export wtime_eobs="00:30:00"
     export wtime_eomg="01:00:00"
     if [ $CASE = "C768" ]; then
-      export npe_eobs=84
+      export npe_eobs=400
     elif [ $CASE = "C384" ]; then
       export npe_eobs=42
```
KateFriedman-NOAA, line 129: Do the lower resolutions need a similar increase for npe_eobs? Going from 400 for C768 to 42 for C384 seems too big of a drop now.

RussTreadon-NOAA (commented):
We need the changes in this branch for GFS v16.   If fv3test reproduces v16rt2, merge.  If fv3test does not reproduce v16rt2, identify steps which differ.  If difference is expected and acceptable, merge.   

GFS v16 components need to be tagged before official GFSv16 parallel begin.    DA tag can not yet be created since some items affecting results remain under active investigation.   This does not prevent this branch from being merged.

modulefiles/module_base.hera
yangfanglin: removed

parm/config/config.anal
yangfanglin: good point.  This is copied over from v16rt2. Need to check when and for how long these two have been used in the parallel.   For now, my purpose if to make sure this branch can reproduce v16rt2

parm/config/config.eupd
yangfanglin: Cory:  these parameters are copied over from v16rt2.

parm/config/config.fcst
yangfanglin: This setting is based on extensive tests made by Jun Wang and Jeff Whitaker.  It is odd, but indeed writing the 2D files using parallel_netcdf is slow than serial netcdf on WCCOSS but faster on HERA.

parm/config/config.eupd
CoryMartin-NOAA: @yangfanglin yes sorry, I meant more for asking @RussTreadon-NOAA if we decided if we need to decrease from 10000 due to runtime concerns from testing Cathy has done.

parm/config/config.gldas
yangfanglin: already removed see the right hand column

parm/config/config.gldas
yangfanglin: already removed in the new branch

parm/config/config.resources
yangfanglin: Reduced to 2 hours.

parm/config/config.resources
yangfanglin: reduced to 8 hrs.  Should use much less, but this is to allow the fcst to finish if low resources are given.

parm/config/config.resources
yangfanglin: Need to test and make changes in the future.

parm/config/config.resources
yangfanglin: changed back to 30 minutes.
Most of  the numbers are copied over from v16rt2, in which some of the larger numbers were set during trouble shootings and testing.  All these numbers need to be revisited before the gfsv16b branches are merged back to develop.

parm/config/config.eupd
RussTreadon-NOAA: Good question.  Let's ask Cathy what the DA team decided.  I don't recall.

parm/config/config.resources
RussTreadon-NOAA: v16rt2 wall times are inflated to minimize failures due to long running jobs.    All GFS v16 resource and wall clock limits need to be revisited.   v16rt2 has been and remains a moving target.  It is not a mature, ready-to-implement package.

parm/config/config.anal
RussTreadon-NOAA: Please contact Haixia to get start dates for AHI and ABI.

parm/config/config.fcst
KateFriedman-NOAA: Got it, thanks!

yangfanglin:
The experiment fv3test1 has been run successfully for all steps of the gdas cycle. Ready to merge this branch gfsv16v_paranetcdf back to gfsv16b

PR #26, opened 2020-03-04
Feature/metplus

malloryprow:
For this pull request, I did a single cycle free-forecast testing on Hera. I would be happy to re-run this test, if needed. I did to a merge with develop that was done since my last test. I did not do any testing on WCOSS since I know resources are limited with GFSv16, but let me know if you would like testing done there.


env/WCOSS_C.env
```diff
@@ -166,6 +166,11 @@ elif [ $step = "vrfy" ]; then
     export IOBUF_PARAMS="*:size=32M:count=4:verbose"
     export APRUNTRACK="$launcher -j1 -n1 -N1 -d1 -cc depth"
 
+elif [ $step = "metp" ]; then
+
+    export IOBUF_PARAMS="*:size=32M:count=4:verbose"
+    export APRUNTRACK="$launcher -j1 -n1 -N1 -d1 -cc depth"
+
```
KateFriedman-NOAA, line 173: Double checking the metp block isn't needed in the other env files. I only see this added to the WCOSS_C env file.

KateFriedman-NOAA (commented):
Changes look good. Left one comment about the env files. This is running in the v16 real-time parallel now, right? If so, that tells me that this works in cycled mode ok.

Hang-Lei-NOAA:
I saw that it was merged with develop. So, it is not working with v16
environment now. The standard branch for v16 should be feature/gfsv16b.
Hi, Kate, Is that correct?

env/WCOSS_C.env
malloryprow: It shouldn't be. It has been working fine in GFSv16, and this env file was the only one with a vrfy block so just to be similar to vrfy I added the metp block.

malloryprow:
Yup, working with GFSv16 in the real-time and the retro that Fanglin had started.

KateFriedman-NOAA:
> Yup, working with GFSv16 in the real-time and the retro that Fanglin had started.

Ok excellent!

Hang-Lei-NOAA:
After talking with Mallory, I noticed that this version was originated from
gfsv16b, and then merged with develop. So, It should be fine with v16 env.

KateFriedman-NOAA (approved):

Issue #3, opened 2019-12-17
Convert checkout.sh to manage_externals


KateFriedman-NOAA:
Original issue on VLab: https://vlab.ncep.noaa.gov/redmine/issues/66742

Replace sorc/checkout.sh with manage_externals:

https://github.com/ESMCI/manage_externals

Gave demo of manage_externals and users seemed interested. Should resolve issues with lack of flexibility in checkout.sh.

- [x] Establish initial Externals.cfg
- [x] Install v1.1.7 manage_externals tag on WCOSS-Dell
- [x] Install v1.1.7 manage_externals tag on WCOSS-Cray
- [x] Install v1.1.7 manage_externals tag on Hera
- [x] Install v1.1.8 manage_externals tag on WCOSS-Dell
- [x] Install v1.1.8 manage_externals tag on WCOSS-Cray
- [x] Install v1.1.8 manage_externals tag on Hera
- [ ] Create manage_externals module on WCOSS-Dell (replaces prior install)
- [ ] Create manage_externals module on WCOSS-Cray (replaces prior install)
- [ ] Create manage_externals module on Hera (replaces prior install)
- [x] Add note about WCOSS-Cray ssh.variant change to wiki post commit
- [x] Complete commit

KateFriedman-NOAA:
Example Externals.cfg from prior tests:

```
# External sub-modules of global-workflow

[NEMSfv3gfs]
tag = nemsfv3gfs_beta_v1.0.18
local_path = sorc/fv3gfs.fd
repo_url = ssh://vlab.ncep.noaa.gov:29418/NEMSfv3gfs
protocol = git
required = True

[GSI]
tag = fv3da.v1.0.43
local_path = sorc/gsi.fd
repo_url = ssh://vlab.ncep.noaa.gov:29418/ProdGSI
protocol = git
required = True

[EMC_post]
tag = ncep_post.v8.0.27e
local_path = sorc/gfs_post.fd
repo_url = ssh://vlab.ncep.noaa.gov:29418/EMC_post
protocol = git
required = True

[UFS_UTILS]
tag = v1.0.0
local_path = sorc/ufs_utils.fd
repo_url = ssh://vlab.ncep.noaa.gov:29418/UFS_UTILS
protocol = git
required = True

[EMC_gfs_wafs]
tag = gfs_wafs.v5.0.9
local_path = sorc/gfs_wafs.fd
[...]

PR #29, opened 2020-03-06
Issue #3 - Introduce manage_externals as replacement for checkout.sh

KateFriedman-NOAA:
Manage_externals is ready for commit consideration into the develop branch. Please review the incoming changes and test on a platform of your choice. Thanks!


aerorahul (approved):
@KateFriedman-NOAA 
Changes look good.  A few minor suggestions in the review.

I would either remove `checkout.sh`, or atleast put in bold letters saying `checkout.sh` is being deprecated and will be removed in X days or on some not-to-distant day in the future.

Aside from that, I would also suggest that in the future, clean-up's be their own PR.  This PR says nothing about clean-up and yet unrelated files are being deleted.  It will speed up review and turn-around time.

README.md
```diff
@@ -0,0 +1,49 @@
...
+## Build global-workflow:
+
+### 1. Check out components
+
+The global-workflow uses the manage_externals utility to handle checking out its components. The manic-v1.1.8 manage_externals tag is supported.
+
+Run manage_externals (checkout_externals) while at top of clone:
+
+```
+$ checkout_externals
```
aerorahul: I suggest replacing:
```$ checkout_externals```
with
```$ checkout_externals -e Externals.cfg```

This makes it explicit that is being checked out and from where.

Comment (KateFriedman-NOAA) [reply README.md]: I like it, I'll make the change, thanks!

File: README.md [hunk 2]
```diff
@@ -0,0 +1,49 @@
+# global-workflow
+Global Superstructure/Workflow currently supporting the Finite-Volume on a Cubed-Sphere Global Forecast System (FV3GFS)
+
+The global-workflow depends on the following prerequisities to be available on the system:
+
+* workload management platform / scheduler - LSF or SLURM
+* workflow manager - ROCOTO (https://github.com/christopherwharrop/rocoto)
+* modules - NCEPLIBS (various), esmf v8.0.0bs48, hdf5, intel/ips v18, impi v18, wgrib2, netcdf v4.7.0, hpss, gempak (see module files under /modulefiles for additional details)
+* manage_externals - A utility from ESMCI to checkout external dependencies. Manage_externals can be obtained at the following address and should be in the users PATH: https://github.com/ESMCI/manage_externals
```
aerorahul, line 9: You might want to add a blurb on what `manage_externals` is replacing (`checkout.sh`).
I also suggest either removing `checkout.sh` or put in big bold letters that support for `checkout.sh` is being deprecated, future PR's will be denied if an inconsistencies are found between `checkout.sh` and `Externals.cfg` and `checkout.sh` will be removed on a not-so-distant future date.

KateFriedman-NOAA:
> I would either remove `checkout.sh`, or atleast put in bold letters saying `checkout.sh` is being deprecated and will be removed in X days or on some not-to-distant day in the future.

@aerorahul The phase out time will be long-ish due to considerations for the implementation branch. There is a side discussion occurring to see how this would transition to NCO so I can't define a date to delete until we get clarification on that. If NCO asks for us to keep checkout.sh I will have to consider our options. I may explore refactoring checkout.sh to source the manage_externals Externals.cfg file so we can avoid duplication with two locations where the checkout values are defined.

Hang-Lei-NOAA:
Add manage_externals would be a great thing for developers. If we are going to remove checkout.sh, we may need to check with not only NCO, but also the GEFS group. I remembered that they said that they need to use checkout.sh.

KateFriedman-NOAA:
Received initial buy-in from NCO. Completing PR.

PR #30, opened 2020-03-06
Move all default values of configurable variables into defaults/

jkuang-exploro:
This pull request is of a clean up. All default values are moved to defaults/ as it should be. 

Historically, a lot of default values are in /schema. This make it an extra burden for developers to check if where are the default values set, if any, and where. This is not right.

In this way, developers only need to go to /defaults to check if the default values are there and with a desired value.

The usage of CROW won't change otherwise.


jkuang-exploro:
@JessicaMeixner-NOAA @binli2337 This is an overdue change that put default values into defaults/. As of now, some of them are in defaults/ while some of them are in schema/, which is hard to manage.

Please make sure it could reproduce the EXPDIR as of now. You may not need to run the entire experiment, just reproduce the EXPDIR would be OK.

JessicaMeixner-NOAA:
@JianKuang-NOAA  I'll test and let you know the results.

binli2337 (approved):

JessicaMeixner-NOAA:
@binli2337 can you share your control directory location so I can use it to test my next round of changes against it?

Issue #34, opened 2020-03-13
Update GDA in operations branch


KateFriedman-NOAA:
Update the DMPDIR paths and associated scripts to use the new WCOSS P3 GDA.

KateFriedman-NOAA:
Branch feature/operations_gda created off operations branch.

KateFriedman-NOAA:
Test to reproduce ops prepbufr file worked with GDA updates. Further test against ops showed we're missing the update to the satinfo file from ops. The DA team created the "DA_GFS_v15.3" tag with the update and tests using it confirm it reproduces ops. Will need to update operations branch GSI tag to "DA_GFS_v15.3".

KateFriedman-NOAA:
Changes merged into operations branch and feature/operations_gda branch deleted.

PR #35, opened 2020-03-16
Issue #34 - update GDA DMPDIR paths and associated scripts

KateFriedman-NOAA:
GDA paths in feature/operations_gda tests on Dell. The operational prepbufr was reproduced and the correct files were used in the DA jobs. Needed to change GSI tag to reproduce ops (satinfo change). That update will come after this.


RussTreadon-NOAA (changes requested):
GDA changes tested on Dell and found to be acceptable.  Need to update sorc/checkout.sh to ProdGSI tag fv3da_gfsv15.2.9

RussTreadon-NOAA (approved):
Look good to me.

Issue #38, opened 2020-03-23
Update EMC_verif-global pointer to new GitHub repo


KateFriedman-NOAA:
The EMC_verif-global repo has moved from VLab to GitHub:

https://github.com/NOAA-EMC/EMC_verif-global

The v1.5.0 tag:

https://github.com/NOAA-EMC/EMC_verif-global/tree/verif_global_v1.5.0

KateFriedman-NOAA:
Created branch feature/verif-tag

KateFriedman-NOAA:
Made changes to Externals.cfg and checkout.sh:

```
[Kate.Friedman@v72a3 global-workflow]$ git clone github:global-workflow ./feature-verif-tag
Cloning into './feature-verif-tag'...
remote: Enumerating objects: 84, done.
remote: Counting objects: 100% (84/84), done.
remote: Compressing objects: 100% (67/67), done.
remote: Total 31896 (delta 29), reused 33 (delta 17), pack-reused 31812
Receiving objects: 100% (31896/31896), 70.12 MiB | 29.46 MiB/s, done.
Resolving deltas: 100% (21951/21951), done.
[Kate.Friedman@v72a3 global-workflow]$ cd feature-verif-tag/
[Kate.Friedman@v72a3 feature-verif-tag]$ git branch
* develop
[Kate.Friedman@v72a3 feature-verif-tag]$ git log -1
commit 057b2a82fa43f7bc36c3e757ca7d48fb86b9541c (HEAD -> develop, origin/develop, origin/HEAD)
Merge: 0377d20f 622167d5
Author: Kate Friedman <kate.friedman@noaa.gov>
Date:   Wed Mar 11 12:00:58 2020 -0400

    Merge pull request #29 from NOAA-EMC/feature/manage_externals

    Issue #3 - Introduce manage_externals as replacement for checkout.sh
[Kate.Friedman@v72a3 feature-verif-tag]$ git checkout -b feature/verif-tag
Switched to a new branch 'feature/verif-tag'
[Kate.Friedman@v72a3 feature-verif-tag]$ git branch
  develop
* feature/verif-tag
[Kate.Friedman@v72a3 feature-verif-tag]$ git push origin feature/verif-tag
Total 0 (delta 0), reused 0 (delta 0)
remote:
remote: Create a pull request for 'feature/verif-tag' on GitHub by visiting:
remote:      https://github.com/NOAA-EMC/global-workflow/pull/new/feature/verif-tag
[...]

PR #39, opened 2020-03-23
Issue #38 - update EMC_verif-global pointer from VLab to GitHub

KateFriedman-NOAA:
Pointer updated in both Externals.cfg and checkout.sh. Both checkout methods tested successfully on Venus.


malloryprow (approved):
Looks good, glad we could get EMC_verif-global on GitHub and checked out in the global workflow so easily!

PR #40, opened 2020-03-25
Issue #21 - Add support for WCOSS phase 3.5

KateFriedman-NOAA:
Issue #21 - add support for WCOSS phase 3.5

Added "--partition" flag to setup_expt*.py scripts. Partition flag not required and does not do anything on WCOSS-Cray or Hera, only does something when invoked on WCOSS-Dell with "3p5" as the partition value. If partition flag is not invoked on WCOSS-Dell then it defaults to phase 3.

To use on WCOSS-Dell: --partition 3p5

When the partition flag is used with a value of "3p5" it will set QUEUE to "dev2" and QUEUE_ARCH to "dev2_transfer". Then when setup_workflow scripts are run it will see that QUEUE is set to a phase 3.5 queue and use the phase 3.5 ppn=40 value for resource calculations instead of ppn=28 which is phase 3's value. Additional phase 3.5 queues that QUEUE can be set to prior to running setup_workflow*.py are "devonprod2" and "devmax2".

Also removed machine if-blocks in config.base and added different machine paths to setup_expt scripts which populate config.base variables.

Example for cycled mode:

```
./setup_expt.py --pslot testcyc --configdir /gpfs/dell2/emc/modeling/save/Kate.Friedman/git/global-workflow/port2wcoss3p5/parm/config --idate 2020010212 --edate 2020010218 --expdir /gpfs/dell2/emc/modeling/save/Kate.Friedman/expdir --comrot /gpfs/dell3/ptmp/Kate.Friedman/comrot --resdet 768 --resens 384 --nens 80 --gfs_cyc 4 --partition 3p5
```

Example for free-forecast mode:

```
./setup_expt_fcstonly.py --pslot testff --configdir /gpfs/dell2/emc/modeling/save/Kate.Friedman/git/global-workflow/port2wcoss3p5/parm/config --idate 2020010212 --edate 2020010218 --expdir /gpfs/dell2/emc/modeling/save/Kate.Friedman/expdir --comrot /gpfs/dell3/ptmp/Kate.Friedman/comrot --res 192 --gfs_cyc 4 --partition 3p5
```

### DIFF
--- a/parm/config/config.base.emc.dyn
+++ b/parm/config/config.base.emc.dyn
@@ -12,19 +12,9 @@ export machine="@MACHINE@"
 export RUN_ENVIR="emc"
 
 # Account, queue, etc.
-if [ $machine = "HERA" ]; then
-
-    export ACCOUNT="fv3-cpu"
-    export QUEUE="batch"
-    export QUEUE_ARCH="service"
-
-elif [ $machine = "WCOSS_C" -o $machine = "WCOSS_DELL_P3" ]; then
-
-    export ACCOUNT="GFS-DEV"
-    export QUEUE="dev"
-    export QUEUE_ARCH="dev_transfer"
-
-fi
+export ACCOUNT="@ACCOUNT@"
+export QUEUE="@QUEUE@"
+export QUEUE_ARCH="@QUEUE_ARCH@"
 
 # Project to use in mass store:
 HPSS_PROJECT=emc-global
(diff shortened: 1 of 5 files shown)


parm/config/config.base.emc.dyn
```diff
@@ -12,19 +12,9 @@ export machine="@MACHINE@"
...
-    export QUEUE_ARCH="service"
-
-elif [ $machine = "WCOSS_C" -o $machine = "WCOSS_DELL_P3" ]; then
-
-    export ACCOUNT="GFS-DEV"
-    export QUEUE="dev"
-    export QUEUE_ARCH="dev_transfer"
-
-fi
+export ACCOUNT="@ACCOUNT@"
```
yangfanglin, line 15: Kate, removing the references to specific machines in config.base is fine, but where will these parameters be defined, including account, queue and directory settings ?  Do  you have a script that defines these parameters for each machine and then call the python setup scripts ?  No one can remember all the parameters for different platforms.  I have been using a script, for example, /gpfs/dell6/emc/modeling/noscrub/emc.glopara/para_gfs/misc/rocoto_cycled_warm.sh, to set up my runs.

ush/rocoto/setup_expt.py
```diff
@@ -149,6 +161,48 @@ def edit_baseconfig():
     nens = args.nens
     cdump = args.cdump
     gfs_cyc = args.gfs_cyc
+    partition = args.partition
+
+    # Set machine defaults
+    if machine is 'WCOSS_DELL_P3':
```
yangfanglin, line 167: OK, I see now you have most of the parameters defined in the setup python script. It is fine.  Still, users may have to edit these settings locally in their EXPRDIR because not all are using or can use the same disk allocation.

yangfanglin (approved):

ush/rocoto/setup_expt.py
KateFriedman-NOAA: The defaults we had in config.base.emc.dyn before were moved into the setup_expt scripts so the final config.base generated will only include the defaults for the machine the user is on. Users will then need to update their config.base before running setup_workflow script, just as they had to before.

ush/rocoto/setup_expt_fcstonly.py
```diff
@@ -115,6 +127,48 @@ def edit_baseconfig():
...
+      dmpdir = '/gpfs/dell3/emc/global/dump'
+      nwprod = '/gpfs/dell1/nco/ops/nwprod'
+      homedir = '/gpfs/dell2/emc/modeling/noscrub/$USER'
+      stmp = '/gpfs/dell3/stmp/$USER'
+      ptmp = '/gpfs/dell3/ptmp/$USER'
+      noscrub = '/gpfs/dell2/emc/modeling/noscrub/$USER'
+      account = 'GFS-DEV'
+      queue = 'dev'
+      queue_arch = 'dev_transfer'
+      if partition is not None and partition in ['3p5']:
```
aerorahul, line 145: `if partition is not None` is not required here, because if it is `None`, the `partition in ['3p5']` will evaluate to `False`.

KateFriedman-NOAA: Thanks, I'll remove that bit.

KateFriedman-NOAA:
Committed removal of unneeded logic at f4ee509.

ush/rocoto/setup_expt.py
aerorahul, line 167: @KateFriedman-NOAA 
You should be using `if machine == 'WCOSS_DELL_P3'` and **not** `if machine is 'WCOSS_DELL_P3'`.
The **is** relational evaluates to `True` if the two variables point to the same object.
The **==** relational evaluates to `True` if the variables are equal.

KateFriedman-NOAA: Gotcha, thanks!

KateFriedman-NOAA:
Committed syntax correction at b133700.

PR #43, opened 2020-03-27
Update README.md

jkuang-exploro:


jkuang-exploro:
@binli2337 I updated readme for feature/coupled-crow branch. Adding a line that requiring people to use FIX_SCRUB: True option.

binli2337 (approved):

JessicaMeixner-NOAA:
Should https://github.com/NOAA-EMC/global-workflow/blob/feature/coupled-crow/workflow/user.yaml.default be updated to set FIX_SCRUB as True?

PR #44, opened 2020-03-27
Update user.yaml.default

jkuang-exploro:
Change default FIX_SCRUB to True


jkuang-exploro:
Default FIX_SCRUB change to True.

JessicaMeixner-NOAA (approved):
Thanks Jian!

PR #45, opened 2020-03-27
Update checkout.sh

jkuang-exploro:
Do not checkout gsi when building coupled.


binli2337 (approved):

JessicaMeixner-NOAA:
I think this is great update, but is this the only repository we do not need to checkout? We only build 2 repositories.

PR #58, opened 2020-04-17
Orion porting 

jkuang-exploro:
This PR enables the features/coupled-crow branch to run on Orion. Technical elements like load_fv3gfs_modules.sh and ORION.env are borrowed from the port2orion branch. CROW configuration files (YAML) are largely maintained the same, except 1, platform file orion.yaml is created; 2, a new option called mom6ic_prepared is created under the "settings" section, allowing the user to skip the IC steps.

The procedure to set up and run workflows are the same as the current features/coupled-crow branch.


KateFriedman-NOAA (approved):
Orion changes look good.

Externals.cfg
```diff
@@ -0,0 +1,46 @@
+# External sub-modules of global-workflow
```
JessicaMeixner-NOAA, line 1: Are we moving to externals instead of checkout.sh?

sorc/checkout.sh
```diff
@@ -1,110 +1,87 @@
...
-model=$1
-fi
-model=${model:-"uncoupled"}
-
-if [ $model = "coupled" ]; then
-echo fv3_mom6_cice5 checkout ...
-rm -f ${topdir}/checkout-fv3_coupled.log
-
-if [[ ! -d fv3_coupled.fd ]] ; then
-    git clone https://github.com/ufs-community/ufs-s2s-model fv3_coupled.fd >> ${topdir}/checkout-fv3_coupled.log 2>&1
```
JessicaMeixner-NOAA: I think we still want to checkout the coupled model?   Or did we move too externals with this commit as well?

JessicaMeixner-NOAA (commented):
@JianKuang-NOAA  does your branch have all the latest changes from NOAA-EMC:feature/coupled-crow merged in?   

I think there are some changes to the checkout.sh scrip that need to remain unchanged or does this commit also move us to the manage_externals feature? If so, we need to update the readme.  

I also noticed a lot of module files scripts got deleted.  I'm assuming we don't need any of those, but haven't re-run on hera to make sure.  Does someone need to do that?

sorc/checkout.sh
jkuang-exploro: Oh. This is wrong...

binli2337 (approved):

PR #70, opened 2020-05-15
Copy stochastic pattern state on restart

WalterKolczynski-NOAA:
To maintain consistency, the stochastic pattern must be preserved
when restarting. Now the stochastic pattern saved at restart times
will be copied for use in the restart.


XianwuXue-NOAA (approved):

ajhenrique (approved):
Looks good!

PR #73, opened 2020-05-19
Fix error message in wave gempak script

WalterKolczynski-NOAA:
One of the error messages in the wave gempak script was not
printed to STDOUT and did not indicate it was a fatal error.


RobertoPadilla-NOAA (approved):

ajhenrique (approved):
Looks good!

PR #77, opened 2020-05-27
automate fix file linking on Orion

jkuang-exploro:


JessicaMeixner-NOAA (approved):
@JianKuang-NOAA  this is for running v3.1 the fix files are correct, we'll need another update for running with the p4 fix files for both Hera and orion.

PR #81, opened 2020-06-03
wave configuration update

jkuang-exploro:
This PR address the following issues:
1, Fixed a bug in modulefile/module-setup.sh.inc causing user-defined modules not purged, on Orion.
2, Built out the configuration system for waves
3, Upgraded modularized forecast script for the handling of additional parameters (ie, PETS parameters) and the generation of nems.configure
4, Moved some of the variables in JWAVE_* scripts into the configuration system


aerorahul:
is adding a ton of wave specific scripts in global-workflow a good idea? Can the wave components be handled the same way the other components are being handled, e.g. GSI, POST, etc.

jkuang-exploro:
@aerorahul That's a good suggestion to bring up. My thinking is that after the contents are fully tested to some level of robustness, they should be moved to a sub-repo and linked from there. And the current emplacement of them should be temporary.

aerorahul:
> @aerorahul That's a good suggestion to bring up. My thinking is that after the contents are fully tested to some level of robustness, they should be moved to a sub-repo and linked from there. And the current emplacement of them should be temporary.

1. If the contents are not tested robustly, they should not be merged.
2. Separation later usually never works as that becomes more difficult, inter-twined and takes lower priority.  If a separation is kept from the beginning, you don't have to worry about it.  Seeing as these are effectively new additions, they can be from a different submodule/repository now itself.

JessicaMeixner-NOAA:
@aerorahul these scripts are all updates from the gfsv16b branch where the waves scripts are not kept in a separate repository. There are also scripts in this branch for ocean and ice models. While there might be merit to having these each in their own repository, we're also moving to a unified coupled model, so don't these scripts belong here along with the other model scripts?

ush/parsing_namelists_FV3.sh
```diff
@@ -216,6 +224,13 @@ if [ $CCPP_SUITE = "FV3_GSD_v0" ]; then
   bl_mynn_tkeadvect=${bl_mynn_tkeadvect:-".true."}	! In config.fcst
   bl_mynn_edmf_mom=${bl_mynn_edmf_mom:-"1"}	! In config.fcst
 EOF
+elif [ $CCPP_SUITE = "S2S" ]; then
+  cat >> input.nml << EOF
+  oz_phys      = ${oz_phys}             ! In config.fcst
+  oz_phys_2015 = ${oz_phys_2015}        ! In config.fcst
+  cplwav       = ${cplwav}              ! CROW configured
```
JessicaMeixner-NOAA: cplwav and cplwav2atm are not CCPP_SUITE specific and are also valid for IPD, they should not be within an if statement.

ush/parsing_namelists_FV3.sh
```diff
@@ -216,6 +224,13 @@ if [ $CCPP_SUITE = "FV3_GSD_v0" ]; then
   bl_mynn_tkeadvect=${bl_mynn_tkeadvect:-".true."}	! In config.fcst
   bl_mynn_edmf_mom=${bl_mynn_edmf_mom:-"1"}	! In config.fcst
 EOF
+elif [ $CCPP_SUITE = "S2S" ]; then
```
JessicaMeixner-NOAA: There is not a CCPP_SUITE at this time defined as "S2S" this is a duplication of the if block above for the CCPP_SUITE that the s2s model is using right now: FV3_GFS_v15p2_coupled

workflow/cases/coupled_free_forecast_s2s.yaml
```diff
@@ -0,0 +1,53 @@
+case:
+  places:
+    workflow_file: layout/free_forecast_gfs.yaml
+    MOM6IC: /scratch1/NCEPDEV/nems/Bin.Li/S2S/FROM_HPSS
```
JessicaMeixner-NOAA: In PR 80 this variable was renamed and the contents of the file was moved to: 
BASE_CPLIC: /scratch2/NCEPDEV/climate/climpara/S2S/IC

workflow/cases/coupled_free_forecast_s2s.yaml
```diff
@@ -0,0 +1,53 @@
+case:
```
JessicaMeixner-NOAA: I am assuming the _s2s means running with waves?  There are no updates in the README that correspond to running a different case file.

jkuang-exploro: @JessicaMeixner-NOAA Yes this is the dedicated case file for wave-included run. I haven't managed to update the README yet. I plan to update it after we could generally agree on these changes.

ush/parsing_namelists_FV3.sh
jkuang-exploro: I will fix it. @JessicaMeixner-NOAA

ush/parsing_namelists_FV3.sh
jkuang-exploro: I will fix it. @JessicaMeixner-NOAA

aerorahul:
> @aerorahul these scripts are all updates from the gfsv16b branch where the waves scripts are not kept in a separate repository. There are also scripts in this branch for ocean and ice models. While there might be merit to having these each in their own repository, we're also moving to a unified coupled model, so don't these scripts belong here along with the other model scripts?

@JessicaMeixner-NOAA I am all for unification.  However, if there is a need/desire to run a single component individually, for development, testing etc., carrying around the whole baggage of the global-workflow is cumbersome and developers are easily overwhelmed.  Besides, unified does not have to be in the same repository or code base. e.g. WW3 code does not reside in the UFS code base. 

One should be able to configure and link what is necessary, not have everything just plopped down into a single source tree.

My 2 cents.

JessicaMeixner-NOAA:
@aerorahul that's 

> > @aerorahul these scripts are all updates from the gfsv16b branch where the waves scripts are not kept in a separate repository. There are also scripts in this branch for ocean and ice models. While there might be merit to having these each in their own repository, we're also moving to a unified coupled model, so don't these scripts belong here along with the other model scripts?
> 
> @JessicaMeixner-NOAA I am all for unification. However, if there is a need/desire to run a single component individually, for development, testing etc., carrying around the whole baggage of the global-workflow is cumbersome and developers are easily overwhelmed. Besides, unified does not have to be in the same repository or code base. e.g. WW3 code does not reside in the UFS code base.
> 
> One should be able to configure and link what is necessary, not have everything just plopped down into a single source tree.
> 
> My 2 cents.

I'm just following what was done in gfsv16 and what we have so far been doing for the s2s application, if a strategy exists for putting other model components in another repository and to be linked, then I'll happily follow.  The wave, ocean and ice components do not have the capabilities to be run as standalone components within this workflow.  This is just for running the fully coupled system.   I'm happy to conform to whatever strategies are decided on, but for now I believe this is following current strategies.

RussTreadon-NOAA:
As Rahul notes the GSI/EnKF DA component does not reside in the
global-workflow repo.  Instead it resides in gerrit:ProdGSI.   This repo is
currently being ported to github.   When we check out the v16 workflow it
does not contain the GSI/EnKF DA component.  sorc/checkout.sh clones the DA
repo and checks out the appropriate branch (tag) into sorc/gsi.fd.
sorc/build_all.sh executes the DA build script in sorc/gsi.fd/ush.
sorc/link_fv3gfs.sh remaps via ln or cp the DA repo structure into the flat
gfs.vX.Y.Z structure NCO wants for operations.

workflow/cases/coupled_free_forecast_s2s.yaml
jkuang-exploro: Thanks. I updated it. @JessicaMeixner-NOAA

jkuang-exploro:
@JessicaMeixner-NOAA I've merged in the latest changes of PR#80 into this PR (one additional nems.configure template file).

jkuang-exploro:
@JessicaMeixner-NOAA I've put the variables requested into the configuration system. Please take a look.

JessicaMeixner-NOAA (approved):
The "base" no waves case should still work.  The waves work except for post which will there will be an update later for.  I approve this, but it should be tested by @jiandewang or @JianKuang-NOAA or others before merged.

jiandewang:
tried twice on Orion, fcst and post finished properly, ocnpost failed due to library reason but it's not within the scope of this PR. So I am approving this PR.

jiandewang (approved):

binli2337 (approved):

PR #83, opened 2020-06-03
revive GFS forecast break-point restart capability with IAU turned on

yangfanglin:
modified:   scripts/exglobal_fcst_nemsfv3gfs.sh and  parm/config/config.fcst

GFS forecast restrat capability from a breakpoint is no longer working
with IAU turned on.  This function has been overhauled to make it more
general and works for cases with and without IAU

So far both atmf and sfcf data reproduced after restart, but netCDF file headers seems to differ. Now running another test with UPP turned on.  May need to run more tests with vrfy turned on as well.

### DIFF
--- a/parm/config/config.fcst
+++ b/parm/config/config.fcst
@@ -214,8 +214,22 @@ elif [[ "$CDUMP" == "gfs" ]] ; then # GFS cycle specific parameters
     fi
 
     # Write gfs restart files to rerun fcst from any break point         
-     export restart_interval=${restart_interval_gfs:-0}
+    export restart_interval_gfs=${restart_interval_gfs:-0}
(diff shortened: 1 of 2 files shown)


yangfanglin:
Russ and Jun, 
The restart capability is now working except REFC in the "master" file does not reproduce.  It needs extra effort to find the cause.  To facilitate the high watermark test we will be perform over the weekend, I'd suggest we merge the changes I made to feature/gfsv16b. Please check and see if you agree.

RussTreadon-NOAA:
Given that the high watermark test begins tomorrow I agree that we need
these changes in the v16 workflow branch sooner rather than later.

junwang-noaa:
Sure, I think it's good to merge the changes.

Jun

PR #93, opened 2020-06-24
fix for exiting properly with error for wave prep/init scripts

JessicaMeixner-NOAA:
The postmsg was deleted because it seems odd to exit with error and have a message "completed normally".  It's my understanding that this is no longer a EE2 requirement.

### DIFF
--- a/scripts/exwave_init.sh
+++ b/scripts/exwave_init.sh
@@ -231,7 +231,6 @@
   echo ' '
   [[ "$LOUD" = YES ]] && set -x
 
-  msg="$job completed normally"
-  postmsg "$jlogfile" "$msg"
+  exit $err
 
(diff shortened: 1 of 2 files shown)


KateFriedman-NOAA (approved):
Fix looks good. Will let Fanglin comment before submitting.

yangfanglin (approved):
Changes look fine.  I thought I had told Henrique not to use jlogfile since NCO does not support it anymore.  Apparently this fell into the crack.

JessicaMeixner-NOAA:
@yangfanglin Does the removal of jlogfiles need to go on the todo list?

yangfanglin:
There are probably still some jlogfiles in the workflow.  Let's wait for Eric to check the EE2 compliance.  I think NCO is not enforcing the removal of jlogfiles  yet.

PR #103, opened 2020-07-17
Update miscellaneous pieces for GEFS v12 NCO parallel

WalterKolczynski-NOAA:
A couple small changes before GEFS heads to NCO's real-time parallel:

1. Removed deprecated USE_PREINST_LIBS, which included specifying non-NCO (and non-WCOSS). NCO disliked these, even if they weren't used

2. Updated the ESMF module to an NCO-installed version for production now that it is available.


XianwuXue-NOAA (approved):
you didn't change esmf/8.0.1 for cray, however, this is not affect for the current use.

PR #113, opened 2020-08-05
make wavepost a singular job

jkuang-exploro:
This is a simple PR to make wave post job a singular job instead of a task array. Please test it on both hera and orion to check if the wavepost could run smoothly.


JessicaMeixner-NOAA (approved):

jiandewang (commented):
look good

PR #115, opened 2020-08-10
add "ic_source" option to control cfsr/gfs IC selection in "settings"

jkuang-exploro:
This PR adds "ic_source" option into "settings" section in the configuration system. Default selection for this is "cfsr"

### DIFF
--- a/workflow/cases/atm_free_forecast.yaml
+++ b/workflow/cases/atm_free_forecast.yaml
@@ -13,7 +13,7 @@ case:
     nems_temp_cold: 'atm'
     mom6ic_prepared: .false.
     KEEPDATA: NO
-    ic_source: cfsr
+    ic_source: coupled_ic
 
   nsst:
     NST_MODEL: 0
--- a/workflow/cases/atm_free_forecast.yaml
+++ b/workflow/cases/atm_free_forecast.yaml
@@ -9,8 +9,8 @@ case:
 
     cplflx: .false.
     print_esmf: .true.
-    nems_temp: 'med_atm_ocn_ice'
-    nems_temp_cold: 'medcold_atm_ocn_ice'
+    nems_temp: 'atm'
+    nems_temp_cold: 'atm'
     mom6ic_prepared: .false.
     KEEPDATA: NO
     ic_source: cfsr
@@ -19,7 +19,6 @@ case:
     NST_MODEL: 0
 
   output_settings:
-    OCN_INTERVAL: 24
     FHOUT_GFS: 6
     FHMIN_GFS: 0
     FHMAX_GFS: 48
@@ -44,9 +43,6 @@ case:
        WGRP_NTASKS: 24
        WRTIOBUF: "32M"
     cdmbgwd: "1.0,1.2"
-
-  ocn_settings:
-    OCNPETS: 120
 
   ice_settings:
     ICEPETS: 48
--- a/workflow/cases/atm_free_forecast.yaml
+++ b/workflow/cases/atm_free_forecast.yaml
@@ -7,7 +7,7 @@ case:
     SDATE: 2013-04-01t00:00:00
     EDATE: 2013-04-01t00:00:00
 
-    cplflx: .true.
+    cplflx: .false.
     print_esmf: .true.
     nems_temp: 'med_atm_ocn_ice'
     nems_temp_cold: 'medcold_atm_ocn_ice'
--- a/workflow/cases/coupled_free_forecast.yaml
+++ b/workflow/cases/coupled_free_forecast.yaml
@@ -13,7 +13,7 @@ case:
     nems_temp_cold: 'medcold_atm_ocn_ice'
     mom6ic_prepared: .false.
     KEEPDATA: NO
-    ic_source: cfsr
+    ic_source: coupled_ic
 
   nsst:
     NST_MODEL: 0
--- a/workflow/defaults/settings.yaml
+++ b/workflow/defaults/settings.yaml
@@ -58,7 +58,8 @@ default_settings: &default_settings
       do: .true.
     - otherwise: .false.
 
-  ic_source: coupled_ic  # choices: coupled_ic, fv3_ic; default: coupled_ic
+  ic_source: coupled_ic  # choices: coupled_ic, fv3_ic; default: coupled_ic. This option controls which
+                         # script will be used for IC preparation: coupled_ic.sh or fv3_ic.sh
 
   mom6ic_prepared: .false. # If set True, User need to manually copy mom6ic files into ROTIDR
 
--- a/workflow/defaults/settings.yaml
+++ b/workflow/defaults/settings.yaml
@@ -58,7 +58,7 @@ default_settings: &default_settings
       do: .true.
     - otherwise: .false.
 
-  ic_source: cfsr  # choices: cfsr, gfs; default: cfsr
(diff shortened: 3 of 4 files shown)


workflow/cases/atm_free_forecast.yaml
```diff
@@ -0,0 +1,56 @@
...
+    nems_temp_cold: 'medcold_atm_ocn_ice'
+    mom6ic_prepared: .false.
+    KEEPDATA: NO
+    ic_source: cfsr
+
+  nsst:
+    NST_MODEL: 0
+
+  output_settings:
+    OCN_INTERVAL: 24
```
JessicaMeixner-NOAA: What is OCN_INTERVAL? Is this an ATM variable? Otherwise this could be removed, along with the ocn_settings and ice_settings

workflow/cases/coupled_free_forecast.yaml
```diff
@@ -13,6 +13,7 @@ case:
     nems_temp_cold: 'medcold_atm_ocn_ice'
     mom6ic_prepared: .false.
     KEEPDATA: NO
+    ic_source: cfsr
```
JessicaMeixner-NOAA: Which initial condition is this for? ATM, OCN, ICE, WAV?

workflow/cases/atm_free_forecast.yaml
```diff
@@ -0,0 +1,56 @@
...
+    FHMAX_GFS: 48
+    FHMAX_HF_GFS: 0
+    FHOUT_HF_GFS: -1
+
+  fv3_gfs_settings:
+    CASE: C384
+    LEVS: 65
+    DELTIM: 450
+    SEEDLET: 10
+    CCPP_SUITE: FV3_GFS_v15p2_coupled
```
benwgreen, line 33: If running uncoupled, might it be better to set the suite as "FV3_GFS_v15p2"? That will add NSST and remove sfc_cice. And then associated namelist changes for NSST would have to be made where appropriate, right?
Debatable on whether to include "FV3_GFS_v15p2" in the list of CCPP suites to build, since that adds to build time and very few people will be using coupled workflow to run uncoupled (at least for a while)

jkuang-exploro: You can of course change the CCPP suite as long as it is currently supported. But then you've added another degree of freedom into your experiment (couple/uncouple, and CCPP_SUITE)

workflow/cases/atm_free_forecast.yaml
jkuang-exploro: Agree. This should be removed.

jkuang-exploro: @JessicaMeixner-NOAA I've also changed the nems.configure templates to "atm" in most recent push.

workflow/cases/coupled_free_forecast.yaml
jkuang-exploro: @JessicaMeixner-NOAA This is the one that we are using for the coupled experiment now, for all components.

JessicaMeixner-NOAA: @JianKuang-NOAA in the current feature/coupled-crow branch, each component has IC variables: 
CPL_ATMIC - atm IC determination for coupled model 
CPL_OCNIC - ocean IC 
CPL_ICEIC - ice IC
CPL_WAVIC - wave IC 
(See: https://github.com/NOAA-EMC/global-workflow/blob/feature/coupled-crow/jobs/rocoto/coupled_ic.sh for their use) 

I see the purpose of this variable, but I think it needs to have a different name or target value.  For example if the "ic_source" was "CPL" to signify that you want to run jobs/rocoto/coupled_ic.sh to source your ICs that makes sense.  But ic_source=cfsr is very misleading because we need and want each component to be able to have a different IC source.

jkuang-exploro: OK. Then let's name it "coupled_ic", just so that it means this will trigger the couple_ic.sh script to take care of the IC stacking, instead of fv3_ic.sh

workflow/cases/atm_free_forecast.yaml
JessicaMeixner-NOAA: You will have to make sure you compile the model with "FV3_GFS_v15p2" when you install the model.  This is not the default behavior and would have to be manually changed. I do think you will need to change this to run coupled versus uncoupled.  As this was copied directly from the coupled mode, you will have to change any variables here to be different if you want (such as turning on NSST, which is set to 0 as of now).

workflow/defaults/settings.yaml
```diff
@@ -58,6 +58,8 @@ default_settings: &default_settings
       do: .true.
     - otherwise: .false.
 
+  ic_source: coupled_ic  # choices: coupled_ic, fv3_ic; default: coupled_ic
```
JessicaMeixner-NOAA: Can we add more of an explanation/definition of the variable here?  It seems obvious now, but adding information that it determines which script is used for ICs would be helpful to those who didn't follow this conversation I think.

workflow/cases/coupled_free_forecast.yaml
JessicaMeixner-NOAA: This should be updated to coupled_ic

JessicaMeixner-NOAA (changes requested):
In addition to updates mentioned in the comments, it would be good to hear how the uncoupled (if it at least gets past the first IC step) and coupled tests go before approving.

workflow/cases/atm_free_forecast.yaml
benwgreen: Okay, I'm fine with it as-is. Better not to add degrees of freedom and fall down a rabbit hole at this point.

workflow/defaults/settings.yaml
jkuang-exploro: Your suggestion is taken

workflow/cases/coupled_free_forecast.yaml
jkuang-exploro: Your suggestion is taken

jkuang-exploro:
@JessicaMeixner-NOAA @benwgreen Just want to briefly follow up, are we able to merge this PR?

workflow/cases/coupled_free_forecast.yaml
JessicaMeixner-NOAA: This was not updated to coupled_ic

JessicaMeixner-NOAA (changes requested):
The code itself looks fine except one more variable (marked) needs to be updated to coupled_ic.  If @benwgreen says it's okay for the standalone (or at least an improvement) and someone tests and it still runs the coupled, I have no issues. I do not have time to test the coupled so someone else will have to do that.

benwgreen:
> @JessicaMeixner-NOAA @benwgreen Just want to briefly follow up, are we able to merge this PR?

Sorry, I had a family issue and wasn't able to get around to this. I won't have time to test this for a while, but the code looks fine with me. I will defer to others.

jkuang-exploro:
I will close this PR since it looks like it's no longer needed and the branch is no longer up-to-date with feature/coupled-crow. I will re-initiate this effort if this feature is requested again.

jkuang-exploro:
I will close this PR since it looks like it's no longer needed and the branch is no longer up-to-date with feature/coupled-crow. I will re-initiate this effort if this feature is requested again.

PR #124, opened 2020-09-02
Splitting post jobs for waves

JessicaMeixner-NOAA:
This adds two additional wave jobs for point processing at the end of the job and then for boundary point jobs removing the points from the side-by-side jobs.   The boundary point job supplies new output needed for downstream jobs and requires a change in fix files.  There will be another change for optimizing these jobs as a follow up to this. 

This also adds the cdo module used in wave prep and small changes related to using this module in ush/wave_prnc_cur.sh
There will be no fix-file changes related to this. 

There are a few clean up/ changes for better error handling. 

Changes were made for several of the resources for wave jobs. There should be no impact in performance. 

Testing the forecast only reproduces the real-time parallel for atm and has the expected changes for waves.


KateFriedman-NOAA:
I will be reviewing the changes and testing them against v16rt2 on WCOSS-dev before approving.

jobs/JGLOBAL_WAVE_POST_BNDPNT
```diff
@@ -0,0 +1,107 @@
...
+export COMOUT=${COMOUT:-${ROTDIR}/${CDUMP}.${PDY}/${cyc}/$COMPONENT}
+
+export COMINice=${COMINice:-${COMROOTp2}/omb/prod}
+export COMINwnd=${COMINwnd:-${COMROOT}/gfs/prod}
+export COMIN_WAV_CUR=${COMIN_WAV_CUR:-${COMROOTp2}/rtofs/prod}
+
+mkdir -p $COMOUT/station
+
+# Set mpi serial command
+export wavempexec=${launcher:-"mpirun -n"}
```
yangfanglin: These should be defined in  the ./env/WCOSS_DELL_P3.env file for WCOSS, and for others platforms as well

jobs/JGLOBAL_WAVE_POST_PNT
```diff
@@ -0,0 +1,107 @@
...
+
+export COMINice=${COMINice:-${COMROOTp2}/omb/prod}
+export COMINwnd=${COMINwnd:-${COMROOT}/gfs/prod}
+export COMIN_WAV_CUR=${COMIN_WAV_CUR:-${COMROOTp2}/rtofs/prod}
+
+mkdir -p $COMOUT/station
+
+# Set mpi serial command
+export wavempexec=${launcher:-"mpirun -n"}
+export wave_mpmd=${mpmd:-"cfp"}
```
yangfanglin: define in  files under ./env

jobs/JGLOBAL_WAVE_POST_BNDPNT
JessicaMeixner-NOAA: launcher is defined in the ./env/*.env files.  Do you want the definition of wavempexec also set in env or the default value removed?

jobs/JGLOBAL_WAVE_POST_PNT
JessicaMeixner-NOAA: launcher and mpmd are defined in env files.  Should wavempexec/wave_mpmd  also defined there?

yangfanglin: We do define these type of commands in ./env using, for instance,  APRUN_WAVE,  APRUN_CFP etc for different platforms

JessicaMeixner-NOAA: done.

jobs/JGLOBAL_WAVE_POST_BNDPNT
JessicaMeixner-NOAA: done.

KateFriedman-NOAA (approved):
Ran updates successfully in test on Mars. Jessica checked output and confirmed jobs ran correctly. Also tested gempak updates from Roberto in the same test and he confirmed those updates ran correctly and that folks at OPC will now examine gempak output once the updates are in the real-time parallel.

PR #137, opened 2020-09-18
Bugfix/coupled crow

jkuang-exploro:
DATADIR re-organize
updated PE allocation


jkuang-exploro:
This PR also address the ic_source option, enabling coupled-ic for standalone atm run

workflow/cases/coupled_free_forecast_wave_exp1.yaml
```diff
@@ -0,0 +1,60 @@
...
+    EDATE: 2013-04-01t00:00:00
+
+    cplwav: .true.
+    cplwav2atm: .true.
+    cplflx: .true.
+    print_esmf: .true.
+    nems_temp: 'med_atm_ocn_ice_wav'
+    nems_temp_cold: 'medcold_atm_ocn_ice_wav'
+    mom6ic_prepared: .false.
+    KEEPDATA: YES
```
JessicaMeixner-NOAA: Is this for the prototypes?   If so can we name this p5 instead of exp1? Also if this is for the prototypes KEEPDATA should be NO,

JessicaMeixner-NOAA: Should we also update the README as part of this PR for the additional case files to explain the additional ones?

jkuang-exploro: OK let me address it. I will name it p5

JessicaMeixner-NOAA (approved):

PR #143, opened 2020-10-02
optimize wave post and EE2

JessicaMeixner-NOAA:
* the wave point jobs are optimized to be parallelized over forecast hour instead of just the buoys.  This way fewer processors will be attempting to read the same file which caused lack of scalability and general slowness of the job before.  The resources for the two points jobs are increased back to 280, but the amount of time is going from 2 hours -> ~30 min and 3 hours -> 1 hour. 
* There are additional updates to address EE2 related issues in the wave init, prep, fcst, wave gridded and point output job & scripts. 

The wave init, prep and point jobs were all tested.


JessicaMeixner-NOAA:
Also for the boundary point wave job, it's now triggered by the f180 instead of f192, which further helped reduce the runtime of the boundary points job, in addition to no longer gzipping the boundary point tar file.  The downstream users of this file have been notified.

KateFriedman-NOAA:
Discussed incoming changes with Jessica and did a quick review. Submitting this.

PR #151, opened 2020-10-07
Update GEFS version of g-w for GFS v16 downstream

WalterKolczynski-NOAA:
Makes the necessary updates for GEFS downstream of GFS v16.

- Adds NetCDF version of recenter copied from GFS v16.
- GSI is updated to the GFS v16 version.
- UFS_UTILS is updated to a custom version with necessary updates to chgres_cube and other changes needed to handle GFS v16 while removing programs not needed by GEFS.
- The model version is updated to replace the bufr module version being retired by NCO (and also module updates to run on Hera).
- Other module loads of the retired bufr module are also updated.
- A lot of components not needed for GEFS are removed from the checkout/build process.


WalterKolczynski-NOAA:
Note: Until WGRIB2 2.0.8 is moved from nwpara to nwprod, before building you have to modify wgrib2_ROOT in the build.wcoss_dell_p3 modulefile of UFS_UTILS to use nwpara instead of nwprod (the line is already there, you just have to toggle which one is commented).

XianwuXue-NOAA (approved):

PR #152, opened 2020-10-07
for rocoto add a dependency to wavepostpnt job on wavepostbndpnt

JessicaMeixner-NOAA:
so that both jobs will not run at the same time which will
slow both jobs down. This is the reason for the dependency, otherwise
there is not a "true" dependency between the jobs

### DIFF
--- a/ush/rocoto/setup_workflow.py
+++ b/ush/rocoto/setup_workflow.py
@@ -591,7 +591,9 @@ def get_gdasgfs_tasks(dict_configs, cdump='gdas'):
         deps = []
         dep_dict = {'type':'task', 'name':'%sfcst' % cdump}
         deps.append(rocoto.add_dependency(dep_dict))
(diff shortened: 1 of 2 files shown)


KateFriedman-NOAA (approved):
Changes look good!

KateFriedman-NOAA:
@RussTreadon-NOAA FYI, please do the following to ingest this change:

1) git pull $HOMEgfs
2) rerun setup_workflow.py or manually add dependency to the gfswavepostpnt job:

```
<task name="gfswavepostpnt" cycledefs="gfs" maxtries="&MAXTRIES;">
...
        <dependency>
                <and>
                        <taskdep task="gfsfcst"/>
                        <taskdep task="gfswavepostbndpnt"/>
                </and>
        </dependency>
```

Thanks!

RussTreadon-NOAA:
Thanks, Kate.  Already done.

PR #153, opened 2020-10-08
update wave post pnt dependency 

JessicaMeixner-NOAA:
Updating the earlier commit that added a dependency for the wavepostpnt on the completion of the wave post bnd pnt job.  This job does not exist for gdas, so I made 2 parts, one for gdas w/no dependency on boundary points and one for gfs that has a dependency on the boundary points. 

This should be double checked before merged,  I made an xml file (/gpfs/dell2/emc/modeling/noscrub/Jessica.Meixner/EXPDIR/exp/exp.xml) that looks good to me, but there were lots of diffs with the parallel xml, so ?


KateFriedman-NOAA (approved):
Thanks Jessica!

PR #157, opened 2020-10-08
Adding wave grids to grib interpolation

JessicaMeixner-NOAA:
This reduces the output from WW3 
It interpolates the grib2 outputs to have the multi_1 masks which are necessary to keep AWIPS file sizes appropriate. 

This requires new fix files which @KateFriedman-NOAA  knows about and will need a subsequent update from @RobertoPadilla-NOAA


KateFriedman-NOAA:
Pulling right in, will test when Roberto's changes come in too. Thanks!

PR #158, opened 2020-10-08
Feature/gfsv16 wave prdgen

RobertoPadilla-NOAA:
Generation of awips files with wave fields that have masks


KateFriedman-NOAA (approved):
Thanks Roberto!

PR #173, opened 2020-11-04
Updates for RTOFS preprocessing for wave model 

JessicaMeixner-NOAA:
In NCO testing it was found that the current way of checking for if RTOFS files existed was insufficient.  
Instead we now look for the files from each day (depending on if it's gdas, the 06z RTOFS files or if gfs the last of each day for the 12z RTOFS files).  

In addition, it was discovered that no matter what the model cycle was, we were starting at the 00z for RTOFS.  The starting time for processing in wave prep of the RTOFS files is now based on the starting time of the forecast (including making sure to use the previous RTOFS forecast if the beginning time is not included) and ending at the end of the forecast time if it happens to be shorter than GFS. 

This means we can reduce the number of nodes gdaswaveprep needs to 1. The gfswaveprep job still needs 65 processors so 3 nodes.   However, the resources have not been changed.   

This uses the new RTOFS naming conventions.

### DIFF
--- a/jobs/JGLOBAL_WAVE_PREP
+++ b/jobs/JGLOBAL_WAVE_PREP
@@ -71,11 +71,7 @@ export COMOUT=${COMOUT:-${ROTDIR}/${CDUMP}.${PDY}/${cyc}/$COMPONENT}
 
 if [ $RUN_ENVIR = "nco" ]; then
   export COMIN_WAV_ICE=${COMIN_WAV_ICE:-$(compath.py gfs/prod)}/${CDUMP}.${PDY}/${cyc}/atmos
-  export COMIN_WAV_CUR=${COMIN_WAV_CUR:-$(compath.py ${WAVECUR_DID}/prod)}/${WAVECUR_DID}.${RPDY}
-  if [ ! -d $COMIN_WAV_CUR ]; then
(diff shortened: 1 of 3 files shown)


KateFriedman-NOAA (approved):
Changes look good. Please note in PR if/where they were tested. Thanks!

JessicaMeixner-NOAA:
These runs were tested in a single gfs forecast run (not complete, but enough to make sure currents were correctly coming in).   This will be tested by NCO in the upcoming 00z and 06z cycles.

JessicaMeixner-NOAA:
Test was on WCOSS-DELL.

KateFriedman-NOAA:
Thanks for adding the testing notes! 

Can we get confirmation when the edits work as anticipated in the NCO cycles? Then I'm happy to submit this. Thanks!

JessicaMeixner-NOAA:
Yes, I will update the PR tomorrow after the NCO tests.

JessicaMeixner-NOAA:
NCO's tests with these new changes also worked.

KateFriedman-NOAA:
> NCO's tests with these new changes also worked.

Excellent, thanks for confirming! Now submitting this PR.

PR #180, opened 2020-11-06
Feature/coupled-crow updates for wave post

JessicaMeixner-NOAA:
This PR updates the wave post to be more inline with GFSv16b.  It solves some, but not all of the current wave post issues.  There are two issues that will persist with this PR: 
* The grib2 files from WW3 are still incorrect/empty  (There is an issue open for this). 
* The resources/timings of the wave jobs will likely need to be updated 

The reason I am not waiting to fix the above is because this fixes some of the post issues, updates the wave model to use the new com directory structure and I think it will help @JianKuang-NOAA as he merges the develop branch back into feature/coupled-crow.  

This was tested on orion, see COM here: /work/noaa/marine/jmeixner/p6ww3post/test01/COMROOT/test01/


JessicaMeixner-NOAA:
The test on hera has also now finished.  You can find the COM for the hera test at: 
/scratch2/NCEPDEV/climate/Jessica.Meixner/P6updates/ww3post/COMROOT/wpost01

jkuang-exploro (approved):

JessicaMeixner-NOAA:
Thanks @JianKuang-NOAA for reviewing!

PR #184, opened 2020-11-12
Feature/ccpp

lgannoaa:
feature/ccpp merge back to development.
This package is tested with a C768 cycled (6 cycle) run on Dell system using FV3_GFS_v16beta (imp_physics=11).


parm/config/config.aeroic
```diff
@@ -0,0 +1,14 @@
...
+########## config.aeroic ##########
+# Append aerosol variables to FV3 initial conditions
+
+echo "BEGIN: config.aeroic"
+
+# Task and thread configuration
+export wtime_aeroic="00:30:00"
+export npe_aeroic=1
+export npe_node_aeroic=3
+export nth_aeroic=8
```
KateFriedman-NOAA: All of this needs to be added to a new "aeroic" block in config.resources. Change this config to match the others with regards to sourcing config.resources with "aeroic" as the argument after moving resources to config.resources. See this config as an example of what will remain in this config:

https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/config/config.analcalc

parm/config/config.base.emc.dyn
```diff
@@ -45,6 +45,10 @@ export NOSCRUB="@NOSCRUB@"
 # Base directories for various builds
 export BASE_GIT="@BASE_GIT@"
 
+#### CCPP Suite
+#### export CCPP_SUITE="FV3_GSD_noah"      # GSDsuite + NOAH LSM  
+export CCPP_SUITE="FV3_GFS_v16beta"        # EMC v16beta
```
KateFriedman-NOAA, line 50: Is this ignored if not running with CCPP?

parm/config/config.resources
```diff
@@ -297,7 +297,8 @@ elif [ $step = "eupd" ]; then
 
     export wtime_eupd="00:30:00"
     if [ $CASE = "C768" ]; then
-      export npe_eupd=150
+####      export npe_eupd=150
```
KateFriedman-NOAA: Remove this comment line.

sorc/build_fv3.sh
```diff
@@ -20,7 +20,11 @@ if [ $target = hera ]; then target=hera.intel ; fi
...
 cd tests/
-./compile.sh "$FV3" "$target" "WW3=Y 32BIT=Y" 1
-mv -f fv3_1.exe ../NEMS/exe/global_fv3gfs.x
+#### Modify to turn on CCPP with four suite 
+#### As of 11/12/2020 The FV3_GFS_v16beta has been tested with C768 cycled run on DELL system
+#./compile.sh "$target" "WW3=Y 32BIT=Y" 1
+# mv -f fv3_1.exe ../NEMS/exe/global_fv3gfs.x
+# mv -f fv3_2.exe ../NEMS/exe/global_fv3gfs_ccpp.x
+./compile.sh "$target" "CCPP=Y 32BIT=Y SUITES=FV3_GFS_v15,FV3_GSD_v0,FV3_GSD_noah,FV3_GFS_v16beta" 2 NO NO
+mv -f fv3_2.exe ../NEMS/exe/global_fv3gfs.x
```
KateFriedman-NOAA: Please clean this up, thanks!

Are we now forcing the CCPP build on users? Specifically line 29. @arunchawla-NOAA

sorc/checkout.sh
```diff
@@ -26,7 +26,8 @@ if [[ ! -d fv3gfs.fd ]] ; then
     rm -f ${topdir}/checkout-fv3gfs.log
     git clone https://github.com/ufs-community/ufs-weather-model fv3gfs.fd >> ${topdir}/checkout-fv3gfs.log 2>&1
     cd fv3gfs.fd
-    git checkout GFS.v16.0.14
+    git checkout d021e7b0395ccac2b7a30b414b58a8c924d2784f
+####    git checkout GFS.v16.0.14
```
KateFriedman-NOAA: Is there no tag for this? Can one be made at that hash perhaps?

Please remove the comment line on line 30.

sorc/link_fv3gfs.sh
```diff
@@ -205,6 +205,7 @@ if [ -d ../sorc/fv3gfs.fd/WW3/exec ]; then # Wave execs
   done
 fi
 
+
```
KateFriedman-NOAA: Please remove extra line.

jobs/rocoto/fcst.sh
```diff
@@ -2,7 +2,8 @@
 
 ###############################################################
 # Source FV3GFS workflow modules
-. $HOMEgfs/ush/load_fv3gfs_modules.sh
+#### . $HOMEgfs/ush/load_fv3gfs_modules.sh
+. $HOMEgfs/ush/load_fv3gfs_modules_fcst.sh
```
KateFriedman-NOAA: Why are we introducing a whole new load module script and module_base.wcoss_dell_p3_fcst? What about other machines?

KateFriedman-NOAA (changes requested):
Many questions:
1) Why is an additional ush/load_fv3gfs_modules_fcst.sh script being added? Is this out of a need of some kind?
2) What are additional module_base files being added? Add to that, why is there not a new one for the other platforms? Please explain reason for additional files (which we should be moving away from, especially when hpc-stack comes in).
3) Does this work on WCOSS, Hera, and Orion?
4) With these updates can users still run without CCPP and reproduce GFSv16 (current develop)?
5) Is there a switch to not run with CCPP and not force the CCPP in the build?
Thanks!

PR #192, opened 2020-11-18
wave parm update for realease/gfsv16

JessicaMeixner-NOAA:
update for the wave parm so that the wave model will look for the correct restart for when gfs is not run every cycle


RussTreadon-NOAA (approved):
The suggested change has been made in the EMC real-time v16 parallel running on Venus and has been found to work as designed.   WAVHCYC=6 for both gdas and gfs cycles.

yangfanglin:
Since Kate is on leave. I (Fanglin) will take the liberty to approve and merge this request.

PR #206, opened 2020-12-08
coupled: updates for merging checkout, build and link with develop branch

JessicaMeixner-NOAA:
This PR updates the checkout, build and link scripts to incorporate updates in the develop branch.  This adds "coupled" to the -c option (currently for CCPP) and enables the build_all.sh to have a coupled option and be used. 

when this commit goes in, please note that the checkout and build process has been changed and so follow the README for instructions.  

At this point the branch forecast job should run and the ocean post will still run.  The atm post still is not working yet.


jkuang-exploro:
Not-inline post processing doesn't work because JJOB and ex-level script names have changed. I will merge it in and solve the following issues in PR#194:
1, test if inline post works
2, re-connect the above scripts for regular UPP.

PR #218, opened 2020-12-15
feature/coupled-crow fractional grid 

JessicaMeixner-NOAA:
This PR mades using fractional grid the default option in: 
cases/coupled_free_forecast_wave.yaml 

and creates a new case file for the non-fractional grid here: 
workflow/cases/coupled_free_forecast_nofrac_wave.yaml 

The following settings change in the case file to turn on fraction grid: 
For turning on fractional grid we need to use different ICs: 
CPL_ATMIC: CFSRfracL64  --- this specifies that we are using the frac grid ICs 
cplmode: nems_frac --- this is for the CMEPS mediator which coupling mode 
FRAC_GRID: .true. --- this is for the input.nml and is used to grab the correct oro data 

Tests for both frac and non frac have been run on hera and orion.  
Orion 
non frac test: /work/noaa/marine/jmeixner/p6frac/nonfrac01
frac grid test: /work/noaa/marine/jmeixner/p6frac/frac01

Hera: 
non frac test: /scratch2/NCEPDEV/climate/Jessica.Meixner/p6workflow/nonfrac01
frac grid test: /scratch2/NCEPDEV/climate/Jessica.Meixner/p6workflow/frac01

This will close issue #171


JessicaMeixner-NOAA:
FYI @jiandewang @LydiaStefanova-NOAA @shansun6 @JianKuang-NOAA

WalterKolczynski-NOAA (approved):

jkuang-exploro:
Thanks for creating this PR. I will test it today.

ShanSunNOAA:
Hi Jessica,

So glad to see it. What is your default "suite"?

I see there is a suite called "suite_FV3_GFS_v16beta_coupled.xml". Is this
the one to test for P6 L127? I am asking because the default physics keeps
crashing with L127, while Fanglin can run L127 with V16. Thus I would like
to run with V16 first then L127 or both together. This is for the
regression test. What do you think?

Thanks,
Shan

JessicaMeixner-NOAA:
@shansun6 this is just to add the fractional grid option for the current CCPP suite which is "FV3_GFS_v15p2_coupled". 

Switching to use "FV3_GFS_v16beta_coupled" is doable but we should do that in a different branch/PR for testing first.

ShanSunNOAA:
Sounds good. Will do. Thanks,  -Shan

jiandewang:
@JessicaMeixner-NOAA from workflow point of view, I think the changes you made in script are correct. I further made a pair of 35day run. The results are on orion:
/work/noaa/stmp/jiwang/PR-FRAC/COMROOT/Frac_on
/work/noaa/stmp/jiwang/PR-FRAC/COMROOT/Frac_off

jkuang-exploro (approved):

jkuang-exploro:
@JessicaMeixner-NOAA There's a small issue in this PR: Additional parameters in the case file need to be added into /default. Not doing so won't affect the runs for this specific test case, but will potentially disable all other cases. I will add them into PR194 instead of revert this one.

JessicaMeixner-NOAA:
@JianKuang-NOAA where can we find instructions about all the proper places to add new parameters?  I'm pretty sure I just grep'd another variable and followed that pattern so there are likely other variables in this same position.  Thanks for adding the fix in your branch!

PR #228, opened 2020-12-23
Reducing bufr output sizes for Release/gfsv16.0.0

GuangPingLou-NOAA:
Per NCO request, the sizes of the BUFR sounding output need to be reduced. The bufr sounding time series output will have 64 levels instead of 127.


GuangPingLou-NOAA:
should be merged to release/gfsv16.0.0

PR #229, opened 2020-12-23
Release/gfsv16.0.0

GuangPingLou-NOAA:
Per NCO request, the sizes of the BUFR sounding output need to be reduced. The bufr sounding time series output will have 64 levels instead of 127.


yangfanglin (approved):
I assume there is no need to make any changes in computing resources. Approved

GuangPingLou-NOAA:
No, Fanglin. The resource requirements are actually reduced: Space reduced by about half, wall time reduced by about 2 minutes.

PR #242, opened 2021-01-19
Update README.md

jkuang-exploro:


JessicaMeixner-NOAA (changes requested):
@JianKuang-NOAA  thanks for adding this into the readme.  Based on this section alone, I'm still not sure how I would add a new variable.  Maybe it's just me @BrianCurtis-NOAA is this sufficient for you? I know you are also trying to add new variables and things right now. 

Also there are a few grammar issues that need to be corrected before merging.

jkuang-exploro:
@JessicaMeixner-NOAA Please ask questions or advices on this. I would like to make it both concise and helpful.

BrianCurtis-NOAA:
This helps. I would like to see something similar to:
The "default" directory yml files set the default variables for all runs
The run settings are set/configured in the "case" directory files
The setting defined in case files overwrite default values and are limited by the settings defined in the "schema" directory

And anything else that may help me understand how everything is set and/or started from beginning to end. I'm happy to look over anything to see if it makes sense to me as I am learning these same things myself.

JessicaMeixner-NOAA:
I think two things could possibly help: 

1. a description of how variables are set in a hierarchical sense: which overwrites what: cases, config, layout, defaults, shcema, platform, etc.    
2. From start to finish what are the steps to add a new crow configured variable.   -- This is still unclear to me.

jkuang-exploro:
@JessicaMeixner-NOAA I updated the last paragraph of the newly added "Managing Default / Adding new configuration variables" section. This includes some more instructions on how to add a new variable. Is this sounds reasonable to you?

And, "hierarchical sense of the variable" is simple, basically user+case file override default, then load into config or layout, as described in that paragraph. However, these instructions only works for variables that won't affect the entire workflow architecture, for example, making one of the ordinary namelist variable into configuration manger. Editing the entire architecture of the workflow, is totally another topic. (ie, add another on/off switch to control if a certain job is running or not)

jkuang-exploro:
My impression is that only basic instructions for configuration update should be included in this README. Since more advanced changes to the workflow often comes with unique demands and requires joint efforts, therefore harder to summarize into a step-by-step instruction.

https://noaa-emc.github.io/CROW/docs/html/pages.html

This is the page of a more in-depth instruction of CROW mechanism

README.md
```diff
@@ -205,4 +205,12 @@ Currently this is set up for benchmark runs which have ICs which are already gen
 
 Prototype 5 was run with global-workflow hash 005468b9299ea6fc9afdbeace33c336c6797833a
 
+## Managing Default / Adding new configuration variables
 
+workflow/default is the central place of managing default values of configuration variables. In principle, every single variable that could be configured by CROW need to have a default value included in workflow/default. A series of YAML files are used to storage these values. During the configuration step, CROW will load values from workflow/default first, and then customized by the overriding values in the YAML file selected by the user within workflow/cases/.
```
JessicaMeixner-NOAA: "CROW need" -> "CROW needs" 
"to storage" -> "to store" 
"then customized" -> "then will be customized"

JessicaMeixner-NOAA (approved):
I'd recommend fixing the minor typos but this is a helpful start to getting more information about adding variables and the configuration system - thanks @JianKuang-NOAA

README.md
jkuang-exploro: OK. I will fix these.

PR #247, opened 2021-01-25
Error checking update for waves

JessicaMeixner-NOAA:
Moving the error check definition variable from script file to jobs for wave prodgen and gempak. This will match with Jen's changes that NCO is using.


KateFriedman-NOAA (approved):
Moving errchk from script level to job level, looks good, approve!

PR #248, opened 2021-01-25
Feature/ww3 modules post

RobertoPadilla-NOAA:
WW3 was producing empty grib2 files.
To solve this issue:
On Orion, the hpc-stack is used but the version of jasper has to be changed from 2.0.15 to 1.900.0
On Hera,  the hpc-stack is not used and and all modules has to be loaded outside of hpc-stack.

Now  WW3 is producing valid grib2 files on Hera and Orion.


JessicaMeixner-NOAA (approved):
Thanks Roberto for sending the email with the output and getting these changes in a PR

PR #249, opened 2021-01-28
feature/coupled-crow: first update for v16 physics options

JessicaMeixner-NOAA:
First round of updates for using v16 physics options and the CCPP suite file "FV3_GFS_v16beta_coupled" 
it also updates the model to a recent version of ufs-weather-model which required updates to nems.configure inputs

The wave ICs are temporarily disabled until new ICs are created and ready for the wave model with the updated code. 

v16 options were updated in the defaults.  Some updates are in non-ideal ways such as the launch_level should be a calculation and the filed table was mostly hard-coded to be the right choice since which field table was determined before some variables that would change it were set.  These changes allow us to move forward though. 

Please note that the settings for p6 are not finalized and  for those who want to run p5 physics settings, the easiest/best thing to do will be to go back to using the p5 hash of the workflow & tag of ufs-model.


WalterKolczynski-NOAA (approved):
Recommend creating a tag for prototype 5 before merging.

JessicaMeixner-NOAA:
The hash that is used is mentioned in the readme for p5.  It's not the top of the feature/coupled-crow branch so merging this has no effect on the creation of a tag for p5 or not.

JessicaMeixner-NOAA:
@WalterKolczynski-NOAA is it okay if I merge this? Or are you going to?

PR #268, opened 2021-02-16
CMake for reg2grb2

BrianCurtis-NOAA:
Add a CMakeLists.txt to be able to create reg2grb2.x for coupled-crow after the merge of CMake work from https://github.com/NOAA-EMC/global-workflow/pull/264

Linked issue:
#263 

Waiting on:
Pull #264

Issue #164, opened 2020-10-21
Updating global workflow to use hpc-stack


arunchawla-NOAA:
The develop branch of ufs-weather-model now uses hpc-stack for all the libraries. 

The following changes need to be made to the global workflow
  1. The codes that are in the global workflow repository should now use hpc-stack
  2. The repositories that global workflow connects to also need to make updates to hpc-stack

We should track the PRs for all the component repositories through here. (This is for the top of develop)

Checklist for global-workflow external components:

- [x] Component uses stack: FV3/ufs-weather-model
- [x] Component uses stack: GSI (Hera only, waiting for other platform updates)
- [x] Component uses stack: GLDAS
- [x] Component uses stack: UPP
- [x] Component uses stack: UFS_UTILS
- [x] Component uses stack: EMC_gfs_wafs

Checklist for non-component/machine-specific stack updates (e.g. module_base, machine-setup.sh):
- [x] Hera-specific changes
- [x] Orion-specific changes
- [x] WCOSS-Dell-specific changes

JessicaMeixner-NOAA:
@arunchawla-NOAA for the wave_prep for GFSv16, we need CDO, do we request this to be added to hpc-stack?  For feature/coupled-crow, the ocean/ice post uses ncl.  Should this also be requested to be added to hpc-stack?

aerorahul:
@JessicaMeixner-NOAA cdo is a very useful tool and we should add it to hpc-stack.  It depends on how many bells and whistles we need to add or use.  For e.g. grib handling is done via grib_api from eccodes.

JessicaMeixner-NOAA:
Made an hpc-stack issue for CDO: https://github.com/NOAA-EMC/hpc-stack/issues/82 

But I guess the question remains, do all modules used in global-workflow need to be added to hpc-stack?

arunchawla-NOAA:
Yes, everything should use hpc-stack. We want this to be our way to make the environment equivalent across different platforms. CDO is available as third party libraries on WCOSS. We should add it to hpc-stack.

JessicaMeixner-NOAA:
@HuiyaChuang-NOAA @WenMeng-NOAA  is there a version of EMC_post that uses hpc-stack?

WenMeng-NOAA:
[...]

PR #273, opened 2021-02-22
Add WCOSS-Dell capability to coupled-crow and migrate to hpc-stack

### LINKED ISSUE #175: Port feature/coupled-crow to WCOSS-Dell
The workflow needs to be ported to all the WCOSS machines. Might just need new modulefiles and an updated linker.

WalterKolczynski-NOAA:
Restores the capability to run the coupled workflow on WCOSS-Dell. Also partially migrates the coupled workflow to hpc-stack.

The capability to run on WCOSS suffered code rot due to an inability of one of the primary maintainers to run there. That capability is now restored for WCOSS-Dell. It is only tested on phase 3; it would likely run on phase 3.5, but resource allocation would be suboptimal due to the larger number of cores. Additional configuration optimized for phase 3.5 can be added at a later time.

This merge also migrates library imports to hpc-stack for programs located within global-workflow (i.e. not the model or post) *and* are currently built in the coupled configuration. There are two that fall into both categories: reg2grb2 and waveprepost. Also included in this migration is the runtime modulefile (module_base). Modulefiles are migrated on WCOSS-Dell, Hera, and Orion.

New commit hashes will be needed to update the components. @KateFriedman-NOAA is working on getting these squared away for the develop branch for component versions using the release version of hpc-stack (#164) (some are currently using a test version). The coupled model will need to be similarly updated (it also currently uses a test version), along with any other coupled-only components.

The hpc-stack changes are also relevant to the cmake migration for reg2grb2 (#268).

Refs: #164, #268
Closes: #175


WalterKolczynski-NOAA:
This should be it, but I am going to reclone and start from scratch on each machine on Monday to confirm everything works. I am testing using the coupled_free_forecast_wave case, assuming that encompasses all programs for now.

JessicaMeixner-NOAA:
Thanks @WalterKolczynski-NOAA I will review this afternoon.  One quick comment thought:  This is just a new port not  "restoring" a capabaility as feature/coupled-crow has never been ported to WCOSS-DELL.

JessicaMeixner-NOAA:
@WalterKolczynski-NOAA one question: since everything is using hpc-stack and the grib issues with WW3 are resolved, the need for the extra WW3 build modules isn't really needed.  Do we want to remove those in this PR or in a subsequent PR?

JessicaMeixner-NOAA:
My test on hera seems to have worked as expected: /scratch2/NCEPDEV/climate/Jessica.Meixner/p6pr273/test01/COMROOT/test01/gfs.20130401

ush/wave_grib2_sbs.sh
```diff
@@ -35,6 +35,8 @@
   export LOUD=${LOUD:-YES}; [[ $LOUD = yes ]] && export LOUD=YES
   [[ "$LOUD" != YES ]] && set +x
 
+  WGRIB2=${WGRIB2:-${WGRIB2_ROOT:-${wgrib2_ROOT:?}}/bin/wgrib2}
```
JessicaMeixner-NOAA: I'm assuming this. was needed because hpc-stack no longer defines this? Should this be added in a job or higher level script instead?

WalterKolczynski-NOAA: Yes, although I should just add it to the modulefile instead (I did this for a different reason on WCOSS). I'll make an update.

WalterKolczynski-NOAA: FWIW, I think hpc-stack is supposed to be adding the WGRIB2 variable at some point.

WalterKolczynski-NOAA:
coupled_free_forecast_wave passed on all three machines (Hera, Orion, WCOSS-Dell). Did encounter some issues on Venus I didn't encounter on Mars (had to use mpiifort for reg2grb2, ocean post hitting wallclock limit). Will keep an eye on that and possible increase the runtime or otherwise optimize for WCOSS.

Also discovered that ocean post does not resume from where it left off; it redoes at least some of the processing from scratch. That may deserve its own issue, but it would be low-priority (mostly needed for operations) and might be mooted when we switch to the unified workflow anyway.

JessicaMeixner-NOAA (approved):

PR #278, opened 2021-02-25
Fixed reg2grb2 build on WCOSS

WalterKolczynski-NOAA:
The build of reg2grb2 was not linking the MPI libraries on WCOSS-Dell,
causing the build to fail (I don't know why this wasn't an issue before).
Modified the modulefile to now set the fortran compiler to the $MPI_FC
set by the impi module (which should be mpiifort) instead of ifort.

Refs: #273


WalterKolczynski-NOAA:
I missed pushing this last-minute commit for PR #273

JessicaMeixner-NOAA (approved):

Issue #283, opened 2021-03-08
coupled: setting the correct npe_node_fcst 


JessicaMeixner-NOAA:
In the feature/coupled-crow branch, the variable npe_node_fcst is set to the maximum pets per node, even when threads are requested.  This means that when the env/${MACHINE}.env file is sourced, that the requested threads gets set back to 1 because npe_node_max=npe_node_fcst

```
    nth_max=$(($npe_node_max / $npe_node_fcst))

    export NTHREADS_FV3=${nth_fv3:-$nth_max}
    [[ $NTHREADS_FV3 -gt $nth_max ]] && export NTHREADS_FV3=$nth_max
```

I have added a workaround in my branch for now, that in ush/forecast_predet.sh I have commented out: 
NTHREADS_FV3=${NTHREADS_FV3:-${NTHREADS_FCST:-${nth_fv3:-1}}}
and replaced it with: 
NTHREADS_FV3=${nth_fv3:-1}

But that is not the correct way to fix this.   I tried a couple of ways of changing the definition of npe_node_fcst but failed to get the correct output for threading.

JessicaMeixner-NOAA:
@WalterKolczynski-NOAA @KateFriedman-NOAA I think this issue can be closed, but wanted to confirm with you first.  This is specific to feature/coupled-crow, and since we are moving forward to using the develop branch it's no longer relevant. Although it might be a good to note somewhere if CROW is being considered for the future workflow.

KateFriedman-NOAA:
I'm good to close this if @WalterKolczynski-NOAA is. Thanks!

PR #284, opened 2021-03-08
Updates for prototype 6

JessicaMeixner-NOAA:
This PR turns wave initial conditions back on.  There were also several bug fixes: 
* for dddmp to get correctly to the input.nml 
* setting FISCL=9999, until bug fixes for using FISCL=0 are in the model code
* a workaround for getting threads set correctly (Issue #283)

### DIFF
--- a/workflow/config/fcst.yaml
+++ b/workflow/config/fcst.yaml
@@ -15,7 +15,6 @@ config_fcst:
         export hord_xx_nh_nonmono={doc.fv3_gfs_settings.phy_dependent_var.hord_xx_nh_nonmono}
         export vtdm4_nh_nonmono={doc.fv3_gfs_settings.phy_dependent_var.vtdm4_nh_nonmono}
         export nord={doc.fv3_gfs_settings.phy_dependent_var.nord}
-        export dddmp={doc.fv3_gfs_settings.phy_dependent_var.dddmp}
         export d4_bg={doc.fv3_gfs_settings.phy_dependent_var.d4_bg}
     - otherwise: ""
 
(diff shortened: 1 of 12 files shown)


workflow/defaults/fv3_gfs.yaml
```diff
@@ -116,7 +116,6 @@ fv3_gfs_defaults: &fv3_gfs_defaults
            hord_xx_nh_nonmono: 5
            vtdm4_nh_nonmono: 0.02
            nord: 2
-           dddmp: 0.1
```
WalterKolczynski-NOAA, line 119: Does it not work if this is set? My understanding is these values should be getting overridden by what is in the case file, and if not I should dig in and figure it out. If the defaults are removed, workflow might fail if dddmp is *not* set.

WalterKolczynski-NOAA (changes requested):
See question about dddmp defaults.

Also, in the future, please follow the [new guidelines](https://github.com/NOAA-EMC/global-workflow/wiki/Development-of-Global-Workflow#commit-message-standards) for all commit messages in branches that will be merged back to global-workflow.

workflow/config/fcst.yaml
```diff
@@ -119,6 +118,7 @@ config_fcst:
     export dnats={doc.fv3_gfs_settings.phy_dependent_var.dnats}
     export do_sat_adj="{tools.fort(doc.fv3_gfs_settings.phy_dependent_var.do_sat_adjust)}"
     export nwat={doc.fv3_gfs_settings.phy_dependent_var.nwat}
+    export dddmp={doc.fv3_gfs_settings.dddmp}
```
JessicaMeixner-NOAA, line 121: @WalterKolczynski-NOAA ... I don't know if it was adding this line or the removal of the other lines or both that are needed for the dddmp to get set to the value that was set in the case file.  But as it was before, even though the case file said dddmp=.2, it was being set as .1

workflow/defaults/fv3_gfs.yaml
JessicaMeixner-NOAA: Replied on another line in response to this.

WalterKolczynski-NOAA (approved):
Going to approve to get this in, but hopefully I'll remember to circle back to the defaults issue.

JessicaMeixner-NOAA:
Thanks @WalterKolczynski-NOAA

PR #307, opened 2021-03-25
Fix cycle date for wave bulletins

JessicaMeixner-NOAA:
This fixed the wave bulletins so that they will have the correct cycle date. 

This was pulled into operations by Jen and started running with on 20210324 for the 12z cycle.


KateFriedman-NOAA (approved):
Change is what was discussed via email with SPA Jen. Looks good. Thanks!

KateFriedman-NOAA:
Reference issue #1.

PR #318, opened 2021-04-30
bug fix for wave post point 

JessicaMeixner-NOAA:
This bug was found while testing the RW-points for NHC issue.  The commented out line of code removes the last buoy from the buoy list and it should be included. 

I have run a test of the point post jobs using the para run data and the only effect is that the last point is actually included as it should be. 

This should be included with the 16.1 (assuming waves continues as part of this). 

FYI @CatherineThomas-NOAA @RussTreadon-NOAA @KateFriedman-NOAA   I have given this to Jen, this was the bug I mentioned that I found in the email thread yesterday.

### DIFF
--- a/scripts/exgfs_wave_post_pnt.sh
+++ b/scripts/exgfs_wave_post_pnt.sh
@@ -306,10 +306,10 @@
     mv buoy_log.tmp buoy_log.dat
 
     grep -F -f ibp_tags buoy_lst.loc >  buoy_tmp1.loc
-    sed    '$d' buoy_tmp1.loc > buoy_tmp2.loc
-    buoys=`awk '{ print $1 }' buoy_tmp2.loc`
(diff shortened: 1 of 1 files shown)


RussTreadon-NOAA:
According to Steven, Simon will be the SPA handling GFS v16.1.  The GFS v16.1 workflow release notes direct the SPA to check out workflow tag EMC-v16.1.0.   After this PR is completed, tag EMC-v16.1.0 will need to be recreated.   Otherwise, Simon may miss this change in the operational snapshot of GFS v16.1.

KateFriedman-NOAA:
@CatherineThomas-NOAA Please sync merge the release/gfsv16.1 branch with the release/gfsv16.0.0_to_ops branch. Let me know when done and I will remake the EMC-v16.1.0 tag off of release/gfsv16.1. Thanks!

PR #319, opened 2021-05-04
Fix indentation in forecast scripts

WalterKolczynski-NOAA:
The forecast scripts were not indented properly, hindering readability.
All program branches are now properly indented.

Other scripts need the same treatment, but this portion is done now to
facilitate coming additions.


JessicaMeixner-NOAA (approved):
All changes are white space, so there should be no impact on the results.

PR #322, opened 2021-05-10
update for ww3 pre and post exec builds for hpc-stack branch

JessicaMeixner-NOAA:
* remove ww3 specific module files in favor of using the base set 
* updated module base file on hera to be in sync with ufs-weather-model (other machines need this same update)
** The only diff is that Jasper is a more recent version which is needed for ww3_grib compilation**
* added extra ww3 execs needed for this branch
* added ww3prepost build to build_all and partial build

The build of ww3 pre and post scripts was tested on hera, further testing is likely needed.


KateFriedman-NOAA (approved):
Updates look good. Will merge, test on Hera, and then apply similar updates for other platforms.

Issue #328, opened 2021-05-25
load module nco too early


jiandewang:
in coupled-crow branch, nco module needs to be loaded after module load hpc-intel, otherwise nco module will not be loaded successfully.
https://github.com/NOAA-EMC/global-workflow/blob/feature/coupled-crow/modulefiles/module_base.hera#L8-L18

For a quick offline test, you can do
module purge
module spider nco/4.9.1

you will get the following:
You will need to load all module(s) on any one of the lines below before the "nco/4.9.1" module is available to load.

      intel/17.0.5.239
      intel/18.0.5.274
      intel/19.0.4.243
      intel/19.0.5.281
      intel/2020
      intel/2020.2

Note this problem only exists on HERA. Module nco can be loaded before module intel on orion.

This issue cause the variable renaming unsuccessful at
https://github.com/NOAA-EMC/global-workflow/blob/feature/coupled-crow/ush/ocnpost.ncl#L142

PR #329, opened 2021-05-25
Fix module load order on Hera

WalterKolczynski-NOAA:
The module load order on Hera was incorrect and tried to load some modules before
the compiler module is loaded.

Closes: #328


jiandewang (approved):
look good

JessicaMeixner-NOAA (approved):

Issue #314, opened 2021-04-17
Enable aerosol cycling


WalterKolczynski-NOAA:
Restart capability is urgently needed for aerosols in order to properly run and evaluate the new system.

Issue #260, opened 2021-02-09
Supporting GFS interval greater than 24 hours


malloryprow:
Issue came up when a user manually change the GFS interval in their XML from to _<!ENTITY INTERVAL "24:00:00">_ to _<!ENTITY INTERVAL "120:00:00">_. However the user left their config.metp unchanged and left VRFYBACK_HRS to its default of 24 hours, meaning the verification would run for the cycle date - 24 hours. This caused problems in the gfsmetp tasks as the GFS cycles would not have been ran for these dates. When the user update VRFYBACK_HRS to 120 the verification ran as expected.

The goal with this issue would be to rework how gfs_cyc gets set and used, and creating a variable that script that connect EMC_verif-global and the global workflow can key off of to get the right verification dates so users don't have to make the changes in config.metp. 

From Kate: "It could become an hour value rather than a toggle value. It would default to "24" (similar to the current gfs_cyc=1). Then users could provide a different value at setup time (or between setup steps by changing config.base) and could set it to a value divided by the assimilation frequency (6hrs)."

We will need to consider how this could impact and affect other places where gfs_cyc is used.

malloryprow:
@KateFriedman-NOAA I just had an idea for this that would potentially fix things for the metp tasks with leaving gfs_cyc alone. If we set up the setup_workflow.py and setup_workflow_fcstonly.py to add INTERVAL_GFS to the environment variables for the "gfsmetp*" tasks, I can use that to adjust the verification date. I understand though wanting to address the bigger picture with gfs_cyc and user manually adjusting the XML to set INTERVAL_GFS. I know you said you plate is pretty full for the rest of the month, but just wanted to get this idea written down.

jkuang-exploro:
@malloryprow Has this issue been addressed?

malloryprow:
@JianKuang-UMD It has not

WalterKolczynski-NOAA:
I had a solution in the coupled-crow branch, but now we've moved away from that.

aerorahul:
Out of scope for current development and capabilities.

WalterKolczynski-NOAA:
This has become relevant again with GEFS reforecast

PR #334, opened 2021-06-07
Add ability to cycle GOCART aerosols

WalterKolczynski-NOAA:
The ability to cycle GOCART aerosols for otherwise cold-start initial conditions is added to the CROW-based coupled workflow.

The following now happen automatically when GOCART coupling is on
- The restart interval is modified to ensure one is created at the forecast cadence frequency in addition to any other times that may be specified.
- A new job, init_chem, is run on cycles after the first that copies the aerosol concentration from the previous cycle's restart files and  adds it to the cold-start initial conditions.

Importantly, the aerosol fields are merely copied to the initial condition files. **They are not modified for mass conservation or any other properties.**

As part of this modification, the method for specifying the forecast cadence was overhauled. gfs_cyc has been eliminated in favor of an explicit specification of the step size for both the GFS and DA system via STEP_GFS and STEP_DA, respectively. This allows for steps that are longer than 24 hours (See #260). Technically, the step sizes can be any arbitrary length, but other downstream settings are only listed in hours, so fractional hours may not work properly.

Setting STEP_GFS to !timedelta "00:00:00" will turn off GFS forecasts (DA cycling-only)
Setting STEP_DA to !timedelta "00:00:00" will turn off DA (free-forecast mode)

Note that setting STEP_DA to zero is now required for free forecasts in addition to using the free forecast CROW layout.

Examples are available in the commit message for c196176.

Refs: #260 
Closes: #314


WalterKolczynski-NOAA:
A couple things that I hope can be updated before merging:
- Currently using a custom installation of hpc-stack to use the NASA version of a module. Hope to move to an official installation.
- Want to eliminate the separate checkout version and repo for the GOCART-enabled UFS.

WalterKolczynski-NOAA:
@rmontuoro in particular needs to confirm these changes are doing what he expected.

workflow/cases/dev/fv3q2fy19retro5-GFS@C768_ENKF@384+80MEM_whatever.yaml
```diff
@@ -186,5 +186,5 @@ case:
     EDATE: 2016-02-14t12:00:00
     chgres_and_convert_ics: No
     realtime: false
-    gfs_cyc: 4 # run gfs at 0z, 12z
+    STEP_GFS: !timedelta "06:00:00"
```
KateFriedman-NOAA, line 189: What does STEP_DA default to if it's not set in the case file?

workflow/config/fcst.yaml
```diff
@@ -231,6 +251,7 @@ config_fcst:
         export CHM_CFGDIR="{doc.chem_settings.CFGDIR}"
         export CHM_INPDIR="{doc.chem_settings.INPDIR}"
         export dnats="$(( $dnats + {doc.chem_settings.ntdiag} ))"
+        export KMP_INIT_AT_FORK=FALSE # Temporary setting
```
KateFriedman-NOAA, line 254: Curious about this variable and how it helps...this looks like something I considered for Comet.

workflow/config/wave.yaml
```diff
@@ -78,12 +78,8 @@ config_wave:
         WAVNCYC=4
         WAVHCYC=6
         FHMAX_WAV_CUR=48 # RTOFS forecasts only out to 8 days
-      elif [ $gfs_cyc -ne 0 ]
-      then
-        FHMAX_WAV_CUR=192 # RTOFS forecasts only out to 8 days
-        WAVHCYC=`expr 24 / $gfs_cyc`
       else
-        WAVHCYC=0
+        WAVHCYC={tools.dt_to_HMS(doc.settings.STEP_GFS).split(':')[0]}
```
KateFriedman-NOAA, line 82: Nice!

workflow/layout/cycled_gfs.yaml
```diff
@@ -731,7 +731,7 @@ suite: !Cycle
 
     jgfs_archive: !Task
       <<: *service_task_template
-      Disable: !calc doc.settings.gfs_cyc == 0
+      Disable: !calc doc.settings.STEP_GFS == doc.zero_dt
```
KateFriedman-NOAA, line 734: Curious what zero_dt is here.

workflow/runtime/suite.yaml
```diff
@@ -2,6 +2,9 @@
 # suites in the workflow/ directory.  Most of these variables are
 # specific to the cycled workflow which is not in this release.
 
+day_dt: !timedelta "24:00:00"
+zero_dt: !timedelta "00:00:00"
```
KateFriedman-NOAA, line 6: I think this answers my question earlier...these are constants (0 and 24) that you use elsewhere?

workflow/runtime/suite.yaml
```diff
@@ -39,31 +42,41 @@ cycled_suite_alarms: &cycled_suite_alarms
...
 # It sets defaults for various variables that are generally not overridden
 suite_defaults: &suite_defaults
   # Clock - the list of cycles to run
   Clock: !Clock
-    # NEED TO FIX FOR LESS FREQUENT GFS
     start: !calc doc.settings.SDATE
     end: !calc doc.settings.EDATE
-    step: !timedelta "6:00:00"
+    step: !FirstTrue
+      - when: !calc doc.settings.STEP_DA == doc.zero_dt
```
KateFriedman-NOAA, line 75: Is this where the default for STEP_DA is set?

KateFriedman-NOAA (approved):
I'm here just to see the revamp of gfs_cyc and related workflow updates. I like!

workflow/runtime/suite.yaml
WalterKolczynski-NOAA: Default is !timedelta 06:00:00

workflow/cases/dev/fv3q2fy19retro5-GFS@C768_ENKF@384+80MEM_whatever.yaml
WalterKolczynski-NOAA: STEP_DA defaults to !timedelta "06:00:00"

workflow/runtime/suite.yaml
WalterKolczynski-NOAA: No, that is set in defaults/settings.yaml. This is determining the master clock for the experiment. Since DA can happen more often than the forecast, STEP_DA is used unless it is off (zero).

workflow/runtime/suite.yaml
WalterKolczynski-NOAA: Yes

workflow/config/fcst.yaml
WalterKolczynski-NOAA: This was suggested by Raffaele for multithread problems in the aerosol model, though I don't know that it helped.

workflow/cases/dev/fv3q2fy19retro5-GFS@C768_ENKF@384+80MEM_whatever.yaml
KateFriedman-NOAA: Gotcha, thanks!

jobs/rocoto/init_chem.sh
```diff
@@ -0,0 +1,56 @@
...
+    [[ $status -ne 0 ]] && exit $status
+done
+
+###############################################################
+# Source machine runtime environment
+. $BASE_ENV/${machine}.env init_aerosol
+status=$?
+[[ $status -ne 0 ]] && exit $status
+
+# export DATAROOT="$RUNDIR/$CDATE/$CDUMP"
```
JessicaMeixner-NOAA, line 38: delete commented out lines?

modulefiles/module_base.hera
```diff
@@ -4,7 +4,8 @@
 ##
 
 # Modules availble through hpc-stack
-module use /scratch2/NCEPDEV/nwprod/hpc-stack/libs/hpc-stack/modulefiles/stack
+# module use /scratch2/NCEPDEV/nwprod/hpc-stack/libs/hpc-stack/modulefiles/stack
+module use /scratch1/NCEPDEV/nems/Raffaele.Montuoro/dev/nasa/libs/hpc-stack/1.1.0-nasa/modulefiles/stack
```
JessicaMeixner-NOAA, line 8: What's the path forward with not relying on Raffaele's stack? What's the timeline for getting what's needed in the emc hpc stack?

modulefiles/module_base.hera
```diff
@@ -32,11 +34,14 @@ module load hdf5/1.10.6
...
+setenv "WGRIB2" "wgrib2"
 
 # Modules not under hpc-stack
 module load hpss/hpss
 module load nco/4.9.1
 module load gempak/7.4.2
 module load ncl/6.5.0
 module load cdo/1.9.5
+module load intelpython/3.6.8
+# netCDF4 is not in the python installation on Hera
```
JessicaMeixner-NOAA, line 46: So same thing here, we're moving away from hpc-stack or official modules.  Is this what we want to do?

modulefiles/module_base.hera
JessicaMeixner-NOAA: These are also only for hera, so if we are doing this, shouldn't we be consistent throughout the platforms.

JessicaMeixner-NOAA: Also in the strategy used here: https://github.com/ufs-community/ufs-weather-model/pull/591 only the modules that are different use the personal stack, perhaps that's a better way forward?  So only: 
module use -a /scratch1/NCEPDEV/nems/Raffaele.Montuoro/dev/nasa/ufs/lib/modulefiles

 module load gftl-shared/v1.3.0
 module load yafyaml/v0.5.1
 module load mapl/v2.7.1-beta

 module load intelpython/3.6.8

While all others use the hpc-stack from nwprod?

JessicaMeixner-NOAA (commented):
I have a couple of comments on the module files as maybe a different path forward.  Otherwise, as long as the ATM-OCN-ICE-WAV test passes, I don't see anything that would conflict with that application so should be okay.

modulefiles/module_base.hera
WalterKolczynski-NOAA: As noted in the commit message, this is definitely a work-around because NetCDF4 is not in the default python build on Hera (it is on WCOSS). We can talk to RDHPCS about having it added, though often their solution is "make your own conda environment", which is pretty much the same problem.

Personally I'd like to have a full python installation/environment in hpc-stack so it is standardized on all machines, but that path is something for @arunchawla-NOAA to determine. Alternatively, just the pyNetCDF4 module could be added there instead of my personal directory.

rmontuoro (approved):
I've confirmed that these changes allow to successfully perform cycled runs with aerosols.

Issue #324, opened 2021-05-14
update feature/coupled-crow to use latest ufs-weather-model 


JessicaMeixner-NOAA:
To update feature/coupled-crow to use the latest develop branch of ufs-weather-model, we need to track the changes in the tests/parm folder in the diffs here: 
https://github.com/ufs-community/ufs-weather-model/compare/Prototype-6.0beta...develop

Updates are needed to: 
* change name of mediator in nems.configure 
* MOM_input has new option ALLOW_LANDMASK_CHANGES (we use the default value which is false, but still should keep in sync)
* ice_in has changes
* the build needs to be updated as well (see https://github.com/JessicaMeixner-NOAA/global-workflow/commit/cdc04db78d946d904f4e1235956e48864c9a3ba3 for needed change)
* use ocean mesh file (a new ocean mesh fix file is not needed as its the same as the ice mesh file)

JessicaMeixner-NOAA:
Another item: 
* After PR https://github.com/ufs-community/ufs-weather-model/pull/586  this line https://github.com/NOAA-EMC/global-workflow/blob/feature/coupled-crow/ush/nems_configure.sh#L158 needs to be changed from: 
`cp $HOMEgfs/sorc/ufs_coupled.fd/CMEPS-interface/CMEPS/mediator/fd_nems.yaml fd_nems.yaml
`
to 
`cp $HOMEgfs/sorc/ufs_coupled.fd/tests/parm/fd_nems.yaml fd_nems.yaml`

JessicaMeixner-NOAA:
The mesh ocn/ice are the same: 
  mesh_ocn = mesh.mx100.nc
same as   mesh_ice = mesh.mx100.nc

WalterKolczynski-NOAA:
The ufs module naming convention has been changed as well.

WalterKolczynski-NOAA:
I'll take care of this after we clear out some of the existing PRs.

JessicaMeixner-NOAA:
With the latest merge into ufs-weather-model today (6/11), https://github.com/NOAA-EMC/global-workflow/issues/325 is also needed when updating to the latest.

WalterKolczynski-NOAA:
I think I've got all of these except the first thing. I don't see any mediator name changes in the current UFS develop nems.configure templates.

WalterKolczynski-NOAA:
So I think I am just waiting on #325 now so I can test?

JessicaMeixner-NOAA:
[...]

PR #338, opened 2021-06-09
Update feature/coupled-crow to run the latest ufs-weather-model with prognostic aerosols

### LINKED ISSUE #336: Update feature/coupled-crow to use the latest coupled aerosols app
Branch `feature/coupled-crow` needs to be updated to run the latest ufs-weather-model app including prognostic aerosols.

### LINKED ISSUE #337: Improperly set NST_SPINUP in FV3 namelist (coupled-crow)
In branch `feature/coupled-crow`, the `NST_SPINUP` flag in the parsed FV3 namelist  is always `0` regardless of the value set in the case YAML file.

rmontuoro:
This PR replaces the original coupled atmosphere-aerosols system (FV3-GOCART) with the latest revision of the ufs-weather-model including prognostic aerosols. The model's code is temporarily retrieved from a forked repository as the corresponding updates are being integrated in the authoritative ufs-weather-model repository. Aerosol test cases are also updated, and the original `cplchem` flag is reused to enable coupling with the aerosol component, removing the additional `cplgocart` flag.

The included changes partially address #324, updating FV3 namelist settings to run the latest FV3 revision with the `FV3_GFS_v16` CCPP suite used with the coupled UFS Aerosols component.
Commit d16630f3b34671e2350e6a88b9df749e4fd58ece also fixes a bug reported in #337.

Closes: #336, #337
Refs: #324


WalterKolczynski-NOAA:
While I'm not necessarily against the changes, there are important changes here (like changing the modulefile used by global-workflow at runtime) that really should've been discussed with the global-workflow team before being implemented. Additionally, the module name change impacts the non-aerosol version, which is still using the old UFS naming convention. Finally, with regard to the modulefile, I don't know that UFS is going to want to keep a separate modulefile for aerosols, given that no other component has its own modulefile in the root modulefile directory.

The change from cplgocart to cplchem in CROW is also going to conflict with PR #334, which makes changes keyed off of cplgocart that will need to be updated.

On the whole, I'd prefer to sit on this one if possible at least until the model changes are merged into the authoritative ufs-weather-model repo and we can coordinate with @JessicaMeixner-NOAA to update the entire coupled workflow to the latest UFS commit at the same time. Ideally merging to a single build command with all the components compiled as well, instead of different build settings.

JessicaMeixner-NOAA:
Since this only partially addresses #324 I'm afraid if we merge this by itself we will break the ATM-OCN-ICE-WAV model workflow.  However, if we only address #324 without this, we'd likely break the aerosols app as well.  So perhaps these will have to be combined into an effort.  I'm still not sure who is going to address #324 in full but hopefully we can discuss that at the tag-up here shortly.  

@WalterKolczynski-NOAA Depending on how long it will take to fully reintegrate GOCART into ufs-weather-model, I don't think waiting until it's one build system should stop us.  However, if we're changing modules for the ATM-OCN-ICE-WAV as well, we really need to consider how that impacts. 

@rmontuoro did you run the coupled_free_forecast_wave.yaml case? Perhaps I'm thinking it conflicts with #324 when it actually doesn't?

sorc/build_ufs_coupled.sh
```diff
@@ -13,15 +13,16 @@ if [ $target = hera ]; then target=hera.intel ; fi
...
-MOD_PATH=$cwd/ufs_coupled.fd/modulefiles/$target
+MOD_PATH=$cwd/ufs_coupled.fd/modulefiles
 
 module purge 
 module use $MOD_PATH 
-module load fv3 
+module load ufs_${target}
 cd ufs_coupled.fd/
 if [[ -d build ]]; then rm -Rf build; fi
 if [[ -d GOCART ]]; then
```
JessicaMeixner-NOAA: So you have switched to using your fork, but as soon as this PR goes into ufs-weather-model this build check will not work anymore.  Do we address that now or later?

ush/forecast_postdet.sh
```diff
@@ -868,7 +864,7 @@ GSD_in() {
 
 GSD_nml() {
 	echo "SUB ${FUNCNAME[0]}: Creating name list for GSD"
-	sh parsing_namelists_GSD.sh
+	# generate GSD-specific input
```
JessicaMeixner-NOAA: should this whole section just be deleted now?

ush/load_fv3gfs_modules.sh
```diff
@@ -24,10 +24,10 @@ elif [[ -d /lfs3 ]] ; then
 	module load module_base.jet 
 elif [[ -d /scratch1 ]] ; then
 	# We are on NOAA Hera
-	module load module_base.hera
+	module use "$HOMEgfs/sorc/ufs_coupled.fd/modulefiles"
```
JessicaMeixner-NOAA: I don't necessarily disagree with this change, but @WalterKolczynski-NOAA needs to be okay with it.  I don't think there are other modules that are needed when running the forecast but I could be wrong.  This is just for the forecast job right? If it's more than just the forecast I don't think this will work.

ush/load_fv3gfs_modules.sh
```diff
@@ -24,10 +24,10 @@ elif [[ -d /lfs3 ]] ; then
...
+	module load ufs_hera.intel
 	if [[ -d $HOMEgfs/sorc/ufs_coupled.fd/GOCART ]] ; then
-		module use "$HOMEgfs/sorc/ufs_coupled.fd/modulefiles/hera.intel"
-		module load fv3
+		module load ufs_aerosols_hera.intel
```
JessicaMeixner-NOAA: Like above, this works for now, but this will not be sufficient after the PR goes in.

ush/nems_configure.sh
```diff
@@ -155,10 +155,10 @@ EOF
 
 echo "$(cat med_modelio.nml)"
 
-cp $HOMEgfs/sorc/ufs_coupled.fd/CMEPS-interface/CMEPS/mediator/fd_nems.yaml fd_nems.yaml
-
 fi 
 
+cp $HOMEgfs/sorc/ufs_coupled.fd/tests/parm/fd_nems.yaml fd_nems.yaml
```
JessicaMeixner-NOAA, line 165: This is rightfully not just under cplflx, but should it be behind a cpl flag? I guess it doesn't hurt standalone atm, its just not really needed?

workflow/cases/aerosol_firex_forecast.yaml
```diff
@@ -5,8 +5,8 @@ case:
   places:
     workflow_file: layout/free_forecast_gfs.yaml
     BASE_CPLIC: /scratch1/NCEPDEV/nems/Raffaele.Montuoro/data
-    OROFIX:     /scratch1/NCEPDEV/nems/emc.nemspara/RT/NEMSfv3gfs/develop-20200914/INTEL/FV3_input_data_c384/INPUT
-    GRDFIX:     /scratch1/NCEPDEV/nems/emc.nemspara/RT/NEMSfv3gfs/develop-20200914/INTEL/FV3_input_data_c384/INPUT
+    OROFIX:     /scratch1/NCEPDEV/nems/emc.nemspara/RT/NEMSfv3gfs/input-data-20210528/FV3_input_data384/INPUT
```
JessicaMeixner-NOAA: Should these be put with the global-workflow fix files instead? @KateFriedman-NOAA would be the person to coordinate that update with.

JessicaMeixner-NOAA (commented):
Made a few comments.  Before this is merged, we need to confirm that the atm-ocn-ice-wav case still runs as is.

sorc/build_ufs_coupled.sh
rmontuoro: In the most recent versions of `ufs-weather-model`all modulefiles have been renamed as: `ufs_<machine>.<compiler_settings>`. This change introduces the update and should work from now on.

ush/forecast_postdet.sh
rmontuoro: Agreed. I was not sure if we wanted to leave it as a template for future development.

ush/load_fv3gfs_modules.sh
rmontuoro: I fully defer to your choice here. My assumption was that platform-specific modulefiles provided with the source code would include all necessary modules to run different coupled configuration but this may not be the case.

While this is mostly a convenience-driven change, I'm concerned about duplicating model settings in the workflow as it will make it harder to maintain.

ush/load_fv3gfs_modules.sh
rmontuoro: This modulefile should go away soon.

workflow/cases/aerosol_firex_forecast.yaml
rmontuoro: I'll follow your guidance on this, too. Note the updated baseline linked to the updated ufs-weather-model.

sorc/build_ufs_coupled.sh
JessicaMeixner-NOAA: But you didn't update the hash of what is checked out for the ufs-weather-model for the coupled case.  Therefore this will break the coupled case as is,  unless I missed that.

workflow/cases/aerosol_firex_forecast.yaml
JessicaMeixner-NOAA: The workflow does not use the rt.sh inputs and instead use the global-workflow set. This is fine for now, but we need to get whatever is needed here in the global-workflow fix files that @KateFriedman-NOAA is maintaining.

sorc/build_ufs_coupled.sh
WalterKolczynski-NOAA: This was my issue as well. The currently non-aerosol coupled-crow ufs model version is still using the old naming convention. So this PR would break that until #324 is also merged.

ush/nems_configure.sh
rmontuoro: This is an interesting question.

Yes, the NUOPC field dictionary is not required by standalone atm. Furthermore, recent FV3 versions no longer rely on the `cpl` flag, formerly in the `model_configure` file, to advertise/realize coupling fields in the NUOPC cap. We could wrap this file copy in a conditional block to check whether any of the physics coupling flags (`cplflx`, `cplwav`, etc.) are true, similarly to what implemented in `cplvalidate.sh`:
https://github.com/NOAA-EMC/global-workflow/blob/3428097dd39934ea648d86e5cdee8d8c2bdb3b14/ush/cplvalidate.sh#L32
However, there is an ongoing effort to refactor the FV3 cap, with the possible (but remote) outcome that other coupling flags may no longer be used, too.

sorc/build_ufs_coupled.sh
WalterKolczynski-NOAA: (And vice versa)

ush/load_fv3gfs_modules.sh
WalterKolczynski-NOAA: The global-workflow team has already discussed moving to this paradigm (using the same script for build and runtime), but haven't gotten to it yet. Here I want to be sure that replacing base with the UFS module here is only impacting the forecast job and not any other jobs.

ush/nems_configure.sh
JessicaMeixner-NOAA: This isn't just used by the atm model in coupling. It's used by any model.  Did the "cpl" flag get removed completely? I believe it's still a variable in the workflow if not in input.nml/model_configure 

At any rate,  it doesn't adversely affect fv3 standalone so it can just be left as is for simplicity.

ush/load_fv3gfs_modules.sh
JessicaMeixner-NOAA: The issue is the if the GOCART directory exists is not going to work once the ufs-weather-model PR is committed and all scenarios have the folder GOCART.  So this will have to not adversely affect the other scenarios.

workflow/cases/aerosol_firex_forecast.yaml
WalterKolczynski-NOAA: Yes, all fix files should be in the fix directory. Nothing should look into the UFS regression test directory as we clean things up. I'm okay addressing this later though.

sorc/build_ufs_coupled.sh
JessicaMeixner-NOAA: Well if the changes were only within the if/else statements it could be fine, but as it is this will break atm-ocn-atm-wav

ush/nems_configure.sh
rmontuoro: Yes, the `cpl` flag was removed and it is no longer needed in `model_configure`.

ush/load_fv3gfs_modules.sh
rmontuoro: Agreed. We should limit use of the additional aerosol modulefile only to coupled configurations building GOCART.

sorc/build_ufs_coupled.sh
rmontuoro: You are right. My assumption was that we were going to update to the latest ufs-weather-model for the coupled case, too, before or at the same time as this PR. I will address this issue by adding a conditional block.

JessicaMeixner-NOAA: I hope we will update ufs-weather-model for the other case very soon, but as of now #324 is remains an unassigned task.

junwang-noaa:
@*rmontuoro *All the UFS applications require this yaml file including the
standalone FV3 since May 21 So it's good to keep the yaml file without any
constraints.

sorc/build_ufs_coupled.sh
rmontuoro: Commit f81741bdb3a8cc221f6e0846c706cfe3cc9c4010 addresses this issue. Updated modulefiles are only loaded to build the ufs-weather-model with prognostic aerosols. This model is identified though a hidden file, created when using option `-a` during checkout.

ush/forecast_postdet.sh
rmontuoro: Commit 6b19a39b6f086ff7eb077c60a3acbc40d6dae5a1 removes such GSD-specific shell functions.

ush/load_fv3gfs_modules.sh
rmontuoro: Commit 0307896af1d309718d6b08d89631f8c080c575e7 refactors this script to:
- use a new shell function (`get_platform`) introduced to return the string identifier for the auto-detected local platform to improve modularity
- load updated modulefiles only for coupled aerosols on Hera/Intel

ush/load_fv3gfs_modules.sh
rmontuoro: See comment above.

sorc/build_ufs_coupled.sh
WalterKolczynski-NOAA: Looks like I'm too late and the complications have already been added, but I'm going to try and tackle #324 after we clear a couple of the other PRs out. So let's cool it with adding more if-block exceptions for aerosols when we want to be removing them.

ush/load_fv3gfs_modules.sh
rmontuoro: Commit a821a1aef761b386beace4efab476495e209adbb fixes additional issues with the script.

rmontuoro:
@JessicaMeixner-NOAA - the `coupled_free_forecast_wave.yaml` case runs successfully after applying the changes included in this PR.

sorc/build_ufs_coupled.sh
WalterKolczynski-NOAA: At least until we can discuss it in the afternoon.

rmontuoro:
This branch now includes the latest updates from PR #334

workflow/cases/aerosol_firex_forecast.yaml
rmontuoro: References to aerosol-specific fix files have been removed.

ush/load_fv3gfs_modules.sh
rmontuoro: Work is ongoing to include all GOCART dependencies into hpc-stack.

WalterKolczynski-NOAA:
This PR is being rolled into PR #359 after all, as I needed the changes to the checkout/build for the updated UFS.

PR #340, opened 2021-06-09
Fix ww3 pre-post modules

### LINKED ISSUE #339: Fix WW3 pre-post build
The build of WW3 pre-post is failing to build ww3_grib. The issue was isolated to the load of an old nceppost module that was overriding the modules we were otherwise loading.

The nceppost is no longer needed, nor are several others currently being loaded to build ww3prepost.

Additionally, the build script does not correctly report build failures by using a non-zero exit code.

WalterKolczynski-NOAA:
Fixes the build failure of ww3_grib in ww3prepost by eliminating unneeded modules from the modulefile. Also adds error checking to the ww3prepost build script.

Closes #339

### DIFF
--- a/modulefiles/modulefile.ww3.hera
+++ b/modulefiles/modulefile.ww3.hera
@@ -13,18 +13,5 @@ module load jasper/2.0.25
 module load zlib/1.2.11
 module load png/1.6.35
 
-module load hdf5/1.10.6
-module load netcdf/4.7.4
-module load esmf/8_1_0_beta_snapshot_36
-
 module load bacio/2.4.1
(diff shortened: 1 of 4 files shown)


JessicaMeixner-NOAA (approved):
@WalterKolczynski-NOAA thanks for cleaning up the modules and adding the error checking in the build so this will not be a silent failure in the end.  I think this looks good as is, but one question: Doesn't a module file.ww3.stampede exists and should we go ahead and clean that up too?

WalterKolczynski-NOAA:
I've made the same changes on stampede, and they will be included if/when I'm ever successful there.

PR #342, opened 2021-06-10
Free-forecast integration with chgres_cube and resolve know mode bugs

### LINKED ISSUE #178: Update free-forecast mode with UFS_UTILS/chgres_cube integration and resolve know mode bugs
Updates in this issue will do the following:
- resolve existing free-forecast mode bugs
- integrate free-forecast mode with chgres_cube gdas_init utility in UFS_UTILS
- produce free-forecast ICs from v16 retrospective parallels or operational output for GFS versions 12, 13, 14, 15, and 16

The following changes are incoming with this work:

- New:
  - add MODE variable (either "free" or "cycled") to config.base and setup scripts
  - add RETRO variable in config.getic (set YES to use v16 pre-implementation parallel output as source for init, NO=ops)
  - add SDATE check in config.base: if earlier than 2019020100 then turn off waves (no rtofs support in GDA before then)

- Rework/adjustments:
  - replace fv3ic job with init job (new init.sh and config.init) - job will run chgres_cube via UFS_UTILS gdas_init utility
  - updated UFS_UTILS version
    - checkout version in checkout.sh (new tag: ufs_utils_1_4_0)
    - update build_ufs_utils.sh to use newer UFS_UTILS build script (build_all_ufs_utils.sh -> build_all.sh)
    - update link_fv3gfs.sh to adjust ufs_utils.fd fix path and remove other exec/script symlinks that are no longer needed
  - rewrite getic job to pull warm starts or chgres_cube inputs for init job off HPSS; includes pull of pgrb anl files for verification (v14+)
  - expand IAU on/off check in config.base to include MODE check
  - low resolution resource adjustments; provided by Fanglin Yang and Jun Wang while testing getic/init jobs for all resolutions
  - remove options in checkout/build for pre-CCPP GFSv16 FV3 (CCPP is now default)
    - checkout:
      - remove "-c" argument from checkout.sh (which set run_ccpp=YES)
      - remove checkout of v16 FV3 ops tag (when run_ccpp=NO)
    - build:
      - remove "-c" argument from build_all.sh, remove check of RUN_CCPP in build_all.sh for build_fv3.sh
      - remove check of RUN_CCPP in build_fv3.sh and option to build v16 ops tag (CCPP build without waves is default now)
  - add hpc/1.1.0 and cmake module loads to machine-setup.sh for Hera and WCOSS-Dell
  - disconnect HPSS access to whether the arch jobs are added the workflow, always have archive jobs now regardless 

- Bug fixes:
  - add check for prior day rtofs symlink in waveprep job (not a bug in cycled mode)
  - adjust gfs archive tarballs depending on MODE; resolves free-forecast archive bug
  - update config.nsst to set NST_SPINUP=1 when CDATE<2017072000; bug fix provided by Xu Li
  - add check of DOBLL_WAV=YES in wavepostpnt job to resolve bug when DOBLL_WAV=NO and a mismatch is produced with the CFP options
  - add missing jobs to free-forecast mode (bufrsnd, awips*, gempak*, wavepost*)

Updated job sequence for free-forecast mode:

1. gfsgetic  - pull warm starts or chgres_cube inputs*
2. gfsinit - will use files pulled by getic job to run UFS_UTILS gdas_init utility to produce user-defined resolution forecast ICs
-- rest of jobs unchanged from current system - except for bug fixes in arch --
3. gfsfcst
4. gfspost*
5. gfsvrfy
6. gfsmetp*
6. additional downstream jobs; switch dependent
7. gfsarch

*not added to workflow on machines without HPSS access (e.g. Orion); users will need to provide inputs for init job

Thank you to @GeorgeGayno-NOAA for the UFS_UTILS updates to support this!

KateFriedman-NOAA:
This PR completes work for issue #178 and will close #178. Branch tested on WCOSS-Dell, Hera, and Orion.

Updates in this PR will do the following:
- resolve existing free-forecast mode bugs
- integrate free-forecast mode with chgres_cube gdas_init utility in UFS_UTILS
- produce free-forecast ICs from v16 retrospective parallels or operational output for GFS versions 12, 13, 14, 15, and 16

The following changes are incoming with this commit:

- New:
  - add MODE variable (either "free" or "cycled") to config.base and setup scripts
  - add RETRO variable in config.getic (set YES to use v16 pre-implementation parallel output as source for init, NO=ops)
  - add SDATE check in config.base: if earlier than 2019020100 then turn off waves (no rtofs support in GDA before then)

- Rework/adjustments:
  - updated UFS_UTILS version
    - checkout version in checkout.sh (new tag: ufs_utils_1_4_0)
    - update build_ufs_utils.sh to use newer UFS_UTILS build script (build_all_ufs_utils.sh -> build_all.sh)
    - update link_fv3gfs.sh to adjust ufs_utils.fd fix path and remove other exec/script symlinks that are no longer needed
  - replace fv3ic job with init job (new init.sh and config.init) - job will run chgres_cube via UFS_UTILS gdas_init utility
  - rewrite getic job to pull warm starts or chgres_cube inputs for init job off HPSS; includes pull of pgrb anl files for verification (v14+)
  - expand IAU on/off check in config.base to include MODE check
  - low resolution resource adjustments; provided by Fanglin Yang and Jun Wang while testing getic/init jobs for all resolutions
  - remove options in checkout/build for pre-CCPP GFSv16 FV3 (CCPP is now default)
    - checkout:
      - remove "-c" argument from checkout.sh (which set run_ccpp=YES)
      - remove checkout of v16 FV3 ops tag (when run_ccpp=NO)
    - build:
      - remove "-c" argument from build_all.sh, remove check of RUN_CCPP in build_all.sh for build_fv3.sh
      - remove check of RUN_CCPP in build_fv3.sh and option to build v16 ops tag (CCPP build without waves is default now)
  - add hpc/1.1.0 and cmake module loads to machine-setup.sh for Hera and WCOSS-Dell
  - disconnect HPSS access to whether the arch jobs are added the workflow, always have archive jobs now regardless 

- Bug fixes:
  - add check for prior day rtofs symlink in waveprep job (not a bug in cycled mode)
  - adjust gfs archive tarballs depending on MODE; resolves free-forecast archive bug
  - update config.nsst to set NST_SPINUP=1 when CDATE<2017072000; bug fix provided by Xu Li
  - add check of DOBLL_WAV=YES in wavepostpnt job to resolve bug when DOBLL_WAV=NO and a mismatch is produced with the CFP options
  - add missing jobs to free-forecast mode (bufrsnd, awips*, gempak*, wavepost*)

Updated job sequence for free-forecast mode:

1. gfsgetic  - pull warm starts or chgres_cube inputs*
2. gfsinit - will use files pulled by getic job to run UFS_UTILS gdas_init utility to produce user-defined resolution forecast ICs
-- rest of jobs unchanged from current system - except for bug fixes in arch --
3. gfsfcst
4. gfspost*
5. gfsvrfy
6. gfsmetp*
6. additional downstream jobs; switch dependent
7. gfsarch

*not added to workflow on machines without HPSS access (e.g. Orion); users will need to provide inputs for init job

Thank you to @GeorgeGayno-NOAA for the UFS_UTILS updates to support this!


jobs/JGLOBAL_WAVE_PREP
```diff
@@ -77,6 +77,10 @@ else
   if [ ! -L $ROTDIR/${WAVECUR_DID}.${RPDY} ]; then # Check if symlink already exists in ROTDIR
     $NLN $DMPDIR/${WAVECUR_DID}.${RPDY} $ROTDIR/${WAVECUR_DID}.${RPDY}
   fi
+  BRPDY=`$NDATE -24 ${RPDY}00 | cut -c1-8`
```
WalterKolczynski-NOAA, line 80: Can we start the transition away from backticks as we make any modifications?

jobs/rocoto/arch.sh
```diff
@@ -182,7 +182,13 @@ if [ $CDUMP = "gfs" ]; then
 
     #for targrp in gfs_flux gfs_netcdf/nemsio gfs_pgrb2b; do
     if [ ${SAVEFCSTNEMSIO:-"YES"} = "YES" ]; then
-        for targrp in gfs_flux gfs_${format}a gfs_${format}b gfs_pgrb2b; do
+        if [ $MODE = "cycled" ]; then
+          targrp_list=`gfs_flux gfs_${format}a gfs_${format}b gfs_pgrb2b`
```
WalterKolczynski-NOAA, line 186: More backticks

jobs/rocoto/arch.sh
```diff
@@ -182,7 +182,13 @@ if [ $CDUMP = "gfs" ]; then
...
+        elif [ $MODE = "free" ]; then
+          targrp_list=`gfs_flux gfs_${format}b gfs_pgrb2b`
```
WalterKolczynski-NOAA, line 188: More backticks

sorc/build_all.sh
```diff
@@ -9,16 +9,6 @@ set +x
 #                   Anything other than "true"  will use libraries locally.
 #------------------------------------
 
-while getopts "c" option;
```
WalterKolczynski-NOAA, line 12: I'm worried when this change gets merged into coupled-crow, it is going to remove the c option we need there.

sorc/build_fv3.sh
```diff
@@ -23,10 +23,9 @@ cd fv3gfs.fd/
...
 if [ ! -d ../NEMS/exe ]; then mkdir ../NEMS/exe ; fi
-if [ ${RUN_CCPP:-${1:-"YES"}} = "NO" ]; then
- ./compile.sh "$FV3" "$target" "WW3=Y 32BIT=Y" 1
- mv -f fv3_1.exe ../NEMS/exe/global_fv3gfs.x
-else
- ./compile.sh "$target" "APP=ATM CCPP=Y 32BIT=Y SUITES=FV3_GFS_v16,FV3_GFS_v16_RRTMGP" 2 YES NO
- mv -f fv3_2.exe ../NEMS/exe/global_fv3gfs.x
-fi
+
+#./compile.sh "$FV3" "$target" "WW3=Y 32BIT=Y" 1
```
WalterKolczynski-NOAA: Just remove the dead code?

sorc/checkout.sh
```diff
@@ -9,10 +9,6 @@ do
    echo "Received -o flag for optional checkout of GTG, will check out GTG with EMC_post"
    checkout_gtg="YES"
    ;;
-  c)
```
WalterKolczynski-NOAA, line 12: Same issue as before. coupled-crow uses the c switch to checkout and build the coupled model.

sorc/checkout.sh
```diff
@@ -31,11 +27,7 @@ if [[ ! -d fv3gfs.fd ]] ; then
     rm -f ${topdir}/checkout-fv3gfs.log
     git clone https://github.com/ufs-community/ufs-weather-model fv3gfs.fd >> ${topdir}/checkout-fv3gfs.log 2>&1
     cd fv3gfs.fd
-    if [ ${run_ccpp:-"NO"} = "NO" ]; then
-     git checkout GFS.v16.0.16
-    else
-     git checkout 9350745855aebe0790813e0ed2ba5ad680e3f75c
-    fi
+    git checkout 9350745855aebe0790813e0ed2ba5ad680e3f75c
```
WalterKolczynski-NOAA, line 30: Would prefer a tag here, if there is one.

ush/rocoto/setup_workflow_fcstonly.py
```diff
@@ -132,7 +132,7 @@ def get_definitions(base):
     strings.append('\t<!ENTITY SCHEDULER  "%s">\n' % scheduler)
     strings.append('\n')
     strings.append('\t<!-- Toggle HPSS archiving -->\n')
-    strings.append('\t<!ENTITY ARCHIVE_TO_HPSS "YES">\n')
+    strings.append('\t<!ENTITY ARCHIVE_TO_HPSS "%s">\n' % base['HPSSARCH'])
```
WalterKolczynski-NOAA, line 135: String modulo has long been discouraged in favor of using format of f-strings. We should start using f-strings whenever possible, as it is the easiest and most readable, thought it doesn't have to be right now.

WalterKolczynski-NOAA (changes requested):
Mostly just format modernization requests, but one possible problem this could cause later when we merge to coupled-crow.

scripts/exgfs_wave_post_pnt.sh
```diff
@@ -624,7 +624,7 @@
 
   if [ ${CFP_MP:-"NO"} = "YES" ]; then nm=0; fi
 
-  if [ ${CFP_MP:-"NO"} = "YES" ]; then
+  if [ ${CFP_MP:-"NO"} = "YES" ] && [ "$DOBLL_WAV" = "YES" ]; then
```
JessicaMeixner-NOAA, line 627: What was the issue here?  What if say, DOBLL_WAV=NO but DOSPC_WAVE=YES, do we get to line 630?

jobs/rocoto/getic.sh
```diff
@@ -36,167 +36,131 @@ status=$?
...
+  # Pull RESTART files off HPSS
+  if [ ${RETRO:-"NO"} = "YES" ]; then # Retrospective parallel input
+
+    # Pull prior cycle restart files
+    htar -xvf ${HPSSDIR}/${GDATE}/gdas_restartb.tar
+    status=$?
+    [[ $status -ne 0 ]] && exit $status
 
-export COMPONENT=${COMPONENT:-atmos}
+    if [ $DO_WAVE = "YES" ]; then
```
JessicaMeixner-NOAA: This only works if you are using v16 does the develop branch still support v16? If you move away from v16, this will not work. WW3 restarts/ICs are WW3 model version specific.

jobs/JGLOBAL_WAVE_PREP
WalterKolczynski-NOAA: Actually, don't worry about all the backticks. I'm going to just go through and replace them in all scripts after this PR is merged (though I will need someone to test afterwards).

sorc/checkout.sh
JessicaMeixner-NOAA: Maybe we can get rid of the -c option in feature/coupled-crow too?  After all you can build the coupled exe and still run atmosphere only.

WalterKolczynski-NOAA: We could do that. My only concern is if/when we merge coupled-crow back to develop we're now locked into using the same ufs commit for both. This may not really be an issue though.

JessicaMeixner-NOAA: That sounds like a positive not a negative to me.   Someone can always manually update the version to be checked out.

JessicaMeixner-NOAA: Having the two options is somewhat of a hold over from the ufs-s2s-model versus ufs-weather-model.  Now the difference is that the previous s2s checks out into a "ufs-coupled.fd" instead of the fv3gfs folder.

sorc/build_all.sh
KateFriedman-NOAA: ~~Want me to leave this line but still remove what that argument does?~~ Nevermind, per discussion about similar change in checkout.sh.

sorc/build_fv3.sh
KateFriedman-NOAA: Removed.

sorc/checkout.sh
KateFriedman-NOAA: Are we ok with me removing the -c option in develop and it's impact on feature/coupled-crow?

sorc/checkout.sh
KateFriedman-NOAA: I prefer tags as well but there aren't any recent ones. Will definitely retest and move to a new tag when one is made.

ush/rocoto/setup_workflow_fcstonly.py
KateFriedman-NOAA: Gotcha. Am happy to redo things to f-strings when we go through to do mass updating like we've discussed, I kept things consistent with how the script was doing this for now.

scripts/exgfs_wave_post_pnt.sh
KateFriedman-NOAA: If CFP_MP=YES we enter the if-block on [line 627](https://github.com/KateFriedman-NOAA/global-workflow/blob/issue178/scripts/exgfs_wave_post_pnt.sh#L627). Then, regardless of what DOBNDPNT_WAV equals, if DOBLL_WAV=NO and DOSPC_WAV=YES your nm counter coming out of that if-block will be "1". Then [wavenproc=1](https://github.com/KateFriedman-NOAA/global-workflow/blob/issue178/scripts/exgfs_wave_post_pnt.sh#L671) and since it's not greater than 1 it won't enter the [CFP_MP part of the next if-block on line 681](https://github.com/KateFriedman-NOAA/global-workflow/blob/issue178/scripts/exgfs_wave_post_pnt.sh#L681). So you set up your cmdfile with the leading nm value (what running with CFP_MP=YES needs for using --multi-prog or equivalent) but when the job goes to [the non-CFP_MP else](https://github.com/KateFriedman-NOAA/global-workflow/blob/issue178/scripts/exgfs_wave_post_pnt.sh#L691) it errors because of the leading nm # in the cmdfile. Essentially, we have to avoid coming out of a CFP_MP block with nm<2.

> What if say, DOBLL_WAV=NO but DOSPC_WAVE=YES, do we get to line 630?

If DOBLL_WAV=NO you won't go through the CFP_MP=YES part of the block, you'll go to the non-CFP_MP part on line 651 and get a cmdfile without the leading # on each line within for CFP_MP running.

Note, this is a highly specialized workaround for this bug. If additional DO*=YES blocks are added to that section that will avoid this situation and the check of DOBLL_WAV=YES can go away. A rework of this may be best in the future.

jobs/rocoto/getic.sh
KateFriedman-NOAA: I added this before we dropped wave support in develop but I kept it for possible future needs. I will remove it given your comment.

KateFriedman-NOAA: Removed.

JessicaMeixner-NOAA: @KateFriedman-NOAA it's not about supporting waves or not, it's about which version of the code/waves you are supporting.  If you are supporting the production branch, then this is necessary, but I believe you dropped support of the production branch which is what makes this unnecessary.

KateFriedman-NOAA: Gotcha, thanks!

sorc/checkout.sh
JessicaMeixner-NOAA: I am

WalterKolczynski-NOAA (approved):

PR #350, opened 2021-06-22
Fix NST_SPINUP being ignored by CROW

WalterKolczynski-NOAA:
The NSST config file was setting NST_SPINUP to use an environment variable rather than using the value set by the yaml, so values set in the case file were ignored.

Refs: #337

### DIFF
--- a/workflow/config/nsst.yaml
+++ b/workflow/config/nsst.yaml
@@ -6,12 +6,12 @@ config_nsst:
   filename: config.nsst
   nst_spinup_logic: !FirstTrue
     - when: !calc doc.nsst.get("NST_SPINUP","")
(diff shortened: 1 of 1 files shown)


WalterKolczynski-NOAA:
Note: I didn't actually run the model, I just confirmed that the config.nsst file is now exporting the value set in the yaml.

KateFriedman-NOAA (approved):
Looks correct based on my understanding of CROW yaml.

rmontuoro (approved):
See also commit d16630f3b34671e2350e6a88b9df749e4fd58ece

WalterKolczynski-NOAA:
Going to merge this one, since it is clearly a bug, then we can deal with any additional changes needed for NSST separately.

PR #362, opened 2021-07-14
Correct wrong settings for high-res warm-start free-forecast

### LINKED ISSUE #353: Wrong settings for high-res warm-start free-forecast
Few bugs in free-forecast mode regarding high res warm start file pulls:

1) EXP_WARM_START needs to be .true. for this mode...force set config.getic value if user sets to run CASE=C768
2) GDATE gdas restart files pulled off tape but not moved into ROTDIR (left out)
3) default C768 node value (218) is too high for Hera (max per job limit now exists, adjust C768 settings to keep it under 200 nodes)

KateFriedman-NOAA:
Changes in this PR:

- fix bug in getic when pulling warm starts off tape
- reduce C768 forecast resources to lower the node value below Hera limit of 210 nodes/job
- update setup_expt_fcstonly.py to better set EXP_WARM_START based on user inputs
- add OPS_RES variable to use for checks that need to know current ops resolution

Close #353

### DIFF
--- a/jobs/rocoto/getic.sh
+++ b/jobs/rocoto/getic.sh
@@ -82,6 +82,7 @@ if [[ $gfs_ver = "v16" && $EXP_WARM_START = ".true." && $CASE = $OPS_RES ]]; the
 
   else # Opertional input - warm starts
 
+    cd $ROTDIR
     # Pull CDATE gfs restart tarball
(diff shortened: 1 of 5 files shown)


WalterKolczynski-NOAA (approved):

KateFriedman-NOAA:

PR #373, opened 2021-07-24
Update to tiled fix files, UGWD v1, NOAH-MP

### LINKED ISSUE #331: coupled: use tiled fix files and oro data
New tiled fix files and oro data have been staged on Mars here: /gpfs/dell2/emc/modeling/noscrub/emc.glopara/git/fv3gfs/fix_NEW

We need to use the new oro data and tiled fix files in the feature/coupled-crow workflow.  An example of how this is used can be seen in the ufs-weather-model rt tests: cpld_bmarkfrac_wave_v16_noahmp

### LINKED ISSUE #346: updates for fix files for uGWD
There are new fix files for uGWD here: 
 /scratch1/NCEPDEV/global/Fanglin.Yang/save/git/gfsv16_rrtmgp2/fix/fix_ugwd 

Namelist changes are needed to use the uGWD which should be in the 35 day tests in the p7b branch: 
https://github.com/ufs-community/ufs-weather-model/tree/release/P7b the code updates themselves should already be in the develop branch of ufs-weather-model (not sure which commit/PR those updates went in for an exact date). 

FYI: @yangfanglin @KateFriedman-NOAA @junwang-noaa @WalterKolczynski-NOAA @jiandewang

WalterKolczynski-NOAA:
Updates model to use tiled fix files. Fix directory is updated to the
fix_NEW location.

Adds the ability to use UGWD v1. Since this capability is tied to the
CCPP suite used, the sutie definition file is grepped to determine whether
UGWD is active. Otherwise, gwd_opt 1 is used. Either way, the appropriate
namelist settings are added to input.nml. For v1, the necessary fix files
are also linked to the run directory. If additional options are supported
in the future, there will need to be more sophisicated parsing.

Adds the ability to use Noah-MP. Like UGWD, this is dictated by the CCPP
suite used, so the suite definition file is grepped to determine whether
to use Noah (lsm=1) or Noah-MP (lsm=2).

Additional CCPP suites are added to the build to allow for the new options.

The two non-fractional coupled cases are updated to use a CCPP suite using
both UGWD v1 and Noah-MP. The two aerosol cases are updated to a suite using
UGWD v1 (there does not appear to be an atm-only suite that has both).

There is also a minor change to the UFS build script to remove any existing
UFS build directory. This prevents problems when attempting to build a
different app after one has already been built.

Closes: #331, #346

### DIFF
--- a/ush/forecast_postdet.sh
+++ b/ush/forecast_postdet.sh
@@ -212,7 +212,13 @@ FV3_GFS_postdet(){
 		done
 	fi
 
+	export CCPP_SUITE=${CCPP_SUITE:-"FV3_GFS_v16"}
 	_suite_file=$HOMEgfs/sorc/ufs_coupled.fd/FV3/ccpp/suites/suite_${CCPP_SUITE}.xml
+
+	if [ ! -f ${_suite_file} ]; then
+		echo "FATAL: CCPP Suite file ${_suite_file} does not exist!"
+		exit 2
+	fi
 
 	# Scan suite file to determine whether it uses Noah-MP
 	if [ $(grep noahmpdrv ${_suite_file} | wc -l ) -gt 0 ]; then
--- a/ush/forecast_postdet.sh
+++ b/ush/forecast_postdet.sh
@@ -198,21 +198,46 @@ FV3_GFS_postdet(){
 		$NLN ${src_grid_spec} $DATA/INPUT/${CASE}_mosaic.nc
 	fi
 
-	if [ -z ${OROFIX} ]; then
-		if [ $FRAC_GRID ]; then
-			FIXoro=${FIXoro:-$FIX_DIR/fix_fv3_fracoro}/${CASE}.mx${OCNRES}_frac/${CASE}.mx${OCNRES}_
-		else
-			FIXoro=$FIXfv3/$CASE/${CASE}_oro_data.
-		fi
+	if [ $FRAC_GRID = ".true." ]; then
+		OROFIX=${OROFIX:-"${FIX_DIR}/fix_fv3_fracoro/${CASE}.mx${OCNRES}_frac"}
+		FIX_SFC=${FIX_SFC:-"${OROFIX}/fix_sfc"}
+		for n in $(seq 1 $ntiles); do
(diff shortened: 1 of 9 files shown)


WalterKolczynski-NOAA:
Confirmed to work for coupled_wave, coupled, and firex. Also tested firex with the old CCPP suite to confirm the old settings still work.

ush/forecast_postdet.sh
```diff
@@ -198,21 +198,46 @@ FV3_GFS_postdet(){
...
+		for n in $(seq 1 $ntiles); do
+			$NLN ${OROFIX}/${CASE}_oro_data.tile${n}.nc $DATA/INPUT/oro_data.tile${n}.nc
+		done
 	fi
-	for n in $(seq 1 $ntiles); do
-		$NLN ${FIXoro}tile${n}.nc $DATA/INPUT/oro_data.tile${n}.nc
-	done
 
-	# GFS standard input data
+	_suite_file=$HOMEgfs/sorc/ufs_coupled.fd/FV3/ccpp/suites/suite_${CCPP_SUITE}.xml
```
rmontuoro, line 216: Should this statement be guarded? What happens if suite_${CCPP_SUITE}.xml does not exist?

rmontuoro (approved):
Confirming that these changes work with aerosols.
One minor comment to be addressed.

ush/forecast_postdet.sh
WalterKolczynski-NOAA: Other portions of the same script already use it, so I assumed it was guaranteed to be set. That does not appear to be the case after all. I will move the default from parsing_namelists_FV3.sh up somewhere to make sure it is set.

WalterKolczynski-NOAA: And yes, should also check that the file exists.

JessicaMeixner-NOAA (approved):
The tiled fix files look good to me and the fix files for uGWD also seem to be moved over as they're in the rt.sh.  I was able to create a successful run on hera.

ush/forecast_postdet.sh
WalterKolczynski-NOAA: Fixed now

PR #375, opened 2021-07-26
Unify build system

### LINKED ISSUE #366: Bring coupled model forecast capabilities into `develop`
This epic will consolidate issues to bring in the capabilities of executing coupled model forecast from `feature/coupled_crow` (without CROW dependency) into `develop` to support coupled prototype efforts and eliminate duplication and redundancies.

Several accompanying issues will be handled in this epic as well:
- Python 2to3 updates for setup scripts. See issue #265
- Modularize exglobal forecast script. See issue #365

### LINKED ISSUE #368: Update free forecast setup for coupled model.
Update free forecast setup scripts (experiment and workflow) to include coupled model mode.

Branch created for this effort: feature/coupled_sprint

### LINKED ISSUE #374: Unify checkout and build systems
The checkout and build systems for the standard GFS, the coupled version, and the aerosol version need to be unified around a single set of scripts.

See also: #366, #368

WalterKolczynski-NOAA:
Combines the checkout and build from the standard and coupled workflows
system that checks out the current UFS.

The model source directory is now ufs_model.fd rather than fv3gfs.fd or
ufs_coupled.fd. The build script is renamed build_ufs.sh and the link
script is renamed to simply link.sh.

The checkout script checks out the same UFS version (current as of
2021 Jul 20) for all applications, removing the coupled option from the
coupled branch. A new option, -m, allows the user to specify a different
model hash to check out. Checkout log files are now placed in the logs
directory like the build logs in order to keep the sorc directory clean.

Two new options are added to the build_all script: -c to build the
coupled (S2SW) app, and -a to build the aerosol (ATMAERO) app. Without
either argument, the traditional (ATMW) app is built. If either option
is provided, the cpl_build.cfg file controls which executables are built;
otherwise fv3gfs_build.cfg is still used. The build_ufs script corre-
spondingly accepts the same options.

The reg2grb2 source directory is copied over from the coupled model and
cmake files for workflow_utils are updated correspondingly to build.
Additionally, the build_workflow_utils modulefiles are updated to use the
ncio module in hpc-stack and the hack that cloned and built it as part of
the workflow_utils build is removed.

The build_ww3prepost script is updated to make the code more easily main-
tained and the corresponding modulefiles are updated to remove unnecessary
modules.

The link script is updated to the new fix location (ending in fix_NEW)
removes links with targets that no longer exist. A new third argument to
link, coupled, is added to additionally link files needed by the coupled
system.

Note that while the checkout, build, and link have all been tested and
complete without error, the workflow is otherwise untested and may not
work until additional updates are made to use the current UFS.

Refs: #366, #368, #374


sorc/build_ufs.sh
```diff
@@ -0,0 +1,49 @@
...
+esac
+
+MOD_PATH=$cwd/ufs_model.fd/modulefiles
+
+module purge
+
+cd ufs_model.fd/
+module use ${MOD_PATH}
+module load ufs_${target}
+CMAKE_FLAGS="-DAPP=${APP} -DCCPP_SUITES=${CCPP_SUITES}" ./build.sh
```
emcbot, line 49: Please don't use the `build.sh` in the UFS weather model.  Its use is for the regression tests.  Users and developers should use `cmake` instructions required to build the model.

```suggestion
mkdir -p build && cd build
cmake -DAPP=${APP} -DCCPP_SUITES=${CCPP_SUITES} ..
OMP_NUM_THREADS=1 make -j ${BUILD_JOBS:-8} VERBOSE=${BUILD_VERBOSE:-} 
```

sorc/link.sh
```diff
@@ -5,19 +5,30 @@ set -ex
```
emcbot, line 5: I don't think just `link.sh` is appropriate as it is too generic.
Please retain a suffix and ensure it is consistent with the `usage` blurb inside the script.

sorc/build_ufs.sh
aerorahul: Please don't use the `build.sh` in the UFS weather model.  Its use is for the regression tests.  Users and developers should use `cmake` instructions required to build the model.

```suggestion
mkdir -p build && cd build
cmake -DAPP=${APP} -DCCPP_SUITES=${CCPP_SUITES} ..
OMP_NUM_THREADS=1 make -j ${BUILD_JOBS:-8} VERBOSE=${BUILD_VERBOSE:-} 
```

sorc/link.sh
aerorahul: I don't think just `link.sh` is appropriate as it is too generic.
Please retain a suffix and ensure it is consistent with the `usage` blurb inside the script.

sorc/build_ufs.sh
WalterKolczynski-NOAA: I am fine making this change, but until now this is how we've been building UFS in both branches.

sorc/checkout.sh
```diff
@@ -21,24 +24,24 @@ do
...
+if [[ ! -d ufs_model.fd ]] ; then
+    git clone https://github.com/ufs-community/ufs-weather-model ufs_model.fd >> ${logdir}/checkout-ufs_model.log 2>&1
+    cd ufs_model.fd
+    git checkout ${ufs_model_hash:-22613e8}
     git submodule update --init --recursive
     cd ${topdir}
 else
-    echo 'Skip.  Directory fv3gfs.fd already exists.'
-fi
+    echo 'Skip.  Directory ufs.fd already exists.'
```
KateFriedman-NOAA: Should be "ufs_model.fd already exists."

sorc/link.sh
WalterKolczynski-NOAA: Updated to link_workflow.sh

sorc/build_ufs.sh
aerorahul: Sure. 
Going forward, if we want to maintain portability.  We should:
- ensure we use the build instructions and not the build script provided by the components
- minimize reliance on modulefiles (this is impossible on supported platforms).  But, if we use build instructions, we take the module load environment out of the equation, since the module files provide paths to the third party libraries.  So, if we land on a machine with no modules but the build environment in the `PATH`, we can build without modulefiles and the component build scripts.

sorc/fv3gfs_build.cfg
```diff
@@ -1,7 +1,7 @@
 #
 # ***** configuration of global-workflow build *****
 
- Building fv3gfs (fv3gfs) .............................. yes
+ Building ufs_model (ufs_model) ........................ yes
```
KateFriedman-NOAA, line 4: Why retain this build config if we have cpl_build.cfg now?

sorc/checkout.sh
WalterKolczynski-NOAA: Yes, it should. Fixed.

sorc/fv3gfs_build.cfg
WalterKolczynski-NOAA: The coupled config doesn't build most of the parts used by GFS (like GSI). Eventually it will, but until then there is no use building parts not needed for the prototypes. There probably should be a separate one for aerosols too (they don't need WW3 pre/post, for example).

sorc/build_ufs.sh
WalterKolczynski-NOAA: Updated to use cmake directly. Per conversation, postponed moving build directory out of UFS repo. A future PR will do that for all components.

sorc/link_workflow.sh
```diff
@@ -68,7 +100,11 @@ cd ${pwd}/../parm               ||exit 8
     $LINK ../sorc/gldas.fd/parm                              gldas
 cd ${pwd}/../scripts            ||exit 8
     $LINK ../sorc/gfs_post.fd/scripts/exgdas_atmos_nceppost.sh .
-    $LINK ../sorc/gfs_post.fd/scripts/exgfs_atmos_nceppost.sh  .
+    if [ $model = "coupled" ]; then
+      $LINK exgfs_nceppost_cpl.sh exgfs_atmos_nceppost.sh
```
JessicaMeixner-NOAA, line 104: This seems to be the last special section for "coupled", I forget what the issue/difference is here, but should we open an issue(s) somewhere to get this resolved so we don't have to have these two different copies?

JessicaMeixner-NOAA: Well, there's one other "coupled" section, but comment still remains.

WalterKolczynski-NOAA: I expect there will be a bunch of little things like this to hammer out as we determine which is because newer UFS and which is because coupled.

WalterKolczynski-NOAA: This script in particular is the one Arun so much wants to move out of post and into workflow anyway (and replace it with one that just does the regridding now that we use inline post).

sorc/link_workflow.sh
```diff
@@ -207,29 +247,29 @@ for workflowexec in fbwndgfs gfs_bufr regrid_nemsio supvit syndat_getjtbul \
...
 $LINK ../sorc/gfs_post.fd/exec/upp.x gfs_ncep_post
 
 if [ -d ${pwd}/gfs_wafs.fd ]; then 
     for wafsexe in \
-          wafs_awc_wafavn.x  wafs_blending.x  wafs_blending_0p25.x \
-          wafs_cnvgrib2.x  wafs_gcip.x  wafs_grib2_0p25.x \
-          wafs_makewafs.x  wafs_setmissing.x ; do
+          wafs_awc_wafavn  wafs_blending  wafs_blending_0p25 \
+          wafs_cnvgrib2  wafs_gcip  wafs_grib2_0p25 \
+          wafs_makewafs  wafs_setmissing; do
```
KateFriedman-NOAA: The hash of WAFS in the develop branch produces execs with ".x" in the name. Does the linking work with this change?

WalterKolczynski-NOAA: I'm going to guess no. I didn't pick up on this in testing because no WAFS links are produced at all, so there were no broken links to see.

KateFriedman-NOAA (changes requested):
Please revert removal of ".x" from WAFS exec linking. Changes look good otherwise, thanks!

KateFriedman-NOAA (approved):

sorc/checkout.sh
```diff
@@ -72,11 +75,10 @@ fi
 echo EMC_post checkout ...
 if [[ ! -d gfs_post.fd ]] ; then
     rm -f ${topdir}/checkout-gfs_post.log
-    git clone https://github.com/NOAA-EMC/EMC_post.git gfs_post.fd >> ${topdir}/checkout-gfs_post.log 2>&1
+    git clone https://github.com/NOAA-EMC/EMC_post.git gfs_post.fd >> ${logdir}/checkout-gfs_post.log 2>&1
     cd gfs_post.fd
-    git checkout ff42e0227d6100285d4179a2572b700fd5a959cb
+    git checkout upp_v10.0.3
```
JessicaMeixner-NOAA: https://github.com/NOAA-EMC/EMC_post/commit/ff42e0227d6100285d4179a2572b700fd5a959cb is from April where as the upp_v10.0.3 is from January.  Should we use the more recent UPP hash?

KateFriedman-NOAA: If the coupled system doesn't specifically need the upp_v10.0.3 tag then I agree that the newer hash should remain. That's the hash that was tested with the hpc-stack updates commit.

modulefiles/modulefile.ww3.wcoss_dell_p3
```diff
@@ -9,20 +9,7 @@ module load jasper/2.0.25
...
-module load g2tmpl/1.9.1
-module load ip/3.3.3
-module load nemsio/2.5.2
-module load sp/2.3.3
-module load w3emc/2.7.3
-module load w3nco/2.4.1
-module load wgrib2/2.0.8
-module load upp/10.0.6
+
+module load netcdf/4.7.4
```
aerorahul, line 16: we should load the `netcdf` dependency `hdf5` before netcdf.

modulefiles/modulefile.ww3.orion
```diff
@@ -14,20 +14,7 @@ module load jasper/2.0.25
...
-module load g2tmpl/1.9.1
-module load ip/3.3.3
-module load nemsio/2.5.2
-module load sp/2.3.3
-module load w3emc/2.7.3
-module load w3nco/2.4.1
-module load wgrib2/2.0.8
-module load upp/10.0.6
+
+module load netcdf/4.7.4
```
aerorahul, line 21: load `hdf5` dependency explicitly before `netcdf`.

modulefiles/workflow_utils.hera
```diff
@@ -20,7 +20,9 @@ module load sp/2.3.3
 module load ip/3.3.3
 module load nemsio/2.5.2
 module load nemsiogfs/2.5.3
-#module load ncio/1.0.0
+module load ncio/1.0.0
```
aerorahul, line 26: Please move `hdf5` and `netcdf` from lines 30-31 above.
`ncio` depends on `netcdf` and order matters.
We get away with it here because we build everything as static libraries.

modulefiles/workflow_utils.orion
```diff
@@ -20,7 +20,9 @@ module load sp/2.3.3
 module load ip/3.3.3
 module load nemsio/2.5.2
 module load nemsiogfs/2.5.3
-#module load ncio/1.0.0
+module load ncio/1.0.0
```
aerorahul, line 26: same comment as for hera

sorc/build_all.sh
```diff
@@ -9,6 +9,25 @@ set +x
...
+   break
+   ;;
+  c)
+   echo "Received -c flag, building ufs-weather-model for S2SW app"
+   echo "skipping builds not needed for prototype runs"
+   _build_ufs_options=-c
+   break
+   ;;
+ esac
+done
```
aerorahul, line 29: We will have to revisit this approach in the future.
Imagine if every component covered in `build_all.sh` has options to build, this argument list will get unmanageable.

sorc/build_ufs.sh
```diff
@@ -0,0 +1,52 @@
...
+  esac
+done
+
+source ./machine-setup.sh > /dev/null 2>&1
+cwd=$(pwd)
+
+
+# Set target platform
+case "${target}" in
+  hera|orion|stampede)
```
aerorahul: might as well throw `jet` and `cheyenne` here.  They follow the same pattern.

sorc/build_ufs.sh
```diff
@@ -0,0 +1,52 @@
...
+    ;;
+esac
+
+MOD_PATH=$cwd/ufs_model.fd/modulefiles
+
+module purge
+
+cd ufs_model.fd/
+module use ${MOD_PATH}
+module load ufs_${target}
```
aerorahul, line 45: please wrap all `module` commands in `set +x` and `set -x`.

sorc/build_ww3prepost.sh
```diff
@@ -35,6 +34,7 @@ export Z_LIB=${Z_LIB:-$ZLIB_ROOT/lib/libz.a}
 export JASPER_LIB=${JASPER_LIB:-$JASPER_ROOT/lib64/libjasper.a}
 export WWATCH3_NETCDF=NC4
 export NETCDF_CONFIG=$NETCDF_ROOT/bin/nc-config
+export PRINTER=${PRINTER:-""}
```
aerorahul, line 37: I wonder what is `PRINTER` doing?

sorc/fv3gfs_build.cfg
aerorahul: Should this file be renamed from `fv3gfs_build.cfg` to `gfs_build.cfg`. I assume `fv3gfs` refers to the application and not just the fv3-based model.

aerorahul: looking at the `link_fv3gfs.sh` file renamed below, another alternative name `workflow_build.cfg`?
Just throwing out suggestions.

sorc/link_workflow.sh
```diff
@@ -30,18 +41,39 @@ pwd=$(pwd -P)
...
-    FIX_DIR="/gpfs/dell2/emc/modeling/noscrub/emc.glopara/git/fv3gfs/fix"
+    FIX_DIR="/gpfs/dell2/emc/modeling/noscrub/emc.glopara/git/fv3gfs/fix_NEW"
 elif [ $machine = "hera" ]; then
-    FIX_DIR="/scratch1/NCEPDEV/global/glopara/fix"
+    FIX_DIR="/scratch1/NCEPDEV/global/glopara/fix_NEW"
 elif [ $machine = "orion" ]; then
-    FIX_DIR="/work/noaa/global/glopara/fix"
+    FIX_DIR="/work/noaa/global/glopara/fix_NEW"
+elif [ $machine = "stampede" ]; then
+    FIX_DIR="/work/07738/jkuang/stampede2/tempFixICdir/fix_UFSp6"
```
aerorahul, line 53: This is pointing to `jkuang`'s personal space.  Does this person know not to delete this path?

aerorahul (commented):
Just a few minor comments.  Otherwise looks good.

modulefiles/modulefile.ww3.wcoss_dell_p3
aerorahul: Also, move this up before any other module load that depends on these, e.g. `upp`

sorc/build_ww3prepost.sh
JessicaMeixner-NOAA: it's a variable that's defined as part of the ww3 build which is getting cleaned out.  So this should go away soon.  I'm assuming Walter added it because a version he was using needed it for some reason?

sorc/link_workflow.sh
JessicaMeixner-NOAA: We shouldn't even be pointing to that fix directory anymore b/c it's now out of date compared to the fix_NEW

sorc/fv3gfs_build.cfg
WalterKolczynski-NOAA: Yeah, in the future we can rename these files so they are consistent/clearer. We can worry about that later though.

sorc/build_all.sh
WalterKolczynski-NOAA: Don't disagree. Right now this is just a shortcut for the APP to build. We could easily just change this to take the app as an argument instead of these switches. Although, hopefully in the near future we can build one app executable and use it for all configurations (at least until regional gets brought in or something, by which time this will be replaced).

sorc/checkout.sh
WalterKolczynski-NOAA: Should we just use the latest UPP tag here?

aerorahul: the atmosphere only does inline post, so I don't think that will break @JianKuang-UMD's branch.

sorc/build_ww3prepost.sh
WalterKolczynski-NOAA: I turned on set -u and the script errored because it wasn't defined. So I defined it.

sorc/build_ufs.sh
WalterKolczynski-NOAA: Done

modulefiles/workflow_utils.orion
WalterKolczynski-NOAA: Done

modulefiles/workflow_utils.hera
WalterKolczynski-NOAA: Done

modulefiles/modulefile.ww3.orion
WalterKolczynski-NOAA: Done

modulefiles/modulefile.ww3.wcoss_dell_p3
WalterKolczynski-NOAA: UPP is no longer loaded for this. The only other modules loaded are the three compression libraries, bacio, and g2.

sorc/link_workflow.sh
WalterKolczynski-NOAA: Stampede users have been left to fend for themselves for the time being. The stampede blocks were brought in because it had partially been added to the coupled branch and I just copied it. I can remove them all, though I am inclined to leave them for at least a little while in case we wind up returning to that effort. But I also don't see a need to update paths until that happens.

sorc/checkout.sh
WalterKolczynski-NOAA: module_base is loading upp/10.0.6, so I propose we check out that version

WalterKolczynski-NOAA: Alternatively, UPP is now up to 10.0.8. I can update all of them to that tag

modulefiles/workflow_utils.hera
aerorahul: I think w3emc/2.7.3 depends on netcdf.
Later versions of w3emc removed that dependency.  These should be moved before the nceplibs list start since these are more closer to third-party than the nceplibs ones.

modulefiles/workflow_utils.orion
aerorahul: same comment as in hera.

sorc/build_ufs.sh
```diff
@@ -0,0 +1,54 @@
...
+# Set target platform
+case "${target}" in
+  hera|orion|stampede)
+    target=${target}.intel
+    ;;
+esac
+
+MOD_PATH=$cwd/ufs_model.fd/modulefiles
+
+module purge
```
aerorahul: please move this inside the `set +x` `set -x`

sorc/link_workflow.sh
aerorahul: Leave for now.

sorc/build_ufs.sh
aerorahul:
```suggestion
  hera|orion|stampede|jet|cheyenne)
```

sorc/build_ufs.sh
aerorahul:
```suggestion
```

sorc/build_ufs.sh
```diff
@@ -0,0 +1,54 @@
...
+
+cd ufs_model.fd/
+set +x
+module use ${MOD_PATH}
```
aerorahul, line 44:
```suggestion
module purge
module use ${MOD_PATH}
```

sorc/checkout.sh
aerorahul: sure.  I don't really have a preference.  It is used for offline post processing, and who is using that now?

WalterKolczynski-NOAA: I think we might be using some of the scripts? But I don't think the version really matters all that much.

WalterKolczynski-NOAA: Updated all to 10.0.8 for consistency

PR #376, opened 2021-07-26
Update free forecast EXPDIR setup for coupled model

### LINKED ISSUE #368: Update free forecast setup for coupled model.

KateFriedman-NOAA:
Bring in new configuration settings/variables from feature/coupled-crow for coupled system. Did not need to change setup_expt*py scripts for this effort.

The following were copied in as is from example coupled EXPDIR config.base but need quick discussion:
```
export FHRLST="[0, 6, 12, 18, 24, 30, 36, 42, 48]"
export STEP_GFS="24"
```

Please review new APP if-blocks in config.base for DO_* switches.

Testing these changes against develop branch and modular forecast script runs now.

Issue #368


parm/config/config.base.emc.dyn
```diff
@@ -136,17 +140,18 @@ export SENDDBN_NTC=${SENDDBN_NTC:-"NO"}
 export SENDDBN=${SENDDBN:-"NO"}
 export DBNROOT=${DBNROOT:-${UTILROOT}/fakedbn}
 
+export print_esmf=".false."
+export esmf_profile=".true."
```
JessicaMeixner-NOAA: Probably should be false as default.

parm/config/config.base.emc.dyn
```diff
@@ -227,11 +233,77 @@ export lobsdiag_forenkf=".true."
...
+
+if [ $APP = "ATM" ]; then
+  export DO_WAVE="NO"
+  export DO_MOM6="NO"
+  export DO_CICE="NO"
+  export DO_GOCART="NO"
+  export CCPP_SUITE="FV3_GFS_v16,FV3_GFS_v16_ugwpv1"
+elif [ $APP = "ATMW" ]; then
+  export DO_WAVE="YES"
+  export DO_MOM6="NO"
```
JessicaMeixner-NOAA: Is DO_<component> other than DO_WAVE a new variable even for f/c-c?

parm/config/config.base.emc.dyn
```diff
@@ -29,6 +29,10 @@ export UTILgfs=$HOMEgfs/util
 export EXECgfs=$HOMEgfs/exec
 export SCRgfs=$HOMEgfs/scripts
 
+export FIXcice=$HOMEgfs/fix/fix_cice
+export FIXmom=$HOMEgfs/fix/fix_mom6
+export OCNFIXDIR=$HOMEgfs/fix/fix_reg2grb2
```
aerorahul: Fine for now, but is there a reason why this is not `FIXreg2grb2` or atleast `FIXocn`

parm/config/config.base.emc.dyn
aerorahul: aren't these used only in the forecast components?
If so, why not in `config.fcst`?

parm/config/config.base.emc.dyn
```diff
@@ -227,11 +233,77 @@ export lobsdiag_forenkf=".true."
...
+  export DO_CICE="YES"
+  export DO_GOCART="NO"
+  export CCPP_SUITE="FV3_GFS_v16,FV3_GFS_v16_coupled_nsstNoahmpUGWPv1"
+elif [ $APP = "ATMAERO" ]; then
+  export DO_WAVE="NO"
+  export DO_MOM6="NO"
+  export DO_CICE="NO"
+  export DO_GOCART="YES"
+  export CCPP_SUITE="FV3_GFS_v16,FV3_GFS_v16_ugwpv1"
+fi
```
aerorahul:
```suggestion
# Defaults:
export DO_WAVE="NO"
export DO_MOM6="NO"
export DO_CICE="NO"
export DO_GOCART="NO"
export CCPP_SUITE="FV3_GFS_v16,FV3_GFS_v16_ugwpv1"

if [ $APP = "ATMW" ]; then
  export DO_WAVE="YES"
elif [ $APP = "S2S" ]; then
  export DO_MOM6="YES"
  export DO_CICE="YES"
  export CCPP_SUITE="FV3_GFS_v16,FV3_GFS_v16_coupled_nsstNoahmpUGWPv1"
elif [ $APP = "S2SW" ]; then
  export DO_WAVE="YES"
  export DO_MOM6="YES"
  export DO_CICE="YES"
  export CCPP_SUITE="FV3_GFS_v16,FV3_GFS_v16_coupled_nsstNoahmpUGWPv1"
elif [ $APP = "ATMAERO" ]; then
  export DO_GOCART="YES"
fi
```

parm/config/config.base.emc.dyn
```diff
@@ -227,11 +233,77 @@ export lobsdiag_forenkf=".true."
...
+
+# Temporarily override settings for aerosol coupling
+if [ $cplgocart = ".true." ]; then
+  export cplflx=".false."
+  export cplice=".false."
+  export cplocn=".false."
+  export cplwav=".false."
+  export cplwav2atm=".false."
+fi
+
```
aerorahul:
```suggestion
# Coupling settings

export cpl=".true."
export cplflx=".true."
export cplice=".false."
export cplgocart=".false."
export cplwav=".false."
export cplwav2atm=".false."

if [ $DO_CICE = "YES" ]; then
  export cplice=".true."
fi
if [ $DO_GOCART = "YES" ]; then
  export cplgocart=".true."
fi
if [ $DO_WAVE = "YES" ]; then
  export cplwav=".true."
  export cplwav2atm=".true."
fi

# Temporarily override settings for aerosol coupling
if [ $cplgocart = ".true." ]; then
  export cplflx=".false."
  export cplice=".false."
  export cplocn=".false."
  export cplwav=".false."
  export cplwav2atm=".false."
fi

```

parm/config/config.fcst
```diff
@@ -133,13 +172,45 @@ elif [ $imp_physics -eq 11 ]; then # GFDL
...
+export SKEB_LSCALE=500.E3,1000.E3,2000.E3,2000.E3,2000.E3
+export SKEBNORM=1
+
+export SPPT=0.8,0.4,0.2,0.08,0.04
+export iseed_sppt=$ISEED_SPPT
+export SPPT_TAU=2.16E4,2.592E5,2.592E6,7.776E6,3.1536E7
+export SPPT_LSCALE=500.E3,1000.E3,2000.E3,2000.E3,2000.E3
+export sppt_logit=".true."
+export sppt_sfclimit=".true."
+export use_zmtnblck=".true."
```
aerorahul: Aren't these relevant to ensemble forecasts?
So, shouldn't they be in `config.efcs`?

parm/config/config.fcst
```diff
@@ -276,6 +349,7 @@ fi
 export REMAPSH="$HOMEgfs/ush/fv3gfs_remap.sh"
 export master_grid="0p25deg"                   # 1deg 0p5deg 0p25deg 0p125deg etc
 export npe_remap=$((npe_fcst < 240 ? npe_fcst : 240))
+export nth_remap=2
```
aerorahul: Is anyone useing `regrid_nemsio` and `nc2nemsio` anymore?

parm/config/config.ocnpost
```diff
@@ -0,0 +1,14 @@
...
+########## config.ocnpost ##########
+
+echo "BEGIN: config.ocnpost"
+
+# Convert nemsio files to grib files using post job
+#-------------------------------------------
+
+# No. of concurrent post jobs [0 implies sequential]
+export NPOSTGRP=1
+export ncks="/apps/intel-2020/nco-4.8.1/bin/ncks"
```
aerorahul: This is hardwired for Orion.
Should be handled by a modulefile and not in `config.ocnpost`

aerorahul (changes requested):
First cut at the review.
Some questions.

parm/config/config.base.emc.dyn
```diff
@@ -227,11 +233,77 @@ export lobsdiag_forenkf=".true."
 export DO_GLDAS="YES"
 export gldas_cyc=00
 
-# run wave component
-export DO_WAVE="NO"
+# APP settings
+export APP="ATM"
+
+if [ $APP = "ATM" ]; then
```
WalterKolczynski-NOAA: This would be better as a case block rather than a bunch of elifs. Otherwise looks fine.

parm/config/config.base.emc.dyn
```diff
@@ -227,11 +233,77 @@ export lobsdiag_forenkf=".true."
 export DO_GLDAS="YES"
 export gldas_cyc=00
 
-# run wave component
-export DO_WAVE="NO"
+# APP settings
+export APP="ATM"
```
WalterKolczynski-NOAA, line 234: How does APP get set? Is this where a user would modified it?

parm/config/config.base.emc.dyn
```diff
@@ -227,11 +233,77 @@ export lobsdiag_forenkf=".true."
...
+  export DO_CICE="NO"
+  export DO_GOCART="YES"
+  export CCPP_SUITE="FV3_GFS_v16,FV3_GFS_v16_ugwpv1"
+fi
+
 if [[ "$SDATE" -lt "2019020100" ]]; then DO_WAVE="NO" ; fi # no rtofs in GDA
 export WAVE_CDUMP="both"
 
+# Coupling settings
+export cpl=".true."
```
WalterKolczynski-NOAA: cpl should be .false. until one of the coupling things is turned on.

parm/config/config.base.emc.dyn
```diff
@@ -227,11 +233,77 @@ export lobsdiag_forenkf=".true."
...
+if [ $DO_WAVE = "YES" ]; then
+  export cplwav=".true."
+  export cplwav2atm=".true."
+else
+  export cplwav=".false."
+  export cplwav2atm=".false."
+fi
+
+# Temporarily override settings for aerosol coupling
+if [ $cplgocart = ".true." ]; then
```
WalterKolczynski-NOAA: This block is probably no longer needed, since the switch on the APP already enforces this.

parm/config/config.base.emc.dyn
```diff
@@ -227,11 +233,77 @@ export lobsdiag_forenkf=".true."
...
+  export cplice=".false."
+fi
+if [ $DO_GOCART = "YES" ]; then
+  export cplgocart=".true."
+else
+  export cplgocart=".false."
+fi
+if [ $DO_WAVE = "YES" ]; then
+  export cplwav=".true."
+  export cplwav2atm=".true."
```
JessicaMeixner-NOAA: cplwav2atm is not a "DO_WAVE" yes/no option, we can have DO_WAVE= true but cplwave2atm false

parm/config/config.fcst
```diff
@@ -133,13 +172,45 @@ elif [ $imp_physics -eq 11 ]; then # GFDL
...
+
+export min_lakeice="0.15"
+export min_seaice="1e-11"
+
+export FSICL="99999"
+
+# Stochastic physics parameters
+# nam_stochy section
+export SEEDLET="10"
+export DO_SKEB="YES"
```
WalterKolczynski-NOAA: Stochastic physics should be off until we start dealing with ensembles.

parm/config/config.base.emc.dyn
```diff
@@ -227,11 +233,77 @@ export lobsdiag_forenkf=".true."
...
+fi
+
 if [[ "$SDATE" -lt "2019020100" ]]; then DO_WAVE="NO" ; fi # no rtofs in GDA
 export WAVE_CDUMP="both"
 
+# Coupling settings
+export cpl=".true."
+export cplflx=".true."
+if [ $DO_CICE = "YES" ]; then
+  export cplice=".true."
```
JessicaMeixner-NOAA: We have the "DO_WAVE" because of choices that were made in f/c-c versus what was in develop.  But DO_<comp> for the other components seem like new variables that are not in f/c-c (at least for what I quickly grep'd maybe I missed something?). If they don't exist already in f/c-c, should we not add them here?

parm/config/config.fcst
```diff
@@ -255,14 +325,17 @@ elif [[ "$CDUMP" == "gfs" ]] ; then # GFS cycle specific parameters
...
 
-
 fi
 
-    ## Debug load balancing
-    # export KEEPDATA="YES"
-    # export ESMF_RUNTIME_PROFILE=ON
-    # export ESMF_RUNTIME_PROFILE_OUTPUT=SUMMARY
-
+if [ $cplgocart = .true. ]; then # temporary settings for aerosol coupling
```
WalterKolczynski-NOAA: Shouldn't this use DO_GOCART instead, per previous conversations?

parm/config/config.fcst
```diff
@@ -255,14 +325,17 @@ elif [[ "$CDUMP" == "gfs" ]] ; then # GFS cycle specific parameters
...
+    export DIAG_TABLE="$HOMEgfs/parm/parm_fv3diag/diag_table_aer"
+    export FIELD_TABLE="$HOMEgfs/parm/parm_fv3diag/field_table_aer"
+    export CHM_CFGDIR="not_set"
+    export CHM_INPDIR="not_set"
+    export dnats="$(( $dnats + 0 ))"
+elif [[ $cpl == .true. ]] ; then # coupled model
```
WalterKolczynski-NOAA: Same here, do we need a DO_COUPLED variable?

parm/config/config.resources
```diff
@@ -47,16 +48,16 @@ if [ $step = "prep" -o $step = "prepbufr" ]; then
 
 elif [ $step = "waveinit" ]; then
 
-    export wtime_waveinit="00:10:00"
+    export wtime_waveinit="00:15:00"
```
JessicaMeixner-NOAA: This could be left at the 10min?

parm/config/config.fv3ic
```diff
@@ -1,19 +1,13 @@
...
-export wtime_fv3ic="00:30:00"
-export npe_fv3ic=1
-export npe_node_fv3ic=1
-export nth_fv3ic=${NTHREADS_CHGRES:-24}
-if [ $machine = HERA ]; then
-  export npe_fv3ic=4
-  export npe_node_fv3ic=4
-  export nth_fv3ic=1
-fi 
+export ORIGIN_ROOT=${BASE_CPLIC:-"/work/noaa/global/wkolczyn/noscrub/global-workflow/IC"}
```
WalterKolczynski-NOAA: I'd prefer the default location to be the one on Hera since that is used for the prototypes, though I know we are working on Orion for this sprint.

parm/config/config.resources
```diff
@@ -47,16 +48,16 @@ if [ $step = "prep" -o $step = "prepbufr" ]; then
...
     export npe_waveinit=12
     export nth_waveinit=1
     export npe_node_waveinit=$(echo "$npe_node_max / $nth_waveinit" | bc)
     export NTASKS=${npe_waveinit}
 
 elif [ $step = "waveprep" ]; then
 
-    export wtime_waveprep="00:30:00"
-    export npe_waveprep=65
+    export wtime_waveprep="00:15:00"
```
JessicaMeixner-NOAA: Hmm... so this is where we have different needs for a GFSv16 wave configuration versus s2s, and actually for s2s this could be 1.

parm/config/config.resources
```diff
@@ -71,8 +72,8 @@ elif [ $step = "wavepostsbs" ]; then
 
 elif [ $step = "wavepostbndpnt" ]; then
 
-    export wtime_wavepostbndpnt="02:00:00"
-    export npe_wavepostbndpnt=280
+    export wtime_wavepostbndpnt="01:00:00"
```
JessicaMeixner-NOAA: We do not use this job at all for s2s yet, so why is this being changed here?

parm/config/config.wave
```diff
@@ -56,6 +62,10 @@ export FHOUT_WAV=${FHOUT_WAV:-3}
 export FHMAX_HF_WAV=${FHMAX_HF_WAV:-120}
 export FHOUT_HF_WAV=${FHOUT_HF_WAV:-1}
 
+# Output stride
+export WAV_WND_HOUR_INC=1 # This value should match with the one used in
```
JessicaMeixner-NOAA: ? Is this from f/c-c?

parm/config/config.wave
```diff
@@ -64,19 +74,24 @@ export FHINCP_WAV=`expr $DTPNT_WAV / 3600`
 # Selected output parameters (gridded)
 export OUTPARS_WAV="WND HS FP DP PHS PTP PDIR"
 
+# Options for point output (switch on/off boundary point output)
+# export DOIBP_WAV='NO'
+
+# Number of cycles to look back for restart files
+export nback=
```
JessicaMeixner-NOAA: Is this from f/c-c?

parm/config/config.wave
```diff
@@ -64,19 +74,24 @@ export FHINCP_WAV=`expr $DTPNT_WAV / 3600`
 # Selected output parameters (gridded)
 export OUTPARS_WAV="WND HS FP DP PHS PTP PDIR"
 
+# Options for point output (switch on/off boundary point output)
```
JessicaMeixner-NOAA: Where is this coming from?

WalterKolczynski-NOAA (changes requested):

parm/config/config.waveprep
```diff
@@ -20,20 +20,24 @@ export POFILETYPE=1     # POFILETYPE=1 one point file per output step
 # Unified output T or F
 export FUNIPNT='T'
 # Output server type (see ww3_multi.inp in WW3 repo)
-export IOSRV='3'
+export IOSRV='1'
```
JessicaMeixner-NOAA: Why is this being changed?

parm/config/config.waveprep
```diff
@@ -20,20 +20,24 @@ export POFILETYPE=1     # POFILETYPE=1 one point file per output step
...
 
 # Wind interval for standalone file-based runs
 # Output stride
 export WAV_WND_HOUR_INC=1      # This value should match with the one used in
                          # the wind update script
 # Intake currents settings
 export WAV_CUR_DT=${WAV_CUR_DT:-3}
-export WAV_CUR_HF_DT=${WAV_CUR_HF_DT:-3} # Defaults to 3h for GEFSv12
-export WAV_CUR_HF_FH=${WAV_CUR_HF_FH:-0} # Constant DT for GFSv16 from getgo
+export WAV_CUR_HF_DT=${WAV_CUR_HF_DT:-1}
```
JessicaMeixner-NOAA: Why is this being changed?

parm/config/config.base.emc.dyn
WalterKolczynski-NOAA: Yes. Rahul doesn't want us using namelist parameters for script control.

JessicaMeixner-NOAA: Okay, fair enough, then why not DO_ICE, DO_OCN instead of specific component names,  After all DO_WAVE is not DO_WW3?

parm/config/config.base.emc.dyn
KateFriedman-NOAA: Took as-is from feature/coupled-crow example configs, happy to change if others agree. I assume changes will need to be made at the script level too if so. @WalterKolczynski-NOAA @JessicaMeixner-NOAA thoughts?

parm/config/config.base.emc.dyn
KateFriedman-NOAA: @WalterKolczynski-NOAA @JessicaMeixner-NOAA please confirm if these are only needed in the forecast job(s), if so I will move these two variables to config.fcst. Thanks!

parm/config/config.base.emc.dyn
JessicaMeixner-NOAA: We should add this to a list somewhere to get fixed. No reason this can't or shouldn't be changed/cleaned up.

parm/config/config.base.emc.dyn
JessicaMeixner-NOAA: Only needed for forecast, and esmf_profile should probably default to false (not saying it'snot true in f/c-c but probably just an oversight).

WalterKolczynski-NOAA: Yes, these only matter when the model is run (so I guess for EnKF as well). The first controls whether the individual PET log files are written into the run directory. The second controls whether the models are profiled to identify bottlenecks in the PET balance. They should probably both be off for ops.

parm/config/config.base.emc.dyn
KateFriedman-NOAA: A user could modify here and/or this could be added to the variables in config.base that the setup script arguments override. The default could be "ATM" but a user provided flag (e.g. --app) given to setup_expt*py could override. Thoughts?

parm/config/config.ocnpost
JessicaMeixner-NOAA: There's a nicks for each platform in for example: https://github.com/NOAA-EMC/global-workflow/blob/feature/coupled-crow/workflow/platforms/orion.yaml#L42 or https://github.com/NOAA-EMC/global-workflow/blob/feature/coupled-crow/workflow/platforms/hera.yaml#L41

parm/config/config.base.emc.dyn
KateFriedman-NOAA: By "one of the coupling things" do you mean the other cpl* variables?

parm/config/config.fv3ic
JessicaMeixner-NOAA: There are variables for this in each platform file in f/c-c, for example here: https://github.com/NOAA-EMC/global-workflow/blob/feature/coupled-crow/workflow/platforms/hera.yaml#L39

parm/config/config.base.emc.dyn
JessicaMeixner-NOAA: yes, cpl is only true if cplwav or cplflx, etc is true. For example this is calculated in crow here: https://github.com/NOAA-EMC/global-workflow/blob/feature/coupled-crow/workflow/defaults/settings.yaml#L57-L66

parm/config/config.ocnpost
WalterKolczynski-NOAA: The nco module in hpc-stack adds the bin directory to the path, so we should be able to just set this to "ncks". Or use $(which ncks) if we really want the full path.

parm/config/config.base.emc.dyn
KateFriedman-NOAA: The cplflx setting is the only one that isn't .false. and wouldn't be set to .false. if I remove that if-block. Otherwise, yes it can go. Should cplflx=.false. for now?

```# Coupling settings

export cpl=".true."
export cplflx=".true."
export cplice=".false."
export cplgocart=".false."
export cplwav=".false."
export cplwav2atm=".false."

if [ $DO_CICE = "YES" ]; then
  export cplice=".true."
fi
if [ $DO_GOCART = "YES" ]; then
  export cplgocart=".true."
fi
if [ $DO_WAVE = "YES" ]; then
  export cplwav=".true."
  export cplwav2atm=".true."
fi

# Temporarily override settings for aerosol coupling
if [ $cplgocart = ".true." ]; then
  export cplflx=".false."
  export cplice=".false."
  export cplocn=".false."
  export cplwav=".false."
  export cplwav2atm=".false."
fi
```

parm/config/config.base.emc.dyn
JessicaMeixner-NOAA: cplwav2atm right now is more appropriately set by APP=ATMW (more/less GFSv16), then cplwav2atm=false, but if APP=S2SW then cplwav2atm=true.

parm/config/config.base.emc.dyn
```diff
@@ -227,11 +233,66 @@ export lobsdiag_forenkf=".true."
...
+if [ $cplgocart = ".true." ]; then
+  export cplflx=".false."
+  export cplice=".false."
+  export cplocn=".false."
+  export cplwav=".false."
+  export cplwav2atm=".false."
+fi
+
+export nems_temp="cpld_wave"
+export nems_temp_cold="atm"
```
JessicaMeixner-NOAA: Hmm... aren't these variables that kind of depend on the app? And then this is also something that could tell you if cplwav2atm=true/false.

parm/config/config.fcst
KateFriedman-NOAA: Updated to set them to NO. Thanks!

parm/config/config.fcst
KateFriedman-NOAA: Yes, have now changed it to DO_GOCART=YES. Thanks!

parm/config/config.base.emc.dyn
KateFriedman-NOAA: Removed that line, it will default to .false. in config.base now. Thanks!

parm/config/config.resources
KateFriedman-NOAA: Changed back to 10 mins. Thanks!

parm/config/config.wave
KateFriedman-NOAA: Yes, it was in the example configs I was given, see: /work/noaa/global/wkolczyn/save/global-workflow/expdir/example/config.wave

Let me know if we don't want this, thanks!

parm/config/config.wave
KateFriedman-NOAA: Yes, see example config.wave: /work/noaa/global/wkolczyn/save/global-workflow/expdir/example/config.wave

parm/config/config.wave
KateFriedman-NOAA: See example config.wave I was provided: /work/noaa/global/wkolczyn/save/global-workflow/expdir/example/config.wave

parm/config/config.waveprep
KateFriedman-NOAA: Example config.waveprep has it set to 1, see: /work/noaa/global/wkolczyn/save/global-workflow/expdir/example/config.waveprep
If it should remain set to 3 let me know, thanks!

parm/config/config.wave
JessicaMeixner-NOAA: Let me look closer at these, I think these are probably variables that didn't get cleaned up.  For the most part for waves, I thought we were sticking to what was in develop.

parm/config/config.waveprep
KateFriedman-NOAA: The example config.wave (/work/noaa/global/wkolczyn/save/global-workflow/expdir/example/config.wave) has `export WAV_CUR_HF_DT=1`. In the develop branch we set that variable in config.waveprep so I changed it in config.waveprep to match the example config.wave setting. Let me know what it should be, thanks!

parm/config/config.base.emc.dyn
KateFriedman-NOAA: These are new to me so I need you and @WalterKolczynski-NOAA to tell me what/where they should be. Thanks!

parm/config/config.base.emc.dyn
WalterKolczynski-NOAA: Yeah, along with the DO_* and a bunch of other settings, everything needs a few consistency updates.

parm/config/config.ocnpost
WalterKolczynski-NOAA: Rahul is right, this should be covered by the module, not a path we set for each platform.

parm/config/config.base.emc.dyn
WalterKolczynski-NOAA: Let's just get it in for now and worry about making it more user-friendly later.

parm/config/config.base.emc.dyn
WalterKolczynski-NOAA: cplflx should only be turned on if one of the coupling components that need it are on (MOM6/CICE, not sure about WW3)

JessicaMeixner-NOAA: cplflx is for ocn/ice, wave is seperate.

WalterKolczynski-NOAA: Here's what my aerosol run has for settings:

in input.nml:
```
cplwav       = .false.
cplchm       = .true.
cplflx       = .false.
cplwav2atm   = .false.
```

in model_configure:
`cpl:                     .true.`

parm/config/config.wave
```diff
@@ -24,6 +29,7 @@ export CDUMPRSTwave="gdas"
 export waveGRD='gnh_10m aoc_9km gsh_15m'
 export waveGRDN='1 2 3' # gridnumber for ww3_multi
 export waveGRDG='10 20 30' # gridgroup for ww3_multi
+export USE_WAV_RMP='NO' #yes/no rmp grid remapping pre-processed coefficients
```
JessicaMeixner-NOAA: Since all other variables for waves are still for the GFSv16/develop case, this variable should be YES

parm/config/config.wave
JessicaMeixner-NOAA: Why don't we just keep this variable in config.waveprep like it is in develop and remove from config.wave.  I don't remember why f/c-c moved this, but keeping it in waveprep and not doubly defining it is probably the way to go.

parm/config/config.wave
JessicaMeixner-NOAA: Let's remove this as it's commented out, let's just not move it in.

parm/config/config.wave
JessicaMeixner-NOAA: Looks like this is something that was brought in from develop at some point but then was removed but not removed from f/c-c.  This variable is not used in develop or f/c-c so it can be removed.

parm/config/config.waveprep
JessicaMeixner-NOAA: Since we were keeping everything as this is the develop/gfsv16 case and not updating to use the "coupled"/s2s case, this should be reverted to what's in develop.

parm/config/config.waveprep
JessicaMeixner-NOAA: Please revert this and the line below to what is in develop as we are using the develop/gfsv16 case

parm/config/config.base.emc.dyn
KateFriedman-NOAA: I have moved these two lines to config.fcst and set both to .false. for now. Thanks!

parm/config/config.fcst
KateFriedman-NOAA: config.efcs has stochastic physics settings already, see these lines: https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/config/config.efcs#L28:L45

parm/config/config.fcst
KateFriedman-NOAA: I had a user contact me recently about fv3nc2nemsio but that's the only person I know for sure is using one of those. I'll remove nth_remap=2 to bring it back to how config.fcst is in develop, that variable should be in config.resources if needed anyway.

parm/config/config.ocnpost
KateFriedman-NOAA: So remove the ncks line completely from config.ocnpost?

parm/config/config.fcst
WalterKolczynski-NOAA: We don't use nemsio any more, so I don't see a reason to maintain it. If others want it, they need to either create their own or use an old version of the workflow.

parm/config/config.ocnpost
WalterKolczynski-NOAA: Set it to ncks. Otherwise you'll have to replace the $ncks wherever it is used in post.

parm/config/config.base.emc.dyn
KateFriedman-NOAA: How about the adjustment below? What should cplchem be by default?

```
# Coupling settings

export cplflx=".false."
export cplflx=".true."
export cplice=".false."
export cplgocart=".false."
export cplwav=".false."
export cplwav2atm=".false."

if [ $DO_CICE = "YES" ]; then
  export cplice=".true."
fi
if [ $DO_GOCART = "YES" ]; then
  export cplgocart=".true."
fi
if [ $DO_WAVE = "YES" ]; then
  export cplwav=".true."
fi

if [ $cplflx = ".true." -o $cplice = ".true." -o $cplwav = ".true." -o $cplchem = ".true." ]; then
  export cpl=".true."
fi
```

parm/config/config.base.emc.dyn
KateFriedman-NOAA: Sounds like I should add DO_OCN and this if-block?

```
export DO_OCN="NO" #default
export cplflx=".false." #default

if [ $DO_CICE = "YES" -o $DO_OCN = "YES" ]; then
  export cplflx=".true."
fi
```

parm/config/config.fcst
WalterKolczynski-NOAA: That's not to say this needs to be addressed now. We can remove it later.

parm/config/config.fv3ic
KateFriedman-NOAA: The orion.yaml has `BASE_CPLIC: "/work/noaa/marine/Partha.Bhattacharjee/IC_Dir"`. Should I add that or stick with Walter's for the Orion path?

I notice they are all personal account areas...we should move these files to a less personal area on all machines at some point...maybe a group account?

parm/config/config.base.emc.dyn
KateFriedman-NOAA: Roger that, thanks! Have set it to default after moving to config.fcst.

parm/config/config.wave
KateFriedman-NOAA: Will do, thanks!

parm/config/config.wave
KateFriedman-NOAA: Will remove, thanks!

parm/config/config.wave
KateFriedman-NOAA: Roger that, will remove, thanks!

parm/config/config.fv3ic
JessicaMeixner-NOAA: Hera is the climate group account.  We don't have group accounts on orion.  Or do we now? We only actively run things on hera/orion.  Walter ported to DELL but prototypes aren't actually being run there really.

parm/config/config.waveprep
KateFriedman-NOAA: Changing back to 3, thanks!

parm/config/config.waveprep
KateFriedman-NOAA: Reverting, thanks!

parm/config/config.wave
KateFriedman-NOAA: Am changing this to be set to YES, thanks!

parm/config/config.ocnpost
KateFriedman-NOAA: `export ncks="ncks"`?

aerorahul: IMO, this line should disappear from config files altogether.

Wherever it is used, it should be resolved as `$(which ncks)` and then used as appropriate.
If this is used in ocnpost, we have the chance to clean than up right now.  

I suggest removing this line completely.

parm/config/config.base.emc.dyn
aerorahul: `DO_ICE`, `DO_OCN`, `DO_WAVE`, `DO_AERO`, and so on.

parm/config/config.fcst
KateFriedman-NOAA: Will remove block.

parm/config/config.ocnpost
KateFriedman-NOAA: Will remove line.

parm/config/config.base.emc.dyn
KateFriedman-NOAA: Replacing `OCNFIXDIR` with `FIXreg2grb2`.

parm/config/config.fcst
KateFriedman-NOAA: Removing remap and regrid blocks in config.fcst.

parm/config/config.base.emc.dyn
JessicaMeixner-NOAA: If you replace that, are you also replacing it where the old variable is used in scripts?

KateFriedman-NOAA: Yes, that was something we discussed in the tag-up this morning. The same variable in the scripts will also get updated when we get to that part.

parm/config/config.fcst
KateFriedman-NOAA: Changed to `elif [[ $DO_COUPLED == "YES" ]] ; then # coupled model`

parm/config/config.base.emc.dyn
KateFriedman-NOAA: I moved the cpl variables into config.fcst.

parm/config/config.base.emc.dyn
KateFriedman-NOAA: Have renamed DO_* variables and reworked if-block (converted into case block).

parm/config/config.fv3ic
KateFriedman-NOAA: We don't have a group account on Orion at the moment but I was told that they are coming. For now I am holding glopara group files in my personal account and will move it all to the group account when available. We have emc.glopara/glopara on WCOSS-Dell, Hera, and Jet.

I have added an if-block based on machine to set BASE_CPLIC for now.

parm/config/config.base.emc.dyn
JessicaMeixner-NOAA: nems_temp_cold is only used in jobs/JGLOBAL_FORECAST_MEDCOLD which is not a job that is used anymore and probably shouldn't be included when we merge things into develop. Therefore I'd recommend just not adding this variable.  

for nems_temp here is what it should be defined on based on the $APP: 

case "${APP}" in
   ATM)
    nems_temp='atm'
  ATMW) 
     nems_temp='leapfrog_atm_wav'
   S2S)
     nems_temp='cpld'
   S2SW)
     nems_temp='cpld_wave'
   ATMAERO)
         nems_temp='atm_aer'

parm/config/config.resources
JessicaMeixner-NOAA: This still needs to be reverted to what's in develop.

parm/config/config.resources
KateFriedman-NOAA: Will revert back to develop settings, thanks!

parm/config/config.resources
KateFriedman-NOAA: Will revert, thanks!

WalterKolczynski-NOAA (approved):

KateFriedman-NOAA:
Final edits have been made. Please perform final review of incoming changes, thanks!

I tested setup_expt_fcstonly.py and setup_workflow_fcstonly.py on Orion successfully with final config versions:

```
Orion-login-2[201] /work/noaa/global/kfriedma/git/fork_feature-coupled_sprint/ush/rocoto$ ./setup_expt_fcstonly.py --pslot fcsdev2 --res 192 --idate 2020090118 --edate 2020090118 --comrot /work/noaa/stmp/kfriedma/comrot --expdir /work/noaa/global/kfriedma/expdir --gfs_cyc 4 --start cold --configdir /work/noaa/global/kfriedma/git/fork_feature-coupled_sprint/parm/config

COMROT already exists in /work/noaa/stmp/kfriedma/comrot/fcsdev2

Do you wish to over-write COMROT [y/N]: y

EXPDIR already exists in /work/noaa/global/kfriedma/expdir/fcsdev2

Do you wish to over-write EXPDIR [y/N]: y

SDATE = 2020-09-01 18:00:00
EDATE = 2020-09-01 18:00:00

EDITED:  /work/noaa/global/kfriedma/expdir/fcsdev2/config.base as per user input.
DEFAULT: /work/noaa/global/kfriedma/expdir/fcsdev2/config.base.default is for reference only.
Please verify and delete the default file before proceeding.

Orion-login-2[202] /work/noaa/global/kfriedma/git/fork_feature-coupled_sprint/ush/rocoto$ ./setup_workflow_fcstonly.py --expdir /work/noaa/global/kfriedma/expdir/fcsdev2
sourcing config.getic
sourcing config.init
sourcing config.waveinit
sourcing config.waveprep
sourcing config.fcst
sourcing config.post
sourcing config.wavepostsbs
sourcing config.wavepostbndpnt
sourcing config.wavepostbndpntbll
sourcing config.wavepostpnt
sourcing config.wavegempak
sourcing config.waveawipsbulls
sourcing config.waveawipsgridded
sourcing config.wafs
sourcing config.wafsgrib2
sourcing config.wafsblending
sourcing config.wafsgcip
sourcing config.wafsgrib20p25
sourcing config.wafsblending0p25
sourcing config.postsnd
sourcing config.gempak
sourcing config.awips
sourcing config.vrfy
sourcing config.metp
sourcing config.arch
```

parm/config/config.fcst
```diff
@@ -255,31 +334,18 @@ elif [[ "$CDUMP" == "gfs" ]] ; then # GFS cycle specific parameters
...
-    export REGRID_NEMSIO_TBL="$HOMEgfs/parm/parm_fv3diag/variable_table_da.txt"
-else
-    export REGRID_NEMSIO_TBL="$HOMEgfs/parm/parm_fv3diag/variable_table_da_nonsst.txt"
+if [ $DO_AERO = "YES" ]; then # temporary settings for aerosol coupling
+    export DIAG_TABLE="$HOMEgfs/parm/parm_fv3diag/diag_table_aer"
+    export FIELD_TABLE="$HOMEgfs/parm/parm_fv3diag/field_table_aer"
+    export CHM_CFGDIR="not_set"
+    export CHM_INPDIR="not_set"
+    export dnats="$(( $dnats + 0 ))"
+elif [[ $DO_COUPLED == "YES" ]] ; then # coupled model
```
aerorahul: I know it works, just making it consistent to the `if` test above on L339
```suggestion
elif [[ $DO_COUPLED = "YES" ]] ; then # coupled model
```

aerorahul (approved):
looks good.  just one nit pick comment.  Feel free to commit or reject.

KateFriedman-NOAA:
Close #368

Issue #379, opened 2021-07-27
Update to MERRA2


WalterKolczynski-NOAA:
Update the aerosol inputs to MERRA2 in the coupled branch.

PR #380, opened 2021-07-27
Update to MERRA2 aerosol climatology

WalterKolczynski-NOAA:
Adds the capability to use MERRA2 aerosol climatology and makes it
the default.
    
An addition to the diag table is required. Rather than continue the
proliferation of diag tables to produce one specifically for coupled
and MERRA2, the existing diag table from the MERRA2 update in develop
is pared down to just the additional fields introduced. This is then
appended to the main diag table if necessary.

Also corrected a related issue in the CROW forecast config. While
IAER and several other similar settings were already present in the
schema, the values were hard-coded in the forecast config instead of
using the value set. The script now correctly uses the values set by
the configuration system. The default for IAER is changed to 1011 for
MERRA2.

Refs: #379


rmontuoro (approved):
Looks good.

JessicaMeixner-NOAA:
My test on hera succeeded.  The forecast run directory is here: /scratch2/NCEPDEV/climate/Jessica.Meixner/p7merra/test01/DATAROOT/test01/2013040100/gfs/fcst.9115 @AnningCheng-NOAA if you'd like to use that to double check the input.nml and run directory are correct for MERRA2 aerosol climatology now.

AnningCheng-NOAA:
The results look good. Approve this pull request.

JessicaMeixner-NOAA (approved):

PR #388, opened 2021-07-29
Update namelist settings for prototype 7c

### LINKED ISSUE #386: P7: Update input nameless
All namelist values need to be checked and confirmed that the correct options are being used for all components. 
Tracking some of this being done in the google doc [here](https://docs.google.com/document/d/1KxrmA74qqu6jMkjQqREbQfuApSksoSozSLKdlv_oGuE/edit?usp=sharing). 

The namelist that we know need updating for now are: 
* LANDICE should be false 
* Check uGWD settings 
* The following for 
ialb=2
iems=2
lsm=2 
iopt_dveg=4
iopt_crs=2
iopt_rad=3
iopt_alb=1
iopt_stc=3

* Use recommended CA settings from Lisa: These are the settings we recommend for the CA, under the namelist  block &gfs_physics_nml. Note that iseed_ca can be set to the ensemble member number if used in an ensemble forecast. 
  do_ca          = .True.
  ca_sgs         = .True.
  nca            = 1
  scells         = 2600
  tlives         = 1800
  nseed          = 1
  nfracseed      = 0.5
  rcell          = 0.72
  ca_trigger     = .True.
  nspinup        = 1
  iseed_ca       = 12345

There might be more as we continue to check the settings for the various different physics options and all components.

WalterKolczynski-NOAA:
Several namelist settings are updated for prototype 7c.

A large number of the settings are tied to Noah-MP, so those settings
are now set alongside the lsm, but mostly accept overrides.

Another group of settings are related to cellular automata and are
only added to the namelist if a new setting, DO_CA, is set to YES.
The setting defaults to on (YES).

Two other settings, IALB and IEMS, are not confirmed to be tied to
other settings, so the defaults for these are just straightfowardly
changed to the new values.

Closes: #386

### DIFF
--- a/ush/forecast_postdet.sh
+++ b/ush/forecast_postdet.sh
@@ -223,8 +223,34 @@ FV3_GFS_postdet(){
 	# Scan suite file to determine whether it uses Noah-MP
 	if [ $(grep noahmpdrv ${_suite_file} | wc -l ) -gt 0 ]; then
 		lsm="2"
+		landice=".false."
+		iopt_dveg=${iopt_dveg:-"4"}
+		iopt_crs=${iopt_crs:-"2"}
+		iopt_btr=${iopt_btr:-"1"}
+		iopt_run=${iopt_run:-"1"}
+		iopt_sfc=${iopt_sfc:-"1"}
+		iopt_frz=${iopt_frz:-"1"}
+		iopt_inf=${iopt_inf:-"1"}
+		iopt_rad=${iopt_rad:-"3"}
+		iopt_alb=${iopt_alb:-"1"}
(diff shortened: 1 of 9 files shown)


WalterKolczynski-NOAA:
input.nml: https://gist.github.com/WalterKolczynski-NOAA/ad1ba47aa768a8b2808bbf3b1b6d6ec4

WalterKolczynski-NOAA:
input.nml gist updated

WalterKolczynski-NOAA:
input.nml gist updated again

JessicaMeixner-NOAA (approved):
Thanks @WalterKolczynski-NOAA

PR #384, opened 2021-07-28
JET update to enable fcst only run:

lgannoaa:
Changes to be committed:
	modified:   env/JET.env
	modified:   modulefiles/module_base.jet
	modified:   sorc/build_gfs_util.sh
	modified:   sorc/build_gldas.sh
	modified:   sorc/build_workflow_utils.sh
	modified:   sorc/link_fv3gfs.sh

PR #391, opened 2021-08-02
Feature/jet fcst only

lgannoaa:
This is a draft PR for running GFS fcst only with global-workflow on JET.


aerorahul:
@lgannoaa 
As mentioned in [this](https://github.com/NOAA-EMC/global-workflow/pull/384#issuecomment-888708951) comment, if there are syntax errors with the python scripts, please open a separate issue to document the errors and how to reproduce them.

Syntax errors in those scripts need to be handled urgently and separately.  The bugfix for syntax errors should not be combined with portability to Jet.

PR #393, opened 2021-08-03
Set seed for Cellular Automata

WalterKolczynski-NOAA:
The seed for the CA is now set in the same manner as the stochastic
physics components. Because the CA seed is only 32-bit (in contrast
to the 64-bit seed used for SP), a modulus is taken to ensure the
number stays in-bounds.

Refs: #386


lisa-bengtsson (approved):
These settings look good to me, thanks.

JessicaMeixner-NOAA (approved):
Approving based on Lisa's approval.

PR #395, opened 2021-08-03
coupled-crow: catch copy errors and bail out if ICs do not exist

JessicaMeixner-NOAA:
In the script for copying in ICs, the script originally did not fail even though ICs did not exist on orion.  This update makes sure that the ic job fails instead of waiting for the forecast job to fail for ICs that do not exist.

### DIFF
--- a/jobs/rocoto/coupled_ic.sh
+++ b/jobs/rocoto/coupled_ic.sh
@@ -53,18 +53,40 @@ fi
 
 # Setup ATM initial condition files
 cp -r $ORIGIN_ROOT/$CPL_ATMIC/$CDATE/$CDUMP/*  $ICSDIR/$CDATE/atmos/
+rc=$?
+if [[ $rc -ne 0 ]] ; then
+  echo "FATAL: Unable to copy $ORIGIN_ROOT/$CPL_ATMIC/$CDATE/$CDUMP/* to $ICSDIR/$CDATE/atmos/ (Error code $rc)"
+  exit $rc    
+fi
+
 
 # Setup Ocean IC files 
 cp -r $ORIGIN_ROOT/$CPL_OCNIC/$CDATE/ocn/$OCNRES/MOM*.nc  $ICSDIR/$CDATE/ocn/
+rc=$?
+if [[ $rc -ne 0 ]] ; then
+  echo "FATAL: Unable to copy $ORIGIN_ROOT/$CPL_OCNIC/$CDATE/ocn/$OCNRES/MOM*.nc to $ICSDIR/$CDATE/ocn/ (Error code $rc)"
+  exit $rc
+fi
 
 #Setup Ice IC files 
 cp $ORIGIN_ROOT/$CPL_ICEIC/$CDATE/ice/$ICERES/cice5_model_${ICERESdec}.res_$CDATE.nc $ICSDIR/$CDATE/ice/cice_model_${ICERESdec}.res_$CDATE.nc
+rc=$?
+if [[ $rc -ne 0 ]] ; then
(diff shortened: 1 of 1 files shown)


jobs/rocoto/coupled_ic.sh
```diff
@@ -53,18 +53,40 @@ fi
 
 # Setup ATM initial condition files
 cp -r $ORIGIN_ROOT/$CPL_ATMIC/$CDATE/$CDUMP/*  $ICSDIR/$CDATE/atmos/
+rc=$?
+if [[ $rc -ne 0 ]] ; then
```
WalterKolczynski-NOAA, line 58: None of these will catch errors if a single file in the group is missing, only if they are all missing. Probably sufficient for now though.

WalterKolczynski-NOAA (approved):

jobs/rocoto/coupled_ic.sh
JessicaMeixner-NOAA: Fair enough, it did catch the current error on orion. Suggestions for updates to improve beyond this are welcome.

WalterKolczynski-NOAA: The best way would be to iterate over the files individually after copy to make sure they are there. The slightly lazier way would be to count the files after copy to make sure the right number are there. Either is a bit annoying to maintain.

I think this is probably good enough for now, as the vast majority of instances will be the whole directory is missing/empty, and this will catch those. We can think about more robust checking as build out the new unified system.

aerorahul: May I suggest one change.

Instead of exiting right after a failed copy, initialize a counter `err_count=0` at the beginning and increment the counter e.g.`err_count` with `$rc`.

Let the script proceed till after all the copies are done.

At the end if `err_count > 0`, exit with a message that something failed to copy.

That way, the user gets a full view of what else is going to be missing, in case they fix this error and move on.

WalterKolczynski-NOAA: Good suggestion. This is how build_all does it.

WalterKolczynski-NOAA (approved):

jobs/rocoto/coupled_ic.sh
JessicaMeixner-NOAA: I updated the script based on your comments.  Let me know what you think.

jobs/rocoto/coupled_ic.sh
```diff
@@ -53,18 +54,40 @@ fi
...
 
 if [ $cplwav = ".true." ]; then
   [[ ! -d $ICSDIR/$CDATE/wav ]] && mkdir -p $ICSDIR/$CDATE/wav
   for grdID in $waveGRD
   do
     cp $ORIGIN_ROOT/$CPL_WAVIC/$CDATE/wav/$grdID/*restart.$grdID $ICSDIR/$CDATE/wav/
+    rc=$?
+    if [[ $rc -ne 0 ]] ; then
+      echo "FATAL: Unable to copy $ORIGIN_ROOT/$CPL_WAVIC/$CDATE/wav/$grdID/*restart.$grdID to $ICSDIR/$CDATE/wav/ (Error code $rc)" 
+      exit $rc
```
aerorahul: I think this is a stale exit and can go.

JessicaMeixner-NOAA: agreed and fixed.

jobs/rocoto/coupled_ic.sh
aerorahul: looks good.

aerorahul (approved):
Script changes look good to me.
However, I did not test.

JessicaMeixner-NOAA:
Just re-tested with the updates and it still fails on orion as expected, but now at the end of the script.

WalterKolczynski-NOAA (approved):

PR #396, opened 2021-08-03
Update wave related scripts 

JessicaMeixner-NOAA:
This is pulling in the updates made on wave scripts in feature/coupled-crow into the feature/coupled_sprint branch.  The main part of these updates are moving templated file from the fix wave directory to a parm/wave directory, so that they can be more easily version controlled.  There are a few other updates such as updates to ww3_multi.inp for a recent update in WW3 and a few other updates that were needed to run feature/coupled-crow.   

This has not yet been tested.


WalterKolczynski-NOAA (approved):
Looks good.

Issue #413, opened 2021-08-17
Review and correct use of $DATA/$DATAROOT in each job


KateFriedman-NOAA:
Review how each job is using $DATA and/or $DATAROOT. Make sure each job is creating/using its own distinct rundir, which utilizes job id (pid). Correct/update as needed to make the DATA/DATAROOT usage consistent across jobs.

Related to issue #401

KateFriedman-NOAA:
Mostly resolved through recent updates to scripts and the addition of `jjob_header.sh`. Review scripts for any lingering cases and resolve. Close issue if none found.

KateFriedman-NOAA:
All JJOBS now invoke `ush/jjob_header.sh` utility script that creates a distinct `DATA` folder based on `DATAROOT` and `jobid`. Snippet from `ush/jjob_header.sh`:
```
##############################################
# make temp directory
##############################################
export DATA=${DATA:-"${DATAROOT}/${jobid}"}
```

Some special jobs (e.g. vrfy) also invoke `ush/jjob_header.sh` within their rocoto job scripts:
```
kate.friedman@clogin05:/lfs/h2/emc/global/noscrub/kate.friedman/git/develop/jobs/rocoto> grep jjob_header *
awips.sh:source "${HOMEgfs}/ush/jjob_header.sh" -e "awips" -c "base awips"
ocnpost.sh:source "${HOMEgfs}/ush/jjob_header.sh" -e "ocnpost" -c "base ocnpost"
prep.sh:source "${HOMEgfs}/ush/jjob_header.sh" -e "prep" -c "base prep"
vrfy.sh:source "${HOMEgfs}/ush/jjob_header.sh" -e "vrfy" -c "base vrfy"
```

Closing this as complete.

Issue #347, opened 2021-06-17
Update Orion support to include rstprod


KateFriedman-NOAA:
A rstprod policy is now on Orion and access to the group is expanding. Update workflow and components that use rstprod to support rstprod on Orion now.

- [x] GDA
- [x] global-workflow flags/variables (e.g. CHGRP_RSTPROD)
- [x] EMC_verif-global (internal settings and on-disk data files)

Will need new tag from @malloryprow for EMC_verif-global.

KateFriedman-NOAA:
@malloryprow Checking back with you on this. Do you have a new EMC_verif-global tag that supports rstprod on Orion? Is there anything you need from me for that? I'm going to make the workflow updates for this now. Thanks!

KateFriedman-NOAA:
Created branch for workflow updates: [feature/rstprod-orion](https://github.com/KateFriedman-NOAA/global-workflow/tree/feature/rstprod-orion)

malloryprow:
@KateFriedman-NOAA The tag verif_global_v2.2.1 has the rstprod support for Orion in it.

KateFriedman-NOAA:
@malloryprow Thanks for the new tag! The global-workflow develop branch is currently using the verif_global_v2.0.2 tag. Can you give me a short summary of all changes from v2.0.2 to v2.2.1? Also, what platforms was this tag tested on? Thanks!

malloryprow:
@KateFriedman-NOAA Yes!

- Support rstprod prepbufr files being used on Orion in grid-to-obs verification
- Support for running on S4 (stand alone and in global workflow)
- Enhancements and clean up of plotting scripts

KateFriedman-NOAA:
> @KateFriedman-NOAA Yes!
> 
> * Support rstprod prepbufr files being used on Orion in grid-to-obs verification
> * Support for running on S4 (stand alone and in global workflow)
> * Enhancements and clean up of plotting scripts

Excellent, thanks @malloryprow for this!

KateFriedman-NOAA:
@malloryprow I have run 4.5 cycles (C192C96L127) on WCOSS-Dell with the new EMC_verif-global tag. Please see my logs and output:

```
logs: /gpfs/dell3/ptmp/Kate.Friedman/comrot/issue347/logs/2020090*/gfsmetp*
output: /gpfs/dell2/emc/modeling/noscrub/Kate.Friedman/archive/metplus_data/by_VSDB/*/*/*/issue347
```

[...]

PR #414, opened 2021-08-17
Update vrfy/metp jobs to use jobid in their DATAROOT folders

### LINKED ISSUE #401: BUG: Fix race condition between vrfy and metp jobs
Discovered that when the vrfy and metp jobs are running at the same time and the metp jobs finish first (usually only during the first full cycle they are invoked) they remove a DATA folder that is incorrectly shared with the vrfy job. Only found because WCOSS-Dell was quiet enough to allow jobs to all start as soon as they were submitted, most users will not encounter this bug.

Easy fix is to change the metp jobs to not use "vrfy" in their DATA paths and instead use something else (e.g. "metp") in the DATA path. Change this line in jobs/rocoto/metp.sh:

from:
`export DATAROOT="$RUNDIR/$CDATE/$CDUMP/vrfy"`
to:
`export DATAROOT="$RUNDIR/$CDATE/$CDUMP/metp"`

https://github.com/NOAA-EMC/global-workflow/blob/develop/jobs/rocoto/metp.sh#L48

KateFriedman-NOAA:
This PR updates how the vrfy and metp* jobs set/use their $DATAROOT rundir folders:

- add pid and jobid variable settings to metp.sh and vrfy.sh (as other jobs do)
- update DATAROOT setting for both jobs to use jobid variable in folder name
- this update stops the vrfy and metp* jobs from all using a generic "vrfy" subfolder under $DATA for $DATAROOT
- the metp* jobs now also use distinct DATAROOTs to avoid race conditions between them
- further DATA/DATAROOT work will happen in issue #413
- tested successfully on WCOSS-Dell with vrfy/metp jobs running in parallel

Refs #401
Close #401

### DIFF
--- a/jobs/rocoto/metp.sh
+++ b/jobs/rocoto/metp.sh
@@ -44,8 +44,10 @@ status=$?
 export COMPONENT=${COMPONENT:-atmos}
 export VDATE="$(echo $($NDATE -${VRFYBACK_HRS} $CDATE) | cut -c1-8)"
 
+export pid=${pid:-$$}
+export jobid=${job}.${pid}
 export COMIN="$ROTDIR/$CDUMP.$PDY/$cyc/$COMPONENT"
-export DATAROOT="$RUNDIR/$CDATE/$CDUMP/vrfy"
+export DATAROOT="$RUNDIR/$CDATE/$CDUMP/metp.${jobid}"
 [[ -d $DATAROOT ]] && rm -rf $DATAROOT
 mkdir -p $DATAROOT
 
--- a/jobs/rocoto/vrfy.sh
+++ b/jobs/rocoto/vrfy.sh
@@ -43,8 +43,10 @@ export COMPONENT=${COMPONENT:-atmos}
 export CDATEm1=$($NDATE -24 $CDATE)
 export PDYm1=$(echo $CDATEm1 | cut -c1-8)
 
+export pid=${pid:-$$}
+export jobid=${job}.${pid}
 export COMIN="$ROTDIR/$CDUMP.$PDY/$cyc/$COMPONENT"
-export DATAROOT="$RUNDIR/$CDATE/$CDUMP/vrfy"
+export DATAROOT="$RUNDIR/$CDATE/$CDUMP/vrfy.${jobid}"
 [[ -d $DATAROOT ]] && rm -rf $DATAROOT
 mkdir -p $DATAROOT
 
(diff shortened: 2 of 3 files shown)


KateFriedman-NOAA:
Completing 4.5 cycle C192C96L127 test on WCOSS-Dell before submitting this PR after approval.

malloryprow:
Thanks for this work Kate! 

One question. On line 48, "jobid" gets set as "${job}.${pid}". I see how/where the variable "pid" is set (line 47), how/where does the variable "job" get set?

KateFriedman-NOAA:
> One question. On line 48, "jobid" gets set as "${job}.${pid}". I see how/where the variable "pid" is set (line 47), how/where does the variable "job" get set?

@malloryprow The machine-specific env files currently set the following:

```
export job=${LSB_JOBNAME:-$step}
export jobid=${job}.${LSB_JOBID:-$$}
```

The env file (e.g. env/WCOSS_DELL_P3.env) is sourced at the start of all jobs so both job and jobid get set there before reaching the DATAROOT setting part of the rocoto job scripts. We may change how we do this when working on issue #413.

FYI, I will have metp logs for you to check for me once I get another couple cycles completed in my test on WCOSS-Dell. The jobs appear to be working but I'd like your confirmation of that once the later cycles have run. Will share the path to those logs soon. Thanks!

KateFriedman-NOAA:
@malloryprow Please see the following metp logs and output from a 4.5 cycle C192C96L127 run on WCOSS-Dell:

logs: /gpfs/dell3/ptmp/Kate.Friedman/comrot/issue401/logs/2020090*/*metp*
metp output: /gpfs/dell2/emc/modeling/noscrub/Kate.Friedman/archive/metplus_data/by_VSDB/*/*/*/issue401

I'm still running the `verif_global_v2.0.2` tag in this test and am just testing a change to the DATAROOT paths. Please confirm you do not see any errors, especially any related to rundir. I will be testing the new EMC_verif-global tag in a separate test for issue #347. Thanks!

malloryprow:
@KateFriedman-NOAA Everything looks good!

KateFriedman-NOAA:
@malloryprow Thanks for checking! Please provide your approval of this PR through the review and I'll submit it. Thanks!

malloryprow (approved):
All good!

KateFriedman-NOAA:
Thanks @malloryprow !

Issue #417, opened 2021-08-17
Partition settings inconsistent


WalterKolczynski-NOAA:
The workflow generated on Orion and Jet have a separate partition definition for each job (PARTITION_<job>_GFS), but the actual task definitions for batch jobs use PARTITION_BATCH directly. This is confusing, particularly when manually editing XML (such as to have some jobs use debug).

Additionally, the PARTITION_BATCH is not defined at all for Hera. While it is not required for Hera, this makes the code more convoluted by checking for the machine beyond whether slurm is required, and also requires the manual addition of these to the task definition if you wish to use a non-standard partition.

PR #418, opened 2021-08-17
Standardize partition settings

WalterKolczynski-NOAA:
The partition settings in setup are updated so all slurm-based
machines are treated the same. Also fixes an issue where task-
specific partitions were created, but the task definition used
PARTITION_BATCH directly instead.

Refs: #417


WalterKolczynski-NOAA:
@KateFriedman-NOAA (and anyone else that is willing to help): can you help test this on the different machines and setups (cycled/fcst-only). The workflows look right, but I haven't actually tried running.

KateFriedman-NOAA:
> @KateFriedman-NOAA (and anyone else that is willing to help): can you help test this on the different machines and setups (cycled/fcst-only). The workflows look right, but I haven't actually tried running.

@WalterKolczynski-NOAA I'll test it everywhere, will take at least a day given queues on some platforms.

One comment after looking at your changes (which look good), the earc section of setup_workflow.py can get the same treatment: https://github.com/NOAA-EMC/global-workflow/blob/93b389e9fbd7cdbef4c822e3435ab2c495efd11c/ush/rocoto/setup_workflow.py#L366:L369

Thanks for this!

KateFriedman-NOAA (approved):
Changes look good and tests of both cycled and free-forecast mode on WCOSS-Dell (Mars), Hera, and Orion were able to successfully submit jobs.

PR #455, opened 2021-10-06
Update rocoto_viewer with python3 version from CROW

WalterKolczynski-NOAA:
There was work on the CROW version of rocoto_viewer to update it to python3, remove the prod_utils dependency, and make some other updates.  This commit brings those changes to global-workflow.


KateFriedman-NOAA:
I tried it out on Orion and it appears to work but I notice that metatasks don't collapse:

image

WalterKolczynski-NOAA:
Weird, metatask folding _does_ work on WCOSS, but apparently not on Hera and Orion. I will look into it.

WalterKolczynski-NOAA:
Metatask folding should work now.

KateFriedman-NOAA (approved):
Works on WCOSS-Dell, Hera, and Orion...including metatasks. Thanks @WalterKolczynski-NOAA !

PR #457, opened 2021-10-11
update metp dependencies, close #437

### LINKED ISSUE #437: MET+ job dependencies
The MET+ job in the first full cycle wont run, because it depends on gfs archive job of the previous cycle which doesn't exist in the first half cycle. Job dependency and/or job scripts need to be updated to get rid of this problem.

jkuang-exploro:
Remove meta dependence of archive jobs.

FIxes #437

### DIFF
--- a/ush/rocoto/setup_workflow.py
+++ b/ush/rocoto/setup_workflow.py
@@ -693,8 +693,6 @@ def get_gdasgfs_tasks(dict_configs, cdump='gdas'):
         deps = []
         dep_dict = {'type':'metatask', 'name':f'{cdump}post'}
         deps.append(rocoto.add_dependency(dep_dict))
-        dep_dict = {'type':'task', 'name':f'{cdump}arch', 'offset':'-&INTERVAL_GFS;'}
-        deps.append(rocoto.add_dependency(dep_dict))
         dependencies = rocoto.create_dependency(dep_condition='and', dep=deps)
         sdate_gfs = rocoto.create_envar(name='SDATE_GFS', value='&SDATE_GFS;')
         metpcase = rocoto.create_envar(name='METPCASE', value='#metpcase#')
--- a/ush/rocoto/setup_workflow_fcstonly.py
+++ b/ush/rocoto/setup_workflow_fcstonly.py
@@ -668,8 +668,6 @@ def get_workflow(dict_configs, cdump='gdas'):
         deps = []
         dep_dict = {'type':'metatask', 'name':f'{cdump}post'}
         deps.append(rocoto.add_dependency(dep_dict))
-        dep_dict = {'type':'task', 'name':f'{cdump}arch', 'offset':'-&INTERVAL;'}
-        deps.append(rocoto.add_dependency(dep_dict))
         dependencies = rocoto.create_dependency(dep_condition='and', dep=deps)
         sdate_gfs = rocoto.create_envar(name='SDATE_GFS', value='&SDATE;')
         metpcase = rocoto.create_envar(name='METPCASE', value='#metpcase#')


malloryprow (approved):

KateFriedman-NOAA (commented):
Changes look good, will approve once testing completes. Thanks!

WalterKolczynski-NOAA:
@malloryprow Could you test this on WCOSS?

malloryprow:
@WalterKolczynski-NOAA How would do you want me to test it?

WalterKolczynski-NOAA:
> @WalterKolczynski-NOAA How would do you want me to test it?

Just clone Jian's branch and set it up like you normally would and check that the prerequisites are correct and MET+ runs when expected. (If you don't have experience setting up global-workflow, let me know and we'll have someone else do it.)

malloryprow:
@WalterKolczynski-NOAA I remember last time I tried to run it I struggled getting the initial conditions set up. It took me a while. If there are some ICs somewhere I can easily use for the test, I can do it.

WalterKolczynski-NOAA:
> @WalterKolczynski-NOAA I remember last time I tried to run it I struggled getting the initial conditions set up. It took me a while. If there are some ICs somewhere I can easily use for the test, I can do it.

I'm copying initial conditions for a canned case now. I'll email you the instructions Kate gave me. If you run into too much trouble, just let us know and don't worry about it.

malloryprow:
So I do have some concerns with this...

I ran a free forecast (like how you showed me to set up @WalterKolczynski-NOAA) I ran 00 cycles only and set it up to run for a few days since it takes the gfsmetp jobs a bit to get spun up.

When I kicked off the experiment, 2020090100 and 2020090200 at the same time. The gfsmetp jobs eventually kicked off for both at the same time. For the gfsmetp jobs, the default for them is to run the verification on a 1 day delay. So 2020090100 gfsmetp jobs run verification for 20200831 (there is a check for the experiment start date so nothing actually runs this time), and 2020090200 run verification for 20200901. The 2020090200 metp jobs weren't able to run because the 2020090100 arch step hadn't run yet to move the files to ARCDIR.

jkuang-exploro:
@malloryprow So metp has a dependency upon the previous cycle, but not specifically gfsarch. Is that correct?

If that's true then I propose making dependency upon the "final" task.

malloryprow:
@JianKuang-UMD No it is gfsarch, as that is the step that the files get copied to ARCDIR (line 57 in jobs/rocoto/arch.sh)

jkuang-exploro:
@malloryprow But what should happen on the second cycle(first full cycle)? Because the first cycle is a half cycle without gfs being run.

malloryprow:
@JianKuang-UMD I'm not sure I am following. The test I ran were free forecast so the gfsfcst was run on the first cycle.

jkuang-exploro:
@malloryprow the problem that I had raised only exist in cycled run. @WalterKolczynski-NOAA free forecast run can't illustrate the problem nor checking if the solution works.

malloryprow:
@JianKuang-UMD So perhaps the dependency should remain for free forecast experiments.

jkuang-exploro:
@malloryprow got it.

WalterKolczynski-NOAA:
Someone else will have to help Mallory set up a test if she needs to test it with cycling. I don't have the knowledge yet.

malloryprow:
@WalterKolczynski-NOAA I talked to Cathy Thomas and got some ICs for a warm start cycled test from her.

malloryprow (changes requested):
Remove changes for free forecast

malloryprow:
I have been thinking about this today. I feel it would make sense for the gfsmetp* tasks to run after gfsarch. The files needed for gfsmetp are put in their ARCDIR location by gfsarch.

Right now, with the gfsmetp tasks running before gfsarch, the verification actually runs for the previous date (ex. for the cycle date 2021110800 the gfsmetp tasks actually verify 20211107). If the gfsmetp* tasks, run after we can actually have the verification not need to do this date adjusting, which is what created the need for the dependency in the first place. So the new dependency would be for the gfsarch of that cycle, not the previous.

jkuang-exploro:
@malloryprow Do you mean running metp after gfsarch of the current cycle? (instead of previous cycle)

aerorahul:
@malloryprow 
The archive job should be the last job in the cycle to ensure all desired artifacts of the cycle are archived.  They should not be a dependency for any data use while in the cycle.
If the `metp` job depends on data that is placed in the archive directory, then the `metp` job should be updated.
The purpose of the archive job is purely to archive (and unfortunately it also does some cleanup).

malloryprow:
@aerorahul The metp (and the vsdb verification in vrfy) uses the files from the ARCDIR because if we used the files from ROTDIRS we would only be able to verify only a handful of days since those get scrubbed in the stmp/ptmp spaces.

aerorahul:
@malloryprow 
Ah, yes.  My apologies.  It's coming back to me.  The `ARCDIR` is the long-term on-disk archive of VSDB statistics.
In either case, we should find a way to separate the functionality into tape-archive and disk-archive.

jkuang-exploro:
Closing this PR. Will draft it again

PR #458, opened 2021-10-12
Update to UFS as of 2021 Oct 07

WalterKolczynski-NOAA:
This update adds several new stochastic physics possibilities, with
associated switched added. There are also some other new namelist
settings.

MOM input template files have been updated to use the @[ ] notation
following same changes in UFS, with corresponding changes to the
replacement code. The new template files were copied from UFS, except
saving of initial conditions is turned on.

Due to changes in how the output cadence is defined in the namelist,
an explicit list of forecast hours is generated in the pre-det script.
To avoid duplication, this list is used later to create the symlinks
for the output files.

### DIFF
--- a/ush/parsing_namelists_CICE.sh
+++ b/ush/parsing_namelists_CICE.sh
@@ -105,7 +105,7 @@ cat > ice_in <<eof
    kdyn            = 1
    ndte            = 120
    revised_evp     = .false.
-   kevp_kernel     = 0
+   evp_algorithm   = 'standard_2d'
    brlx            = 300.0
    arlx            = 300.0
    ssh_stress      = 'coupled'
@@ -148,6 +148,10 @@ cat > ice_in <<eof
    rfracmin        = 0.15
    rfracmax        = 1.
    pndaspect       = 0.8
+/
+
+&snow_nml
+   snwredist       = 'none'
 /
 
 &forcing_nml
--- a/ush/parsing_namelists_CICE.sh
+++ b/ush/parsing_namelists_CICE.sh
@@ -16,6 +16,7 @@ cat > ice_in <<eof
    year_init      = $year
    month_init     = $month
    day_init       = $day
+   sec_init       = $sec
    dt             = $ICETIM
    npt            = $npt
    ndtd           = 1
--- a/modulefiles/module_base.hera
+++ b/modulefiles/module_base.hera
@@ -26,8 +26,8 @@ module load png/1.6.35
 module load hdf5/1.10.6
 module load netcdf/4.7.4
 module load pio/2.5.2
-module load esmf/8_1_1
-module load fms/2020.04.03
+module load esmf/8_2_0_beta_snapshot_14
+module load fms/2021.03
 
 module load bacio/2.4.1
 module load g2/3.4.1
--- a/modulefiles/module_base.jet
+++ b/modulefiles/module_base.jet
@@ -19,6 +19,7 @@ module load prod_util/1.2.2
 module load grib_util/1.2.2
 module load g2tmpl/1.10.0 
 module load crtm/2.3.0
-module load esmf/8_1_1
+module load esmf/8_2_0_beta_snapshot_14
+module load fms/2021.03
 module load netcdf/4.7.4
 module load rocoto
--- a/modulefiles/module_base.orion
+++ b/modulefiles/module_base.orion
@@ -25,8 +25,8 @@ module load png/1.6.35
 module load hdf5/1.10.6
 module load netcdf/4.7.4
 module load pio/2.5.2
-module load esmf/8_1_1
-module load fms/2020.04.03
+module load esmf/8_2_0_beta_snapshot_14
+module load fms/2021.03
 
 module load bacio/2.4.1
 module load g2/3.4.1
--- a/modulefiles/module_base.wcoss_dell_p3
+++ b/modulefiles/module_base.wcoss_dell_p3
@@ -33,8 +33,8 @@ module load png/1.6.35
 module load hdf5/1.10.6
 module load netcdf/4.7.4
 module load pio/2.5.2
-module load esmf/8_1_1
-module load fms/2020.04.03
+module load esmf/8_2_0_beta_snapshot_14
+module load fms/2021.03
 
 module load bacio/2.4.1
 module load g2/3.4.1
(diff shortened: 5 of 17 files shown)


WalterKolczynski-NOAA:
Both a full coupled and an atm-only, forecast-only test ran to completion on Orion.

KateFriedman-NOAA:
@WalterKolczynski-NOAA Need to remove duplicate if-block in build_all.sh:

https://github.com/WalterKolczynski-NOAA/global-workflow/blob/feature/update_ufs/sorc/build_all.sh#L61:L71

WalterKolczynski-NOAA:
> @WalterKolczynski-NOAA Need to remove duplicate if-block in build_all.sh:
> 
> https://github.com/WalterKolczynski-NOAA/global-workflow/blob/feature/update_ufs/sorc/build_all.sh#L61:L71

I didn't touch build_all this PR. I'll fix it after.

KateFriedman-NOAA (approved):
Looks good but I'll need others to review the science aspects.

JessicaMeixner-NOAA:
Two questions as I'm reviewing: 
1. The updating of the Mom input is great! Would this be a good time to just point to the ufs-weather-model/test/parm/MOM* input files as well?  Which could help simplify updates in the future? 
2.  Are you updating the nameless based on a specific rt test in ufs-weather-model? And if so, which one?

WalterKolczynski-NOAA:
> Two questions as I'm reviewing:
> 
> 1. The updating of the Mom input is great! Would this be a good time to just point to the ufs-weather-model/test/parm/MOM* input files as well?  Which could help simplify updates in the future?

We talked about this on the call today, and we're looking towards that. Right now, I made one manual change for saving the initial files, which is hardcoded to false in the RTs.

> 2. Are you updating the namelist based on a specific rt test in ufs-weather-model? And if so, which one?

I used defaults from the coupled control as much as possible, though I leave the new options (like ocean perturbation) off for now. So the default settings for all the new SP pieces are there, but they default to off. Turning them on will bring all the default settings with it, unless they are overridden.

WalterKolczynski-NOAA:
Basically, I made hooks for everything but kept things as much the same as before as possible. But for the options that have a bunch of settings that come along when they are turned on, all the defaults from the coupled RT are sitting there ready for when they are turned on.

ush/parsing_namelists_CICE.sh
```diff
@@ -105,7 +105,7 @@ cat > ice_in <<eof
    kdyn            = 1
    ndte            = 120
    revised_evp     = .false.
-   kevp_kernel     = 0
+   evp_algorithm   = 'standard_2d'
```
JessicaMeixner-NOAA, line 109: &setup_nml
sec_init = @[SECS]
was added in tests/parm/ice_in_template but not here.

JessicaMeixner-NOAA (changes requested):
Looking at changes that were updated in ufs-weahter-model: 
https://github.com/ufs-community/ufs-weather-model/compare/b26a896f2c9bada438414a51218bc72236925b8c...develop
I saw that ufs-weather-model modules are updated to use hpc stack 1.2.0 instead of 1.1.0 and that change is not reflected in the module files here.  I'm not sure if that's intentional or not?  In addition, the ESMF and FMS modules have been updated: module load esmf/8_2_0_beta_snapshot_14
module load fms/2021.03which is also not reflected in the module files in this update. 


Someone else would be a better person to review the updates related to the output frequency.    

I'm still reviewing the namelist changes, do you have output from your run or should I just make a new run myself?  I can tell the changes for MOM6 look okay, I've made one comment on CICE, but for input.nml - I'm still following all the diffs and it might be easier if there's output to just look at instead of just looking at the code.

ush/parsing_namelists_CICE.sh
WalterKolczynski-NOAA: > Looking at changes that were updated in ufs-weahter-model:  [ufs-community/ufs-weather-model@b26a896...develop](https://github.com/ufs-community/ufs-weather-model/compare/b26a896f2c9bada438414a51218bc72236925b8c...develop) I saw that ufs-weather-model modules are updated to use hpc stack 1.2.0 instead of 1.1.0 and that change is not reflected in the module files here.  I'm not sure if that's intentional or not?  In addition, the ESMF and FMS modules have been updated: module load esmf/8_2_0_beta_snapshot_14 module load fms/2021.03which is also not reflected in the module files in this update. 

The hpc-stack update seems to only be on wcoss2.

For esmf and fms, I'm not sure if it currently even matters, since it is correct at compile time (do we even need to be loading these at runtime?). But I will update them for consistency. In the slightly wider view: with the new dictates from NCO, module updates for ALL workflow components need to start being coordinated because NCO wants a single top-level versions file to control all modules.

> 
> Someone else would be a better person to review the updates related to the output frequency.
> 
> I'm still reviewing the namelist changes, do you have output from your run or should I just make a new run myself? I can tell the changes for MOM6 look okay, I've made one comment on CICE, but for input.nml - I'm still following all the diffs and it might be easier if there's output to just look at instead of just looking at the code.

/work/noaa/global/wckolczyn/noscrub/global-workflow/update_ufs
/work/noaa/global/wckolczyn/noscrub/global-workflow/update_ufs_atm

If you need the run directories, you'll have to rerun it, as I neglected to change KEEPDATA. I'm still following the directions I listed in PR #430 to run the coupled runs.

WalterKolczynski-NOAA: Actually, I'm going to rerun at least the coupled with these new changes, and I'm turning KEEPDATA on if you want to wait for that.

WalterKolczynski-NOAA: The run directory will be in /work/noaa/stmp/wkolczyn/RUNDIRS/update_ufs/2013040100/gfs when it is created.

I had to scrub the output from the rotdir, so that will be regenerated as well.

JessicaMeixner-NOAA: Thanks Walter! I will look at your new output.

JessicaMeixner-NOAA:
Thanks for sharing the output.  From what I saw comparing input.nml files with regression tests, we could add 
&gfs_physics_nml
use_cice_alb = @[USE_CICE_ALB]
which is currently set to its default value false, but this could always be added in a separate PR.

WalterKolczynski-NOAA:
Coupled run with the new changes completed successfully. Due to some configuration change on Orion, I had to add `unset SLURM_MEM_PER_CPU` to the end of config.base in order for srun to work. I have an email into the Orion helpdesk on this issue.

JessicaMeixner-NOAA (approved):
Thanks for the updates @WalterKolczynski-NOAA!

PR #462, opened 2021-10-13
Add WCOSS2 support in configs and setup scripts

KateFriedman-NOAA:
- Add needed WCOSS2 support in configs and setup scripts. Tested setup scripts successfully on Cactus (have not yet tested xml).
- Created and added WCOSS2.env file. Have not yet tested at runtime.

Refs: #399

### DIFF
--- a/env/WCOSS2.env
+++ b/env/WCOSS2.env
@@ -1,0 +1,232 @@
+#!/bin/ksh -x
+
+if [ $# -ne 1 ]; then
+
+    echo "Must specify an input argument to set runtime environment variables!"
+    echo "argument can be any one of the following:"
+    echo "anal fcst post vrfy metp"
+    echo "eobs eupd ecen esfc efcs epos"
+    echo "postsnd awips gempak"
+    exit 1
+
+fi
+
+step=$1
+
+# WCOSS2 information
+export launcher="mpiexec -l"
+export mpmd="--cpu-bind verbose,core cfp"
+
+export npe_node_max=128
+
+# Configure MPI environment
+export MPI_LABELIO=YES
+export MP_STDOUTMODE="ORDERED"
+export KMP_STACKSIZE=2048M
+export KMP_AFFINITY=scatter
+export job=${LSB_JOBNAME:-$step}
+export jobid=${job}.${LSB_JOBID:-$$}
+
+if [ $step = "prep" -o $step = "prepbufr" ]; then
+
+    nth_max=$(($npe_node_max / $npe_node_prep))
+
+    export POE=${POE:-"YES"}
+    export BACK=${BACK:-"off"}
+    export sys_tp="wcoss2"
+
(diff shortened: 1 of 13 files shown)


env/WCOSS2.env
```diff
@@ -0,0 +1,232 @@
...
+
+export npe_node_max=128
+
+# Configure MPI environment
+export MPI_LABELIO=YES
+export MP_STDOUTMODE="ORDERED"
+export KMP_STACKSIZE=2048M
+export KMP_AFFINITY=scatter
+export job=${LSB_JOBNAME:-$step}
+export jobid=${job}.${LSB_JOBID:-$$}
```
aerorahul: Are you should this is `LSB_JOBID`?  This is a PBS scheduler.

env/WCOSS2.env
```diff
@@ -0,0 +1,232 @@
...
+export mpmd="--cpu-bind verbose,core cfp"
+
+export npe_node_max=128
+
+# Configure MPI environment
+export MPI_LABELIO=YES
+export MP_STDOUTMODE="ORDERED"
+export KMP_STACKSIZE=2048M
+export KMP_AFFINITY=scatter
+export job=${LSB_JOBNAME:-$step}
```
aerorahul: Are you should this is `LSB_JOBNAME`? This is a PBS scheduler.

aerorahul (commented):
looks ok to me.
just a couple of comments.

env/WCOSS2.env
KateFriedman-NOAA: Good catch, this should be a PBS variable, have now fixed, thanks!

env/WCOSS2.env
KateFriedman-NOAA: Fixed, thanks!

KateFriedman-NOAA:
@aerorahul I believe I have addressed your comments, thanks!

env/WCOSS2.env
```diff
@@ -0,0 +1,232 @@
...
+export launcher="mpiexec -l"
+export mpmd="--cpu-bind verbose,core cfp"
+
+export npe_node_max=128
+
+# Configure MPI environment
+export MPI_LABELIO=YES
+export MP_STDOUTMODE="ORDERED"
+export KMP_STACKSIZE=2048M
+export KMP_AFFINITY=scatter
```
aerorahul, line 26: This is fine for now, but we should ask the system owner if these are advantageous or not needed on this machine.

aerorahul (approved):
looks good to me.

env/WCOSS2.env
KateFriedman-NOAA: Definitely...these are what is used on the Dells right now so I need inputs from others on what (if any) to set here.

KateFriedman-NOAA:
Thanks @aerorahul !

PR #464, opened 2021-10-18
Make switching to coupled defaults easier

WalterKolczynski-NOAA:
In order to facilitate running coupled experiments without having to
modify several config files, some updates are made to the config files.

A new file, config.defaults.s2sw is created to contain the alternate
defaults for the coupled model, based on prototype 7. This file is
sourced whenever the app is S2SW.

The UFS app is made into a variable that is now set by the setup scripts
via the --app switch. Allowed values are ATM, ATMW, S2S, S2SW. If no
argument is provided, the default is ATM (so it doesn't change existing
behavior).

The section in config.base that sets all the components based on the app
is moved earlier in order to allow the new defaults to be set.

Settings in some config files have been modifyied to now accept values
that have been set previously to allow the alternate settings for the
coupled model. In the future, this could be extended to additional config
files to allow more settings to be overridden.

waveGRDN and waveGRDG are updated to be based on waveGRD.

Updated instructions to setup prototype analog:

1. Clone and run checkout, build_all, and link as normal for coupled:
```
./checkout.sh
./build_all.sh -c
./link_workflow emc orion coupled
```

2. In ush/rocoto, run
```
./setup_expt_fcstonly.py --app S2SW --res 384 --pslot $PSLOT --comrot $COMROT --expdir $EXPDIR --idate 2013040100 --edate 2013040100 --configdir ../../parm/config
./setup_workflow_fcstonly.py --expdir $EXPDIR/$PSLOT
```

3. In $EXPDIR/$PSLOT, modify $PSLOT.xml as follows:
- Remove or comment out the entire gfsinit task
- Remove or comment out gfsinit as a dependency for gfswaveprep
- Optionally reduce the walltime for the forecast job to 1 hour to move through the queue faster
-- The forecast job uses too many nodes for the debug queue, though you can use the debug queue for the wave prep jobs

4. In $ROTDIR, do the following:
```
cd $ROTDIR
ln -s /work/noaa/global/wkolczyn/noscrub/global-workflow/FV3ICS FV3ICS

mkdir -p $ROTDIR/$PSLOT/gfs.20130401/00/atmos
cd $ROTDIR/$PSLOT/gfs.20130401/00/atmos
ln -s $ROTDIR/FV3ICS/2013040100/atmos/C384/INPUT INPUT

mkdir -p $ROTDIR/$PSLOT/gdas.20130331/18/wave/restart
cd $ROTDIR/$PSLOT/gdas.20130331/18/wave/restart
ln -s $ROTDIR/FV3ICS/2013040100/wav/20130401.000000.restart.gwes_30m .
```

Refs: #368


WalterKolczynski-NOAA:
Produces bitwise-identical results for the coupled model. Going to test atm-only now.

parm/config/config.fcst
```diff
@@ -93,45 +93,45 @@ export new_o3forc="YES"
...
+export d2_bg_k2=${d2_bg_k2:-0.04}
+export dz_min=${dz_min:-2}
+export n_sponge=${n_sponge:-42}
 
 # PBL/turbulance schemes
-export hybedmf=".false."
-export satmedmf=".true."
-export isatmedmf=1
-tbf=""
+export hybedmf=${hybedmf:-".false."}
```
aerorahul: `variable=${variable:-default_value}`
implies one can define this variable before this script.
This being a configuration script then looses the point of being a configuration script, rather become a configurable script.

The syntax at this level should remain at key-value pairs.

aerorahul (changes requested):
The config scripts should remain `key=value`.

parm/config/config.fcst
WalterKolczynski-NOAA: This was (to me) the least objectionable method to handle multiple classes of default settings:
- If we source the coupled defaults afterwards, you have to worry about also redoing any calculations that were made based on the original defaults, plus you run into that problem where a user will change a setting in one place, only to not have it do anything because it is being overwritten elsewhere.
- If we insert different defaults where they appear, you wind up with a bunch of if/switch blocks based on the app scattered all over the config files.
- We already agreed keeping an additional set of config files was a bad idea, because of the bloat and duplication

This method also has the advantage of being able to be easily extended more broadly to something sorta like the case files from the CROW workflow, where the config files _would_ basically become a configurable script where you just have one file where you make any overrides. Even if we don't do that (since we're building a new system), this way is the cleanest option in the interim, IMHO. Unless there is an option I'm missing?

We can discuss it in the tag-up tomorrow if need be.

parm/config/config.defaults.s2sw
```diff
@@ -0,0 +1,37 @@
...
+FHMAX_GFS_06=48
+FHMAX_GFS_12=48
+FHMAX_GFS_18=48
+FHOUT_GFS=6
+FHOUT_HF_GFS=-1
+
+# config.fcst
+IALB=2
+IEMS=2
+DO_CA="YES"
```
KateFriedman-NOAA, line 17: What does the "CA" in `DO_CA` stand for?

parm/config/config.fv3
```diff
@@ -42,110 +42,110 @@ fi
...
-        export layout_y_gfs=2
-        export nth_fv3=1
-        export nth_fv3_gfs=1
-        export cdmbgwd="0.071,2.1,1.0,1.0"  # mountain blocking, ogwd, cgwd, cgwd src scaling
-        export WRITE_GROUP=1
-        export WRTTASK_PER_GROUP=$npe_node_max
-        export WRITE_GROUP_GFS=1
-        export WRTTASK_PER_GROUP_GFS=$npe_node_max
-        export WRTIOBUF="1M"
+        export DELTIM=${DELTIM:-1200}
```
KateFriedman-NOAA: Why have overrides in this config? This is where these values get set...so what would potentially override these values?

ush/rocoto/setup_expt.py
```diff
@@ -135,6 +136,7 @@ def edit_baseconfig():
 link initial condition files from $ICSDIR to $COMROT'''
 
     parser = ArgumentParser(description=description, formatter_class=ArgumentDefaultsHelpFormatter)
+    parser.add_argument('--app', help='UFS application', type=str, choices=['ATM', 'ATMW', 'S2S', 'S2SW'], required=False, default='ATM')
```
KateFriedman-NOAA, line 139: Not request for a change, just a thought...I imagine this list of possible APPs will show itself in multiple places, is there a way to consolidate all of that? For example, the value of `choices` here could be sourced from some sort of common file/place earlier in the script. Have been contemplating something similar for a while to handle the list of supported machines/target that's listed/used through the system and other similar lists that will grow/shrink over time as we support more/less. Food for thought, thanks!

KateFriedman-NOAA (commented):
Changes look good from a technical standpoint...but want to discuss the addition of overrides in the configs (see my comment in the review). Love the addition of the `--app` argument, thanks for adding that!

parm/config/config.defaults.s2sw
WalterKolczynski-NOAA: Cellular Automata

parm/config/config.fv3
WalterKolczynski-NOAA: Prototype 7 is using a different timestep and layout than listed here for C384

KateFriedman-NOAA: Gotcha. Should we just pull these values right from wherever those values are set then? Skip the middle man. Or is that the next step in moving towards the model template files?

ush/rocoto/setup_expt.py
WalterKolczynski-NOAA: Right now it is just three places (the two py scripts and config.base). And in config.base, it takes the form of case statements that might not be wise to abstract. The two py scripts have a LOT of duplicated code (it probably should be just a single script with fcst-only and DA options); app choices aren't unique in that respect.

parm/config/config.fv3
WalterKolczynski-NOAA: See the above response to Rahul. Once you start chopping it up, you're going to have if or switch blocks all over the place where these settings pop up to pick the wrong ones.

KateFriedman-NOAA (approved):

WalterKolczynski-NOAA:
ATM-only fcst-only completed successfully.

parm/config/config.wave
```diff
@@ -26,28 +26,29 @@ export CDUMPRSTwave="gdas"
...
 # Grids for wave model
 # GFSv16
-export waveGRD='gnh_10m aoc_9km gsh_15m'
-export waveGRDN='1 2 3' # gridnumber for ww3_multi
-export waveGRDG='10 20 30' # gridgroup for ww3_multi
-export USE_WAV_RMP='YES' #yes/no rmp grid remapping pre-processed coefficients
+export waveGRD=${waveGRD:-'gnh_10m aoc_9km gsh_15m'}
+nGRD=$(echo $waveGRD | wc -w)
+export waveGRDN=$(seq $nGRD)               # gridnumber for ww3_multi
+export waveGRDG=$(seq -f "%g0" $nGRD)      # gridgroup for ww3_multi
```
JessicaMeixner-NOAA: While this works for this setting, this theoretically is not the formula for all sets of grids, so I actually don't think we should put this formula in.

JessicaMeixner-NOAA: Here's a way I started to make this easier for configuring waves but never finished testing: 
https://github.com/JessicaMeixner-NOAA/global-workflow/blob/coupled-sprint/runwithwaves/parm/config/config.wave#L97-L171   I don't know if this or what you have is preferable.  They both work, I think we just need to set waveGRDN, waveGRDG in your parm/config/config.defaults.s2sw file

WalterKolczynski-NOAA: Backed this change out. I'd still love to have a way to auto-generate these if possible in the future.

parm/config/config.fcst
WalterKolczynski-NOAA: Per our conversation, all of the overrides except those necessary for the alternate settings were removed.

parm/config/config.wave
JessicaMeixner-NOAA: You cannot autogenerate them because they are specific to the combination of grids you choose and how you are running them.

JessicaMeixner-NOAA (approved):
Thanks for updating the wave variables.

parm/config/config.base.emc.dyn
```diff
@@ -180,15 +233,15 @@ export gfs_cyc=@gfs_cyc@ # 0: no GFS cycle, 1: 00Z only, 2: 00Z and 12Z only, 4:
...
 export FHMIN_GFS=0
 
-export FHMAX_GFS_00=384
-export FHMAX_GFS_06=384
-export FHMAX_GFS_12=384
-export FHMAX_GFS_18=384
+export FHMAX_GFS_00=${FHMAX_GFS_00:-384}
+export FHMAX_GFS_06=${FHMAX_GFS_06:-384}
+export FHMAX_GFS_12=${FHMAX_GFS_12:-384}
+export FHMAX_GFS_18=${FHMAX_GFS_18:-384}
```
aerorahul, line 239: I have always wondered, is there ever a practical reason why a user would set different forecast lengths for different cycles?

parm/config/config.base.emc.dyn
```diff
@@ -180,15 +233,15 @@ export gfs_cyc=@gfs_cyc@ # 0: no GFS cycle, 1: 00Z only, 2: 00Z and 12Z only, 4:
...
 export FHMAX_GFS=$(eval echo \${FHMAX_GFS_$cyc})
 
-export FHOUT_GFS=3
-export FHMAX_HF_GFS=0
-export FHOUT_HF_GFS=1
+export FHOUT_GFS=${FHOUT_GFS:-3}
+export FHMAX_HF_GFS=${FHMAX_HF_GFS:-0}
+export FHOUT_HF_GFS=${FHOUT_HF_GFS:-1}
```
aerorahul, line 244: I think the current develop of the UFS now no longer has `HF` as an input.
I think they want all forecast hours explicitly defined as a list.
Please check though.

aerorahul (approved):
Looks good.
Thanks for taking in the suggestions.
Also, see a couple of comments.

parm/config/config.base.emc.dyn
WalterKolczynski-NOAA: I believe there have been times in the past when 00z or 00z & 12z have been run longer than other cycles.

KateFriedman-NOAA: Yeah it's a legacy-type thing to let users run different lengths for different cycles. From a technical standpoint I dislike this and would prefer to merge this back into a single forecast length variable and just let users define which cycles to run gfs instead of having different forecast lengths per cycle which messes with the verification.

parm/config/config.base.emc.dyn
WalterKolczynski-NOAA: This is true, but we're currently still using these settings to build the output list later in the workflow: https://github.com/NOAA-EMC/global-workflow/blob/feature/coupled_sprint/ush/forecast_predet.sh#L57-L68 (this was in the PR I made last week to update the UFS version).

PR #467, opened 2021-10-19
Fix resource assignment issue found in analysis job running global_cycle #466

### LINKED ISSUE #466: Fix resource assignment issue found in analysis job running global_cycle
The GDAS and GFS analysis job running in rocoto workflow with full resolution parallels sometime create zombie job make MARS/Venus Phase-3 Compute Nodes Unresponsive.
Looks like the resource assignment is inconsistent when the analysis job running executable global_cycle.
global_cycle requires 4 threads but was mismatched to 28 in runtime. This caused the WCOSS_d nodes become unrespondable

Tested with a driver running with OMP_NUM_THREADS=28 get 3 zombie jobs in 35 repeat runs.
Locally fixed this issue by modifying OMP_NUM_THREADS=4 in v16iaiscris2 parallel. 
Here is to coordinate the fix into code management commit.

lgannoaa:
The GDAS and GFS analysis job running in rocoto workflow with full resolution parallels sometime create zombie job make MARS/Venus Phase-3 Compute Nodes Unresponsive.
The resource assignment is inconsistent when the analysis job running executable global_cycle.
global_cycle requires 4 threads but was mismatched to 28 in runtime. This caused the WCOSS_d nodes become unrespondable

nth_anal_gfs was not defined in config.resources when config.anal executed it.
nth_anal_gfs now defaults to nth_anal if nth_anal_gfs is not defined.
set nth_cycle=nth_anal in config.resources


parm/config/config.resources
```diff
@@ -137,9 +138,15 @@ elif [ $step = "anal" ]; then
       export npe_anal=84
       export npe_anal_gfs=84
     fi
-    if [[ "$machine" = "WCOSS_DELL_P3" ]]; then export nth_anal=7; fi
+    if [[ "$machine" = "WCOSS_DELL_P3" ]]; then 
+      export nth_anal=4
+      if [ "$QUEUE" = "dev2" -o "$QUEUE" = "devonprod2" -o "$QUEUE" = "devmax2" ]; then # WCOSS Dell 3.5
+        export nth_anal=5
+      fi
```
aerorahul: Why is there a `QUEUE` dependence here?

parm/config/config.resources
```diff
@@ -129,6 +129,7 @@ elif [ $step = "anal" ]; then
     export npe_anal=1000
     export nth_anal=5
     export npe_anal_gfs=1000
+    export nth_anal_gfs=$nth_anal
```
aerorahul: I don't think you need this line, since Line 147 will take care of that.

parm/config/config.resources
```diff
@@ -137,9 +138,15 @@ elif [ $step = "anal" ]; then
...
       export npe_anal_gfs=84
     fi
-    if [[ "$machine" = "WCOSS_DELL_P3" ]]; then export nth_anal=7; fi
+    if [[ "$machine" = "WCOSS_DELL_P3" ]]; then 
+      export nth_anal=4
+      if [ "$QUEUE" = "dev2" -o "$QUEUE" = "devonprod2" -o "$QUEUE" = "devmax2" ]; then # WCOSS Dell 3.5
+        export nth_anal=5
+      fi
+    fi
+    export nth_anal_gfs=${nth_anal_gfs:-nth_anal}
```
aerorahul: Also, after removing L132, this should be:
`export nth_anal_gfs=$nth_anal`

aerorahul (commented):
See comments.

parm/config/config.resources
lgannoaa: v16iasicris2 parallel used this rule when running with these QUEUE.

parm/config/config.resources
lgannoaa: Modified

parm/config/config.resources
lgannoaa: removed

parm/config/config.resources
aerorahul: I suggest remove the `QUEUE` logic and replace 4 with 5 on Line 141.

lgannoaa: I will remove line 140-145 because line 130 has already set nth_anal=5.

aerorahul: yep.  That was going to be my next review comment.

parm/config/config.resources
```diff
@@ -137,9 +137,9 @@ elif [ $step = "anal" ]; then
       export npe_anal=84
       export npe_anal_gfs=84
     fi
-    if [[ "$machine" = "WCOSS_DELL_P3" ]]; then export nth_anal=7; fi
```
KateFriedman-NOAA, line 140: Why is this line that sets `nth_anal` for the Dells removed?

KateFriedman-NOAA (commented):
See review comment about change to remove nth_anal setting for WCOSS-Dell in config.resources.

parm/config/config.resources
lgannoaa: It is because @aerorahul requested " I suggest remove the QUEUE logic and replace 4 with 5 on Line 141." above.

KateFriedman-NOAA: `if [[ "$machine" = "WCOSS_DELL_P3" ]]; then export nth_anal=7; fi`
The line you removed is not QUEUE logic, it's machine logic. Please put it back.

aerorahul: I did suggest to remove the `QUEUE` logic.
I also suggested to replace 4 with 5, because that is what was effectively been done.  
I am not sure why 4/5 and not 7 as is the case in `develop`.
I imagine it was so in the parallel.
I will leave it to the person who wants to test and determine what it should be; 7 as is in `develop` or 4/5 based on updated evidence of performance.

lgannoaa: To my understand, @aerorahul requested this change. "replace 4 with 5 on Line 141" is in place. Please get approve from @aerorahul if you want to back out this change.

aerorahul: @lgannoaa 
This is the original logic:
```
if [[ "$machine" = "WCOSS_DELL_P3" ]]; then export nth_anal=7; fi
```

This is the logic in your original PR at commit ef6c0462 when the conversation started:
```
  export nth_anal=4
    if [ "$QUEUE" = "dev2" -o "$QUEUE" = "devonprod2" -o "$QUEUE" = "devmax2" ]; then # WCOSS Dell 3.5
      export nth_anal=5
    fi
fi
```

This commit shows `nth_anal=4`, which after removing the `QUEUE` logic became 5, and eventually disappeared because `nth_anal=5` is set on line 130.

So, you changed the `nth_anal=7` for `WCOSS_DELL_P3` even before I commented on the PR.
Can you explain what is wrong with `nth_anal=7` on `WCOSS_DELL_P3`?  I believe that is what @KateFriedman-NOAA is asking.

lgannoaa: Based on /gpfs/dell2/emc/modeling/noscrub/emc.gloparadev/para_fv3gfs/v16iasicris2/config.resources. Line 125 was removed from DA team. export nth_anal=4 is used in v16iasicris2 final configuration. That is the version I purposed originally. Here, @aerorahul requested to use nth_anal=5 instead. Thus, the final became nth_anal=5.
Logic nth_anal inside WCOSS_DELL_P3 if statement is equal to the setting before on line 130. Therefore, it is not need to have extra assignment.

aerorahul: Ok.  This is new information and good to know.
Based on this information, I would revise my suggestion to say:
```
if [[ "$machine" = "WCOSS_DELL_P3" ]]; then export nth_anal=4; fi
```
This is accurate and consistent with the `v16iasicris2` final configuration.

If `nth_anal=5` is acceptable for the DA group, then we can remove the `WCOSS_DELL_P3` specific logic.
If not, we could replace `nth_anal=5` with `nth_anal=4` as the default for all machines and remove the `WCOSS_DELL_P3` logic.

What do you want to do?

KateFriedman-NOAA: Lin, what machines was this `nth_anal` change tested on? If it was tested on every supported platform then I am ok with removing the machine logic and setting `nth_anal` to the same value everywhere (whatever value the DA team prefers). If this was only tested on WCOSS-Dell phase 3 then please use what @aerorahul suggested:
```
if [[ "$machine" = "WCOSS_DELL_P3" ]]; then export nth_anal=4; fi
```

lgannoaa: I backout the change and restored nth_anal=4 within WCOSS_DELL_P3 if statement.

KateFriedman-NOAA (approved):

KateFriedman-NOAA:
Resolves #466

aerorahul (approved):
This PR essentially boils down to:
```
if [[ "$machine" = "WCOSS_DELL_P3" ]]; then export nth_anal=4; fi
export nth_anal_gfs=$nth_anal
export nth_cycle=$nth_anal
```
The first line changes 7 threads to 4 on WCOSS_DELL_P3 for analysis
The second line sets the same number of threads for the GFS cycle for analysis
The third line sets the number of threads for global cycle to be the same for analysis.

I assume this has been tested adequately on platforms other than WCOSS_DELL_P3.

KateFriedman-NOAA:
@lgannoaa Has this change been tested on the other platforms?

lgannoaa:
> @lgannoaa Has this change been tested on the other platforms?

No.

KateFriedman-NOAA:
@lgannoaa Please test this resource change on at least one of the SLURM R&D platforms (Hera or Orion) for 2.5 cycles. A low res test is fine. Confirm the resource adjustment works without issue on whichever platform you choose and then we can submit this PR. Thanks!

lgannoaa:
@KateFriedman-NOAA 
Will try to test with C192C96 on Orion.

KateFriedman-NOAA:
> @KateFriedman-NOAA Will try to test with C192C96 on Orion.

Great, thanks! If you need C192C96 ICs I have a set here: /work/noaa/stmp/kfriedma/ICs/2020090118_C192C96L127

lgannoaa:
Running a cycled low resolution C192C96 parallel PSLOT=naf19296 on ORION. Jobs completed without issue.
HOMEgfs=/work/noaa/global/lgan/git/nth_anal_fix
COM=/work/noaa/stmp/lgan/comrot/naf19296
expdir=/work/noaa/global/lgan/expdir/naf19296
SDATE=2020090118
EDATE=2020090218

aerorahul:
@RussTreadon-NOAA 
@lgannoaa says in [this comment](https://github.com/NOAA-EMC/global-workflow/pull/467#discussion_r732239700) the DA team replaced the number of threads for `global_cycle` from 7 to 4 in the final configuration for v16iasicris2. 
 Looking at `/gpfs/dell2/emc/modeling/noscrub/emc.gloparadev/para_fv3gfs/v16iasicris2/config.resources`, some logic was added based on `QUEUE`.  
Question is why does `QUEUE` have a bearing on number of threads used in a job?  The PE count allocation is the same regardless of the `QUEUE`.

Is this changed supposed to be migrated back to the repository or was this a temporary fix.
Please review at your convenience.

RussTreadon-NOAA:
Lin reported 10/8 on tests he ran to look at anal job hangs.   He found
that global_cycle did not hang when he used 4 threads.   This is the same
number of threads used by global_gsi.x in the v16iasicris2 configuration.
We accepted Lin's recommendation.

NCO runs JGLOBAL_ATMOS_ANALYSIS with 7 threads for global_gsi.x and 28
threads for global_cycle.   I do not recall the operational global analysis
jobs hanging when global_cycle begins.  JGLOBAL_ATMOS_ANALYSIS  runs
global_gsi.x with 1000 tasks on 250 phase 3 nodes (ptile=4).   Each node is
fully used by global_gsi.x (4 tasks with each task running 7 threads).  The
v16iasicris2 ran global_gsi.x with 400 tasks, ptile=7 and threads=4.   The
xml file requested 58 nodes.  57 of these nodes are fully utilized by the
gsi (57*7=399 task with 4 threads per task).  The last task runs on the
58th node with 4 threads.   The rest of the 58th node is not used by
global_gsi.x.   Perhaps the fact that the last node is not fully utilized
by the global_gsi.x results in global_cycle occasionally hanging.

As to the QUEUE logic I think it's origins go back to the fact that
developers can run WCOSS_DELL_P3 parallels on either phase 3 or phase 3.5
nodes.   Phase 3 nodes have 28 physical cores.   Phase 3.5 nodes have 40
physical cores.   The QUEUE logic for several jobs (prep, anal, post, vrfy,
eobs, eupd, ecen, epos) may be an attempt to have npe_node_$step =
npe_node_max / nth_$step be an integer.

I do not know who in the DA team maintains or supports global_cycle.   Is
global_cycle a land team app?_cycle.   Their take on all this could prove
useful.

aerorahul:
`global_cycle` is run after the `global_gsi.x`.
Both are executed as part of the same job in `JGLOBAL_ANALYSIS`.
Is it possible, that an instance of `global_cycle` is landing on the last node that is not fully utilized?
Is it possible that there is a resource mismatch between GSI resources (pe/node, threads) and global_cycle resources (1 node, threads)? 
Since this is part of the larger analysis job, who from the `global_cycle` should this be addressed to?

RussTreadon-NOAA:
Good question.   Who maintains global_cycle?   Does the land team maintain global_cycle?

aerorahul:
This will likely be moot once `global_cycle` is pulled out of the `JGLOBAL_ATMOS_ANALYSIS` into its own job to manage resources optimally.

If this has been tested and works, CM's please take appropriate action.

RussTreadon-NOAA:
I will be working with the Land DA team to make changes in the GLDAS repo
which make this change in the workflow possible.   A corresponding change
will also be needed in the GSI repo.   Is there an existing workflow branch
to which we can make the necessary changes on the workflow side.

Russ

KateFriedman-NOAA:
Approved in a side conversation. Submitting! Thanks @lgannoaa !

PR #473, opened 2021-10-25
modify set -x settings to skip load_fv3gfs_modules.sh

### LINKED ISSUE #389: wrap `module` commands with `set +x` and `set -x`
Through out our scripts, we load and unload modules while enabling tracing `set -x`.
While it is extremely useful for logging the progression of the script, the trace output from `module` commands is extremely long, cryptic and completely useless to the general user/developer.

`module` commands should be expected to work as is without the need for tracing their inner workings.

Identify all instances where `module` commands exist and wrap them inside `set +x` and `set -x` such that:

```
set +x
module purge
module use /path/to/modulefiles
module load compiler
module load mpi
module load third-party-library
module list
set -x
```

@DanIredell Would you be willing to take this on?  This seems like a gentle introduction to global workflow functionality.   If so, please assign yourself on the right hand side under `Assignees`.

jkuang-exploro:
Changes are made in jobs/rocoto scripts. "-x" no longer in effect for load_fv3gfs_modules.sh


jkuang-exploro:
To address issue #389

WalterKolczynski-NOAA (changes requested):
Instead of changing 40 files, why not just change load_fv3gfs_modules.sh to have set +x?

jkuang-exploro:
I will delete this PR and start over

Issue #367, opened 2021-07-19
Identify and bring waves, ocean, and ice pre and post processing scripts to develop


aerorahul:
Identity the pre and post processing scripts for waves, ocean and ice used in coupled prototypes and integrate with develop.

JessicaMeixner-NOAA:
So in terms of ocean/ice pre-processing jobs, there is only the staging of the initial conditions.  The ocean/ice post job and the initial condition staging job have been described in detail in a comment on #369, which I am not repeating here for now. 


For waves, all jobs are in the develop branch for pre-and post-processing.  However, the develop branch and the feature/coupled-crow branch are out of sync and need to be reconciled. One of the bigger differences is in the fix file.  Recently in https://github.com/NOAA-EMC/global-workflow/pull/343 in feature/coupled-crow fix files that should actually be parm files were updated accordingly.  Most of this work cannot be done until we can either 1. run atm-ocn-ice-wav coupled in develop or 2. Issue #256 is resolved so we can run atm->wav in the v16 like configuration in develop.    In the meantime I can attempt to determine what I think should work but you can't test or confirm until we have 1 or 2.  Another major difference between the two branches is that in develop for v16 we process boundary points and we do not do those jobs for featue/coupled-crow for the prototypes.  This will be needed for a future GFS, so some sort of option to do or not needs to at some point be made.  Lastly another big difference is that the gridded post processing in develop runs side-by-side with the forecast job (which we would want in a future GFS scenario) but for the prototype runs in feature/coupled-crow this job is dependent on the forecast job completing.

PR #488, opened 2021-11-04
Add coupled tasks to workflow and fix dependencies

### LINKED ISSUE #191: Update ocn/ice com restructure
At some point the ocean and ice com directories should be restructured to follow the NCO structure: 

<COM>/atmos
<COM>/wave 
<COM>/chem 

etc. 

One trick with this will be that the resulting grib files for ocean and ice are actually combined.  So I'm not sure how that should be handled.  Otherwise, it should be pretty straight froward.

### LINKED ISSUE #461: ocnpost jobs time out on Orion
When running feature/coupled-crow on Orion, the gfs.post.ocnpost tasks all time out (the time limit is set to 30 minutes). This started happening on Wednesday 6 October 2021, consistent with a change in Orion -- see the email below. Just by increasing the time limit, jobs finish anywhere from 45 to 75 minutes now; I have the time limit set to 90 minutes to be safe.

-------------------
_On Wednesday, October 6, beginning at 1:00pm CDT, changes to orion's slurm configuration will be made.
These changes will be completed live and a short window of service interruption for job submissions may occur while the scheduler is restarted._

_These changes should help with job throughput._

_Users should note a change to the default memory allocation.
The default memory allocation per core will be changed from using all available memory to being based on cores requested per node.
Standard compute nodes w/ 192GB will default to 4608 Mb per core
Big mem nodes w/ 384GB will default to 9472 Mb per core
Users can change these defaults by using the --mem or --exclusive options._

_Running jobs should not be affected, and queued jobs may have their priority/fairshare adjusted after the
reconfiguration._

### LINKED ISSUE #481: Add job to stage coupled prototype ICs
Need to copy the job from coupled-crow that stages the initial conditions for the prototype to develop.

### LINKED ISSUE #482: Add coupled jobs to workflow generation
Missing jobs for coupled model (coupled ICs and ocean post) need to be added to the setup workflow scripts.

### LINKED ISSUE #483: Add ocean post
The ocean post scripts were brought over from coupled-crow, but they haven't been run yet.

### LINKED ISSUE #484: Better labeling of pooled post tasks in rocoto
Right now the task names for the pooled post jobs are just indexes, so you can't tell by looking what forecast hours are covered by each task. It would be better to have more informative labels that tell which forecast hours are included for each task.

### LINKED ISSUE #485: Link ocean/ice output from forecast directory instead of copying
Right now, ocean and ice output files are written directly to the forecast working directory, then copied to COM when the forecast is complete. Other components create symlinks for the expected files before the forecast is run so the output can be processed immediately before the forecast is complete.

### LINKED ISSUE #486: Add ability to turn off boundary point jobs for wave post
The current workflow always runs the boundary point jobs for wave post, but not all wave configurations (like that of the coupled prototype) need them.

### LINKED ISSUE #487: Use waveGRD to determine workflow dependencies for wave post jobs
The data dependencies for wave post are currently hard-wired to the three wave grids used for GFS v16 (gnh_10m, aoc_9km, gsh_15m). However, if a different WW3 grid configuration is used, the dependencies will be incorrect.

WalterKolczynski-NOAA:
Adds remaining coupled jobs to the free forecast workflow, and additionally addresses a number of other issues out of convenience during the process.

There are two new jobs: one to stage coupled ICs for the prototype, and one to run ocean/ice post. The initialization jobs are now controlled by the APP setting: if the app is S2S or S2SW, the coupled IC job will run in place of the getic and init jobs. Ocean/ice post is a metatask that pools forecast hours in the same manner as atm post, with the number of tasks controlled by the setting in config.ocnpost. These new jobs are only added for the forecast-only workflow.

As part of the addition, a couple of changes were made to how the model output is handled. First, symbolic links are now set up for each of the ocean and ice output files that point to the location on COM, in the same way that is done for atm and wave output, rather than copying the data when the forecast is complete. This allows the data to become available immediately and post to begin while the forecast is still running. Additionally, the output location on COM is moved to component files in the same manned as atm and wave. The joint ocn-ice files are currently located in the ocean directory; a final disposition of these files is under discussion.

While the ocean post jobs were being created, the method of naming the tasks is updated to include the first and last forecast hour processed by the task, rather than a plain index. This change was extended to the regular atm post, including for the full cycled workflow.

To fix the wave post dependencies for coupled mode, two changes were made. First, the wave post data dependency is updated from being hard-coded to the GFS v16 wave grids to reading the wave grid names from waveGRD. Second, a new setting, DOBNDPNT_WAVE, is brought over from coupled-crow to turn off the boundary point post jobs. Since these are changes to existing jobs, the changes were made to the full cycled workflow as well (but need to be tested).

Closes #191, #367, #461, #481, #482, #483, #484, #485, #486, #487


WalterKolczynski-NOAA:
A full coupled experiment was tested on Orion. A workflow was generated and manually inspected but not run for atm-only. I'm more concerned about the cycling being bugged by these changes, but cycled is not yet confirmed to work from this branch yet, so that testing is deferred. In particular, need to make sure the analysis job is picked up now that the FHRGRP should be anl instead of 0.

jobs/rocoto/coupled_ic.sh
```diff
@@ -0,0 +1,111 @@
...
+###############################################################
+
+###############################################################
+# Source FV3GFS workflow modules
+. $HOMEgfs/ush/load_fv3gfs_modules.sh
+status=$?
+[[ $status -ne 0 ]] && exit $status
+err=0
+
+export DATAROOT="$RUNDIR/$CDATE/$CDUMP"
```
aerorahul: Do you need `export` here?

jobs/rocoto/ocnpost.sh
```diff
@@ -124,25 +121,25 @@ for fhr in $fhrlst; do
   
 
     #break up ocn netcdf into multiple files:  
-    if [ -f $COMOUT/ocn_2D_$VDATE.$ENSMEM.$IDATE.nc ]; then 
-      echo "File $COMOUT/ocn_2D_$VDATE.$ENSMEM.$IDATE.nc already exists" 
+    if [ -f $COMOUT/ocean/ocn_2D_$VDATE.$ENSMEM.$IDATE.nc ]; then 
+      echo "File $COMOUT/ocean/ocn_2D_$VDATE.$ENSMEM.$IDATE.nc already exists"
     else
-      $ncks -x -v vo,uo,so,temp $COMOUT/ocn$VDATE.$ENSMEM.$IDATE.nc $COMOUT/ocn_2D_$VDATE.$ENSMEM.$IDATE.nc
+      ncks -x -v vo,uo,so,temp $COMOUT/ocean/ocn$VDATE.$ENSMEM.$IDATE.nc $COMOUT/ocean/ocn_2D_$VDATE.$ENSMEM.$IDATE.nc
```
aerorahul, line 127: can we do a status check here and everywhere we perform this `ncks` business?
```
err=$?
[[ $err-ne 0 ]] && exit $err
```

jobs/rocoto/ocnpost.sh
```diff
@@ -89,13 +89,10 @@ else
     export COMIN="$ROTDIR/$CDUMP.$PDY/$cyc"
     export COMOUT="$ROTDIR/$CDUMP.$PDY/$cyc"
 fi
-[[ ! -d $COMOUT ]] && mkdir -m 775 -p $COMOUT
+[[ ! -d $COMOUT/ocean ]] && mkdir -m 775 -p $COMOUT/ocean
+[[ ! -d $COMOUT/ice ]] && mkdir -m 775 -p $COMOUT/ice
```
aerorahul: why is the `-m 755` necessary?

jobs/rocoto/ocnpost.sh
```diff
@@ -124,25 +121,25 @@ for fhr in $fhrlst; do
...
-    if [ -f $COMOUT/ocn-uo-EQ_$VDATE.$ENSMEM.$IDATE.nc ]; then 
-       echo "File $COMOUT/ocn-uo-EQ_$VDATE.$ENSMEM.$IDATE.nc already exists" 
+    if [ -f $COMOUT/ocean/ocn-uo-EQ_$VDATE.$ENSMEM.$IDATE.nc ]; then 
+       echo "File $COMOUT/ocean/ocn-uo-EQ_$VDATE.$ENSMEM.$IDATE.nc already exists" 
     else 
-      $ncks -v uo -d yh,503 -d xh,-299.92,60.03 $COMOUT/ocn_3D_$VDATE.$ENSMEM.$IDATE.nc $COMOUT/ocn-uo-EQ_$VDATE.$ENSMEM.$IDATE.nc
+      ncks -v uo -d yh,503 -d xh,-299.92,60.03 $COMOUT/ocean/ocn_3D_$VDATE.$ENSMEM.$IDATE.nc $COMOUT/ocean/ocn-uo-EQ_$VDATE.$ENSMEM.$IDATE.nc
     fi
     status=$?
     [[ $status -ne 0 ]] && exit $status
```
aerorahul: This will only trap tha last execution and is either inadequate or unnecessary if every file is present.

jobs/rocoto/ocnpost.sh
```diff
@@ -153,7 +150,7 @@ done
 # Restore CDATE to what is expected
 export CDATE=$IDATE
 echo $pwd
-$NMV ocn_ice*.grb2 $COMOUT
+$NMV ocn_ice*.grb2 $COMOUT/ocean/
 status=$?
 [[ $status -ne 0 ]] && exit $status
```
aerorahul, line 160: we are checking on every CP and MV now?

jobs/rocoto/ocnpost.sh
```diff
@@ -153,7 +150,7 @@ done
 # Restore CDATE to what is expected
 export CDATE=$IDATE
 echo $pwd
```
aerorahul: remove.

aerorahul: This line has no meaning.

ush/forecast_postdet.sh
```diff
@@ -797,57 +786,77 @@ MOM6_out() {
     export COMIN="$ROTDIR/$CDUMP.$PDY/$cyc"
     export COMOUT="$ROTDIR/$CDUMP.$PDY/$cyc"
   fi
-  [[ ! -d $COMOUT ]] && mkdir -m 775 -p $COMOUT
+  [[ ! -d $COMOUT/ocean ]] && mkdir -m 775 -p $COMOUT/ocean
```
aerorahul: `-m 775`, why?

ush/forecast_postdet.sh
```diff
@@ -916,60 +925,51 @@ CICE_postdet() {
...
-    $NCP -p $DATA/ice_in $COMOUT/
-    if [ $FHRGRP -eq 0 ]; then
-      fhrlst="anl"
-    else
-      fhrlst=$(echo $FHRLST | sed -e 's/_/ /g; s/\[/ /g; s/\]/ /g; s/f/ /g; s/,/ /g')
-    fi
+  # Link output files
+  export ENSMEM=${ENSMEM:-01}
+  export IDATE=$CDATE
+  [[ ! -d $COMOUT/ice ]] && mkdir -m 775 -p $COMOUT/ice
```
aerorahul: `-m 775`, why?

ush/forecast_postdet.sh
```diff
@@ -797,57 +786,77 @@ MOM6_out() {
...
+  [[ ! -d $COMOUT/ocean ]] && mkdir -m 775 -p $COMOUT/ocean
 
-  if [ $inistep = 'cold' ]; then
-    cp $DATA/ufs.cpld.cold.cpl.r.*.nc $COMOUT/
-    cp $DATA/rpointer.cpl $COMOUT/
+  if [ ${inistep:-"restart"} = 'cold' ]; then
+    $NLN $COMOUT/ocean/ufs.cpld.cold.cpl.r.*.nc $DATA/ufs.cpld.cold.cpl.r.*.nc
+    $NLN $COMOUT/ocean/rpointer.cpl $DATA/rpointer.cpl
     status=$?
     exit $status
```
aerorahul: Linking a non-existent file will create a dead link.  The status will always be 0, no?

ush/forecast_postdet.sh
```diff
@@ -797,57 +786,77 @@ MOM6_out() {
...
-        year=`echo $m_date | cut -c1-4`
-        month=`echo $m_date | cut -c5-6`
-        day=`echo $m_date | cut -c7-8`
-        hh=`echo $m_date | cut -c9-10`
-
-        export ocnfile=ocn_${year}_${month}_${day}_${hh}.nc
-
-        echo "$NCP -p $ocnfile $COMOUT/ocn$p_date.$ENSMEM.$IDATE.nc"
-        $NCP -p $ocnfile $COMOUT/ocn$p_date.$ENSMEM.$IDATE.nc
+      if [ $fhr = 'anl' ]; then
```
aerorahul, line 798: Suggestion, may be for future if it is too much in this PR.
Can we move this to an e.g. `forecast_postdet_ocean.sh`?

aerorahul (changes requested):
I made a pass through the shell scripts.  I did not review the python scripts and will do so in a bit.
Few comments:
1. `ncks` commands could be checked for return status.
2. `-m 775` is unnecessary and if it is, then should be done everywhere.
3.  Some minor points.

All-in-all, looking good.  Thanks for doing this "overnight"

ush/rocoto/workflow_utils.py
```diff
@@ -113,6 +113,9 @@ def source_configs(configs, tasks):
         elif task in ['efcs']:
             files.append(find_config('config.fcst', configs))
             files.append(find_config('config.efcs', configs))
+        elif task[0:4] in ['wave']:
```
aerorahul: hmm ... substring?

aerorahul:
```suggestion
        elif 'wave' in task:
```

ush/rocoto/setup_workflow_fcstonly.py
```diff
@@ -250,55 +254,97 @@ def get_workflow(dict_configs, cdump='gdas'):
...
-      dep_dict = {'type':'data', 'data':data}
-      deps.append(rocoto.add_dependency(dep_dict))
-      dependencies = rocoto.create_dependency(dep_condition='nor', dep=deps)
+    if app in ['S2S', 'S2SW']:
+        # Copy prototype ICs
+        deps = []
+        base_cplic = dict_configs['coupled_ic']['ORIGIN_ROOT']
+
+        # ATM ICs
+        for file in ['gfs_ctrl.nc'] + [f'{a}_data.tile{n}.nc' for a in ['gfs', 'sfc'] for n in range(1, 7)]:
```
aerorahul: I am guilty of this comment, but I have learned that use of single character variables e.g. `a`  and `n` is generally not a good practice.
Going forward, I think we should avoid them.  Replacing `a` with `aa` and `n` with `nn` or something like that does not add too much burden, but helps `grepping` later.

aerorahul: On another note, I assume `S2S` variants are only starting from an atmospheric cold-start.
Perhaps that is known.

ush/rocoto/setup_workflow_fcstonly.py
```diff
@@ -311,8 +357,9 @@ def get_workflow(dict_configs, cdump='gdas'):
         deps = []
         dep_dict = {'type': 'task', 'name': f'{cdump}waveinit'}
         deps.append(rocoto.add_dependency(dep_dict))
-        dep_dict = {'type': 'task', 'name': f'{cdump}init'}
-        deps.append(rocoto.add_dependency(dep_dict))
+        if app not in ['S2S', 'S2SW']:
```
aerorahul, line 360: I don't understand this conditional.
Can you please explain what is the objective here?

aerorahul (changes requested):
A coupled of comments and recommendation on the python scripts.

jobs/rocoto/ocnpost.sh
WalterKolczynski-NOAA: I was just following what was here before.

jobs/rocoto/ocnpost.sh
WalterKolczynski-NOAA: Agree this should happen. This wasn't something I changed.

jobs/rocoto/ocnpost.sh
WalterKolczynski-NOAA: This wasn't something I added.

jobs/rocoto/ocnpost.sh
WalterKolczynski-NOAA: This either.

ush/forecast_postdet.sh
WalterKolczynski-NOAA: Just following what was here before.

ush/forecast_postdet.sh
WalterKolczynski-NOAA: I think so. The check is left over from it being a copy. Will remove.

ush/forecast_postdet.sh
WalterKolczynski-NOAA: I have no problem with that, but we should break the file up for each component to be consistent.

ush/forecast_postdet.sh
WalterKolczynski-NOAA: As before.

ush/rocoto/setup_workflow_fcstonly.py
WalterKolczynski-NOAA: I generally avoid single-character comments in general unless it is a one-line for loop like this. But I have no problem changing it. I'm pretty sure you are correct on the cold-start, at least for the time being.

ush/rocoto/setup_workflow_fcstonly.py
WalterKolczynski-NOAA: The {cdump}init job does not run for S2S modes, so waveinit can't depend on it. 

I'd love for any future workflow generation utilities to have a validation step that just removes dependencies on non-existent jobs like this.

ush/rocoto/workflow_utils.py
WalterKolczynski-NOAA: Didn't occur to me I could do it this way. My brain clearly still hasn't fully adapted to the pythonic ways.

WalterKolczynski-NOAA (commented):
I knew I wasn't going to be able to sleep before I got it all committed anyway.

Most of the comments seem to be about stuff I didn't actually change, but I have no problem fixing. Probably should be a new PR though, this one is already bloated with too many fixes.

ush/forecast_postdet.sh
WalterKolczynski-NOAA: I guess it will be non-zero if it fails to create the link. Not sure if that is worth checking though, since if it can't write to $DATA, there are loads of other problems.

jobs/rocoto/ocnpost.sh
WalterKolczynski-NOAA: Fixed

jobs/rocoto/ocnpost.sh
WalterKolczynski-NOAA: Fixed

ush/forecast_postdet.sh
WalterKolczynski-NOAA: Fixed

ush/rocoto/workflow_utils.py
WalterKolczynski-NOAA: Fixed

jobs/rocoto/ocnpost.sh
aerorahul: I suggest removing it.

jobs/rocoto/ocnpost.sh
aerorahul: true.
But you can remove.

jobs/rocoto/ocnpost.sh
aerorahul: I leave it to you.  I would suggest removing this instance of status check for move.
OTOH, would creating soft-links help "making these files available faster".
Probably can defer that for a later PR.

ush/forecast_postdet.sh
aerorahul: remove.

ush/forecast_postdet.sh
aerorahul: remove.

ush/rocoto/setup_workflow_fcstonly.py
aerorahul: Ok.

aerorahul (commented):
getting closer.

jobs/rocoto/ocnpost.sh
WalterKolczynski-NOAA: Going to wait for @JessicaMeixner-NOAA to chime in just in case the more open permissions are needed for the prototype work. Even if it is needed, this is probably not the best ones to use. For reference, atmos has the following permissions: `rwxr-s---`, compared to the `rwxrwsr-x` for the other directories (wave also uses these broader permissions).

ush/rocoto/setup_workflow_fcstonly.py
WalterKolczynski-NOAA: Replaced with more verbose indicies

jobs/rocoto/ocnpost.sh
JessicaMeixner-NOAA: I don't know of an issue/reason for needing this. Thanks for checking w/me.

WalterKolczynski-NOAA: Done, though FWIW the wave mkdirs also set the permissions, so we'll want to revisit that later.

jobs/rocoto/ocnpost.sh
WalterKolczynski-NOAA: Removed

ush/forecast_postdet.sh
WalterKolczynski-NOAA: done

ush/forecast_postdet.sh
WalterKolczynski-NOAA: done

jobs/rocoto/ocnpost.sh
WalterKolczynski-NOAA: Now that there is proper exit code checking for all the ncks calls, I'm going to leave this so everything is being checked.

aerorahul (approved):
Thanks for addressing the comments.
Code changes look good.
Lets see what the tests reveal.

jobs/rocoto/coupled_ic.sh
WalterKolczynski-NOAA: Actually, we don't need this at all. The directory is created and then never used.

JessicaMeixner-NOAA (approved):
@WalterKolczynski-NOAA I don't think I'm going to get a chance to test this today - but all the changes look good to me so I'm going to go ahead and approve.  I'll try to test next week when I get a chance.  Also --- THANK YOU! Splitting the ocn/ and ice/ to their respective com directories is awesome - as is allowing ocean post to run as the model progresses.  Great updates along with just bringing over these jobs.

WalterKolczynski-NOAA:
Retested coupled on both orion and hera, and also atm-only free forecast on orion. All looks good.

PR #490, opened 2021-11-08
Updates for version files

KateFriedman-NOAA:
Several updates related to new versions files:
* Add build.ver and run.ver to new versions subfolder
* Adjust path to run.ver in downstream scripts (did not test change)
* Change module/library versions in module_base.wcoss2 to variables that come from above
* A few updates to run.ver based on code manager feedback
* Update $FIX_DIR to newer temporary global location (not enough space in EIB space to hold all needed fix sets)
* Update to new GLDAS tag that includes version variable overrides


BoiVuong-NOAA (approved):
look good

KateFriedman-NOAA:
@aerorahul @WalterKolczynski-NOAA I further updated this PR to include updates to convert module_base.wcoss2 into LUA format (now named module_base.wcoss2.lua). Please review this PR and let me know how the LUA changes look. I tested the new module_base.wcoss2.lua in a gfspost job on Cactus and it correctly loaded the runtime modules. Thanks!

@aerorahul The component code managers still need to be poked about converting to LUA as well, thanks!

modulefiles/module_base.wcoss2.lua
```diff
@@ -0,0 +1,62 @@
+help([[
+Load environment to run GFS on WCOSS2
+]])
+
+PrgEnv_intel_ver=os.getenv("PrgEnv_intel_ver")
```
WalterKolczynski-NOAA: Since these are each only being used once, do we want to just use os.getenv in-line in the load statements instead of loading them to variables first?

WalterKolczynski-NOAA:
When this comes back to develop, we should update all the modules to be lua.

modulefiles/module_base.wcoss2.lua
KateFriedman-NOAA: Addressed.

KateFriedman-NOAA:
> When this comes back to develop, we should update all the modules to be lua.

Definitely! I plan to take lessons learned and results of the changes for the ops side into the same updates for develop. Haven't started that effort yet.

WalterKolczynski-NOAA (approved):

KateFriedman-NOAA:

PR #497, opened 2021-11-12
Update ush/ script files with "python3" in shebang

### LINKED ISSUE #471: Update shebang for python3 scripts
We updated the python scripts to python3, but did not update the shebang at the top to ensure python3 is used. This causes an issue for users whose python defaults to python2. We should update the shebang to `/usr/bin/env python3`.

jkuang-exploro:
We updated the python scripts to python3, but did not update the shebang at the top to ensure python3 is used. This causes an issue for users whose python defaults to python2. This PR updates ush/ scripts to include "python3" in the shebang.

### DIFF
--- a/ush/rocoto/rocoto.py
+++ b/ush/rocoto/rocoto.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 '''
     MODULE:
--- a/ush/rocoto/setup_expt.py
+++ b/ush/rocoto/setup_expt.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 ###############################################################
 # < next few lines under version control, D O  N O T  E D I T >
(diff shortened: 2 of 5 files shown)


WalterKolczynski-NOAA (changes requested):
Code looks fine, but there's no commit message for the PR and the title needs work.

WalterKolczynski-NOAA:
Please compare this title and commit message with the issue that it is fixing. The issue has a description of _why_ the change is being made. Commit messages are supposed to describe the why as well. We can tell the what by looking at the diff, the commit needs to tell us why we bothered.

The title should begin with a capital letter, and the grammar is ambiguous; it can be read as you updated files that _had_ python3 in the shebang, rather than updating them so they now have it.

jkuang-exploro:
@WalterKolczynski-NOAA Does it looks OK now?

WalterKolczynski-NOAA:
Much better

WalterKolczynski-NOAA (approved):

PR #499, opened 2021-11-12
Only read from wavepostsbs when waves are on

WalterKolczynski-NOAA:
Workflow generation would fail when waves weren't on because the
key 'wavepostsbs' wasn't defined when reading the whethern to run
wave boundary point jobs from the configs. Reading of other compon-
ent specific settings is not an issue because they are all behind
checks that the relevant component is on.

Refs: #486


WalterKolczynski-NOAA:
Though this change is just fixing an unrelated bug I introduced last week, after this change cycled mode looks to be working based on a 2.5 cycle test on Orion. @KateFriedman-NOAA is going to double-check the output to confirm.

KateFriedman-NOAA (approved):
Looks good, thanks!

PR #503, opened 2021-11-15
feature/coupled-crow: Updating p7.2 case file settings 

JessicaMeixner-NOAA:
Reverting FSICL and FSICS to 99999 from 0 as this was determined to be the source of the earlier model crashes.

### DIFF
--- a/workflow/cases/prototype7.2-aero.yaml
+++ b/workflow/cases/prototype7.2-aero.yaml
@@ -49,8 +49,8 @@ case:
     nst_anl: yes
     psm_bc: 1
     dddmp: 0.1
-    FSICL: 0
-    FSICS: 0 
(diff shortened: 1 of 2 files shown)


aerorahul (approved):
Assuming this was tested and model is happy w/ the new setting.

aerorahul:
@JessicaMeixner-NOAA Will this need to be covered/reproduced in #500?

JessicaMeixner-NOAA:
Yes, it has been tested -- thanks @aerorahul

JessicaMeixner-NOAA:
I'll double check, but I think this was actually not ever pushed to #500, so it actually makes it match better

JessicaMeixner-NOAA:
Confirmed that this is not in PR#500

PR #506, opened 2021-11-17
Update readme with coupled instructions

WalterKolczynski-NOAA:
Add alternative commands for coupled mode to quick-start instructions.
Also add link to full instructions in wiki.


WalterKolczynski-NOAA:
Nevermind, can do this directly

PR #509, opened 2021-11-19
update so that coupled matches p7.2

JessicaMeixner-NOAA:
* Update ATM and WAV IC variables. 
* Update cdmbgwd for C384 (@yangfanglin said make this the new default, this is what's being used in p7.2) 
* Updates for waves to produce restart file at the end of the run and fix ice/current grids

### DIFF
--- a/parm/config/config.wave
+++ b/parm/config/config.wave
@@ -82,22 +82,18 @@ fi
 
 # Restart timing business
 
-
+export RSTTYPE_WAV='T'          # generate second tier of restart files
 if [ "${CDUMP}" != gfs ]; then    # Setting is valid for GDAS and GEFS 
-  export RSTTYPE_WAV='T'          # generate second tier of restart files
   export DT_1_RST_WAV=10800       # time between restart files, set to DTRST=1 for a single restart file
   export DT_2_RST_WAV=43200       # restart stride for checkpointing restart
   export RSTIOFF_WAV=0            # first restart file offset relative to model start
 else                              # This is a GFS run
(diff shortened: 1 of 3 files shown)


JessicaMeixner-NOAA:
One remaining issue for the forecast job to match p7.2 (+ CA which was unintentionally turned off in feature/coupld-crow) is that the IC is not copied in for waves. I'm still working on this but it's not going to be a quick fix so I thought I'd push these updates now.

parm/config/config.wave
```diff
@@ -93,9 +91,16 @@ if [ "${CDUMP}" != gfs ]; then    # Setting is valid for GDAS and GEFS
 else                              # This is a GFS run
   rst_dt_gfs=$(( restart_interval_gfs * 3600 ))
   export RSTTYPE_WAV='F'                 # generate second tier of restart files
-  if [ $rst_dt_gfs -gt 0 ]; then export RSTTYPE_WAV='T' ; fi
-  export DT_1_RST_WAV=${rst_dt_gfs:-0}   # time between restart files, set to DTRST=1 for a single restart file
-  export DT_2_RST_WAV=${rst_dt_gfs:-0}   # restart stride for checkpointing restart
+  if [ $rst_dt_gfs -gt 0 ]; then 
+    export RSTTYPE_WAV='T'
```
WalterKolczynski-NOAA: This is in both branches of the if, so it should be placed before the branch instead of L93

WalterKolczynski-NOAA (changes requested):

WalterKolczynski-NOAA (approved):

PR #511, opened 2021-11-24
Updates for coupled forecast

JessicaMeixner-NOAA:
This updates: 
* WW3 IC locations for S2S experiments 
* removes some old logic for the two-step cold starts that were also causing the "cannot copy" messages in the log files that @KateFriedman-NOAA pointed out 
* Also adds a protection for copying in the wave rmp*.nc files, again to avoid "cannot copy" messages 
* Most changes are in `whitespace`, so hide them to see actual script changes.


JessicaMeixner-NOAA:
Checking the post is still required - but this should get the coupled model tests very close to ready to go.  I'll finish up the tests Nov 29th

WalterKolczynski-NOAA (approved):

Issue #397, opened 2021-08-03
Cleanup and standardize shell scripts


WalterKolczynski-NOAA:
The style of the shell scripts is haphazard and difficult to read. Indentation is inconsistent and where it is correct is often broken by heredocs. There is also no standard pre-able/post-amble to scripts or PS1 specifications. Even if these scripts will be replaced in the future, it would aid the porting process for these scripts to be more readable.

These lists are drafts and also open to suggestions.

First-pass style updates:
- [x] Convert all scripts to bash
- [x] Replace all backticks with $( )
- [x] Use standard pre-amble/post-amble to all scripts (see below)

Suggested pre-amble:
```
echo "${date -u) begin $(basename $BASH_SOURCE)"

set -x
if [[ ${STRICT:-NO} == "YES" ]]; then
    # Exit on error and undefined variable
    set -eu
fi
# Print the script name and line number of each command as it is executed
export PS4='$SECONDS + $(basename $(basename $BASH_SOURCE)[$LINENO] '
```

Suggested post-amble:
```
echo "$(date -u) end $(basename $BASH_SOURCE)"
```

Second-pass style updates:
- [ ] Convert all shell scripts to [Google style](https://google.github.io/styleguide/shellguide.html)
- [ ] Use printf instead of other methods to zero-pad

KateFriedman-NOAA:
@DavidHuber-NOAA FYI, I have addressed shellcheck concerns in the `/env` files for all platforms in `develop`. Here is the compare: https://github.com/NOAA-EMC/global-workflow/compare/develop...KateFriedman-NOAA:global-workflow:feature/shellcheck
Note there are also changes to the `JET.env` and `S4.env`. I'm currently testing these updates on Orion (so far so good) and building my branch on Hera and WCOSS2 for tests there before opening a PR into `develop`. Let me know if you have any concerns about changes in the Jet and S4 env files. Thanks!

DavidHuber-NOAA:
@KateFriedman-NOAA Thanks for making the S4 and Jet changes!  They look good to me.

PR #526, opened 2021-12-11
Replace all backticks for command substitution

WalterKolczynski-NOAA:
All instances where backticks are used for command substitution are
replaced with $( ). This standardizes usage around $( ), which can be 
nested and does not risk being confused with single quotation marks.

Refs: #397


WalterKolczynski-NOAA:
Coupled and cycled both pass on Dell except archive jobs, which are failing for an unrelated reason.

WalterKolczynski-NOAA:
Can we get this approved today?

DanIredell-NOAA (approved):

PR #489, opened 2021-11-04
GFSv16.1.5 - GSI update to include IASI Metop-C in operations

KateFriedman-NOAA:
RFC 8823 - On WCOSS, update GFS to v16.1.5 which includes:
* Turn on the active assimilation of Metop-C IASI (Infrared Atmospheric Sounding Interferometer) data with correlated observation errors
* Update correlated observation errors for Metop-B IASI, N20 CrIS (Cross-track Infrared Sounder), and NPP CrIS
* Put in a minimum threshold for saturation specific humidity to avoid potential minimization issues.
* Add the 2022 CO2 projection files.

Implemented on October 26 at 1430Z.

Resolves #439
Close #439

PR #530, opened 2021-12-17
Update GSI version to 2021 Dec 14

WalterKolczynski-NOAA:


WalterKolczynski-NOAA:
All tasks completed successfully on Dell.

Output: `/gpfs/dell2/emc/modeling/noscrub/Walter.Kolczynski/global-workflow/gsi_update`

WalterKolczynski-NOAA:
Would appreciate if this can get verified and approved today, before I'm technically on leave. Code is only one line, mostly want quick verification of output.

CatherineThomas-NOAA:
> Would appreciate if this can get verified and approved today, before I'm technically on leave. Code is only one line, mostly want quick verification of output.

Understood.  I'll get to this today.

CatherineThomas-NOAA:
Thanks @WalterKolczynski-NOAA for running both the gsi update test as well as pointing me to an appropriate control.  As stated, the code differences for the workflow are very minor, but the impact of changing a tag could be large.

First outer loops are identical for many variables (ps, T, q).  There are no resets and the increments look reasonable.  Differences that I found, which are expected with the GSI update:

- qcpen in first outer loop is different for uv 
- Top 5 layers of ompsnp are on
- Ompstc8 n20 on
- IASI-A is replaced with IASI-C
- CrIS NPP is off
- More observation in lower levels of GPSRO
- More SST obs

One thing I realized though is that some of these observation changes probably shouldn’t have been used for this set of initial conditions (2020083000).  Looking at parm/config in develop as well as the PR, the config.anal changes from v16.1.5 are not there, nor are v16.1.3 and v16.1.4.  It looks like these changes were merged to the operations branch (#489), but not develop.  Was there a reason for this?   This is a problem for outside this PR, so I won't block it but it should be addressed.

CatherineThomas-NOAA (approved):

WalterKolczynski-NOAA:
> As stated, the code differences for the workflow are very minor, but the impact of changing a tag could be large.

Sorry, I didn't mean to imply otherwise. Obviously updating the GSI could have big impacts; I was just saying that the changes in the g-w code itself were small because no changes beyond the updated hash were needed, thus the code review is simple.

> One thing I realized though is that some of these observation changes probably shouldn’t have been used for this set of initial conditions (2020083000). Looking at parm/config in develop as well as the PR, the config.anal changes from v16.1.5 are not there, nor are v16.1.3 and v16.1.4. It looks like these changes were merged to the operations branch (#489), but not develop. Was there a reason for this? This is a problem for outside this PR, so I won't block it but it should be addressed.

This is a question for @KateFriedman-NOAA as to why these changes weren't brought back to develop. We'll make sure to get any needed updates in after the holidays.

CatherineThomas-NOAA:
@WalterKolczynski-NOAA @KateFriedman-NOAA 

Following up on a thread from before the holidays.  The config updates from v16.1.3, v16.1.4, and v16.1.5 were only merged into the operations branch and not develop.  Was this intentional or an oversight?

aerorahul:
@CatherineThomas-NOAA 
It was likely an oversight or something else became higher priority and this slipped under the radar.
Once we have WCOSS2 port complete and handed off, we will find a time to prioritize the `config` updates from the missed implementations.

WalterKolczynski-NOAA:
@CatherineThomas-NOAA this was just me not reading your comment carefully enough in the rush to complete before the holidays.

CatherineThomas-NOAA:
Thanks @aerorahul and @WalterKolczynski-NOAA.  Understood concerning WCOSS2 priorities.

PR #538, opened 2021-12-23
Updates for P8a

### LINKED ISSUE #525: Updates to run mini-P8A
Update the global-workflow to enable P8-A, which includes: 
* Updating the ufs-weather-model version (after PR https://github.com/ufs-community/ufs-weather-model/pull/937 update: merged 12/13)
* Turn on ice-albedo feedback in atm (Requires changing input.nml  to set use_cice_alb=true in &gfs_physics_nml ) 
* Updates to CA namelists 
```
 DO_CA=.true.
CA_SGS=.true.
CA_GLOBAL=.false.
NCA=1
NCELLS=5
NLIVES=12
NTHRESH=18
NSEED=1
NFRACSEED=0.5
CA_TRIGGER=.true.
NSPINUP=1
ISEED_CA=12345
```
* Updates for NOAH-MP which require input.nml update for to export iopt_sfc="3" and modifying parm_fv3diag to include pahi, pah_ave, ecan_acc, etran_acc,edir_acc,wa_acc, lfrac to the grib outputs "gfs_phys",    "pahi",          "pahi",          "fv3_history2d",  "all",  .false.,  "none",  2

JessicaMeixner-NOAA:
This PR updates ufs-weather-model to the 12/32/21 commit and the UPP hash.  Coupled settings are updated to run the P8a mini prototype.  

fixes #525

Updates include: 
* Turn on ice-albedo feedback in atm (Requires changing input.nml  to set use_cice_alb=true in &gfs_physics_nml ) 
* Updates to CA namelists 
* Updates for NOAH-MP which require input.nml update for to export iopt_sfc="3" and modifying parm_fv3diag to include pahi, pah_ave, ecan_acc, etran_acc,edir_acc,wa_acc, lfrac to the grib outputs "gfs_phys",    "pahi",          "pahi",          "fv3_history2d",  "all",  .false.,  "none",  2


parm/config/config.defaults.s2sw
```diff
@@ -15,14 +15,16 @@ FHOUT_HF_GFS=-1
 IALB=2
 IEMS=2
 DO_CA="YES"
-min_seaice="1e-11"
+min_seaice="1e-6"
+use_cice_alb=".true."
+export iopt_sfc="3"
```
WalterKolczynski-NOAA, line 20: Is there a reason this needs an export when the others don't?

ush/parsing_namelists_FV3.sh
```diff
@@ -327,8 +328,9 @@ EOF
 if [ $cpl = .true. ]; then
   cat >> input.nml << EOF
   frac_grid    = ${FRAC_GRID:-".true."}
-  cplchm = ${cplchem:-".false."}
+  cplchm       = ${cplchem:-".false."}
   cplflx       = $cplflx
+  cplice       = ${cplice}
```
WalterKolczynski-NOAA, line 333: this should be controlled by $DO_ICE, which is set in config.base based on the app

ush/parsing_model_configure_FV3.sh
```diff
@@ -19,7 +19,7 @@ fi
 
 rm -f model_configure
 cat >> model_configure <<EOF
-print_esmf:              ${print_esmf:-.true.}
+print_esmf:              .true.
```
WalterKolczynski-NOAA: This should remain an option and not be hard-coded. Many users do not need the PET logs.

ush/nems_configure.sh
```diff
@@ -48,6 +48,8 @@ else
   EPS_IMESH='1.0e-1'
 fi
 
+ATMTILESIZE=$(echo $CASE | cut -c2-)
```
WalterKolczynski-NOAA: A variable with the same function is already available: $RESTILE (set in config.fcst). If the variable isn't currently available to this script, add an export there an switch to that variable.

ush/nems_configure.sh
```diff
@@ -86,6 +88,7 @@ if [ $cplflx = .true. ]; then
   sed -i -e "s;@\[RESTART_N\];$restart_interval_nems;g" tmp1
   sed -i -e "s;@\[use_mommesh\];$USE_MOMMESH;g" tmp1
   sed -i -e "s;@\[eps_imesh\];$EPS_IMESH;g" tmp1
+  sed -i -e "s;@\[ATMTILESIZE\];$ATMTILESIZE;g" tmp1
```
WalterKolczynski-NOAA: Change as above.

WalterKolczynski-NOAA (changes requested):
A couple of small things. Also, I know 2021 felt long, but I don't think Dec had 32 days ;) Looks like it should be the 23rd.

parm/config/config.defaults.s2sw
JessicaMeixner-NOAA: Some of the other variables are exported.  I did this because it wasn't over-writing a value in another config file, so I wanted to make sure it'd be exported so it would be used when creating the input.nml file.  I can try to remove the export statement and see if this still gets set appropriately though.

ush/nems_configure.sh
JessicaMeixner-NOAA: Thanks! I looked everywhere for an existing variable and didn't find one.

ush/parsing_model_configure_FV3.sh
JessicaMeixner-NOAA: Thanks for catching this! Will be reverted, this was unintentional.

ush/parsing_namelists_FV3.sh
JessicaMeixner-NOAA: This is controlled by DO_ICE, here: https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/config/config.fcst#L61-L63

JessicaMeixner-NOAA:
@WalterKolczynski-NOAA I'm checking if we can remove the export line otherwise, I think I have addressed your comments.

WalterKolczynski-NOAA:
I'm going to set up a cycled test just to make sure it still works.

WalterKolczynski-NOAA:
One thing that occurred to me overnight that I missed in the original review: you're updating the post hash, but the post being used by UFS is not changed.

JessicaMeixner-NOAA:
It makes it consistent with the hash used by the UFS model though: 
UFS commit: 
https://github.com/ufs-community/ufs-weather-model/tree/803e03adcf14f2b108b013fd1b54055ec2124caa 
FV3: 
https://github.com/NOAA-EMC/fv3atm/tree/8afe51e12a6f93b022bf1d2cf97b4d0784748f7f
UPP: 
https://github.com/NOAA-EMC/UPP/tree/c939eae6bacb3c2a93753bba54b8646f32a0a7ab

Which is what the UPP module was updated to.

WalterKolczynski-NOAA:
Oh I see. I didn't see it at the top level, so I thought the module in ufs.common was still being used. But it is under FV3.

JessicaMeixner-NOAA:
UPP is now a submodule of FV3 instead of a module  (see commit: https://github.com/ufs-community/ufs-weather-model/commit/9c78af19559e96d833de831e76898e5819fc87b3#diff-604fbc7b745d7249a571c8484dbe95132d8b915a7e726c72365ed799ef2895b8)

WalterKolczynski-NOAA (approved):

PR #540, opened 2021-12-28
Retire jlogfile from scripts

KateFriedman-NOAA:
This PR retires the reference and usage of the jlogfile from scripts owned by global-workflow. This is per a request from NCO during the WCOSS2 port.

- jlogfile postmsg replaced by echo to stdout/stderr (NCO request)
- cleanup surrounding lines and messaging as needed
- add a few missing error exits after fatal errors

Refs: #399

### DIFF
--- a/ush/syndat_qctropcy.sh
+++ b/ush/syndat_qctropcy.sh
@@ -90,12 +90,6 @@ files_override=${files_override:-""}
 cd $DATA
 
 echo "Tropical Cyclone tcvitals QC processing has begun"
-set +x
-echo
-echo $msg
-echo
-set -x
-echo $msg >> $pgmout
 
 if [ "$#" -ne '1' ]; then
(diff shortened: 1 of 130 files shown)


ush/syndat_qctropcy.sh
```diff
@@ -336,16 +281,12 @@ echo
 set -x
 
 if [ -s current ]; then
-   msg="program  SYNDAT_QCTROPCY  completed normally - tcvitals records \
+   echo"program  SYNDAT_QCTROPCY  completed normally - tcvitals records \
```
WalterKolczynski-NOAA: missing space after echo

WalterKolczynski-NOAA (changes requested):

ush/syndat_qctropcy.sh
KateFriedman-NOAA: fixed, thanks for catching!

WalterKolczynski-NOAA (approved):

PR #542, opened 2021-12-30
Fix determination of GDAS forecast decomposition for workflow

### LINKED ISSUE #541: Overwriting GDAS forecast properties in config.resources
There is a small bug in config.resources that causes the default properties of `layout_x`, `layout_y`, `WRITE_GROUP`, and `WRTTASK_PER_GROUP` to be overwritten by the GFS properties when the config file is referenced by `setup_workflow.py` ([see lines 189-197](https://github.com/NOAA-EMC/global-workflow/blob/033b1d8ef29b7c829e8131758135b372ae61de26/parm/config/config.resources#L189-L197)).  Since `CDUMP` is not defined when `setup_workflow.py` is executed, `CDUMP_LIST` is defined as `"gfs gdas"`.  So running through the `for` loop, `CDUMP` is defined as `gfs` on the first iteration, then `gdas` on the second.  Thus, the aforementioned variables are overwritten before proceeding with `gdas`.  The simple fix is to rearrange `CDUMP_LIST` to `"gdas gfs"`.

DavidHuber-NOAA:
By determining the resources for GDAS forecast first, the GDAS grid decomposition is no longer overwritten before it is used.

Closes #541

### DIFF
--- a/parm/config/config.resources
+++ b/parm/config/config.resources
@@ -187,7 +187,7 @@ elif [ $step = "fcst" ]; then
     fi
 
     # During workflow creation, we need resources for all CDUMPs and CDUMP is undefined
-    CDUMP_LIST=${CDUMP:-"gfs gdas"}
+    CDUMP_LIST=${CDUMP:-"gdas gfs"}
     for CDUMP in $CDUMP_LIST; do
         if [[ "$CDUMP" == "gfs" ]]; then
             export layout_x=$layout_x_gfs


WalterKolczynski-NOAA (approved):
This works for now, though really the whole thing needs to be overhauled so this can't happen.

WalterKolczynski-NOAA:
I'll merge this on Monday since I'm still on leave and I don't feel like writing the announcement message right now.

PR #544, opened 2022-01-03
Series of small WCOSS2 updates - round 3

KateFriedman-NOAA:
This PR includes the following WCOSS2 updates:

1. add NODESIZE to setup scripts and hardcode to use on only WCOSS2 for now (value=128)
2. WCOSS2.env updates for the gldas, efcs, and waveawipsbulls jobs
3. add `memory_wafsgcip="50GB"` to config.resources
4. setup_expt bug fixes for waveinit/waveprep jobs (missing syntax from python2 to python3 conversion)
5. convert gfs_bufr, fbwndgfs, fv3nc2nemsio, ,regrid_nemsio, storm_reloc, enkf_chgres_recenter, and gaussian_sfcanl builds to LUA module format
6. fix for adding awips jobs to xml when DO_AWIPS=YES (fix by @WalterKolczynski-NOAA )

Refs: #399


KateFriedman-NOAA:
I will be adding a few more changes to this PR today (at least the remaining LUA format changes).

ush/rocoto/rocoto.py
```diff
@@ -68,6 +68,7 @@ def create_task(task_dict):
     jobname = task_dict.get('jobname', 'demojob')
     account = task_dict.get('account', 'batch')
     queue = task_dict.get('queue', 'debug')
+    nodesize = task_dict.get('nodesize', '128')
```
aerorahul: Can we set this default to 1?

aerorahul (commented):
All changes are good.
Just 1 suggestion.  See below.

ush/rocoto/rocoto.py
WalterKolczynski-NOAA: Not sure it matters. This branch is only going to be used for ops. And at any rate, usually defaults are to the ops configuration.

ush/rocoto/setup_workflow.py
```diff
@@ -197,6 +197,8 @@ def get_definitions(base):
     if scheduler in ['slurm']:
         strings.append(f'''\t<!ENTITY PARTITION_SERVICE "{base['QUEUE_SERVICE']}">\n''')
     strings.append(f'\t<!ENTITY SCHEDULER  "{scheduler}">\n')
+    if machine in ['WCOSS2']:
+       strings.append(f'\t<!ENTITY NODESIZE  "128">\n')
```
WalterKolczynski-NOAA: Should this be based on the nodesize set above?

ush/rocoto/setup_workflow_fcstonly.py
```diff
@@ -130,6 +130,8 @@ def get_definitions(base):
     if scheduler in ['slurm']:
        strings.append(f'''\t<!ENTITY PARTITION_SERVICE "{base['QUEUE_SERVICE']}">\n''')
     strings.append(f'\t<!ENTITY SCHEDULER  "{scheduler}">\n')
+    if machine in ['WCOSS2']:
+       strings.append(f'\t<!ENTITY NODESIZE  "128">\n')
```
WalterKolczynski-NOAA: Should this be based on the nodesize set above?

WalterKolczynski-NOAA (changes requested):

KateFriedman-NOAA:
@WalterKolczynski-NOAA @aerorahul I have committed the remaining WCOSS2 LUA format modulefile updates that needed doing. Please also see those changes in this PR now.

Regarding NODESIZE in the setup scripts...the addition is not complete. For time and quick testing I did the latter half of the work to add NODESIZE into the rocoto xmls on WCOSS2. I hardcoded it for just WCOSS2 for now but need to do the first half of the work still = have NODESIZE/nodesize set based on the machine (parse the value from the env file or configs?). Will need to finish this on the dev side (if no time to finish for ops).

ush/rocoto/setup_workflow_fcstonly.py
KateFriedman-NOAA: It should...this is just hardcoded for ease/time. See comment below in PR.

ush/rocoto/setup_workflow.py
KateFriedman-NOAA: It should...this is just hardcoded for ease/time. See comment below in PR.

WalterKolczynski-NOAA:
> @WalterKolczynski-NOAA @aerorahul I have committed the remaining WCOSS2 LUA format modulefile updates that needed doing. Please also see those changes in this PR now.
> 
> Regarding NODESIZE in the setup scripts...the addition is not complete. For time and quick testing I did the latter half of the work to add NODESIZE into the rocoto xmls on WCOSS2. I hardcoded it for just WCOSS2 for now but need to do the first half of the work still = have NODESIZE/nodesize set based on the machine (parse the value from the env file or configs?). Will need to finish this on the dev side (if no time to finish for ops).

My issue is just that it is hardcoded in several places, instead of just one.

ush/rocoto/rocoto.py
KateFriedman-NOAA: Changed default to 1, thanks!

KateFriedman-NOAA:
> My issue is just that it is hardcoded in several places, instead of just one.

It's now hardcoded only once for each setup_workflow*.py script:
```
    if machine in ['WCOSS2']:
       strings.append(f'\t<!ENTITY NODESIZE  "128">\n')
```
If I find time I will try to get it parsing from a file and un-hardcode it. Will definitely resolve in the dev work if not for ops (where it isn't used anyway).

KateFriedman-NOAA:
@aerorahul @WalterKolczynski-NOAA I've committed the final changes I wanted to get into this PR (e.g. final LUA format conversions). Please continue your reviews, thanks!

WalterKolczynski-NOAA (approved):

ush/rocoto/setup_workflow_fcstonly.py
aerorahul: What is the else value for this `NODESIZE`.
If some one works with this branch on orion or any other, this will fail to put a `NODESIZE` value.  It is required now in the task.

KateFriedman-NOAA: Can one of you guys help finish adding NODESIZE to the setup scripts then? I don't think I'm going to have time before hand-off. If it's not remedied before hand-off then it can be sorted out in the feature/ops-orion branch when I update that (that's where we currently support ops on the R&Ds).

aerorahul:
> > My issue is just that it is hardcoded in several places, instead of just one.
> 
> It's now hardcoded only once for each setup_workflow*.py script:
> 
> ```
>     if machine in ['WCOSS2']:
>        strings.append(f'\t<!ENTITY NODESIZE  "128">\n')
> ```
> 
> If I find time I will try to get it parsing from a file and un-hardcode it. Will definitely resolve in the dev work if not for ops (where it isn't used anyway).

@KateFriedman-NOAA 
Please apply this patch.  I have not checked the nodesizes for all machines, since I don't know that off the top of my head.
This should atleast keep this generic.
[diff.txt](https://github.com/NOAA-EMC/global-workflow/files/7803753/diff.txt)

WalterKolczynski-NOAA:
npe_node_max is defined in the env file for every machine. Why are we not using that?

aerorahul:
> npe_node_max is defined in the env file for every machine. Why are we not using that?

We should grab the value for `nodesize` from `npe_node_max`, but it will require more work to make it available in the function `get_definitions`.  I leave it to you to accept/reject the patch.

ush/rocoto/setup_workflow_fcstonly.py
KateFriedman-NOAA: Incorporated @aerorahul patch, thanks! That gets us most of the way there and removes the WCOSS2 hardcoding. We will just need to get it reading from the env files later.

KateFriedman-NOAA:
Committed the NODESIZE patch (tested, works) and included a NATIVE update to add `-l place=vscatter` into the rocoto NATIVE xml settings. Also fixed a small syntax error (an errant f caused it to fail).

KateFriedman-NOAA:
Included fix for adding awips jobs to this PR, thanks for the fix @WalterKolczynski-NOAA ! Will now submit.

PR #546, opened 2022-01-05
Series of small WCOSS2 updates - round 5

KateFriedman-NOAA:
This PR contains the following WCOSS2 updates:

1. update WAFS tag to gfs_wafs.v6.2.4
2. update workflow_utils.py to a) not add vscatter native setting on WCOSS2 for archive jobs and b) correct bug that ignored nth_*_gfs variables (needed to set threads differently between gdasfcst and gfsfcst jobs)
3. update config.fv3.nco.static and config.resources.nco.static with resource settings based on WCOSS2 testing (further optimization still to come)
4. update FIX_DIR path in link_fv3gfs.sh to more official emc.global space (/lfs/h2/emc/global/save/emc.global/FIX/fix_nco_gfsv16)

Refs: #399

### DIFF
--- a/parm/config/config.resources.nco.static
+++ b/parm/config/config.resources.nco.static
@@ -73,7 +73,7 @@ elif [ $step = "wavepostsbs" ]; then
 elif [ $step = "wavepostbndpnt" ]; then
 
     export wtime_wavepostbndpnt="01:00:00"
-    export npe_wavepostbndpnt=280
+    export npe_wavepostbndpnt=448
     export nth_wavepostbndpnt=1
     export npe_node_wavepostbndpnt=$(echo "$npe_node_max / $nth_wavepostbndpnt" | bc)
     export NTASKS=${npe_wavepostbndpnt}
@@ -81,7 +81,7 @@ elif [ $step = "wavepostbndpnt" ]; then
 elif [ $step = "wavepostbndpntbll" ]; then
 
     export wtime_wavepostbndpntbll="01:00:00"
-    export npe_wavepostbndpntbll=280
+    export npe_wavepostbndpntbll=448
     export nth_wavepostbndpntbll=1
     export npe_node_wavepostbndpntbll=$(echo "$npe_node_max / $nth_wavepostbndpntbll" | bc)
     export NTASKS=${npe_wavepostbndpntbll}
@@ -89,7 +89,7 @@ elif [ $step = "wavepostbndpntbll" ]; then
 elif [ $step = "wavepostpnt" ]; then
 
     export wtime_wavepostpnt="01:30:00"
-    export npe_wavepostpnt=280
+    export npe_wavepostpnt=448
     export nth_wavepostpnt=1
     export npe_node_wavepostpnt=$(echo "$npe_node_max / $nth_wavepostpnt" | bc)
     export NTASKS=${npe_wavepostpnt}
@@ -122,8 +122,9 @@ elif [ $step = "anal" ]; then
 
     export wtime_anal="00:45:00"
     export npe_anal=1000
-    export nth_anal=4  
+    export nth_anal=12
     export npe_anal_gfs=1000
+    export nth_anal_gfs=12
     if [ $CASE = "C384" ]; then
       export npe_anal=160
       export nth_anal=10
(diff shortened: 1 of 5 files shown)


parm/config/config.resources.nco.static
```diff
@@ -133,6 +134,7 @@ elif [ $step = "anal" ]; then
     export npe_node_anal=$(echo "$npe_node_max / $nth_anal" | bc)
     export nth_cycle=$npe_node_max
     export npe_node_cycle=$(echo "$npe_node_max / $nth_cycle" | bc)
+    if [[ "$machine" = "WCOSS2" ]]; then export memory_anal="500GB"; fi
```
aerorahul, line 137: Is this needed? 
The total memory on the WCOSS2 node is 512GB.  If this resource request is not passed, does it explicitly request the full memory or does it do some calculation based on the number of cores requested?

@GeorgeVandenberghe-NOAA might know the answer.

parm/config/config.resources.nco.static
```diff
@@ -152,17 +154,17 @@ elif [ $step = "analdiag" ]; then
     export nth_analdiag=1
     export npe_node_analdiag=$npe_node_max
     if [[ "$machine" == "WCOSS_C" ]]; then export memory_analdiag="3072M"; fi
+    if [[ "$machine" = "WCOSS2" ]]; then export memory_analdiag="500GB"; fi
```
aerorahul, line 157: Same question as above.
Seeing this is a single node job, may be request max memory instead of something that is arbitrarily smaller than the max.

WalterKolczynski-NOAA (approved):
The resource difference for some of these jobs is remarkable. When support is added to develop, we may need to change how resources are computed so that it is by-machine.

ush/rocoto/workflow_utils.py
```diff
@@ -304,7 +304,10 @@ def get_resources(machine, cfg, task, reservation, cdump='gdas'):
 
     if machine in [ 'WCOSS2', 'WCOSS_DELL_P3', 'HERA', 'ORION' ]:
         try:
-            threads = cfg[f'nth_{ltask}']
+            if cdump in ['gfs'] and f'nth_{task}_gfs' in cfg.keys():
+                threads = cfg[f'nth_{ltask}_gfs']
+            else:
+                threads = cfg[f'nth_{ltask}']
         except KeyError:
             threads = cfg["',)nth_epos"]
```
aerorahul: This is a bug from the python2 to python3 conversion PR.
I don't know what the `',)`  is doing here.

aerorahul (commented):
Changes look good.
I have some minor comments about the memory request to `500GB`.  Why not request the full memory of `512GB` rather than the arbitrary `500GB`.

Also I spotted a bug in `workflow_utils.py` that probably got carried from python2 to python3 conversion.

KateFriedman-NOAA:
> When support is added to develop, we may need to change how resources are computed so that it is by-machine.

Yes please. Note: I haven't changed anything in config.fv3.emc.dyn nor config.resources.emc.dyn since some more thought and testing on the R&D machines will be needed to update them. I've focused on the ops versions of those configs for now.

ush/rocoto/workflow_utils.py
KateFriedman-NOAA: I have now removed the errant `',)` text from `threads = cfg["',)nth_epos"]`. Committed @ 6239494f. Thanks!

WalterKolczynski-NOAA: I don't think that was errant. I remember fixing something with that at one point in develop. Let me see if I can track down that commit.

aerorahul (approved):

ush/rocoto/workflow_utils.py
WalterKolczynski-NOAA: Found it. It was fixed in develop as part of PR #500: 68385d92

KateFriedman-NOAA: Thanks @WalterKolczynski-NOAA , I have further cleaned up that section of workflow_utils.py based on your change @ 68385d9. Committed to my branch @ https://github.com/NOAA-EMC/global-workflow/pull/546/commits/521bad764590bf07d8c15f81ba710838724eea41.

PR #549, opened 2022-01-06
Move directory from ecflow/ecf to one level up

lgannoaa:
Move directory from ecflow/ecf to one level up without change


aerorahul (approved):
ecflow related files were moved as per NCO instructions.
There has been no change in content in this PR.

PR #553, opened 2022-01-06
Update ecflow suite definition files for WCOSS2

lgannoaa:
This PR:
- updates the 4 suite definition files `prodXX.def`; `XX=00,06,12,18` per NCO instructions to define variable `PACKAGEHOME`.
- includes updates to facilitate developer testing by specifying developer values for variables such as `QUEUE`, `QUEUE_TRANSFER`, `QUEUE_SHARED` and `COM`

This work is part of #398

### DIFF
--- a/ecf/defs/prod00.def
+++ b/ecf/defs/prod00.def
@@ -3,30 +3,38 @@ extern /prod18/gdas/atmos/post
 extern /prod18/gdas/enkf/post
 #### ecen
 extern /prod18/gdas/atmos/post_processing/jgdas_atmos_chgres_forenkf
+extern /prod06/gdas/atmos/analysis/jgdas_atmos_analysis
+extern /prod06/gfs/atmos/analysis/jgfs_atmos_analysis
+extern /prod06/gdas/enkf/analysis/recenter/ecen
 
 suite prod00
   repeat day 1
   edit ECF_TRIES '1'
   edit CYC '00'
   edit ENVIR 'prod'
-  edit PROJ 'OPS'
-  edit E 'j'
-  edit QUEUE 'prod'
+  edit PROJ 'GFS'
+  edit PROJENVIR 'DEV'
+  edit E ''
+  edit QUEUE 'dev'
   edit QUEUESHARED 'dev_shared'
   edit QUEUESERV 'dev_transfer'
-  edit PROJENVIR 'OPS'
-  edit MACHINE_SITE 'production'
+#### Developer overwrite
+  edit ECF_INCLUDE '/lfs/h2/emc/global/noscrub/Lin.Gan/git/feature-ops-wcoss2/ecf/include'
+  edit MACHINE_SITE 'development'
+####
 
   family gfs
     edit CYC '00'
-    edit ECF_FILES '/ecf/ecfnets/scripts/gfs_FV3'
-    edit PROJ 'GFS-OPS'
+    edit PACKAGEHOME '/lfs/h2/emc/global/noscrub/Lin.Gan/git/feature-ops-wcoss2'
+    edit ECF_FILES '%PACKAGEHOME%/ecf/scripts/gfs'
+    edit PROJ 'GFS'
+    edit PROJENVIR 'DEV'
     edit NET 'gfs'
     edit RUN 'gfs'
-    edit COM '/gpfs/dell1/nco/ops/com'
-    edit QUEUESHARED 'prod_shared'
-    edit QUEUESERV 'prod_transfer'
-    edit QUEUE 'prod'
+    edit COM '/lfs/h2/emc/ptmp/Lin.Gan/ecfops/com'
+    edit QUEUE 'dev'
+    edit QUEUESHARED 'dev_shared'
+    edit QUEUESERV 'dev_transfer'
     family atmos
       family obsproc
         family dump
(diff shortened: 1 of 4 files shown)


ecf/defs/prod00.def
```diff
@@ -3,30 +3,38 @@ extern /prod18/gdas/atmos/post
 extern /prod18/gdas/enkf/post
 #### ecen
 extern /prod18/gdas/atmos/post_processing/jgdas_atmos_chgres_forenkf
+extern /prod06/gdas/atmos/analysis/jgdas_atmos_analysis
+extern /prod06/gfs/atmos/analysis/jgfs_atmos_analysis
+extern /prod06/gdas/enkf/analysis/recenter/ecen
```
WalterKolczynski-NOAA: Repeating the questions from the previous attempted PR:
- Why are there new triggers being added?
- Why are there triggers based on three cycles prior?

ecf/defs/prod00.def
```diff
@@ -3,30 +3,38 @@ extern /prod18/gdas/atmos/post
...
 
 suite prod00
   repeat day 1
   edit ECF_TRIES '1'
   edit CYC '00'
   edit ENVIR 'prod'
-  edit PROJ 'OPS'
-  edit E 'j'
-  edit QUEUE 'prod'
+  edit PROJ 'GFS'
```
WalterKolczynski-NOAA, line 12: All of these settings have been changed from ops settings to dev, which is required to run, but not correct for the hand-off.

ecf/defs/prod00.def
```diff
@@ -3,30 +3,38 @@ extern /prod18/gdas/atmos/post
 extern /prod18/gdas/enkf/post
 #### ecen
 extern /prod18/gdas/atmos/post_processing/jgdas_atmos_chgres_forenkf
+extern /prod06/gdas/atmos/analysis/jgdas_atmos_analysis
```
aerorahul: What extra triggers have we added in this porting exercise?

aerorahul (commented):
Please explain the new extern statements.

WalterKolczynski-NOAA (changes requested):
The PR title needs to be redone (it doesn't really tell me what the PR is at all), and the description is missing (what was changed? why were these changes made? what are the related issue numbers? is there anything that needs additional explanation?). Remember that the PR title/description become the commit message. All commit messages should follow the guidelines given in the wiki.

ecf/defs/prod00.def
lgannoaa: That was for my development removed.

ecf/defs/prod00.def
lgannoaa: That was for my development removed.

ecf/defs/prod00.def
lgannoaa: NCO will change that. NCO accepted this kind deliver before.

ecf/defs/prod00.def
```diff
@@ -9,24 +9,29 @@ suite prod00
...
+#### Developer overwrite
+  edit ECF_INCLUDE '/lfs/h2/emc/global/noscrub/Lin.Gan/git/feature-ops-wcoss2/ecf/include'
+  edit MACHINE_SITE 'development'
+####
 
   family gfs
     edit CYC '00'
-    edit ECF_FILES '/ecf/ecfnets/scripts/gfs_FV3'
-    edit PROJ 'GFS-OPS'
+    edit PACKAGEHOME '/lfs/h2/emc/global/noscrub/Lin.Gan/git/feature-ops-wcoss2'
```
aerorahul, line 25: This variable definition should be outside the `gfs` family as it is a constant.  Same applies for this variable defined on line 2257.

ecf/defs/prod00.def
```diff
@@ -2249,14 +2254,16 @@ suite prod00
   endfamily
   family gdas
     edit CYC '00'
-    edit ECF_FILES '/ecf/ecfnets/scripts/gdas_FV3'
-    edit PROJ 'GDAS-OPS'
+    edit PACKAGEHOME '/lfs/h2/emc/global/noscrub/Lin.Gan/git/feature-ops-wcoss2'
```
aerorahul, line 2257: This variable needs to be defined outside the `gdas` family as it is constant for the package.

ecf/defs/prod00.def
```diff
@@ -9,24 +9,29 @@ suite prod00
...
     edit NET 'gfs'
     edit RUN 'gfs'
-    edit COM '/gpfs/dell1/nco/ops/com'
-    edit QUEUESHARED 'prod_shared'
-    edit QUEUESERV 'prod_transfer'
-    edit QUEUE 'prod'
+    edit COM '/lfs/h2/emc/ptmp/Lin.Gan/ecfops/com'
+    edit QUEUE 'dev'
+    edit QUEUESHARED 'dev_shared'
+    edit QUEUESERV 'dev_transfer'
```
aerorahul, line 34: These are already defined as part of the suite on lines 15, 16, 17.  They don't need to be redefined here as well as in the GDAS family on lines 2264, 2265, 2266.

aerorahul (changes requested):
I have reviewed `prod00.def` and the comments are applicable for the cycle suite definitions as well.
This is unnecessary redefinition of variables that don't change.

ecf/defs/prod00.def
lgannoaa: Steven said in an email the PACKAGEHOME should to be inside family level.

ecf/defs/prod00.def
lgannoaa: As you can see on the left, the current production has it. I will keep this as is.

ecf/defs/prod00.def
lgannoaa: Explained in the previous similar comment. It should be inside family level.

ecf/defs/prod00.def
aerorahul: His team will move all common variables from family to suite level. That includes this variable and the other QUEUE variables.

aerorahul (approved):
Since it is too late to do any meaningful changes, approving this PR.

PR #564, opened 2022-01-12
Incorporate ecflow feedback from NCO - part 1

aerorahul:
This PR incorporates feedback from NCO on scripts in `ecf/include`.  Specifically:
- reverts `head.h` to NCO version.
- removes module loads from `head.h` (leaves `module reset`)
- reverts `tail.h` to NCO version
- reverts `envir-p1.h` to NCO version
- deletes obsolete `envir-p3.h`
- deletes obsolete `model_ver.h`

This work is part of #398 

Including @WeiWei-NCO to gather feedback.


KateFriedman-NOAA (approved):
Requested head.h changes look good. Can't comment on other changes but my understanding is that these files won't be used by NCO anyway so it's for dev use only.

PR #566, opened 2022-01-12
Add PBS debug directive.

aerorahul:
This PR:
- adds `#PBS -l debug=true` to all `.ecf` files as per NCO guidance.


KateFriedman-NOAA (approved):
Changes look good and take care of request by NCO to add debug line to PBS directives. Thanks @aerorahul !

PR #568, opened 2022-01-12
Load compiler env. and modules in the ecf scripts.

aerorahul:
This PR:
- loads the `PrgEnv-intel`, `craype` and `intel` module in the `.ecf` scripts as required by NCO.  Every single `.ecf` script needs these modules to be loaded.
- Ignores `.swp` files through an entry in `.gitignore`.

This work is part of #398, #399

Including @WeiWei-NCO for awareness and comment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ __pycache__
 *.pyc
 *.[aox]
 *.mod
+*.sw[a-p]
 
 # Ignore folders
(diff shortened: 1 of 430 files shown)


KateFriedman-NOAA (approved):
Looks good and addresses the particular NCO request, thanks @aerorahul !

PR #576, opened 2022-01-13
Series of small WCOSS2 updates - round 7

KateFriedman-NOAA:
This PR includes the following WCOSS2 changes:

* update release notes to specify newer hand-off tag (to be cut: `EMC-v16.2.0.1`)
* correct post job resources in ecf script PBS directive and change them to use 2 nodes instead of 10
* remove now-unneeded memory settings in `config.resources.nco.static` (jobs that now run exclusively)
* add needed memory settings in `config.resources.nco.static` for jobs that will run shared
* add `export imagemagick_ver=7.0.8-7` to `run.ver`
* update `workflow_utils.py` to add `excl` to NATIVE entity in xml when memory is not specified (job running exclusive)


aerorahul (approved):
looks good.

PR #579, opened 2022-01-14
Remove redundant toggle for gldas build

### LINKED ISSUE #578: Redundant gldas build toggle
The cfg files that determine which components are built and the partial_build.sh script all have a redundant copy of the gldas build toggle. That means if the first one is changed, it does nothing.

WalterKolczynski-NOAA:
The build cfg files controlling which components are built and the partial_build.sh script all have two instances trying to set the
value for gldas. This means one of the settings did nothing when changed.

Fixes: #578

### DIFF
--- a/sorc/cpl_build.cfg
+++ b/sorc/cpl_build.cfg
@@ -7,7 +7,6 @@
  Building gldas (gldas) ................................ no
  Building ncep_post (ncep_post) ........................ no
  Building ufs_utils (ufs_utils) ........................ no
- Building gldas (gldas) ................................ no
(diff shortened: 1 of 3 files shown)


KateFriedman-NOAA (approved):

aerorahul (approved):
I didn't test. Changes look good.

KateFriedman-NOAA:
> I didn't test. Changes look good.

The removal of the same GLDAS duplicate has worked fine in feature/ops-wcoss2, so it should be good here too.

PR #580, opened 2022-01-14
Fix toggle for building workflow_utils

### LINKED ISSUE #577: workflow_utils builds even if set to no in fv3gfs_build.cfg

WalterKolczynski-NOAA:
The fv3gfs_build.cfg did not have a space between the periods and the setting for workflow_utils, which means the setting was ignored and the default of yes was always used.

Fixes: #577


KateFriedman-NOAA (approved):
Nice simple fix, thanks @WalterKolczynski-NOAA !

aerorahul (approved):

PR #581, opened 2022-01-14
Stop archiving gfsarch.log as it is being written

### LINKED ISSUE #558: gfsarch failing due to mismatched log file size
The gfsarch job is failing on the second and subsequent cycles when `htar`ing the `gfsa` group which includes the `gfsarch.log` log file.  This is due to the file changing during compression.  This file's `gdas` compliment (`gdasarch.log`) is not included in the `htar` list.  If it is intentional to include `gfsarch.log`, then perhaps a check should be added so the job does not fail.  Otherwise, `gfsarch.log` should probably not be included in the archive list.

WalterKolczynski-NOAA:
The gfs archive job was failing because it was attempting to archive its own log file into gfsa.tar while it was being written. To exclude that file pattern, bash extended globbing is turned on, which allows the use of a negating group.

Fixes: #558

### DIFF
--- a/jobs/rocoto/arch.sh
+++ b/jobs/rocoto/arch.sh
@@ -1,4 +1,4 @@
-#!/bin/ksh -x
+#!/bin/bash -x
 
 ###############################################################
 ## Abstract:
@@ -72,7 +72,7 @@ if [ $CDUMP = "gfs" ]; then
         fhr2=$(printf %02i $fhr)
         fhr3=$(printf %03i $fhr)
         $NCP ${APREFIX}pgrb2.1p00.f$fhr3 $ARCDIR/pgbf${fhr2}.${CDUMP}.${CDATE}.grib2
-        (( fhr = $fhr + $FHOUT_GFS ))
+        (( fhr = 10#$fhr + 10#$FHOUT_GFS ))
     done
 fi
 if [ $CDUMP = "gdas" ]; then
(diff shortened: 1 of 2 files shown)


WalterKolczynski-NOAA:
@DavidHuber-NOAA Can you test this on one of your failures so I don't have to run several cycles just to test this?

DavidHuber-NOAA (changes requested):
This isn't working for me on S4.  Since these are `ksh` scripts, shopt is not a valid command.  However, extended globbing should already be part of the available syntax in `ksh`.  I've verified this on the terminal, so I'm not sure why it fails in the script.  I've uploaded the log file to Hera: /scratch1/NESDIS/nesdis-rdo2/David.Huber/s4_logs/384_p8a/gfsarch.log.

aerorahul:
@WalterKolczynski-NOAA 
We want to know exactly what is being archived.
Globbing can lead to archiving stale or irrelevant files or remnants.

aerorahul (approved):
Since this is a bugfix, I am fine with this change.
@WalterKolczynski-NOAA 
Mind if you change the interpreter in `ush/hpssarch_gen.sh` to `#/bin/bash` instead of `ksh`?

WalterKolczynski-NOAA:
@DavidHuber-NOAA I changed the shell for the archive job to bash and it solves the problem for me on Hera. Please confirm it works for you.

DavidHuber-NOAA:
> @DavidHuber-NOAA I changed the shell for the archive job to bash and it solves the problem for me on Hera. Please confirm it works for you.

@WalterKolczynski-NOAA Yes, it works for me now.  Thanks!

DavidHuber-NOAA (approved):
This works for me now.

PR #589, opened 2022-01-18
Explicitly set default netCDF atm and sfc filetypes

### LINKED ISSUE #588: Model 'atm' output filetypes are always netcdf_parallel
The `OUTPUT_FILETYPES` set in config.fcst are being overwritten by parsing_model_configure_FV3.sh.  This happens between [lines 15-18](https://github.com/NOAA-EMC/global-workflow/blob/86d4b305a070c080dcfd44a3f9a31ca2280759fa/ush/parsing_model_configure_FV3.sh#L15-L18).  For S4/ivy at all resolutions as well as Hera and Jet at C768, the `atm` files often fail to write in parallel and so this option should be preserved.

DavidHuber-NOAA:
This retains the `OUTPUT_FILETYPES` option set in config.fcst while parsing the model_configure FV3 configuration file.  This fixes #588.

### DIFF
--- a/parm/config/config.fcst
+++ b/parm/config/config.fcst
@@ -205,6 +205,7 @@ export deflate_level=1
 
 export OUTPUT_FILETYPES="$OUTPUT_FILE"
 if [[ "$OUTPUT_FILE" == "netcdf" ]]; then
+    export OUTPUT_FILETYPES=" 'netcdf_parallel' 'netcdf' "
     export  ichunk2d=0; export jchunk2d=0
     export  ichunk3d=0; export jchunk3d=0;  export kchunk3d=0
     RESTILE=$(echo $CASE |cut -c 2-)
@@ -214,18 +215,10 @@ if [[ "$OUTPUT_FILE" == "netcdf" ]]; then
             export ichunk3d=$((4*RESTILE)) 
             export jchunk3d=$((2*RESTILE))
             export kchunk3d=1
-        else
-            export OUTPUT_FILETYPES=" 'netcdf_parallel' 'netcdf' "
         fi
-    fi
-    if [[ "$machine" == "HERA" ]]; then
-        export OUTPUT_FILETYPES=" 'netcdf_parallel' 'netcdf' "
     fi
     if [[ "$machine" == "ORION" ]]; then
         export OUTPUT_FILETYPES=" 'netcdf_parallel' 'netcdf_parallel' "
-        if [ $RESTILE -le 192 ]; then
-            export OUTPUT_FILETYPES=" 'netcdf_parallel' 'netcdf' "
-        fi
     fi
 fi
 
--- a/parm/config/config.fcst
+++ b/parm/config/config.fcst
@@ -218,7 +218,9 @@ if [[ "$OUTPUT_FILE" == "netcdf" ]]; then
         fi
     fi
     if [[ "$machine" == "ORION" ]]; then
-        export OUTPUT_FILETYPES=" 'netcdf_parallel' 'netcdf_parallel' "
+        if [ $RESTILE -gt 192 ]; then
+           export OUTPUT_FILETYPES=" 'netcdf_parallel' 'netcdf_parallel' "
+        fi
     fi
 fi
 
--- a/ush/parsing_model_configure_FV3.sh
+++ b/ush/parsing_model_configure_FV3.sh
@@ -12,9 +12,9 @@
 
 FV3_model_configure(){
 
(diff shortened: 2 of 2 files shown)


ush/parsing_model_configure_FV3.sh
```diff
@@ -12,9 +12,9 @@
 
 FV3_model_configure(){
 
-export OUTPUT_FILETYPES="$OUTPUT_FILE"
+export OUTPUT_FILETYPES=${OUTPUT_FILETYPES:-${OUTPUT_FILE:-"netcdf"}}
 if [[ "$OUTPUT_FILE" == "netcdf" ]]; then
-   export OUTPUT_FILETYPES=" 'netcdf_parallel' 'netcdf' "
+   export OUTPUT_FILETYPES=${OUTPUT_FILETYPES:-" 'netcdf_parallel' 'netcdf' "}
```
WalterKolczynski-NOAA: This will never run the way you have changed it. OUTPUT_FILETYPES is always already set, so it is never changed. Either this if-block needs to go, or something needs to be changed here.

WalterKolczynski-NOAA (changes requested):

ush/parsing_model_configure_FV3.sh
DavidHuber-NOAA: You're correct.  As-is, this if block is necessary for Jet which does not change `OUTPUT_FILETYPES` from `"netcdf"` in config.fcst.  Instead, I've deleted this if block and added a default setting in config.fcst.

parm/config/config.fcst
```diff
@@ -214,17 +215,11 @@ if [[ "$OUTPUT_FILE" == "netcdf" ]]; then
...
         fi
     fi
-    if [[ "$machine" == "HERA" ]]; then
-        export OUTPUT_FILETYPES=" 'netcdf_parallel' 'netcdf' "
-    fi
     if [[ "$machine" == "ORION" ]]; then
-        export OUTPUT_FILETYPES=" 'netcdf_parallel' 'netcdf_parallel' "
-        if [ $RESTILE -le 192 ]; then
-            export OUTPUT_FILETYPES=" 'netcdf_parallel' 'netcdf' "
+        if [ $RESTILE -gt 192 ]; then
```
aerorahul, line 221: Do we need this `if` statement?  Why not just use `netcdf_parallel` for both atm and sfc files for all resolutions?
Is there a problem?

parm/config/config.fcst
```diff
@@ -205,6 +205,7 @@ export deflate_level=1
 
 export OUTPUT_FILETYPES="$OUTPUT_FILE"
 if [[ "$OUTPUT_FILE" == "netcdf" ]]; then
+    export OUTPUT_FILETYPES=" 'netcdf_parallel' 'netcdf' "
```
aerorahul, line 208: Is there a problem with having `netcdf_parallel` for the surface files?
Unless there is, I suggest the defaults to be `netcdf_parallel`.

WalterKolczynski-NOAA: I'm not sure. That's why I called you and Kate in to help assess these changes. They are almost functionally equivalent to before, but that doesn't mean we actually need these resolution distinctions.

junwang-noaa:
'parallel_netcdf' works for sfc files, but it is much slower than using
serial 'netcdf'. You may have to reevaluate the write tasks to catch up
with forecast grid comp.

parm/config/config.fcst
JessicaMeixner-NOAA: So I think I was the one who originally put in the ush/parsing_model_configure_FV3.sh statement that's being taken out here.  I did this to get around issues that I was having when switching to use parallel netcdf in feature/coupled-crow.  Perhaps we wont have those issues in the develop branch now.  (And sorry I can't remember the exact issue).

WalterKolczynski-NOAA: It looked like it was duplicating what was being done in config.fcst (probably that update came after coupled-crow forked from develop, then you added it in a different location since config files work differently in CROW).

WalterKolczynski-NOAA:
Okay, we're going to go ahead and accept this PR, then we're going to have another that changes how we are setting these, since that is outside the scope of what @DavidHuber-NOAA is trying to fix.

WalterKolczynski-NOAA (approved):
Okay, we're going to go ahead and accept this PR, then we're going to have another that changes how we are setting these, since that is outside the scope of what @DavidHuber-NOAA is trying to fix.

aerorahul (approved):
Looks good.

Issue #601, opened 2022-01-19
Remove support for nemsio


WalterKolczynski-NOAA:
**Description**
For a while global-workflow was supporting both the legacy nemsio filetype as well as NetCDF. There is no longer any need to support the older filetype, and UFS doesn't even support it anymore.

**Requirements**
Remove any code branches that depend on $OUTPUT_FILE, keeping only the code relevant for netcdf.
Any variables that are no longer needed should be removed, and any variables that no longer change should be replaced with their static value. 

**Acceptance Criteria (Definition of Done)**
No code relevant only to nemsio remains.

**Dependencies**
None

PR #602, opened 2022-01-20
Split output filetype variable into atm and sfc

### LINKED ISSUE #600: Improve organization of FV3 output filetype settings
**Description**
The determination of the FV3 model output filetype is a bit unwieldy because it is combining two filetype (atm and sfc) that need determined separately based on resolution and machine performance.

Separate output filetype variables for atm and sfc that are set with other resolution-dependent settings and later combined to form the appropriate namelist entry.

**Dependencies**
None

WalterKolczynski-NOAA:
The determination of the FV3 output filetype was a bit unwieldy as it
contained two different settings (one for the atmosphere and one for
the surface) that were not necessarily changed at the same time. They
were also being determined in a different location than other settings
based on the model resolution.

Now the old OUTPUT_FILETYPES variable has been split into two different
variables, OUTPUT_FILETYPES_ATM and OUTPUT_FILETYPES_SFC. The determi-
nation was also moved into the config.fv3 file, where other resolution-
dependent computational settings are set.

This has resulted in some functional change on WCOSS Dell, as the EnKF
was using different chunking settings and never used parallel output
for the surface likely because wasn't being run at a low enough resolu-
tion to consider it. However, there is no reason to believe the two
forecast modes should have different chunk or output settings. So, I've
used the EnKF chunking settings and the free forecast switchover point
for the surface output mode. However, on Hera the EnKF still forces 
serial output, as there is some issue with parallel netcdf and ensembles 
on that machine.

The new filetype settings are now also divorced from the OUTPUT_FILE
setting. However, that setting will soon be unnecessary as nemsio
is removed from the code as an option (see Issue #601).

Fixes: #600
Refs: #601

**Type of change**

**How Has This Been Tested?**
- [x] Cycled test on Hera C384/C196
- [x] Cylced test on Orion C384/C196
- [x] Cycled test on WCOSS Dell C768/C384
  
**Checklist**

### DIFF
--- a/parm/config/config.fv3
+++ b/parm/config/config.fv3
@@ -38,6 +38,10 @@ elif [[ "$machine" = "ORION" ]]; then
    export npe_node_max=40
 fi
 
+export OUTPUT_FILETYPE_ATM="netcdf_parallel"
+export OUTPUT_FILETYPE_SFC="netcdf"
+export ichunk2d=0; export jchunk2d=0
+export ichunk3d=0; export jchunk3d=0; export kchunk3d=0
 
 # (Standard) Model resolution dependent variables
 case $case_in in
@@ -101,6 +105,17 @@ case $case_in in
         export WRITE_GROUP_GFS=${WRITE_GROUP_GFS:-2}
         export WRTTASK_PER_GROUP_GFS=${WRTTASK_PER_GROUP_GFS:-$npe_node_max}
         export WRTIOBUF=${WRTIOBUF:-"16M"}
+        if [[ $machine == "ORION" ]]; then
+            export OUTPUT_FILETYPE_SFC="netcdf_parallel"
+        fi
+        if [[ $machine == "WCOSS_DELL_P3" ]]; then
+            RESTILE=$(echo $CASE |cut -c 2-)
+            export ichunk2d=$((4*RESTILE))
+            export jchunk2d=$((2*RESTILE))
+            export ichunk3d=$((4*RESTILE))
+            export jchunk3d=$((2*RESTILE))
+            export kchunk3d=1
+        fi
         ;;
     "C768")
         export DELTIM=150
@@ -116,6 +131,15 @@ case $case_in in
         export WRITE_GROUP_GFS=4
         export WRTTASK_PER_GROUP_GFS=$(echo "2*$npe_node_max" |bc)
         export WRTIOBUF="32M"
+        export OUTPUT_FILETYPE_SFC="netcdf_parallel"
+        if [[ $machine == "WCOSS_DELL_P3" ]]; then
+            RESTILE=$(echo $CASE |cut -c 2-)
+            export ichunk2d=$((4*RESTILE))
+            export jchunk2d=$((2*RESTILE))
+            export ichunk3d=$((4*RESTILE))
+            export jchunk3d=$((2*RESTILE))
+            export kchunk3d=1
+        fi
         ;;
     "C1152")
         export DELTIM=120
@@ -131,6 +155,15 @@ case $case_in in
         export WRITE_GROUP_GFS=4
         export WRTTASK_PER_GROUP_GFS=$(echo "2*$npe_node_max" |bc)
         export WRTIOBUF="48M"
+        export OUTPUT_FILETYPE_SFC="netcdf_parallel"
+        if [[ $machine == "WCOSS_DELL_P3" ]]; then
+            RESTILE=$(echo $CASE |cut -c 2-)
+            export ichunk2d=$((4*RESTILE))
+            export jchunk2d=$((2*RESTILE))
+            export ichunk3d=$((4*RESTILE))
+            export jchunk3d=$((2*RESTILE))
+            export kchunk3d=1
+        fi
         ;;
     "C3072")
         export DELTIM=90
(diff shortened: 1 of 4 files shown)


WalterKolczynski-NOAA:
Leaving as draft until I repeat the test on WCOSS Dell.

Adding GDAS team as reviewers since this technically effects them.

WalterKolczynski-NOAA:
Actually, I think I need to run this again at a higher resolution since these are so resolution dependent. I only ran the same C384/C196 test I've run previously. I don't anticipate any code changes though.

ush/parsing_model_configure_FV3.sh
```diff
@@ -38,7 +38,7 @@ write_nsflip:            ${WRITE_NSFLIP:-".false."}
 num_files:               ${NUM_FILES:-2}
 filename_base:           'atm' 'sfc'
 output_grid:             $OUTPUT_GRID
-output_file:             $OUTPUT_FILETYPES
+output_file:             '$OUTPUT_FILETYPE_ATM' '$OUTPUT_FILETYPE_SFC'
```
junwang-noaa, line 41: These two settings are resolution dependent, "parallel_netcdf" is not recommended to lower resolution than C384L127 as it takes longer. For C384L127, the "parallel_netcdf" runs faster for atmf files, but not for sfcf file for C768L127, it is recommended to used "parallel_netcdf" for both atmf and sfcf files. If the workflow will also be used by low resolution runs, especially long forecast run, then using "netcdf" will be a better choice.

parm/config/config.fv3
```diff
@@ -101,6 +105,17 @@ case $case_in in
         export WRITE_GROUP_GFS=${WRITE_GROUP_GFS:-2}
         export WRTTASK_PER_GROUP_GFS=${WRTTASK_PER_GROUP_GFS:-$npe_node_max}
         export WRTIOBUF=${WRTIOBUF:-"16M"}
+        if [[ $machine == "ORION" ]]; then
+            export OUTPUT_FILETYPE_SFC="netcdf_parallel"
+        fi
+        if [[ $machine == "WCOSS_DELL_P3" ]]; then
+            RESTILE=$(echo $CASE |cut -c 2-)
+            export ichunk2d=$((4*RESTILE))
```
junwang-noaa: May I ask what is the reason to choose these specific chunksizes? What benefit do we get?

aerorahul: They are just being moved from `config.fcst` to `config.fv3`

junwang-noaa:
I remember I had a discussion with Kate. The ichunk2d should be same as the
output domain IMO/JMO, which is equal to the (4*RESTILE) and (2*RESTILE)
at this time, but it would be good to set inchunk2d/ichunk3d as "IMO"
directly., same for JMO

ush/parsing_model_configure_FV3.sh
aerorahul: @junwang-noaa 
We are choosing to defining these variables based on the resolution in `config.fv3`. 
There is a `default` also set at the top there.

parm/config/config.fv3
WalterKolczynski-NOAA: All chunk settings were pulled from what was already being used in the EnKF or free forecast previously.

WalterKolczynski-NOAA:
> I remember I had a discussion with Kate. The ichunk2d should be same as the output domain IMO/JMO, which is equal to the (4*RESTILE) and (2*RESTILE) at this time, but it would be good to set inchunk2d/ichunk3d as "IMO" directly., same for JMO

If this is true for all resolutions, we can just pull all that out of the CASE block and set it in one place.

WalterKolczynski-NOAA:
System now runs on all three core machines.

aerorahul (approved):
Looks good. 
Approve pending testing.

WalterKolczynski-NOAA:
> Looks good. Approve pending testing.

Testing is already complete. Still hoping someone from DA will sign off since there is a difference in chunking and serial/parallel at some resolutions.

WalterKolczynski-NOAA:
@RussTreadon-NOAA @CatherineThomas-NOAA can one of you give this a quick look before I merge?

CoryMartin-NOAA:
Were any timing tests performed with new vs old chunk sizes for both forecast and GSI? I seem to remember a great deal of variability in both write-grid component write speed and GSI read speed based on the choices selected for chunk size.

See: https://docs.google.com/spreadsheets/d/1DPXJ6-3OXzn67-7ZK_VRttDl1kwYPhf-FBndtg9ZAnM/edit?usp=sharing from @RussTreadon-NOAA

WalterKolczynski-NOAA:
> Were any timing tests performed with new vs old chunk sizes for both forecast and GSI? I seem to remember a great deal of variability in both write-grid component write speed and GSI read speed based on the choices selected for chunk size.
> 
> See: https://docs.google.com/spreadsheets/d/1DPXJ6-3OXzn67-7ZK_VRttDl1kwYPhf-FBndtg9ZAnM/edit?usp=sharing from @RussTreadon-NOAA

I didn't do any timing tests, but the 3d chunking should be unchanged at operational resolutions. This is mostly extending the same chunking rubric to coarser resolutions and using the 2d chunk rubric EnKF was already doing in other forecast jobs.

CatherineThomas-NOAA (commented):
The changes look okay to me, though I'm not knowledgeable about chunking and will defer to someone who knows more from the DA side (@CoryMartin-NOAA? @jswhit2?)

PR #605, opened 2022-01-25
Series of small WCOSS2 updates - round 8

KateFriedman-NOAA:
**Description**

This PR brings in some changes for the WCOSS2 port:

1. Add `NTASKS_gfs` support in `WCOSS2.env` (for waveprep job that needs to run with different # of tasks in gfs suite compared to gdas suite - need to update wave scripts to not use NTASKS in future work, this is just a patch for WCOSS2 ops).
2. Remove unneeded `COMINice` and `COMINwnd` from `jobs/JGLOBAL_WAVE_POST_SBS`.
3. Add support for gfs suite memory settings in `ush/rocoto/workflow_utils.py`, previously only accepted one memory setting for both gdas and gfs suite versions of a job, now users can set memory separately between the gdas and gfs suite versions of the same job. This aligns with similar support for the other resource variables.
4. Add missing `get_awipsgroups` function to setup_workflow_fcstonly.py.
5. Add `COMIN_OBS` and `COMIN_GES_OBS` in config.base.emc.dyn to properly set both variables when run in emc/dev mode. These two variables will be set via compath.py in NCO mode so this change is only to support emc/dev mode users. Will be working to change this in the future, want to use compath.py everywhere and not need emc/nco mode checks.
6. Add rocoto xml envars for `GDATE`, `GDUMP`, `gPDY`, and `gcyc`. This supports the new `COMIN_GES_OBS` variable, which needs to be built off of these variables before being used. This aligns with similar support in the rocoto xml for setting the `CDATE`, `CDUMP`, `PDY`, `cyc` variables used by all jobs.

Refs: #398, #399

**Type of change**

Port updates and maintenance.

**How Has This Been Tested?**

By rerunning affected jobs in test on Dogwood.

### DIFF
--- a/ush/rocoto/setup_workflow.py
+++ b/ush/rocoto/setup_workflow.py
@@ -397,6 +397,10 @@ def get_gdasgfs_tasks(dict_configs, cdump='gdas'):
     envars.append(rocoto.create_envar(name='CDUMP', value=f'{cdump}'))
     envars.append(rocoto.create_envar(name='PDY', value='<cyclestr>@Y@m@d</cyclestr>'))
     envars.append(rocoto.create_envar(name='cyc', value='<cyclestr>@H</cyclestr>'))
+    envars.append(rocoto.create_envar(name='GDATE', value='<cyclestr offset="-6:00:00">@Y@m@d@H</cyclestr>'))
+    envars.append(rocoto.create_envar(name='GDUMP', value='gdas'))
+    envars.append(rocoto.create_envar(name='gPDY', value='<cyclestr offset="-6:00:00">@Y@m@d</cyclestr>'))
+    envars.append(rocoto.create_envar(name='gcyc', value='<cyclestr offset="-6:00:00">@H</cyclestr>'))
 
     base = dict_configs['base']
     gfs_cyc = base.get('gfs_cyc', 0)
--- a/ush/rocoto/setup_workflow_fcstonly.py
+++ b/ush/rocoto/setup_workflow_fcstonly.py
@@ -673,6 +673,40 @@ def get_workflow(dict_configs, cdump='gdas'):
     return ''.join(tasks)
 
 
+def get_awipsgroups(awips, cdump='gfs'):
+
+    fhmin = awips['FHMIN']
+    fhmax = awips['FHMAX']
+    fhout = awips['FHOUT']
+
+    # Get a list of all forecast hours
+    if cdump in ['gdas']:
+        fhrs = range(fhmin, fhmax+fhout, fhout)
+    elif cdump in ['gfs']:
+        fhmax = np.max([awips['FHMAX_GFS_00'],awips['FHMAX_GFS_06'],awips['FHMAX_GFS_12'],awips['FHMAX_GFS_18']])
+        fhout = awips['FHOUT_GFS']
+        fhmax_hf = awips['FHMAX_HF_GFS']
+        fhout_hf = awips['FHOUT_HF_GFS']
+        if fhmax > 240:
+            fhmax = 240
+        if fhmax_hf > 240:
+            fhmax_hf = 240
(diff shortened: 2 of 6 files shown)


ush/rocoto/setup_workflow.py
```diff
@@ -397,6 +397,10 @@ def get_gdasgfs_tasks(dict_configs, cdump='gdas'):
     envars.append(rocoto.create_envar(name='CDUMP', value=f'{cdump}'))
     envars.append(rocoto.create_envar(name='PDY', value='<cyclestr>@Y@m@d</cyclestr>'))
     envars.append(rocoto.create_envar(name='cyc', value='<cyclestr>@H</cyclestr>'))
+    envars.append(rocoto.create_envar(name='GDATE', value='<cyclestr offset="-6:00:00">@Y@m@d@H</cyclestr>'))
+    envars.append(rocoto.create_envar(name='GDUMP', value='gdas'))
+    envars.append(rocoto.create_envar(name='gPDY', value='<cyclestr offset="-6:00:00">@Y@m@d</cyclestr>'))
+    envars.append(rocoto.create_envar(name='gcyc', value='<cyclestr offset="-6:00:00">@H</cyclestr>'))
```
aerorahul, line 403: Do all jobs now need these variables?

ush/rocoto/setup_workflow_fcstonly.py
```diff
@@ -239,6 +239,10 @@ def get_workflow(dict_configs, cdump='gdas'):
     envars.append(rocoto.create_envar(name='CDUMP', value='&CDUMP;'))
     envars.append(rocoto.create_envar(name='PDY', value='<cyclestr>@Y@m@d</cyclestr>'))
     envars.append(rocoto.create_envar(name='cyc', value='<cyclestr>@H</cyclestr>'))
+    envars.append(rocoto.create_envar(name='GDATE', value='<cyclestr offset="-6:00:00">@Y@m@d@H</cyclestr>'))
+    envars.append(rocoto.create_envar(name='GDUMP', value='gdas'))
+    envars.append(rocoto.create_envar(name='gPDY', value='<cyclestr offset="-6:00:00">@Y@m@d</cyclestr>'))
+    envars.append(rocoto.create_envar(name='gcyc', value='<cyclestr offset="-6:00:00">@H</cyclestr>'))
```
aerorahul, line 245: Same comment as above.

aerorahul (commented):
most changes look good.
I have a couple of questions.

ush/rocoto/setup_workflow.py
KateFriedman-NOAA: Any jobs that use COMIN_GES_OBS need these, so many of the GSI/EnKF jobs. Not all jobs need them but I didn't want to set a job list in the setup scripts to only add them for some. Per a discussion with Russ about COMIN_OBS/COMIN_GES_OBS this is just a patch for dev users running the WCOSS2 ops version. This patch could go away once efforts are taken to investigate using compath.py more widely and remove the existing emc mode workarounds, like this.

ush/rocoto/setup_workflow_fcstonly.py
KateFriedman-NOAA: See reply to prior comment.

aerorahul (approved):

ush/rocoto/setup_workflow.py
RussTreadon-NOAA: These changes have already been merged.  That's OK.   I do not favor the short cut approach but I understand the 1/28 deadline.   Hopefully the patch can be removed before the WCOSS2 port branch is merged into develop.

KateFriedman-NOAA: @RussTreadon-NOAA Agreed, will be looking to rework this during the dev side of this port (next few months).

PR #621, opened 2022-01-31
Fix separate threading for GFS and GDAS forecasts.

### LINKED ISSUE #610: Separate threading options for gdasfcst and gfsfcst
**Expected behavior**

When `nth_fv3` is different from `nth_fv3_gfs` in config.fv3, config.resources should reference those options when calculating forecast GFS and GDAS forecasts.

**Current behavior**

Only the `nth_fv3` variable is referenced on [line 211](https://github.com/NOAA-EMC/global-workflow/blob/d3028b9d8268028226f9c27800fcd6655e9e4bb8/parm/config/config.resources#L211), resulting in the same threading option for both GDAS and GFS forecasts.

**Machines affected**

S4, which will use different threading options for GDAS and GFS forecasts.

**To Reproduce**

Manually set `nth_fv3` to 2, `nth_fv3_gfs` to 1, then run `setup_expt.py` and `setup_workflow.py`.  The resulting .xml will use 2 threads for both the GFS and GDAS forecasts.

**Context**

This should only immediately affect users on S4.

**Detailed Description**

Moving the `nth_fcst` assignment into the `if` block [starting on line 192 of config.resources](https://github.com/NOAA-EMC/global-workflow/blob/d3028b9d8268028226f9c27800fcd6655e9e4bb8/parm/config/config.resources#L192) and adding an `else` will allow users to specify different threading options.

**Additional Information**

**Possible Implementation**

I will add the suggested fix momentarily.

DavidHuber-NOAA:
**Description**
This PR fixes separate threading options for GDAS and GFS forecasts (#610).  This is performed by keeping `nth_fcst_gfs` separate from `nth_fcst` and declaring the new variable `npe_node_fcst_gfs`.

**Type of change**

**How Has This Been Tested?**

- [X] Cycled test on S4 using different threading options in config.fv3
- [X] Created XML workflows on Orion and Hera using different threading options in config.fv3

I have not tested this on WCOSS as I do not have access.
  
**Checklist**

### DIFF
--- a/ush/forecast_predet.sh
+++ b/ush/forecast_predet.sh
@@ -122,7 +122,12 @@ FV3_GFS_predet(){
   #file and the value of npe_node_fcst is not correctly defined when using more than
   #one thread and sets NTHREADS_FV3=1 even when the number of threads is appropraitely >1
   #NTHREADS_FV3=${NTHREADS_FV3:-${NTHREADS_FCST:-${nth_fv3:-1}}}
-  NTHREADS_FV3=${nth_fv3:-1}
+  #Threading can be different for GFS and GDAS forecasts
+  if [[ $CDUMP == "gdas" ]]; then
+     NTHREADS_FV3=${nth_fv3:-1}
+  else
+     NTHREADS_FV3=${nth_fv3_gfs:-1}
+  fi
   cores_per_node=${cores_per_node:-${npe_node_fcst:-24}}
   ntiles=${ntiles:-6}
   if [ $MEMBER -lt 0 ]; then
--- a/ush/forecast_predet.sh
+++ b/ush/forecast_predet.sh
@@ -122,12 +122,6 @@ FV3_GFS_predet(){
   #file and the value of npe_node_fcst is not correctly defined when using more than
   #one thread and sets NTHREADS_FV3=1 even when the number of threads is appropraitely >1
   #NTHREADS_FV3=${NTHREADS_FV3:-${NTHREADS_FCST:-${nth_fv3:-1}}}
-  #Threading can be different for GFS and GDAS forecasts
-  if [[ $CDUMP == "gdas" ]]; then
-     NTHREADS_FV3=${nth_fv3:-1}
-  else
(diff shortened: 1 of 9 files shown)


ush/forecast_predet.sh
```diff
@@ -122,7 +122,12 @@ FV3_GFS_predet(){
   #file and the value of npe_node_fcst is not correctly defined when using more than
   #one thread and sets NTHREADS_FV3=1 even when the number of threads is appropraitely >1
   #NTHREADS_FV3=${NTHREADS_FV3:-${NTHREADS_FCST:-${nth_fv3:-1}}}
-  NTHREADS_FV3=${nth_fv3:-1}
+  #Threading can be different for GFS and GDAS forecasts
```
WalterKolczynski-NOAA: This should be set in the env file with the others. This whole block can go, as NTHREADS_FV3 is already set there too. I don't know why someone added it back in here, contrary to the comment directly above.

WalterKolczynski-NOAA (changes requested):
The WCOSS env file needs to be updated the same way as the others, even if you can't test it yourself.

ush/forecast_predet.sh
DavidHuber-NOAA: OK, I've deleted the block and retested on S4.

DavidHuber-NOAA:
> The WCOSS env file needs to be updated the same way as the others, even if you can't test it yourself.

I have added the threading information to WCOSS_C.env and WCOSS_DELL_P3.env.

ush/forecast_predet.sh
WalterKolczynski-NOAA: You need to add a new entry to the block in the env files you moved up to set nth_fv3=nth_fv3_gfs as well.

DavidHuber-NOAA: I've added the nth_fv3_gfs assignment to all of the env files.  Thanks for the catch.

WalterKolczynski-NOAA (approved):
Confirmed this works on WCOSS-Dell

Issue #536, opened 2021-12-23
Unify `setup_expt.py` and `setup_expt_fcstonly.py` 


aerorahul:
In the present state, there are 2 entry points to the workflow setup.
One for `cycled` and another for `forecast-only` modes of operation.

As more "modes" are added, this adds duplication of code fragments and burden to the maintainers.

This issue documents the unification of `setup_expt.py` and `setup_expt_fcstonly.py`

PR #626, opened 2022-02-02
Only copy cycled ICs if specified

WalterKolczynski-NOAA:
In cycled mode when --icsdir was not specified, setup would create
a bunch of links to the default location, $HOME, inside of $COMROT.
Since most people don't keep initial conditions there, this resulted
in a bunch of dead links. This may also cause problems when people
go to stage their ICs from elsewhere, as they may get written to
$HOME via the links and on most HPC $HOME space is very limited.

Now the default is changed back to None, and links are only created
if a value is specified. Dead links will still be created if the
target does not exist, but at least users will not unwittingly be
creating them.

Also, the links are changed from INPUT is to RESTART.

Refs: #536


WalterKolczynski-NOAA:
We could also create the directory structure but not the links, but that requires more ifs or slight rearrangement.

aerorahul (approved):

Issue #571, opened 2022-01-13
Check FRAC_GRID for uncoupled runs


JessicaMeixner-NOAA:
In the creation of input.nml frac_grid is behind a if cpl switch, but it should be available for uncoupled runs as well (see: https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/parsing_namelists_FV3.sh#L330).  Also need to double check that other uses of FRAC_GRID are not behind "if cpl" statements.

lgannoaa:
@JessicaMeixner-NOAA 
Looks like config.fcst wired in export FRAC_GRID=".true."
ush/parsing_namelists_FV3.sh is the only script check in cpl is set to true before add frac_grid into input.nml.
Is moving frac_grid    = ${FRAC_GRID:-".true."} to the cat block above satisfy your request (parsing_namelists_FV3.sh line 325)?
cat >> input.nml <<EOF
  do_sppt      = ${do_sppt:-".false."}
  do_shum      = ${do_shum:-".false."}
  do_skeb      = ${do_skeb:-".false."}
  frac_grid    = ${FRAC_GRID:-".true."}
EOF

JessicaMeixner-NOAA:
Yes, I actually just told @WalterKolczynski-NOAA that I have this update in a branch.

lgannoaa:
Which branch should I do the PR in?

JessicaMeixner-NOAA:
I already have the code I need.  I'll submit it with a PR to develop with the other changes I'm doing.

lgannoaa:
Ok

PR #638, opened 2022-02-08
Turn on fractional grid by default for uncoupled forecasts

JessicaMeixner-NOAA:
**Description**

Updates input.nml so that there is more consistency between the cpl and standalone atm input.nml.  In particular this addresses the fact that frac_grid was only being set if cpl was true.   

Fixes Issue #571

**Type of change**

**How Has This Been Tested?**

Testing is ongoing 

**Checklist**

### DIFF
--- a/ush/parsing_namelists_FV3.sh
+++ b/ush/parsing_namelists_FV3.sh
@@ -86,16 +86,8 @@ cat > input.nml <<EOF
   npy = $npy
   ntiles = $ntiles
   npz = $npz
-EOF
-
-if [ $cpl = .true. ]; then
-  cat >> input.nml << EOF
-  dz_min =  ${dz_min:-"6"}    ! no longer in develop branch
-  psm_bc = ${psm_bc:-"0"}    ! no longer in develop branch
-EOF
-fi
-
-cat >> input.nml << EOF
+  dz_min =  ${dz_min:-"6"} 
+  psm_bc = ${psm_bc:-"0"} 
   grid_type = -1
   make_nh = $make_nh
   fv_debug = ${fv_debug:-".false."}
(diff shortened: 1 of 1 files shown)


JessicaMeixner-NOAA:
Current testing of this is ongoing with other updates, but this is a PR that only addresses #571 -- I did not find any other area for frac_grid where a relevant piece of code was behind an "if cpl" statement.

ush/parsing_namelists_FV3.sh
```diff
@@ -323,17 +315,12 @@ cat >> input.nml <<EOF
...
   do_shum      = ${do_shum:-".false."}
   do_skeb      = ${do_skeb:-".false."}
-EOF
-
-if [ $cpl = .true. ]; then
-  cat >> input.nml << EOF
   frac_grid    = ${FRAC_GRID:-".true."}
   cplchm       = ${cplchem:-".false."}
   cplflx       = $cplflx
   cplice       = ${cplice}
```
WalterKolczynski-NOAA: Are these last three guaranteed to be set even if cpl is ".false."? If not, they should have defaults of ".false.".

WalterKolczynski-NOAA (changes requested):

ush/parsing_namelists_FV3.sh
JessicaMeixner-NOAA: It looked to be like they are set: https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/config/config.fcst#L47-L51 but happy to add the false if that's what is wanted.

JessicaMeixner-NOAA: Made the update @WalterKolczynski-NOAA  -- always better to have a default after all.

WalterKolczynski-NOAA (approved):
Code approved pending testing.

WalterKolczynski-NOAA:
Cycled test passes.

PR #628, opened 2022-02-03
Add LOCALARCH option

DavidHuber-NOAA:
**Description**

This adds the option to archive the contents of the `ROTDIR` during arch and earc jobs for systems that do not have access to HPSS (i.e. Orion and soon S4).  Fixes #624.

Defaults have been added to setup_expt.py to point to a local `ATARDIR` and `LOCALARCH` (currently =NO).

**Type of change**

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

My suggestion on the documentation would be to add the following line to the [configure your run table](https://github.com/NOAA-EMC/global-workflow/wiki/Configure-your-run):

| Switch | What | Default | Adjusts XML | More Details |
| -- | -- | -- | -- | -- |
| LOCALARCH | Archive the ROTDIR to a local directory | NO | Possibly | Instead of pushing the contents of the ROTDIR to HPSS, archive them to a local directory, specified by ATARDIR.  If LOCALARCH=YES, then HPSSARCH must =NO.  Changing HPSSARCH from YES to NO will adjust the XML. |

**How Has This Been Tested?**

- [ ] Cycled test on WCOSS
- [X] Cycled test on Hera and S4
  
**Checklist**

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes need updates to the documentation and those changes have been made
- [X] My changes generate no new warnings
- [ ] New and existing tests pass with my changes

Issue #138, opened 2020-09-23
Port global-workflow to S4


DavidHuber-NOAA:
The global-workflow should be ported to S4.

Branch: [DavidHuber-NOAA:port2s4](https://github.com/DavidHuber-NOAA/global-workflow/tree/port2s4)

Below is a checklist taken from #5 for porting to Orion which I will use as a template:

- [x]  Test: connection to GitHub
- [x]  Build: hpc-stack (personal area)
- [x]  Build: hpc-stack (official area)
- [x]  Install: manage_externals manic-v1.1.8 tag (personal area)
- [x]  Build: FV3GFS (w/WW3)
- [x]  Build: ProdGSI
- [x]  Build: GLDAS
- [x]  Build: UFS_UTILS
- [x]  Build: gfs_wafs
- [x]  Build: obsproc
- [x]  Build: chgres_recenter
- [x]  Build: chgres_recenter_ncio
- [x]  Build: fbwndgfs
- [x]  Build: fv3nc2nemsio
- [x]  Build: gaussian_sfcanl
- [x]  Build: gfs_bufr
- [x]  Build: regrid_nemsio
- [x]  Build: supvit
- [x]  Build: syndat_getjtbul
- [x]  Build: syndat_maksynrc
- [x]  Build: syndat_qctropcy
- [x]  Build: tave
- [x]  Build: tocsbufr
- [x]  Build: vint
- [x]  Test: fv3gfs_chgres.sh
- [x]  Establish: "glopara" space
- [x]  Establish: FIX_DIR under "glopara"
- [x]  Establish: BASE_GIT under "glopara"
- [x]  Establish: GDA under "glopara"
- [x]  Establish: obsproc_prep and obsproc_global under "glopara"
- [x]  Establish: Fit2Obs under "glopara"
- [x]  Establish: METplus
- [x]  Establish: ens_tracker under "glopara"
- [x]  Create: standard S4 environment
- [x]  Add: system paths to config.base.emc.dyn
- [x]  Add: system to machine-setup.sh
- [x]  Add: system to modulefiles/module_setup.sh[csh].inc
- [x]  Add: system to setup_expt.py
- [x]  Add: system to setup_expt_fcstonly.py
- [x]  Add: system to setup_workflow.py
- [x]  Add: system to setup_workflow_fcstonly.py
- [x]  Add: system to workflow_utils.py
- [x]  Review: S4 module files for final modules and hpc-stack
- [x]  Test: run setup_expt.py
[...]

PR #649, opened 2022-02-10
Port to S4

DavidHuber-NOAA:
**Description**

This ports the global workflow to S4, completing #138.  S4 is capable of supporting experiments up to C384, so C768 and finer resolutions are disabled.  Being a smaller system with up to 40 nodes available to a user at a given time, config.resources and config.fv3 have been heavily modified to optimize performance.

During testing, it was determined that C48 ensemble forecasts required more memory on all systems, so the threads were increased to 2 for that purpose.

Suggested documentation updates:
[HPC Helpdesks](https://github.com/NOAA-EMC/global-workflow/wiki/HPC-Settings-&-Help#hpc-helpdesks), add an entry for S4: https://groups.ssec.wisc.edu/groups/S4/s4-user-guides/support-on-s4-cardinal

[Restricted data access](https://github.com/NOAA-EMC/global-workflow/wiki/HPC-Settings-&-Help#restricted-data-access): a note should be added that restricted data is not allowed on S4.

After 'Restricted data access,' add a subsection titled "Additional considerations for running on S4" with the following information:
S4 can support deterministic/ensemble resolutions of C384/C192, C192/C96, and C96/C48.  The C192/C96 resolution works well for testing purposes while the C384/C192 resolution run should be reserved for experiments that are at the appropriate stage.  Any higher resolution runs (e.g. 768/384) will require too many nodes to run on S4 and will not process.

Similarly, users should weigh whether the use of the 4DVAR data assimilation (i.e. with ensembles) is necessary for their experiment.  For instance, if you are testing the incorporation of a new temperature data source, 3DVAR assimilation may be adequate, while a new wind data source likely necessitates 4DVAR assimilation.  If running 4DVAR, a minimum of 40 ensembles is encouraged, though all 80 may be required to maintain stability when running at C192/C96.  If you are running without ensembles, when following the instructions in this guide, be sure to set the variable DOHYBVAR="NO" in the experiment's config.base before running setup_workflow.py.

Lastly, if you are performing data analysis experiments, the GFS forecast may not need to be run.  If that is the case, be sure to set `--gfs_cyc 0` when running setup_expt.py.  If the GFS forecast does need to be run for your experiment, note that GFS data sources (i.e. the global dump archive) are only available for the 00Z cycle, and so setup_expt.py should be called with `--gfs_cyc 1`.

[Prepare initial conditions](https://github.com/NOAA-EMC/global-workflow/wiki/Prepare-initial-conditions): add the following information to the introduction
"Initial conditions must be created/obtained from a system that has access to HPSS.  If you are working on S4 or Orion, these initial conditions must be transferred from one of the systems.  For those working on S4 without access to system with HPSS access, contact David Huber (david.huber@noaa.gov) with the resolution(s) and cycle(s) that you need and he will generate them for you."

[Free-forecast coupled](https://github.com/NOAA-EMC/global-workflow/wiki/Prepare-initial-conditions#free-forecast-coupled): Add that users wishing to run coupled forecasts on S4 should contact David Huber (david.huber@noaa.gov) with their requests.

[Run setup scripts](https://github.com/NOAA-EMC/global-workflow/wiki/Run-setup-scripts-to-generate-experiment): Add the following command for S4 to the introduction
for Python: `load miniconda/3.8-s4`
for rocoto: `load rocoto/1.3.2`

Also in this section, under the $EXPDIR bullet, add this information: "On S4, users should choose a directory within their home directory."

[Set user and experiment settings](https://github.com/NOAA-EMC/global-workflow/wiki/Run-setup-scripts-to-generate-experiment#2-set-user-and-experiment-settings-1): Add three variables to possibly change in config.base
PARTITION_BATCH (S4 only, may be s4 or ivy)
QUEUE (S4 only, must match PARTITION_BATCH)
FHMAX_GFS_XX (on S4, C384 forecasts should be limited to 168 hours)

[HPCs with access to directly edit your crontab files](https://github.com/NOAA-EMC/global-workflow/wiki/Start-your-run#hpcs-with-access-to-directly-edit-your-crontab-files-wcoss-cray-hera-jet-orion): add S4

Dependency: #628

**Type of change**

**How Has This Been Tested?**

- [X] Cycled C96/C48, C192/C96, C384/C192 tests on S4
- [X] Free-forecast, S2SW test on S4 at C384
- [x] Cycled test on Orion ~or Hera~
- [ ] Cycled test on WCOSS
- [x] Free-forecast on Orion ~or Hera~
  
**Checklist**


DavidHuber-NOAA:
I just started a cycled test on Orion.

DavidHuber-NOAA:
I am seeing an error on Orion that I haven't seen on S4 in the analysis jobs.  For instance, in /work/noaa/nesdis-rdo2/dhuber/para/com/p_2_s4_192/logs/2020080100/gdasanal.log:
`41:  Remove_AntCorr(FAILURE) : Input iFOV inconsistent with AC data`

However, I am also seeing this error in a run from the develop branch (/work/noaa/nesdis-rdo2/dhuber/para/com/test_gldas/logs/2020080100/gdasanal.log).  I'm not sure if this is an Orion issue or if it is related to a restricted data source available on Orion that is not available on S4.  Can someone confirm this?  @KateFriedman-NOAA @CoryMartin-NOAA

CoryMartin-NOAA:
@ADCollard @CatherineThomas-NOAA any insight on Dave's error? Seems unlikely it is restricted data related since this should be satellite obs (field of view). Did something change in the GSI namelist or the config.anal file?

ADCollard:
This looks like an error related to how we use direct broadcast microwave data, possibly related to corrupted input data.  Can you change the permissions so I can see the output.

Copying @emilyhcliu for awareness

DavidHuber-NOAA:
@ADCollard Apologies, I forgot to add +x to directories.  The permissions should be open now.

ADCollard:
This is going to take some debugging, I think.  Something is being corrupted but I'm not sure why.

In read_bufrtovs.f90 around line 690 are these lines:
!          Read data record.  Increment data counter
!          TMBR is actually the antenna temperature for most microwave 
!          sounders.
           if (llll == 1) then
              call ufbrep(lnbufr,data1b8,1,nchanl,iret,'TMBR')
           else     ! EARS / DB
              call ufbrep(lnbufr,data1b8,1,nchanl,iret,'TMBRST')
              if ( amsua .or. amsub .or. mhs )then
                 data1b8x=data1b8
                 data1b4=data1b8
                 call remove_antcorr(sc(instrument)%ac,ifov,data1b4)
                 data1b8=data1b4
                 do j=1,nchanl
                    if(data1b8x(j) > r1000)data1b8(j) = 1000000._r_kind
                 end do
              end if
         endif

Can you print out obstype, sc(instrument)%ac%n_fovs and ifov?

That might give a clue as to what is going on.

RussTreadon-NOAA:
`gdasanal.log` shows 
```
CRTM_FIX=/apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/crtm/2.4.0/fix
RTMFIX=/apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/crtm/2.4.0/fix
```
but 

`GSIEXEC=/work/noaa/nesdis-rdo2/dhuber/gw_port_2_s4/exec/global_gsi.x`

is built with `crtm/2.3.0`.    

Module `module_base.orion` contains
```
module load crtm/2.3.0
setenv CRTM_FIX /apps/contrib/NCEPLIBS/orion/fix/crtm_v2.3.0
... 
module load upp/10.0.8
```
While `crtm/2.3.0` is initially loaded, `upp/10.0.8` unloads `crtm/2.3.0` and loads `crtm/2.4.0`.   `module show upp/10.0.8` returns

```
-------------------------------------------------------------------------------------------------------------------------
   /apps/contrib/NCEP/libs/hpc-stack/modulefiles/mpi/intel/2018.4/impi/2018.4/upp/10.0.8.lua:
--------------------------------------------------------------------------------------------------------------------------
help([[]])
conflict("upp")
load("bacio","g2","g2tmpl","ip","sp","w3nco","w3emc","crtm")
prereq("bacio","g2","g2tmpl","ip","sp","w3nco","w3emc","crtm")
setenv("upp_ROOT","/apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/impi-2018.4/upp/10.0.8")
setenv("upp_VERSION","10.0.8")
setenv("NCEPPOST_INC","/apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/impi-2018.4/upp/10.0.8/include")
setenv("NCEPPOST_LIB","/apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/impi-2018.4/upp/10.0.8/lib/libupp.a")
setenv("POST_INC","/apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/impi-2018.4/upp/10.0.8/include")
setenv("POST_LIB","/apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/impi-2018.4/upp/10.0.8/lib/libupp.a")
setenv("UPP_INC","/apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/impi-2018.4/upp/10.0.8/include")
setenv("UPP_LIB","/apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/impi-2018.4/upp/10.0.8/lib/libupp.a")
whatis("Name: upp")
whatis("Version: 10.0.8")
whatis("Category: library")
whatis("Description: upp library")
```
File `amsua_metop-c.SpcCoeff.bin` differs between `/apps/contrib/NCEPLIBS/orion/fix/crtm_v2.3.0` and `/apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/crtm/2.4.0/fix`

```
Orion-login-4[90] rtreadon$ ls -l /apps/contrib/NCEPLIBS/orion/fix/crtm_v2.3.0/amsua_metop-c.SpcCoeff.bin /apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/crtm/2.4.0/fix/amsua_metop-c.SpcCoeff.bin
-rw-r----- 1 gkyle noaa-hpc  1252 Feb  4 20:03 /apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/crtm/2.4.0/fix/amsua_metop-c.SpcCoeff.bin
-rw-r----- 1 hlei  noaa-hpc 12196 Jun 11  2020 /apps/contrib/NCEPLIBS/orion/fix/crtm_v2.3.0/amsua_metop-c.SpcCoeff.bin
```
Is this a problem?

DavidHuber-NOAA:
@ADCollard @RussTreadon-NOAA OK, the job is running now with the print statement indicating the problematic instrument is amsua.

ADCollard:
Did you print out sc(instrument)%ac%n_fovs and ifov.

If they are outside of the range 1-30 there is a problem.   If it's sc(instrument)%ac%n_fovs that is wrong it could be the CRTM issue that @RussTreadon-NOAA noted.

DavidHuber-NOAA:
@ADCollard Yes, this is the line I added: `print*, obstype, sc(instrument)%ac%n_fovs,ifov` at line 686, just before `call remove_antcorr(sc(instrument)%ac,ifov,data1b4)`.

DavidHuber-NOAA:
I will try rearranging the module loads to make sure that crtm/2.3.0 is loaded and see if that fixes the issue.

WalterKolczynski-NOAA:
Sounds like we really should just be updating crtm to be 2.4.0

DavidHuber-NOAA:
@WalterKolczynski-NOAA I believe the plan is to upgrade CRTM as there is currently an open [GSI issue](https://github.com/NOAA-EMC/GSI/issues/188) to do so.

Rearranging does fix the issue, so it does indeed appear to be related to using the wrong CRTM fix files.  Since the GSI currently uses CRTM 2.3.0, I suspect this is how we will want to proceed for at least this PR.  Would you like for me to add this change to the modulefile into this PR or address it in a new issue?  @KateFriedman-NOAA @WalterKolczynski-NOAA

DavidHuber-NOAA:
I'm going to restart this test run on Orion and have moved the current log files to /work/noaa/nesdis-rdo2/dhuber/SAVELOGS/p_2_s4_192_orig.

DavidHuber-NOAA:
@RussTreadon-NOAA Thanks for pointing out the CRTM 2.4.0 module load.  Everything seems to have run well on Orion now.

I do notice some differences between S4 and Orion:

- Both S4 and Orion module files specify that g2/3.4.1 should be loaded though 3.4.2 and 3.4.3 are loaded, respectively.  Similarly, g2tmpl/1.10.0 is loaded instead of 1.9.1 on both systems.  Both of these are additional upp load casualties.  I have rearranged the S4 load order and rerun one gdasanal successfully.
- `BACK=YES` vs `BACK=NO` in prep jobs.  Can someone enlighten me to the differences between running the prep jobs with `BACK=YES` and `BACK=NO`?  Is it simply reading in the prepbufr files in parallel?  For S4 and Hera, `BACK=NO`, while for Orion, `BACK=YES`.
- There are differences in the number of measurements ingested due to data restrictions (expected)
- Also in the prep step, S4 must create its own NSSTBUFR file since it is not available due to data restrictions (expected).
- There are data quality warnings on Orion for SAPHIR on the 2020080200 cycle during the analysis jobs, which is not available on S4 due to data restrictions.

The Orion test thus seems to have been successful based on the current module files.

parm/config/config.fv3
```diff
@@ -61,6 +67,11 @@ case $case_in in
         export layout_y=4
         export layout_x_gfs=6
         export layout_y_gfs=4
+        if [[ "$machine" == "S4" ]]; then
```
WalterKolczynski-NOAA: I'd rather not add all these machine-specific layouts if we can at all help it. It adds a lot of ugly code for a machine that isn't even a tier-1 supported machine. I know S4 is small, but is there a reason S4 can't just use the same layout as the other machines for most of these?

Besides that, this script shouldn't be looking at $CASE_ENKF. First, there is little reason the layout should be different for EnKF than for any other forecast at the same resolution. Second, this construct will not behave as you want if EnKF were ever run at the same resolution as deterministic.

sorc/machine-setup.sh
```diff
@@ -56,6 +56,25 @@ elif [[ -d /scratch1 ]] ; then
...
+elif [[ -d /data/prod ]] ; then
+    # We are on SSEC S4
+    if ( ! eval module help > /dev/null 2>&1 ) ; then
+	     echo load the module command 1>&2
+        source /usr/share/lmod/lmod/init/$__ms_shell
+    fi
+    target=s4
+
+    module purge
+    module load license_intel/S4
```
WalterKolczynski-NOAA, line 69: Note this is going to change very soon as we move this all to the modulefiles where it belongs and convert them all to lua (see the work that was done on the ops branch for wcoss2). Not saying you have to change this right now, but that will be coming soon and we'll need you to do the conversion for S4.

README.md
```diff
@@ -13,6 +13,7 @@ The global-workflow current supports the following machines:
 * WCOSS-Cray
 * Hera
 * Orion
+* S4
```
WalterKolczynski-NOAA, line 17: I think we need to distinguish between tier-1 support and tier-2 support here (we also don't support the Cray anymore, but that's not your fault).

WalterKolczynski-NOAA (changes requested):
A few things while I free up the bandwidth to test this.

DavidHuber-NOAA:
Comparisons of a 10-day experiment run on both S4 and Orion showed significant differences.  I have traced these back to two issues on Orion.  The first is that while I copied over the GDA from S4 and pointed to it on Orion, I failed to specify `DO_MAKE_NSSTBUFR=YES" in config.prep.   Since the nsstbufr files are not available in S4's GDA due to the restricted data, they must be created on the fly with the available unrestricted bufr files.

The second issue was related to other module loading order problems.  In particular, when the UPP loads in module_base.orion, g2/3.4.3 and g2tmpl/1.10.0 were being loaded.  After fixing these, the two experiments are now much better aligned after one cycle.  I'm not sure which of these was causing the more significant differences, but my guess is g2tmpl.

Another 10-day run has started now that I will let run over the weekend and then perform another comparison on Monday.

parm/config/config.fv3
DavidHuber-NOAA: Most of these changes are here to improve cycle completion times and minimize resource usage on the system so it isn't too bogged down.  In particular, the EnKF is problematic with the default layouts.  S4 is restricted to 40 nodes per user.  Each efcs instance, by default, would use 6 nodes (i.e. 6 instances at once).  With the 4x4 tiling, it is possible to push 10 instances through at a time.  In timing experiments with 80 ensembles, this resulted in mean cycle timings of ~1.5 hours (5.5 vs 4) when running at C384/C192.  Similar speedups were observed for C192/C96.

That said, perhaps this could be addressed a different way.  I could try modifying the S4.env file to specify the layouts to use, and rewrite the layouts in config.fv3 to accept a predefined value, e.g.
`export layout_x=${layout_x:-6}`.  I'm not sure off the top of my head if this would work with the setup scripts.

Alternatively, I may be able to add some magic to setup_expt.py and config.base to specify layouts, but this could get messy, too.

If you have another suggestion, I would be willing to try it.

sorc/machine-setup.sh
DavidHuber-NOAA: Noted, it is on my radar and will be one of my first tasks after this port.

README.md
DavidHuber-NOAA: Good point, I've specified S4 as a tier-2 machine, and deleted Cray.

DavidHuber-NOAA:
@WalterKolczynski-NOAA @KateFriedman-NOAA The cycled testing on Orion is complete.  Comparisons between Orion and S4 out to 6 days show very good agreement and this port should now be ready for testing on WCOSS.

parm/config/config.fv3
DavidHuber-NOAA: Adding something to S4.env will not work since it is not loaded by setup_expt.py.

Looking through the scripts and configs, I see one other relatively clean option:

1. Add an S4-specific if-block to config.efcs that exports `layout_x` for different `CASE_ENKF`s before sourcing config.fv3.
2. As discussed above, modify the layout assignments in config.fv3 (`export layout_x=${layout_x:-6}`).

@WalterKolczynski-NOAA Would you prefer this solution?

WalterKolczynski-NOAA: Yes, I think I'd like putting it in config.efcs better.

DavidHuber-NOAA: Done.  I will test that this works on Orion and run single gdasefcs and gdasfcst jobs at 384, 192, and 96 on both Orion and S4.

parm/config/config.efcs
```diff
@@ -5,6 +5,19 @@
...
+unset -v layout_x layout_y nth_fv3
+if [[ $machine = "S4" ]]; then
+   if [[ "$CASE_ENKF" == "C96" ]]; then
+      export layout_x=4
+      export layout_y=4
+   elif [[ "$CASE_ENKF" == "C192" ]]; then
+      export layout_x=4
+      export layout_y=4
+      export nth_fv3=1
+   fi
```
aerorahul, line 18: I am not sure I agree with this in `config.efcs`

parm/config/config.fcst
```diff
@@ -7,6 +7,19 @@ echo "BEGIN: config.fcst"
...
+if [[ "$machine" == "S4" ]]; then
+   if [[ "$CASE" == "C384" ]]; then
+      export layout_x=8
+      export layout_y=8
+      export layout_x_gfs=8
+      export layout_y_gfs=8
+      export nth_fv3=1
+      export nth_fv3_gfs=1
+   fi
+fi
```
aerorahul, line 21: Or this.

parm/config/config.fv3.nco.static
```diff
@@ -28,6 +28,12 @@ elif [[ "$machine" = "JET" ]]; then
    export npe_node_max=24
 elif [[ "$machine" = "HERA" ]]; then
    export npe_node_max=40
+elif [[ "$machine" = "S4" ]]; then
+   if [[ "$PARTITION_BATCH" = "s4" ]]; then
+      export npe_node_max=32
+   elif [[ "$PARTITION_BATCH" = "ivy" ]]; then
+      export npe_node_max=20
+   fi
```
aerorahul, line 36: I realize this is for consistency, but, `config.fv3.nco.static` is for a specific application and machine

aerorahul (changes requested):
Please see comments re. fv3 configurations for a specific machine.

WalterKolczynski-NOAA:
@DavidHuber-NOAA The global-workflow team is going to get together and discuss this so we're not pulling you all over the place with edits. Might be a couple days before we can though.

DavidHuber-NOAA:
@WalterKolczynski-NOAA @aerorahul Sounds good, thanks.  I will finish the tests on this round of edits just to verify it _could_ work.

parm/config/config.resources
```diff
@@ -178,10 +213,14 @@ elif [ $step = "gldas" ]; then
...
 
     export wtime_fcst="00:30:00"
+    if [[ "$machine" == "S4" ]]; then
+       export wtime_fcst="01:00:00"
+    fi
     if [ $CASE = "C768" ]; then
       export wtime_fcst_gfs="06:00:00"
     elif [ $CASE = "C384" ]; then
       export wtime_fcst_gfs="06:00:00"
+      if [[ "$machine" == "S4" && "$APP" == "S2SW" ]]; then export wtime_fcst_gfs="04:00:00"; fi
```
aerorahul, line 223: I think remove this line and reduce the default to 4hrs.

parm/config/config.resources
```diff
@@ -335,6 +385,7 @@ elif [ $step = "vrfy" ]; then
     export wtime_vrfy="03:00:00"
     export wtime_vrfy_gfs="06:00:00"
     export npe_vrfy=3
+    [[ "$machine" == "S4" ]] && export npe_vrfy=1
```
aerorahul, line 388: We really need to know what is the drive behind some of these special S4 resource requests.

parm/config/config.resources
```diff
@@ -422,6 +473,12 @@ elif [ $step = "eobs" -o $step = "eomg" ]; then
     export nth_eobs=2
     if [[ "$machine" = "WCOSS_DELL_P3" ]]; then export nth_eobs=7; fi
     export npe_node_eobs=$(echo "$npe_node_max / $nth_eobs" | bc)
+    #The number of tasks and number of CPUs used must be the same for eobs
+    #For S4, this is accomplished by running 10 tasks/node
+    if [[ "$machine" = "S4" ]]; then
+       export npe_node_eobs=10
+       export wtime_eobs="00:30:00"
+    fi
```
aerorahul, line 481: based on the calculation on Line 475, for the different partitions of S4, the result on 479 will either be 16 or 10.  Whats wrong with those?
Is the 15min for eobs too short on S4?

aerorahul (changes requested):

parm/config/config.prep
```diff
@@ -65,4 +65,9 @@ else
 fi
 export DO_MAKE_NSSTBUFR="NO"   # if NO, will copy nsstbufr from globaldump
 
+#The NSST bufr files are not available on S4, so we need to make them
+if [[ "$machine" = "S4" ]]; then
+   DO_MAKE_NSSTBUFR="YES"
+fi
```
KateFriedman-NOAA, line 71: Is this because the S4 GDA doesn't include restricted datasets (e.g. nsstbufr)?

aerorahul:
This PR is likely out of date and will likely remain so till @DavidHuber-NOAA recovers.  
Closing PR now and will ask to reopen when ready.

PR #658, opened 2022-02-17
Correct MODE comparisons for forecast only

### LINKED ISSUE #657: MODE comparisons are incorrect
When the setup_expt scripts were combined, the forecast mode became a mandatory argument. The value of this option is then directly used for the MODE variable, but the argument name (forecast-only) does not match what was previously used for MODE in the forecast-only script (free) and some scripts are still testing against the old value instead of the new one.

WalterKolczynski-NOAA:
**Description**

When the setup_expt scripts were combined, the forecast mode became a mandatory argument. The value of this option is then directly used for the MODE variable, but the argument name (forecast-only) does not match what was previously used
for MODE in the forecast-only script (free) and some scripts were still testing against the old value instead of the new one. Those comparisons have now been updated to use the new MODE name.

Fixes #657

**Type of change**

**How Has This Been Tested?**

- [x] Forecast-only test on Hera
  
**Checklist**


KateFriedman-NOAA (approved):
Fix looks good, thanks @WalterKolczynski-NOAA !

Issue #599, opened 2022-01-19
gfs_atmos_gempak_meta silent fail


lgannoaa:
Silent fail found in gfs gempak meta job. Here is the ecflow job log from NCO wcoss2 para:
```
/lfs/h1/ops/para/output/20220116/gfs_atmos_gempak_meta_00.o526230 (line 2068)
/lfs/h1/ops/para/packages/gfs.v16.2.0/scripts/exgfs_atmos_gempak_meta.sh[86]: [: argument expected
grep "argument expected" gfs_atmos_gempak_meta_00.o526230
/lfs/h1/ops/para/packages/gfs.v16.2.0/scripts/exgfs_atmos_gempak_meta.sh[86]: [: argument expected
/lfs/h1/ops/para/packages/gfs.v16.2.0/scripts/exgfs_atmos_gempak_meta.sh[111]: [: argument expected
/lfs/h1/ops/para/packages/gfs.v16.2.0/scripts/exgfs_atmos_gempak_meta.sh[130]: [: argument expected
```

This issue also exist in current operation:
```
[Lin.Gan@v71a1 today]$ pwd
/gpfs/dell1/nco/ops/com/output/prod/today
[Lin.Gan@v71a1 today]$ grep "argument expected" gfs_atmos_gempak_meta_.o
gfs_atmos_gempak_meta_00.o82082149:/gpfs/dell1/nco/ops/nwprod/gfs.v16.1.5/scripts/exgfs_atmos_gempak_meta.sh[88]: [: argument expected
gfs_atmos_gempak_meta_00.o82082149:/gpfs/dell1/nco/ops/nwprod/gfs.v16.1.5/scripts/exgfs_atmos_gempak_meta.sh[113]: [: argument expected
gfs_atmos_gempak_meta_00.o82082149:/gpfs/dell1/nco/ops/nwprod/gfs.v16.1.5/scripts/exgfs_atmos_gempak_meta.sh[134]: [: argument expected
gfs_atmos_gempak_meta_06.o82101027:/gpfs/dell1/nco/ops/nwprod/gfs.v16.1.5/scripts/exgfs_atmos_gempak_meta.sh[88]: [: argument expected
gfs_atmos_gempak_meta_06.o82101027:/gpfs/dell1/nco/ops/nwprod/gfs.v16.1.5/scripts/exgfs_atmos_gempak_meta.sh[113]: [: argument expected
[...]

PR #661, opened 2022-02-18
Series of small WCOSS2 updates - round 11

KateFriedman-NOAA:
**Description**

This PR folds in final updates from @WeiWei-NCO's ops install for gfs.v16.2:

1. update hand-off tag in release notes to `EMC-v16.2.0.4` (will cut tag after this PR goes in)
2. a number of resource adjustments based on NCO/EMC testing on Dogwood, particularly for the wave jobs
3. move the `excl` tag from it's own line in the ecf script PBS directives into the same line as the other resources settings (right after `vscatter`)
4. add `export OMP_PLACES=cores` to fcst block in `WCOSS2.env`
5. a few small script updates from Wei based on testing and errors
6. move transfer list parm files into new `transfer/` folder and update from NCO copies (bulk of the files changed in this PR)
6. update NETCDF flags in `gaussian_sfcanl.fd/makefile.sh` to match what is needed on R&D machines for correct compilation
7. updates to `run.ver`, including moving some version variables into the new `wcoss2.ver`
8. create `wcoss2.ver` to support developers who don't set a few version variables via ecflow because they will be running with rocoto (matches similar updates happening on dev side of this port)
9. update load_fv3gfs_modules.sh to source `wcoss2.ver`
10. update WAFS tag to `gfs_wafs.v6.2.8` (after final WAFS script edit was folded in)

Refs: #399


aerorahul (changes requested):
I don't agree with combining excl on the same line.

KateFriedman-NOAA:
> I don't agree with combining excl on the same line.

@WeiWei-NCO @aerorahul Please hash this out and tell me the final decision on where `excl` will be in the ecf. I've gotten two different instructions. Thanks!

parm/transfer/transfer_gdas_enkf_enkf_05.list
```diff
@@ -24,8 +24,8 @@
 # directory are included, so if no exclude patterns match that file, it will be
 # tranatmferred.
 
-#com/gfs/_ENVIR_/enkfgdas._PDY_/
-/gpfs/dell1/nco/ops/com/gfs/_ENVIR_/enkfgdas._PDY_/
+#_COMROOT_/gfs/_SHORTVER_/enkfgdas._PDY_/
```
WalterKolczynski-NOAA: I know these are ops-only, but can we delete all these commented-out lines from these .list files? This one-liner should do it (but do it in a clean repo so you can reset if not):
`perl -i -p -e "s/^#_.*$//g *.list`

scripts/exgfs_atmos_gempak_meta.sh
```diff
@@ -62,7 +71,22 @@ do
       rm $DATA/poescript
 #   fi
 
-   awk '{print $1}' $FIXgempak/gfs_meta > $DATA/tmpscript
+   if [ $fhr -lt 100 ] ; then
+      typeset -Z2 fhr
```
WalterKolczynski-NOAA, line 75: 1) This does not need to be in an if-block. If fhr > 100, it will simply do nothing.
2) This is a functional change from before, which zero-padded to three places. This is only zero-padding to two
3) If the if-block mattered and we kept the padding as-is, zero-padding would only be needed < 10.

scripts/exgfs_atmos_gempak_meta.sh
```diff
@@ -97,11 +126,12 @@ do
   $APRUNCFP $DATA/poescript
   export err=$?; err_chk
 
-  if [ $fhr -eq 126 ] ; then
-    fhr=`expr $fhr + 6`
-  else
-    fhr=`expr $fhr + $fhinc`
-  fi
+      typeset -Z3 fhr
```
WalterKolczynski-NOAA, line 129: And now we zero-pad again, but to three places? There should probably be one of these at the very top of the loop, and that's it.

scripts/exgfs_atmos_gempak_meta.sh
```diff
@@ -97,11 +126,12 @@ do
...
+      if [ $fhr -eq 126 ] ; then
+        let fhr=fhr+6
```
WalterKolczynski-NOAA, line 131: Use arithmetic expansion instead: `(( fhr = 10#fhr + 6 ))`
(The 10# is necessary because otherwise zero-padded numbers are treated as octals)

scripts/exgfs_atmos_gempak_meta.sh
```diff
@@ -97,11 +126,12 @@ do
...
+      else
+	let fhr=fhr+fhinc
```
WalterKolczynski-NOAA, line 133: Use arithmetic expansion instead

WalterKolczynski-NOAA (dismissed):
A few things, plus the excl changes previously noted.

parm/transfer/transfer_gdas_enkf_enkf_05.list
KateFriedman-NOAA: Double checked with @WeiWei-NCO that those commented out lines weren't necessary and then removed them when he confirmed they weren't. Committed at f3ce093f.

KateFriedman-NOAA:
Updated ecf script PBS directives with regard to `select` and `place` settings. After discussion between NCO and EMC the ecf scripts will now have one `select` line with those resource values and a separate `place` line for setting `vscatter` and `excl`. Example:

Prior (all on single line):
```
#PBS -l place=vscatter:excl,select=...
```
Now (two separate lines):
```
#PBS -l select=...
#PBS -l place=vscatter:excl
```

Committed @ 9a74a804.

KateFriedman-NOAA:
@BoiVuong-NOAA @WeiWei-NCO Please see Walter's comments above about the `scripts/exgfs_atmos_gempak_meta.sh` script. Thanks!

BoiVuong-NOAA:
Kate, 
Do I need to make changes as Walter's comments in the script scripts/exgfs_atmos_gempak_meta.sh  ? or leave it as is.

WeiWei-NCO:
> @BoiVuong-NOAA @WeiWei-NCO Please see Walter's comments above about the `scripts/exgfs_atmos_gempak_meta.sh` script. Thanks!

In #599 , Boi provided a fixed script. However, the new script does a lot more dbnet alerts comparing to WCOSS1. After discussed with Boi, I reverted the script back to the WCOSS1 version.

KateFriedman-NOAA:
> Kate, Do I need to make changes as Walter's comments in the script scripts/exgfs_atmos_gempak_meta.sh ? or leave it as is.

@BoiVuong-NOAA At this point I think it depends on whether @WeiWei-NCO is open to testing another updated copy of this script before we make our final tag for the GFS on WCOSS2. We definitely want to get this updated for the `develop` branch side of things (for future GFS implementations) but not sure if we have time to do this now. @WeiWei-NCO thoughts?

BoiVuong-NOAA:
NCO commended to fix this issue in GFS v17

WeiWei-NCO:
> > Kate, Do I need to make changes as Walter's comments in the script scripts/exgfs_atmos_gempak_meta.sh ? or leave it as is.
> 
> @BoiVuong-NOAA At this point I think it depends on whether @WeiWei-NCO is open to testing another updated copy of this script before we make our final tag for the GFS on WCOSS2. We definitely want to get this updated for the `develop` branch side of things (for future GFS implementations) but not sure if we have time to do this now. @WeiWei-NCO thoughts?

It depends on how soon we can get this fixed and tested...

Since it's working fine in ops for months, I am okay to leave it as is at this point. Please fix it in the future upgrade.

KateFriedman-NOAA:
> NCO commended to fix this issue in GFS v17

@BoiVuong-NOAA @WeiWei-NCO Thanks, we are agreed to make this change in the `develop` branch for GFSv17 but will leave the `exgfs_atmos_gempak_meta.sh` script as is for GFSv16.

scripts/exgfs_atmos_gempak_meta.sh
KateFriedman-NOAA: Will update in `develop` branch.

scripts/exgfs_atmos_gempak_meta.sh
KateFriedman-NOAA: Will update in `develop` branch.

scripts/exgfs_atmos_gempak_meta.sh
KateFriedman-NOAA: Will update in `develop` branch.

scripts/exgfs_atmos_gempak_meta.sh
KateFriedman-NOAA: Will update in `develop` branch.

KateFriedman-NOAA:
@WalterKolczynski-NOAA @aerorahul Any further comments on this PR? @WeiWei-NCO just let me know there may be another update coming in. I can merge this and submit an additional small PR for that change should it come in. Am waiting to tag until that additional change is sorted.

aerorahul (approved):
looks good.

WalterKolczynski-NOAA (approved):

Issue #673, opened 2022-03-02
Implementation of EFSOI in global-workflow develop


AndrewEichmann-NOAA:
**Description**

This is an implementation in global-workflow of Ensemble Forecast Sensitivity to Observation Impacts (EFSOI), a method to determine the positive or negative impact of observations assimilated for the initial conditions for a forecast. 

At the time of this writing most of the necessary code is in the GSI master branch.

**Requirements**

When switched on at the time of workflow generation (running setup_workflow.py), the experiment will have the extra workflow components and tasks to apply the EFSOI calculation and generate a file with the sensitivities of the observations assimilated by GDAS, and run them all successfully.

The implementation has been developed with the global-workflow develop branch and kept up-do-date with the same, and so should requires UFS, GSI, and their dependencies. No additional modules are required. As it is in the current (as of this writing) hash of GSI, the GSI would require the checkout of the master branch, or at least a more recent hash than is currently used in global-workflow develop, before compiling it.

**Acceptance Criteria (Definition of Done)**

When a clone of global-workflow develop can run EFSOI successfully with minimal intervention (switching on before workflow generation, and at most the change of GSI hash).

**Dependencies**

Files to be added: 
config files in parm/config for additional tasks
rocoto scripts in jobs/rocoto for additional tasks

Files to be modified:
env/*env files for additional tasks
parm/config/config.resources for additional tasks
parm/config/config.base.emc.dyn for addition of EFSOI option
parm/config/config.earc for specification of time to sweep EFSOI files
sorc/link_workflow.sh to link additional files from GSI
ush/forecast_predet.sh for if/then block to work with efsoi directory tree instead of enkf
ush/hpssarch_gen.sh for saving EFSOI files
ush/rocoto/setup_workflow.py for generation workflow with EFSOI tasks 
and .gitignore for the files in jobs and scripts that are linked from GSI

These files are intended to be submitted piecewise in multiple pull requests to allow for relaxed review.

AndrewEichmann-NOAA:
@CatherineThomas-NOAA and @KateFriedman-NOAA suggested for review

AndrewEichmann-NOAA:
See also closed issues https://github.com/NOAA-EMC/global-workflow/issues/925 and https://github.com/NOAA-EMC/global-workflow/issues/478

aerorahul:
This capability is now in a frozen branch in global-workflow https://github.com/NOAA-EMC/global-workflow/tree/feature/efsoi

PR #674, opened 2022-03-03
Additions and modifications for EFSOI

AndrewEichmann-NOAA:
Issue #673

**Description**

Changes to files files in parm/config to enable EFSOI (Issue #673)

Files added:

- config.ecenfsoi
- config.efcsfsoi
- config.efsoi
- config.eposfsoi
- config.esfcfsoi
- config.eupdfsoi
(For their respective EFSOI-specific tasks)

Files changed:
config.base.emc.dyn (addition of commented EFSOI option, to be uncommented to run EFSOI. Leaving it as is will allow setup of regular cycling experiment)
config.earc (addition of variable RMOLDEND_EFSOI to determine how long to retain EFOSI files)
config.resources (addition of blocks for EFSOI tasks)

**Type of change**

**How Has This Been Tested?**

Branch has been tested by running regular (non-EFSOI) cycled experiments on Orion and Hera with 20 ensemble members without any changes other than queue specification. Experiments were run for a few cycles with successfully completed tasks.

- [ ] Clone and Build tests on WCOSS Dell P3
- [ ] Cycled test on Orion
- [ ] Cycled test on Hera
  
**Checklist**


parm/config/config.base.emc.dyn
```diff
@@ -11,6 +11,7 @@ export machine="@MACHINE@"
 # EMC parallel or NCO production
 export RUN_ENVIR="emc"
 
+
```
KateFriedman-NOAA: Remove extra line added, thanks!

parm/config/config.resources
```diff
@@ -467,6 +467,81 @@ elif [ $step = "eupd" ]; then
...
+    elif [ $CASE = "C192" -o $CASE = "C96" -o $CASE = "C48" ]; then
+      export npe_eupdfsoi=42
+      export nth_eupdfsoi=2
+      if [[ "$machine" = "HERA" ]]; then
+        export npe_eupdfsoi=40
+        export nth_eupdfsoi=40
+      fi
+    fi
+    export npe_node_eupdfsoi=$(echo "$npe_node_max / $nth_eupdfsoi" | bc)
+    if [[ "$machine" == "WCOSS_C" ]]; then
```
KateFriedman-NOAA: We've dropped support for the Crays (WCOSS_C) so please remove the new WCOSS_C lines that have been added. Thanks!

parm/config/config.vrfy
```diff
@@ -20,8 +20,8 @@ export VSDB_STEP1="YES"           # Populate VSDB database
 export VSDB_STEP2="NO"
 export VRFYG2OBS="YES"            # Grid to observations, see note below if turning ON
 export VRFYPRCP="YES"             # Precip threat scores
-export VRFYRAD="YES"              # Radiance data assimilation monitoring
-export VRFYOZN="YES"              # Ozone data assimilation monitoring
+export VRFYRAD="NO"              # Radiance data assimilation monitoring
+export VRFYOZN="NO"              # Ozone data assimilation monitoring
```
KateFriedman-NOAA: Curious, why changed to "NO"?

KateFriedman-NOAA (changes requested):
Since a number of the new configs contain either minimal settings or duplicate some settings also found in their non-efsoi counterparts, please merge needed settings into the existing configs and wrap settings with the `DO_EFSOI` switch as much as possible. This helps reduce config explosion and maintenance down the road (for the duplicate settings). Thanks!

parm/config/config.vrfy
AndrewEichmann-NOAA: This was left in by mistake - at one point I had to turn these off due to linking failures caused by cross-hash incompatibilities. Will correct.

parm/config/config.base.emc.dyn
AndrewEichmann-NOAA: Will do. To clarify, the `DO_EFSOI` switch is to determine if the entire experiment is doing EFSOI. A separate `EFSOI_TASK` is used to flag tasks that share code with regular tasks. This is used elsewhere, but can be used to advantage as you suggest.

parm/config/config.ecenfsoi
```diff
@@ -0,0 +1,21 @@
+#!/bin/ksh -x
```
aerorahul: What is the purpose of copying `config.ecen` into `config.ecenfsoi` and duplication everything in here?
I also notice that the resources of for `ecen` and `ecenfsoi` are identical.

This comment likely looks relevant to `config.efcsfsoi` and others as well where an `fsoi` version of the config is generated.

aerorahul (commented):
@AndrewEichmann-NOAA 
I see additions to config files that are large duplications of the config files.
Can we instead add logic to handle efsoi in their respective jobs?

I might be missing some bigger motivation here, so feel free to add comments here or discuss offline.

parm/config/config.ecenfsoi
WalterKolczynski-NOAA: Agreed, seems like most of these should just be reusing the existing config files with a couple of switches as necessary.

parm/config/config.efsoi
```diff
@@ -0,0 +1,39 @@
...
+export analpertwt=0.85           # relaxation to prior spread inflation factor
+export readin_localization_enkf=".false." # Don’t read in localization scales from file (doesn’t make
+                                          # sense for LETKF if model space localization on and nobsl_max>0)
+export corrlength=1250           # Horizontal localization scale (max horizontal distance to search for nobsl_max local obs)
+export lnsigcutoff=2.75          # ignored if modelspace_vloc=.true.
+
+export lobsdiag_forenkf=".true." # use jacobian.  must be .true. if modelspace_vloc=".true."
+                                 # need to specify .true. setting since config.anal sets to .false.
+
+#export EFSOIEXEC="/scratch1/NCEPDEV/da/Andrew.Eichmann/gsidev/afe-efsoi/GSI/exec/global_efsoi.x"
```
WalterKolczynski-NOAA: Remove any code that has been commented out.

WalterKolczynski-NOAA (changes requested):
I don't see any changes that would actually make use of these new settings. The attached issue refers to additional jobs, but none are added to the job mesh in this PR. Do j-job scripts hosted by GSI contain all the necessary changes to use these new settings, or is there more to be added to run EFSOI? Does the GSI version in checkout.sh need to be updated to a version containing those changes? Is there a related GSI PR for this, or have they already been merged?

AndrewEichmann-NOAA:
@KateFriedman-NOAA @aerorahul @WalterKolczynski-NOAA 

@CatherineThomas-NOAA and I discussed various approaches to the problem of the proliferation of config files and duplication within them. One of the reasons for the duplicated config content is that the rocoto scheme and setup_workflow seem to expect tasks to have matching files and variables. One approach I am trying now is to have the EFSOI tasks source their regular equivalents (ecen instead of ecenfsoi, for example), and have logic like the following in them:

```
# Get task specific resources
if [[ "$EFSOI_TASK" = "YES" ]]; then
   . $EXPDIR/config.resources ecenfsoi
else
   . $EXPDIR/config.resources ecen
fi
```
This may work when the config file is sourced during execution, but the problem here is that in setup_workflow.py, where I have a block in source_configs like so:

```
        elif task in ['eupdfsoi', 'ecenfsoi', 'esfcfsoi', 'efcsfsoi', 'eposfsoi']:
            configtask = task.replace('fsoi','') # makes 'ecenfsoi' into 'ecen'
            files.append(find_config(f'config.{configtask}', configs))
```
...so that the regular config file gets sourced for the workflow, but here the EFSOI-block in the config file gets ignored, and the resources for the EFSOI task get skipped. [Edit: Is there a way to pass the environment variable from python to the sourced script?]

Another approach would be to retain the EFSOI-specific config files, and just have them source the regular config files, but retain their own sections in the resources, and whatever particular settings they need. (Another problem is that some but not all EFSOI task resources can use the same settings as the regular tasks.) This would have the disadvantage of the proliferation of config files, but minimize the duplication without requiring special cases in setup_workflow.py.

What do you think should be prioritized?

parm/config/config.efsoi
AndrewEichmann-NOAA: @WalterKolczynski-NOAA  There are modifications to GSI (job files and script files among them) that are mostly already merged to GSI master. As this was last done last week, global-workflow would require a different GSI hash to be checked out and built. At present my global-workflow checkout.sh simply clones the EFSOI development branch of my GSI fork, but a solution for that would have to be considered until global-workflow checks out a late enough GSI hash.

KateFriedman-NOAA:
@AndrewEichmann-NOAA Where are the associated setup script updates and rocoto job updates? Need to see all of the needed EFSOI updates together to help determine the best path forward. Thanks!

AndrewEichmann-NOAA:
@KateFriedman-NOAA Everything is in the feature/EFSOI branch of my global-workflow fork. Would it make more sense to submit everything in the same PR? I submitted this partial PR because of previous complaints about large PRs submitted to GSI, but there is considerably less code difference between develop and feature/EFSOI than there was in the diffs with GSI master.

KateFriedman-NOAA:
> Everything is in the feature/EFSOI branch of my global-workflow fork. Would it make more sense to submit everything in the same PR? I submitted this partial PR because of previous complaints about large PRs submitted to GSI, but there is considerably less code difference between develop and feature/EFSOI than there was in the diffs with GSI master.

@AndrewEichmann-NOAA Yeah, let's submit your changes in a single PR. We also strive to reduce the size of our PRs but for this one it's simple enough to do as one, it's only 25 files now that I'm looking at it. We are going to have questions about it once we see the full set of changes and we may set up a Meet with you to discuss a few things already hinted at in the current PR comments. Thanks!

AndrewEichmann-NOAA:
I've updated the feature/EFSOI_A branch with everything except the changes to checkout.sh, and with updated EFSOI task config files that call the config files for the regular tasks. One approach that occurred to me to minimize the number of config files would be to have "efsoi" as an option for the setup_exp script that would copy the extra ones.

parm/config/config.base.emc.dyn
```diff
@@ -63,6 +63,8 @@ export DO_VRFY="YES"       # VRFY step
 #  use RUNMOS flag (currently in config.vrfy)
 export REALTIME="YES"
 
+#export DO_EFSOI="YES"
```
WalterKolczynski-NOAA: This should be uncommented and NO. User can then change it to YES if they want to turn EFSOI on.

parm/config/config.efcsfsoi
```diff
@@ -0,0 +1,96 @@
...
+. $EXPDIR/config.resources efcsfsoi
+
+export npe_fv3=$npe_efcsfsoi
+
+if [ $QUILTING = ".true." ]; then
+    export npe_fv3=$(echo " $npe_fv3 + $WRITE_GROUP * $WRTTASK_PER_GROUP" | bc)
+    export npe_efcsfsoi=$npe_fv3
+fi
+
+### Number of enkf members per fcst job
```
WalterKolczynski-NOAA: Why is there this huge block of commented out code?

parm/config/config.efsoi
```diff
@@ -0,0 +1,37 @@
...
+echo "BEGIN: config.efsoi"
+
+# Get task specific resources
+. $EXPDIR/config.resources efsoi
+
+export npe_enkf=$npe_eupd
+
+
+# Use NAM_ENKF below for serial EnKF
+##export NAM_ENKF="analpertwtnh=0.9,analpertwtsh=0.9,analpertwttr=0.9"
```
WalterKolczynski-NOAA: This should be in a branching structure that turns on when serial, rather than relying on users to uncomment and comment blocks of code.

parm/config/config.eposfsoi
```diff
@@ -0,0 +1,23 @@
...
+echo "BEGIN: config.eposfsoi"
+
+# Get regular epos configs
+. $EXPDIR/config.epos
+
+# Get task specific resources
+. $EXPDIR/config.resources eposfsoi
+
+## No. of concurrent epos jobs [1 implies sequential]
+#export NEPOSGRP=7
```
WalterKolczynski-NOAA: Remove commented-out code

parm/config/config.esfcfsoi
```diff
@@ -0,0 +1,22 @@
...
+. $EXPDIR/config.esfc
+
+# Get task specific resources
+. $EXPDIR/config.resources esfcfsoi
+
+## With IAU only need surface analysis at start of IAU window.
+## Set DOSFCANL_ENKF=NO to prevent creation of sfcanl at 
+## center of analysis window.  
+#
+#if [ $DOIAU_ENKF = "YES" ]; then
```
WalterKolczynski-NOAA: Remove commented-out code

parm/config/config.eupdfsoi
```diff
@@ -0,0 +1,38 @@
...
+. $EXPDIR/config.eupd
+
+# Get task specific resources
+. $EXPDIR/config.resources eupdfsoi
+
+export npe_enkf=$npe_eupd
+
+
+# Use NAM_ENKF below for serial EnKF
+export NAM_ENKF="analpertwtnh=0.9,analpertwtsh=0.9,analpertwttr=0.9"
```
WalterKolczynski-NOAA, line 18: Same as above: put this in a control structure

parm/config/config.resources
```diff
@@ -468,6 +468,75 @@ elif [ $step = "eupd" ]; then
...
+      export nth_eupdfsoi=6
+      if [[ "$machine" = "WCOSS_DELL_P3" ]]; then
+        export nth_eupdfsoi=7
+      fi
+      if [[ "$machine" = "HERA" ]]; then
+        export npe_eupdfsoi=150
+        export nth_eupdfsoi=40
+      fi
+    elif [ $CASE = "C384" ]; then
+      #export npe_eupdfsoi=270
```
WalterKolczynski-NOAA: Remove commented-out code

ush/hpssarch_gen.sh
```diff
@@ -590,5 +590,177 @@ if [ $type = "enkfgdas" -o $type = "enkfgfs" ]; then
...
 #-----------------------------------------------------
 
+#-----------------------------------------------------
+if [ $type = "efsoigdas" ]; then
+#-----------------------------------------------------
+
+  #IAUFHRS_ENKF=${IAUFHRS_ENKF:-6}
+  IAUFHRS_ENKF="6,12,18,24,30"
+  lobsdiag_forenkf=${lobsdiag_forenkf:-".false."}
+  nfhrs=`echo $IAUFHRS_ENKF | sed 's/,/ /g'`
```
WalterKolczynski-NOAA: Please use $( ) and not backticks

ush/hpssarch_gen.sh
```diff
@@ -590,5 +590,177 @@ if [ $type = "enkfgdas" -o $type = "enkfgfs" ]; then
...
+        if [ -s $ROTDIR/${dirpath}${head}atma${FHR}.ensmean${SUFFIX} ]; then
+	    echo  "${dirname}${head}atma00${FHR}.ensmean${SUFFIX}      " >>efsoi${CDUMP}.txt
+        fi
+        if [ -s $ROTDIR/${dirpath}${head}atmi${FHR}.ensmean${SUFFIX} ]; then
+            echo  "${dirname}${head}atmi00${FHR}.ensmean${SUFFIX}      " >>efsoi${CDUMP}.txt
+        fi
+     fi 
+  done # loop over FHR
+  for fstep in ecenfsoi esfcfsoi eupdfsoi efcsfsoi eposfsoi ; do
+   echo  "logs/${CDATE}/${CDUMP}${fstep}*.log        " >>efsoi${CDUMP}.txt
```
WalterKolczynski-NOAA: This doesn't look to be indented properly

ush/hpssarch_gen.sh
```diff
@@ -590,5 +590,177 @@ if [ $type = "enkfgdas" -o $type = "enkfgfs" ]; then
...
+	     echo  "${dirname}${head}atmf${fhr}.ensspread${SUFFIX}     " >>efsoi${CDUMP}.txt
+          fi
+      fi
+      fh=$((fh+6))
+  done
+
+  #...........................
+  n=1
+  while [ $n -le $NTARS ]; do
+  #...........................
```
WalterKolczynski-NOAA: This while block is not indented

ush/rocoto/setup_workflow.py
```diff
@@ -65,6 +65,10 @@ def main():
     steps = steps + wav_steps_gempak if _base.get('DO_GEMPAK', 'NO') == 'YES' else steps
     steps = steps + wav_steps_awips if _base.get('DO_AWIPS', 'NO') == 'YES' else steps
 
+# for EFSOI
+    efsoi_steps = ['eupdfsoi', 'esfcfsoi', 'ecenfsoi', 'efcsfsoi', 'eposfsoi','efsoi']
```
WalterKolczynski-NOAA: Missing space after last comma

ush/rocoto/setup_workflow.py
```diff
@@ -161,8 +165,10 @@ def get_definitions(base):
 
     machine = base.get('machine', wfu.detectMachine())
     scheduler = wfu.get_scheduler(machine)
+    do_efsoi = base.get('DO_EFSOI', 'NO').upper()
     hpssarch = base.get('HPSSARCH', 'NO').upper()
 
+
```
WalterKolczynski-NOAA: remove extra blank line

ush/rocoto/setup_workflow.py
```diff
@@ -316,11 +325,21 @@ def get_hyb_resources(dict_configs):
 
     dict_resources = OrderedDict()
 
+    do_efsoi = base.get('DO_EFSOI', 'NO').upper()
+
     # These tasks can be run in either or both cycles
     if lobsdiag_forenkf in ['.T.', '.TRUE.']:
-        tasks1 = ['eobs', 'ediag', 'eupd', 'echgres']
+
```
WalterKolczynski-NOAA: Remove extra blank line

ush/rocoto/setup_workflow.py
```diff
@@ -354,7 +373,14 @@ def get_hyb_resources(dict_configs):
 
     # These tasks are always run as part of the GDAS cycle
     cdump = 'gdas'
-    tasks2 = ['ecen', 'esfc', 'efcs', 'epos', 'earc']
+
+    do_efsoi = base.get('DO_EFSOI', 'NO').upper()
+
+    if do_efsoi in ['Y', 'YES']:
+        #tasks2 = ['ecen', 'ecenfsoi', 'esfc', 'esfcfsoi', 'efcs', 'efcsfsoi', 'epos', 'eposfsoi', 'earc', 'efsoi']
```
WalterKolczynski-NOAA: Remove commented-out code

ush/rocoto/setup_workflow.py
```diff
@@ -354,7 +373,14 @@ def get_hyb_resources(dict_configs):
...
+        tasks2 = ['ecenfsoi', 'esfcfsoi', 'efcsfsoi', 'eposfsoi', 'efsoi', 'ecen', 'esfc', 'efcs', 'epos', 'earc' ]
```
WalterKolczynski-NOAA: Remove extra space after last item

ush/rocoto/setup_workflow.py
```diff
@@ -1030,6 +1058,25 @@ def get_hyb_tasks(dict_configs, cycledef='enkf'):
 
         dict_tasks[f'{cdump}eupd'] = task
 
+        # # eupdfsoi
+        do_efsoi = base.get('DO_EFSOI', 'NO').upper()
+
+        if do_efsoi in ['Y', 'YES']:
+
```
WalterKolczynski-NOAA: Remove extra blank line

ush/rocoto/setup_workflow.py
```diff
@@ -1030,6 +1058,25 @@ def get_hyb_tasks(dict_configs, cycledef='enkf'):
...
+            deps = []
+            if lobsdiag_forenkf in ['.F.', '.FALSE.']:
+                dep_dict = {'type': 'metatask', 'name': '%seomn' % cdump}
+            else:
+                dep_dict = {'type': 'task', 'name': '%sediag' % cdump}
+            deps.append(rocoto.add_dependency(dep_dict))
+            data = '&ROTDIR;/gdas.@Y@m@d/@H/atmos/gdas.t@Hz.atmanl.ensres.nc'
+            dep_dict = {'type': 'data', 'data': data }
+            deps.append(rocoto.add_dependency(dep_dict))
+            dependencies = rocoto.create_dependency(dep_condition='and',dep=deps)
```
WalterKolczynski-NOAA: Add space after comma

ush/rocoto/setup_workflow.py
```diff
@@ -1030,6 +1058,25 @@ def get_hyb_tasks(dict_configs, cycledef='enkf'):
...
         dict_tasks[f'{cdump}eupd'] = task
 
+        # # eupdfsoi
+        do_efsoi = base.get('DO_EFSOI', 'NO').upper()
+
+        if do_efsoi in ['Y', 'YES']:
+
+            deps = []
+            if lobsdiag_forenkf in ['.F.', '.FALSE.']:
+                dep_dict = {'type': 'metatask', 'name': '%seomn' % cdump}
```
WalterKolczynski-NOAA: Please use modern python string formatting (preferably f-strings) and not the old % notation. Here and throughout the rest of the script.

ush/rocoto/setup_workflow.py
```diff
@@ -1062,6 +1109,34 @@ def get_hyb_tasks(dict_configs, cycledef='enkf'):
 
     dict_tasks[f'{cdump}ecmn'] = task
 
+    if do_efsoi in ['Y', 'YES']:
+    # ecmnfsoi, ecenfsoi
+        deps1 = []
+	#data = '&ROTDIR;/%s.@Y@m@d/@H/%s.t@Hz.loganl.txt' % (cdump, cdump)
```
WalterKolczynski-NOAA: Please remove unused code

ush/rocoto/setup_workflow.py
```diff
@@ -1111,6 +1206,29 @@ def get_hyb_tasks(dict_configs, cycledef='enkf'):
...
+         dep_dict = {'type': 'metatask', 'name': '%secmnfsoi' % cdump}
+         deps1.append(rocoto.add_dependency(dep_dict))
+         dep_dict = {'type': 'task', 'name': '%sesfcfsoi' % cdump}
+         deps1.append(rocoto.add_dependency(dep_dict))
+         dependencies1 = rocoto.create_dependency(dep_condition='and', dep=deps1)
+
+         deps2 = []
+         deps2 = dependencies1
+
+         # liaofan: remove the option to run efcsfsoi in the cold start (2020.05.26)
```
WalterKolczynski-NOAA: Remove unused code

ush/rocoto/setup_workflow.py
```diff
@@ -1127,11 +1245,54 @@ def get_hyb_tasks(dict_configs, cycledef='enkf'):
...
+         varval1, varval2, varval3 = get_eposfsoigroups(dict_configs['eposfsoi'], cdump=cdump)
+         vardict = {varname2: varval2, varname3: varval3}
+         task = wfu.create_wf_task('eposfsoi', cdump=cdump, envar=eposenvars, dependency=dependencies,
+                                   metatask='epmnfsoi', varname=varname1, varval=varval1, vardict=vardict)
+
+         dict_tasks['%sepmnfsoi' % cdump] = task
+
+         # efsoi 
+         deps = []
+     #    dep_dict = {'type': 'metatask', 'name': '%sepmnfsoi' % cdump}
```
WalterKolczynski-NOAA: Remove unused code

ush/rocoto/setup_workflow.py
```diff
@@ -1302,6 +1463,27 @@ def dict_to_strings(dict_in):
 
     return ''.join(strings)
 
+def get_eposfsoigroups(epos, cdump='gdas'):
+
+   # fhmin = 0
```
WalterKolczynski-NOAA: Remove unused code

ush/rocoto/setup_workflow.py
```diff
@@ -1334,19 +1516,41 @@ def create_xml(dict_configs):
...
+                         'gdasediag':'gdasediag',
+                         'gdaseomg':'gdaseomn',
+                         'gdaseupd':'gdaseupd',
+                         'gdasecen':'gdasecmn',
+                         'gdasesfc':'gdasesfc',
+                         'gdasefcs':'gdasefmn',
+                         'gdasepos':'gdasepmn',
+                         'gdasearc':'gdaseamn',
+                         'gdasechgres':'gdasechgres'}
+        #for each_task, each_resource_string in dict_hyb_resources.iteritems():
```
WalterKolczynski-NOAA: Remove unused code

ush/rocoto/setup_workflow.py
```diff
@@ -1334,19 +1516,41 @@ def create_xml(dict_configs):
...
         for each_task, each_resource_string in dict_hyb_resources.items():
-            #print(each_task,hyp_tasks[each_task])
-            #print(dict_hyb_tasks[hyp_tasks[each_task]])
+            #print each_task,hyp_tasks[each_task]
```
WalterKolczynski-NOAA: The removal of the parentheses would makes these lines not work if uncommented, but they should probably just be removed. If they remain for debugging, they should be modified so we aren't defining whole dictionaries just to serve this debug print.

ush/rocoto/setup_workflow.py
```diff
@@ -1334,19 +1516,41 @@ def create_xml(dict_configs):
...
-                     'gdasecen':'gdasecmn',
-                     'gdasesfc':'gdasesfc',
-                     'gdasefcs':'gdasefmn',
-                     'gdasepos':'gdasepmn',
-                     'gdasearc':'gdaseamn',
-                     'gdasechgres':'gdasechgres'}
+        do_efsoi = base.get('DO_EFSOI', 'NO').upper()
+
+        if do_efsoi in ['Y', 'YES']:
+            hyp_tasks = {'gdaseobs':'gdaseobs',
```
WalterKolczynski-NOAA: This variable doesn't appear to be used in active code, perhaps we should just remove it.

WalterKolczynski-NOAA (changes requested):

parm/config/config.efcsfsoi
AndrewEichmann-NOAA: removed

parm/config/config.esfcfsoi
AndrewEichmann-NOAA: removed

parm/config/config.efsoi
AndrewEichmann-NOAA: Lines removed

parm/config/config.eupdfsoi
AndrewEichmann-NOAA: The layout of this file is a reflection of the existing parm/config/config.eupd . I can remove all of the non-serial lines since for the foreseeable future EFSOI will use only that

parm/config/config.resources
AndrewEichmann-NOAA: done

ush/hpssarch_gen.sh
AndrewEichmann-NOAA: fixed

ush/hpssarch_gen.sh
AndrewEichmann-NOAA: fixed

ush/hpssarch_gen.sh
AndrewEichmann-NOAA: fixed

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: added

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: removed

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: done

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: done

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: done

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: done

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: added

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: done

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: done

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: done

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: done

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: done

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: done

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: removed

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: @WalterKolczynski-NOAA Which variable do you mean?

ush/rocoto/setup_workflow.py
```diff
@@ -1159,21 +1154,21 @@ def get_hyb_tasks(dict_configs, cycledef='enkf'):
     if do_efsoi in ['Y', 'YES']:
     # esfcfsoi
          deps1 = []
-         data = '&ROTDIR;/%s.@Y@m@d/@H/%s.t@Hz.loganl.txt' % (cdump, cdump)
+         data = '&ROTDIR;/{cdump}.@Y@m@d/@H/{cdump}.t@Hz.loganl.txt'
```
WalterKolczynski-NOAA: This one is missing the f in front, so the variables won't be substituted

ush/rocoto/setup_workflow.py
WalterKolczynski-NOAA: Nevermind, I didn't look far enough down behind the cut for it.

WalterKolczynski-NOAA (changes requested):

parm/config/config.eposfsoi
AndrewEichmann-NOAA: done

ush/rocoto/setup_workflow.py
AndrewEichmann-NOAA: Fixed. You have a better eye than me. Now I'm puzzled why the xml file worked...

WalterKolczynski-NOAA (approved):

WalterKolczynski-NOAA:
Code approved pending testing, though testing is probably hampered right now by the inability to run cycled experiments with the current develop.

WalterKolczynski-NOAA:
Currently blocked by inability to test due to cycled mode not working (Issue #711).

WalterKolczynski-NOAA:
The PR should be unblocked now that develop can run in cycled mode again. Update the branch with the develop changes. Hopefully by the time you get things updated with that, PR #859 will be merged so you can then redo the setup workflow changes in the refactored scripts.

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA I've updated the branch with a merge from develop as of b61c2375a05daff20805d0216ade8201192c0199

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA I've updated this branch to  e2869a1247ad2ba72c1bfe82f7682323d5128f4c and removed setup_workflow - will require work on the xml script, and eventually checking out a later version of gsi-utils

AndrewEichmann-NOAA:
Updated with merge from e2869a1247ad2ba72c1bfe82f7682323d5128f4c and additions to workflow setup scripts

AndrewEichmann-NOAA:
Backed out changes to build_gsi_utils.sh because they were tripping over CRTM problem in current hash of gsi-utils

AndrewEichmann-NOAA:
EFSOI_A merged with develop f04f3ba4dfd5b358aea425f65f82d40917597776

parm/config/config.ecenfsoi
AndrewEichmann-NOAA: Where possible, config files for EFSOI tasks source the config files for their respective regular tasks (eg config.ecenfsoi sources config.ecen)

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA @KateFriedman-NOAA  @aerorahul 
Wanted to check in to see if any major revisions were wanted on this - the reused code on the in the config files has been reduced by having the EFSOI versions source the standard versions where possible. I also tried to eliminate some of them entirely with EFSOI-specific code blocks, but this becomes complicated during the xml generation. It may be possible, but would probably be fragile in an unhelpful way. Are there existing examples of different tasks using the same config file? 
Let me know if anything is preventing going forward.

WalterKolczynski-NOAA:
> @WalterKolczynski-NOAA @KateFriedman-NOAA @aerorahul Wanted to check in to see if any major revisions were wanted on this - the reused code on the in the config files has been reduced by having the EFSOI versions source the standard versions where possible. I also tried to eliminate some of them entirely with EFSOI-specific code blocks, but this becomes complicated during the xml generation. It may be possible, but would probably be fragile in an unhelpful way. Are there existing examples of different tasks using the same config file? Let me know if anything is preventing going forward.

Multiple GSI-DA jobs load `config.anal`, and many (all?) the wave jobs load `config.wave`. They also then load additional job-specific configs where necessary.

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA I mean do there exist tasks that have no dedicated config file at all, and load a shared one instead (and work that way also in the workflow generation)? That's what I would need to implement to completely eliminate existing config files.

WalterKolczynski-NOAA:
> @WalterKolczynski-NOAA I mean do there exist tasks that have no dedicated config file at all, and load a shared one instead (and work that way also in the workflow generation)? That's what I would need to implement to completely eliminate existing config files.

Just make a do-nothing config file for each job. Several jobs already do that.

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA Ok, cool - that was my approach.

WalterKolczynski-NOAA:
@AndrewEichmann-NOAA Can we get this PR updated and conflicts resolved so we can get this in?

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA I expect to be testing a fresh merge tomorrow

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA Merged with da164f8b00c68d1309759645a57f0df744d46ef2 , will test regular cycling on Orion

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA Test failed on gdasvrfy task, testing develop

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA This appears to be failing in ush/ozn_xtrct.sh - log in /work/noaa/stmp/aeichman/comrot/rival/logs/2021060906/gdasvrfy.log

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA Just so I'm not barking up the wrong tree, what do I need to address to allow completion of the PR?

.gitignore
```diff
@@ -172,6 +172,7 @@ ush/global_cycle.sh
 ush/global_cycle_driver.sh
 ush/mkwfsgbl.sh
 ush/ufsda
+ush/rstprod.sh
```
WalterKolczynski-NOAA: This file shouldn't be ignored anymore, since it is in the workflow repo.

jobs/rocoto/earc.sh
```diff
@@ -119,6 +128,33 @@ if [[ $ENSGRP -gt 0 ]] && [[ $HPSSARCH = "YES" || $LOCALARCH = "YES" ]]; then
        fi
      fi
 
+     if [ $DO_EFSOI = "YES" ]; then
+       htar -P -cvf $ATARDIR/$CDATE/efsoi${CDUMP}_grp${ENSGRP}.tar `cat $ARCH_LIST/efsoi${CDUMP}_grp${n}.txt`
```
WalterKolczynski-NOAA: All these `htar` commands should use `$TARCMD` instead

jobs/rocoto/earc.sh
```diff
@@ -202,10 +253,49 @@ if [ $ENSGRP -eq 0 ]; then
...
-
-    done
+       if [ $DO_EFSOI = "YES" ]; then
+            COMIN_ENS="$ROTDIR/efsoigdas.$gPDY/$gcyc/$COMPONENT"
+            if [ -d $COMIN_ENS ] ; then
+                rm -rf $COMIN_ENS/*f012*nc
+                rm -rf $COMIN_ENS/*f018*nc
+                for imem in $(seq 1 $NMEM_ENKF); do
+                    memchar="mem"$(printf %03i $imem)
+                    for file in `ls $COMIN_ENS/$memchar |grep -v atmf024`; do
```
WalterKolczynski-NOAA: Please use `$( )`, not backticks.

jobs/rocoto/ecenfsoi.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash -x
```
WalterKolczynski-NOAA: All of these should be `#! /usr/bin/env bash` and be followed by `source "$HOMEgfs/ush/preamble.sh"` as the first command.

jobs/rocoto/efcsfsoi.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -x
```
WalterKolczynski-NOAA: Same

jobs/rocoto/efsoi.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -x
```
WalterKolczynski-NOAA: Same

jobs/rocoto/eposfsoi.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash -x
```
WalterKolczynski-NOAA: Same

jobs/rocoto/esfcfsoi.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -x
```
WalterKolczynski-NOAA: Same

jobs/rocoto/eupdfsoi.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -x
```
WalterKolczynski-NOAA: Same

parm/config/config.ecenfsoi
```diff
@@ -0,0 +1,14 @@
...
```
WalterKolczynski-NOAA: If these are staying, they should use bash: `#! /usr/bin/env bash`.

parm/config/config.efcsfsoi
```diff
@@ -0,0 +1,25 @@
+#!/bin/ksh -x
```
WalterKolczynski-NOAA: Same

parm/config/config.efsoi
```diff
@@ -0,0 +1,33 @@
+#!/bin/ksh -x
```
WalterKolczynski-NOAA: Same

parm/config/config.eposfsoi
```diff
@@ -0,0 +1,17 @@
+#!/bin/ksh -x
```
WalterKolczynski-NOAA: Same

parm/config/config.esfcfsoi
```diff
@@ -0,0 +1,22 @@
+#!/bin/ksh -x
```
WalterKolczynski-NOAA: Same

parm/config/config.eupdfsoi
```diff
@@ -0,0 +1,23 @@
+#!/bin/ksh -x
```
WalterKolczynski-NOAA: Same

ush/hpssarch_gen.sh
```diff
@@ -584,5 +584,175 @@ if [ $type = "enkfgdas" -o $type = "enkfgfs" ]; then
...
+
+  IAUFHRS_ENKF="6,12,18,24,30"
+  lobsdiag_forenkf=${lobsdiag_forenkf:-".false."}
+  nfhrs=$(echo $IAUFHRS_ENKF | sed 's/,/ /g')
+  NMEM_ENKF=${NMEM_ENKF:-80}
+  NMEM_EARCGRP=${NMEM_EARCGRP:-10}               ##number of ens memebers included in each tarball
+  NTARS=$((NMEM_ENKF/NMEM_EARCGRP))
+  [[ $NTARS -eq 0 ]] && NTARS=1
+  [[ $((NTARS*NMEM_EARCGRP)) -lt $NMEM_ENKF ]] && NTARS=$((NTARS+1))
+##NTARS2=$((NTARS/2))  # number of earc groups to include analysis/increments
```
WalterKolczynski-NOAA: Remove dead code.

ush/hpssarch_gen.sh
```diff
@@ -584,5 +584,175 @@ if [ $type = "enkfgdas" -o $type = "enkfgfs" ]; then
...
+  fi
+  if [ -s $ROTDIR/${dirpath}${head}oznstat.ensmean ]; then
+       echo  "${dirname}${head}oznstat.ensmean       " >>efsoi${CDUMP}.txt
+  fi
+  if [ -s $ROTDIR/${dirpath}${head}radstat.ensmean ]; then
+       echo  "${dirname}${head}radstat.ensmean       " >>efsoi${CDUMP}.txt
+  fi
+  for FH in $nfhrs; do  # loop over analysis times in window
+     FHR=$(printf %03i $FH)
+     if [ $FHR -eq 6 ]; then
```
WalterKolczynski-NOAA: Using the zero-padded number here may cause hidden errors because bash likes to think numbers starting with 0 are octals. It would still evaluate to false though, so it wouldn't change the results.

ush/hpssarch_gen.sh
```diff
@@ -584,5 +584,175 @@ if [ $type = "enkfgdas" -o $type = "enkfgfs" ]; then
...
+        nm=$(((n-1)*NMEM_EARCGRP+m))
+        mem=$(printf %03i $nm)
+        dirpath="efsoi${CDUMP}.${PDY}/${cyc}/atmos/mem${mem}/"
+        dirname="./${dirpath}"
+        head="${CDUMP}.t${cyc}z."
+    
+        #---
+        for FH in $nfhrs; do  # loop over analysis times in window
+           FHR=$(printf %03i $FH)
+           if [ $FHR -eq 6 ]; then
```
WalterKolczynski-NOAA: Same

ush/hpssarch_gen.sh
```diff
@@ -584,5 +584,175 @@ if [ $type = "enkfgdas" -o $type = "enkfgfs" ]; then
...
+                      echo "${dirname}${head}ratmi${FHR}${SUFFIX}      " >>efsoi${CDUMP}_grp${n}.txt
+                  fi
+              fi
+              if [ -s $ROTDIR/${dirpath}${head}ratmi${FHR}${SUFFIX} ] ; then
+                  echo "${dirname}${head}ratmi${FHR}${SUFFIX}      " >>efsoi${CDUMP}_restarta_grp${n}.txt
+              fi
+     
+           fi 
+           echo "${dirname}${head}atmf${FHR}${SUFFIX}       " >>efsoi${CDUMP}_grp${n}.txt
+           if [ $FHR -eq 6 ]; then
```
WalterKolczynski-NOAA: Same

workflow/applications.py
```diff
@@ -193,6 +194,9 @@ def _cycled_configs(self):
                 configs += ['eobs', 'eomg', 'ediag', 'eupd']
             configs += ['ecen', 'esfc', 'efcs', 'echgres', 'epos', 'earc']
 
+        if self.do_efsoi:
+            configs += ['eupdfsoi','ecenfsoi','esfcfsoi','efcsfsoi','eposfsoi','efsoi']
```
WalterKolczynski-NOAA: There should be a space after each comma to comply with PEP-8.

workflow/applications.py
```diff
@@ -349,6 +355,7 @@ def _get_cycled_task_names(self):
         gdas_gfs_common_tasks_before_fcst += ['sfcanl', 'analcalc']
 
         gldas_tasks = ['gldas']
+        efsoi_tasks = ['eupdfsoi','ecenfsoi','esfcfsoi','efcsfsoi','eposfsoi','efsoi']
```
WalterKolczynski-NOAA: Same

workflow/rocoto/workflow_tasks.py
```diff
@@ -15,6 +15,7 @@ class Tasks:
                    'atmanalprep', 'atmanalrun', 'atmanalpost',
                    'earc', 'ecen', 'echgres', 'ediag', 'efcs',
                    'eobs', 'eomg', 'epos', 'esfc', 'eupd',
+                   'eupdfsoi','ecenfsoi','esfcfsoi','efcsfsoi','eposfsoi','efsoi'
```
WalterKolczynski-NOAA: Same

workflow/rocoto/workflow_xml.py
```diff
@@ -77,7 +77,10 @@ def _get_workflow_header(self):
         """
 
         scheduler = self._app_config.scheduler
-        cyclethrottle = self._base.get('ROCOTO_CYCLETHROTTLE', 3)
+        if self._app_config.do_efsoi:
+            cyclethrottle = self._base.get('ROCOTO_CYCLETHROTTLE', 7)
```
WalterKolczynski-NOAA: Why is the cycle throttle being changed here for EFSOI?

WalterKolczynski-NOAA (changes requested):
Found some more things that I either missed before or have been standardized/changed in global workflow since this PR was first reviewed.

WalterKolczynski-NOAA:
> @WalterKolczynski-NOAA Just so I'm not barking up the wrong tree, what do I need to address to allow completion of the PR?

I don't think the gdas_vrfy problem is on you, but there are things I just found in a re-review.

WalterKolczynski-NOAA:
> @WalterKolczynski-NOAA This appears to be failing in ush/ozn_xtrct.sh - log in /work/noaa/stmp/aeichman/comrot/rival/logs/2021060906/gdasvrfy.log

@EdwardSafford-NOAA Can you help us with this issue? `oznmon_time.x` is failing due to a file being missing, but I'm not sure which or why:
```
 &INPUT
 SATNAME = aura_ges            ,
 IYY     =        2021,
 IMM     =           6,
 IDD     =           9,
 IHH     =           6,
 IDHH    =        -720,
 INCR    =           6,
 NREGION =           6,
 REGION  = global                                  70N-90N                                 20N-70N                                 20
 S-20N                                 20S-70S                                 70S-90S                                               
                                                                                                                                     
                                                                                                                                     
                                                                                                                                     
                                                                                                                                     
                                                                                                                                     
                                                                                       ,
 RLONMIN = 6*-180.0000      , 19*0.0000000E+00  ,
 RLONMAX = 6*180.0000       , 19*0.0000000E+00  ,
 RLATMIN =  -90.00000    ,   70.00000    ,   20.00000    ,  -20.00000    ,  -70.00000    ,  -90.00000    ,
  19*0.0000000E+00  ,
 RLATMAX = 2*90.00000       ,   70.00000    ,   20.00000    ,  -20.00000    ,  -70.00000    , 19*0.0000000E+00  ,
 VALIDATE        = T,
 NEW_HDR = F,
 PTYPE   = ges       ,
 NETCDF  = T
 /
  
 diag_oz =
 aura_ges.ges                                                                   
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                           
 grad_file=aura_ges.ges.2021060906.ieee_d          
 ctl_file =aura_ges.ges.ctl                        
 bad_pen_file =bad_pen.2021060906                      
 bad_cnt_file =bad_cnt.2021060906                      
 netcdf       = T
 **   ERROR: No such file or directory
 ** Failed to read NetCDF4.
```
```
(base) Orion-05-53 vrfy.42427/>pwd
/work/noaa/stmp/aeichman/RUNDIRS/rival/2021060906/gdas/vrfy.vrfy.42427/vrfy.42427
(base) Orion-05-53 vrfy.42427/>ls
bad_diag.2021060906                diag_ompsnp_n20_ges.2021060906.gz   diag_ompstc8_npp_ges.2021060906.gz  input
diag_omi_aura_anl.2021060906.gz    diag_ompsnp_npp_anl.2021060906.gz   diag_sbuv2_n19_anl.2021060906.gz    omi_aura.base
diag_omi_aura_ges.2021060906.gz    diag_ompsnp_npp_ges.2021060906.gz   diag_sbuv2_n19_ges.2021060906.gz    oznmon_horiz.x
diag_ompslp_npp_anl.2021060906.gz  diag_ompstc8_n20_anl.2021060906.gz  gdas_oznmon_base.tar                oznmon_time.x
diag_ompslp_npp_ges.2021060906.gz  diag_ompstc8_n20_ges.2021060906.gz  gome_metop-a.base                   sbuv2_n19.base
diag_ompsnp_n20_anl.2021060906.gz  diag_ompstc8_npp_anl.2021060906.gz  gome_metop-b.base                   stdout.time.aura_ges.ges
```

workflow/rocoto/workflow_xml.py
AndrewEichmann-NOAA: The efsoi task for a given cycle CDATE runs when the verifying analysis for the cycle CDATE+24h is present, so the cycle for CDATE and the cycles thereafter remain uncompleted until the efsoi task runs.

WalterKolczynski-NOAA: Okay, that's not really the purpose of the throttle. The task should probably be moved to the cycle where it will actually runs and look backwards to where it needs to. MET+ jobs already do that. We might be able to let it go for now and have it fixed in a follow-up PR. @KateFriedman-NOAA @aerorahul thoughts?

ush/hpssarch_gen.sh
AndrewEichmann-NOAA: Changing to reflect the scheme used by regular enkf in same file

ush/hpssarch_gen.sh
AndrewEichmann-NOAA: Changing to reflect the scheme used by regular enkf in same file

ush/hpssarch_gen.sh
AndrewEichmann-NOAA: Changing to reflect the scheme used by regular enkf in same file

workflow/rocoto/workflow_xml.py
AndrewEichmann-NOAA: I think that's doable

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA I changed everything I marked as resolved, though I'm not seeing the changes to ush/hpssarch_gen.sh reflected here.

workflow/rocoto/workflow_xml.py
AndrewEichmann-NOAA: I remembered the complication with this. So right now the efsoi task that produces the sensitivities for CDATE waits until the analysis for CDATE+24hrs is available. If it's swapped around so that it runs on CDATE to produce the sensitivities for CDATE-24hrs, then there are 4 efsoi tasks at the beginning of the experiment that don't complete - to calculate the sensitivities for CDATE-24hrs, you need the forecasts from CDATE-24hrs and CDATE-30hrs. One way to approach this is have the task run when the verifying analysis is available, then check if the forecasts exist, and exit and return 0 if not. Are there examples of such behavior? Should it be done at the jjob level?

WalterKolczynski-NOAA: We have several jobs that do not run during the first cycle, so this would be no different except for more cycles. You can add a cycleexist dependency for 24 and 30 hours prior (see jobs like aerosol_init and gldas).

AndrewEichmann-NOAA: So does an unmet cycleexist dependency mean that that the uncompleted task does not count towards whether its cycle is uncompleted? A rocotocheck on the gdasgldas from the first cycle indicates that the cycle remains active:
```
Cycle: 202106090000
  Valid for this task: YES
  State: active
  Activated: 2022-08-24 13:03:05 UTC
  Completed: -
  Expired: -
```
I thought the standard throttle value took this into account - that an extra was needed to allow for the first cycle never completing.

WalterKolczynski-NOAA: It shouldn't. If it does, there are ways around that (we can make the last job "final").

AndrewEichmann-NOAA: I guess I don't understand then what the cycleexistdep provides that the existing current file and task dependencies with cycle_offset values wouldn't. Either way with running the task on the cycle in which the verifying analysis is produced, you have five efsoi tasks in the beginning that never complete (the first cycle and then four regular ones). That's why I suggest doing "stub" instances of the efsoi tasks in the first four complete cycles after the first cycle.

AndrewEichmann-NOAA: Sorry, this might have been confusing: by "first four complete cycles after the first cycle" I mean the cycles where all of the tasks are expected to complete, as opposed to the first cycle.

env/ORION.env
```diff
@@ -204,6 +228,21 @@ elif [ $step = "efcs" ]; then
     [[ $NTHREADS_REGRID_NEMSIO -gt $nth_max ]] && export NTHREADS_REGRID_NEMSIO=$nth_max
     export APRUN_REGRID_NEMSIO="$launcher $LEVS"
 
+elif [ $step = "efcsfsoi" ]; then
+
+    nth_max=$(($npe_node_max / $npe_node_efcsfsoi))
+
+    export NTHREADS_FV3=${nth_efcfsois:-$nth_max}
```
aerorahul: There is likely a typo here.

aerorahul (commented):
I am still not convinced with the way the configs are being handled for this application.
It is adding a capability that is immediately going to pose a conflict when the infrastructure to run ensemble forecasts is added.

AndrewEichmann-NOAA:
@aerorahul I'm not familiar with any pending infrastructure for ensemble forecasts. What about the configs conflicts with it?

env/ORION.env
AndrewEichmann-NOAA: thank you, fixed

AndrewEichmann-NOAA:
@aerorahul In discussion with @CatherineThomas-NOAA , we wondered if the objection is with sourcing config files from config files, like in config.ecenfsoi:

```
#! /usr/bin/env bash

########## config.ecen ##########
# Ensemble recentering specific

echo "BEGIN: config.ecenfsoi"

# get regular ecen configs
. $EXPDIR/config.ecen

# Get task specific resources
. $EXPDIR/config.resources ecenfsoi

echo "END: config.ecenfsoi"
```
...as opposed to sourcing them separately in the jjob like so:

```
#############################
# Source relevant config files
#############################
export EXPDIR=${EXPDIR:-$HOMEgfs/parm/config}
configs="base ecen ecenfsoi"
config_path=${EXPDIR:-$NWROOT/gfs.${gfs_ver}/parm/config}
for config in $configs; do
    . $config_path/config.$config
    status=$?
    [[ $status -ne 0 ]] && exit $status
done

```

TBH I forget if there was a reason not to do it this way.

workflow/rocoto/workflow_xml.py
AndrewEichmann-NOAA: @WalterKolczynski-NOAA @CatherineThomas-NOAA pointed out that gldas does something similar to the stub task instance idea, like this snippet from jobs/JGDAS_ATMOS_GLDAS

```
xtime=$((gldas_spinup_hours+12))
if [ $CDATE -le $($NDATE +$xtime $SDATE) ]; then
    echo "GLDAS needs fluxes as forcing from cycles in previous $xtime hours"
    echo "starting from $SDATE. This gldas cycle is skipped" 
    exit 0
fi
```
I was thinking something similar, but for checking to see if the 24hr/30hr ensemble forecasts exist.

CatherineThomas-NOAA: @AndrewEichmann-NOAA 
Rather than checking if the forecasts exist, can you check the CDATE/SDATE as it does in gldas?  For EFSOI, this is only meant to catch the first handful of cycles and there are other cases that could trigger a forecast file check, for instance if you're running slowly on WCOSS and the ptmp scrubs.  You don't want later cycles saying the efsoi job completed successfully because of missing files.

AndrewEichmann-NOAA: @CatherineThomas-NOAA Yes, what you suggest makes sense.

EdwardSafford-NOAA:
@WalterKolczynski-NOAA Sorry for a late reply -- I was on leave last week.  The issue with `oznmon_time.x` is due to an issue in how `ush/ozn_xtrct.sh` is handling file names.  It manifests on orion only, and is unrelated to changes in this issue.  The problem is that the diagnostic file names in the oznstat are not getting parsed correctly.  It's on me and I'll get it fixed.

WalterKolczynski-NOAA:
@aerorahul @AndrewEichmann-NOAA Do we need to have a meeting to resolve the config issue? Don't want this PR to languish too much longer.

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA @aerorahul
A meeting would be good, but I wonder if hashing it out here might be more concise.  

Would it be acceptable if, to pick ecenfsoi as an example, config.ecenfsoi lost the sourcing of config.ecen, and the jjob had ecen then ecenfsoi on its list of config files to source? I think that's what how it was before, but with more duplication of config.ecen in config.ecenfsoi.

I'm also testing a solution to the throttling issue, using the approach in gldas.

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA @aerorahul 

I've merged f517d48dddbe2e03a3b9758c902a5d1474af6c26 with feature/EFSOI_A and changed the throttle situation.

AndrewEichmann-NOAA:
Are the linters messages anything to take action over right now? The numbers in the messages don't seem to line up with relevant line numbers in the files.

WalterKolczynski-NOAA:
> Are the linters messages anything to take action over right now? The numbers in the messages don't seem to line up with relevant line numbers in the files.

The `SC####` is the check being violated (like the E### or W### of PEP-8). See https://github.com/koalaman/shellcheck/wiki/Checks for more info. The PR comments are attached to the appropriate lines, or if you look at the full action log, it also has the line numbers.

workflow/rocoto/workflow_xml.py
WalterKolczynski-NOAA: I'd still prefer this be done at the workflow mesh level so the job doesn't even run (and sit in the queue) when it doesn't run, but this will do for now. If gldas is doing it, that should be changed too. But, we may need to keep it this way for ecflow; not sure.

AndrewEichmann-NOAA: I'm unfamiliar with workflow mesh.

WalterKolczynski-NOAA: At the workflow level (e.g. rocoto).

WalterKolczynski-NOAA: At the workflow level (e.g. rocoto).

AndrewEichmann-NOAA:
Regarding the script checks, it is flagging things like:
"npe_node_eupdfsoi is referenced but not assigned."
in the env files, but this variable is assigned in config.resources and I try to handle it the same way as npe_node_eupd. Checking the logs, it is assigned. So I'm not sure what is supposed to be fixed.

It's also flagging lines in the env files that I haven't added.

Also, I don't understand the consistency of some these warnings. The above example is line 173 in HERA.env, and that's the only complaint about that section as well as in efsoi (185), efcsfsoi (233), ecenfsoi (277), but then in esfcfsoi (305) it has a bunch more, even though it's in the same style.  These are also seem contradictory, advising to add curly brackets around variables but also saying that they are not necessary for arithmetic variables.

AndrewEichmann-NOAA:
I can't test this on Hera now anyway because the serial filter blows up there, so I'm just going to take all the Hera parts out.

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA @KateFriedman-NOAA

I'm figuring out what linter wants, but as I said it's flagging lines that I have not added. Do you want me to fix those too? Many of these look like they are amenable to global search/replace.

WalterKolczynski-NOAA:
We don't expect you to fix lines you didn't change, but we wouldn't be unhappy if you fixed those either.

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA 

I've worked through the shellcheck flags and this is what remains:

- Lines that I hadn't changed - I corrected a number of these, but corrections tend to beget more flags and this thing doesn't need a wider scope
- Complaints about unassigned variables and non-existing files which are evidently incorrect
- A couple cases that I think need some outside attention, annotated in reviews above

I've also taken out the config files calling each other, and instead put them in the series of config files sourced at the beginning of jjobs. They should be in line with all the other tasks now.

AndrewEichmann-NOAA:
Superseded by https://github.com/NOAA-EMC/global-workflow/pull/1074

PR #682, opened 2022-03-11
Revert "Add single forecast GOCART support"

WalterKolczynski-NOAA:
Reverts NOAA-EMC/global-workflow#659


JessicaMeixner-NOAA:
Is it necessary to revert?  Can we just make a subsequent PR fixing issues?

Issue #695, opened 2022-03-28
DO_GLDAS=YES requires FHOUT=1


RussTreadon-NOAA:
GLDAS requires hourly forecast model output from the gdas cycle.   Checks to ensure FHOUT=1 when DO_GLDAS=YES are missing from g-w.     

**Expected behavior**
If FHOUT=1 is a requirement when DO_GLDAS=YES, logic in config.base or workflow generation scripts should ensure this condition is met.

**Current behavior**
FHOUT remains set to 3 if either DOHYBVAR="NO" or l4densvar=".false.".   The gdasgldas job fails because it can not find required hourly model output from the gdas cycle.

**Machines affected**
All platforms which g-w supports.

**To Reproduce**
Set DOHYBVAR="NO" or l4densvar=".false." along with DO_GLDAS="YES" and cycle sufficient days for the 00Z gdasgldas job run.   The job will fail with a message similar to the one below
```
0.022 + [ ! -s /work/noaa/stmp/rtreadon/comrot/prufsda/gdas.20211221/12/atmos/gdas.t12z.sfluxgrbf001.grib2 ]
0.023 + echo 'WARNING: GLDAS MISSING /work/noaa/stmp/rtreadon/comrot/prufsda/gdas.20211221/12/atmos/gdas.t12z.sfluxgrbf001.grib2, WILL NOT RUN.'
WARNING: GLDAS MISSING /work/noaa/stmp/rtreadon/comrot/prufsda/gdas.20211221/12/atmos/gdas.t12z.sfluxgrbf001.grib2, WILL NOT RUN.
0.023 + exit 2
0.104 + err=2
0.104 + export err
0.104 + eval '[[' '$err' '=' 0 ']]'
0.104 + [[ 2 == 0 ]]
0.104 + exit 2
0.711 + status=2
0.711 + [[ 2 -ne 0 ]]
0.711 + exit 2
+ status=2
+ exit 2
```

**Context**
Developers set FHOUT=3 to reduce disk usage unaware of the FHOUT=1 requirement for GLDAS.    This requirement becomes apparent after several days of cycling.   Developers then either need to set DO_GLDAS=NO or rerun gdasfcst and gdaspost for previous cycles with FHOUT=1.

**Detailed Description**
[...]

PR #740, opened 2022-04-19
Modify config.base.emc.dyn to ensure FHOUT is 1 when DO_GLDAS is set

lgannoaa:
**Description**
Exception handling added to the config.base.emc.dyn file to address issue #695 

**How Has This Been Tested?**
Similar solution has been tested by developer before.

  
**Checklist**

### DIFF
--- a/parm/config/config.base.emc.dyn
+++ b/parm/config/config.base.emc.dyn
@@ -293,6 +293,11 @@ export lobsdiag_forenkf=".true."
 export DO_GLDAS="YES"
 export gldas_cyc=00
 
+# Exception handling that when DO_GLDAS is set, the FHOUT must be 1
+if [ $DO_GLDAS = "YES" ]; then
+  export FHOUT=1
+fi
+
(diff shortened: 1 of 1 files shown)


parm/config/config.base.emc.dyn
```diff
@@ -293,6 +293,11 @@ export lobsdiag_forenkf=".true."
 export DO_GLDAS="YES"
 export gldas_cyc=00
 
+# Exception handling that when DO_GLDAS is set, the FHOUT must be 1
+if [ $DO_GLDAS = "YES" ]; then
+  export FHOUT=1
```
WalterKolczynski-NOAA, line 298: Please add a comment up on L227 where this is originally set that the value may be overridden if GLDAS is on.

WalterKolczynski-NOAA (changes requested):

parm/config/config.base.emc.dyn
lgannoaa: It is done.

WalterKolczynski-NOAA (approved):

Issue #736, opened 2022-04-13
Update wave related settings and code for P8c 


JessicaMeixner-NOAA:
**Description**
A new mesh based cap (work by @DeniseWorthen) will be used in p8c. We will be updating which grid the wave model is using.  Updates for the new grid and new code will be required. 
 
**Requirements**
This will require an update to ufs-weather-model (PR is still under review), will require a new input grid, and various setting updates.  There will likely also be build updates that are required. 

**Acceptance Criteria (Definition of Done)**
Done means we can run prototype 8c from the workflow with the new wave cap and grids, setting up the input files correctly (if we do not use multi, a shel input file will have to be created), setting updates will need to be made in addition to creation of regular lat/lon gridded output. 

**Dependencies**
https://github.com/ufs-community/ufs-weather-model/pull/1131

JessicaMeixner-NOAA:
@WalterKolczynski-NOAA @KateFriedman-NOAA @aerorahul One question with this upcoming work: Right now there is a different build for "ATMW" or "S2SW".  What are the requirements about maintaining the ATMW capabilities?  Many of those features are broken at this point, but if that flexibility for making that an option in the future is desired (my guess is yes), I want to know in advance so I can build in the potential flexibility as I go instead of retroactively.

JessicaMeixner-NOAA:
@DeniseWorthen I want to confirm with you that we'll be using ww3_shel.inp for p8c.

WalterKolczynski-NOAA:
@JessicaMeixner-NOAA My position is to maintain (or fix) all current capabilities until told otherwise that they are no longer needed. So I've been dragging ATMW along with the assumption one of us will fix it at some point. I'm not even sure (or at least, don't remember) what is even broken with it.

JessicaMeixner-NOAA:
@WalterKolczynski-NOAA thank you.  I will keep the flexibility.

JessicaMeixner-NOAA:
The ufs-weather-model was [updated](https://github.com/ufs-community/ufs-weather-model/pull/1221) to use intel 2021/2022.  Should the base set of modules be updated to use this version or should a set of modules for the forecast be added?

WalterKolczynski-NOAA:
> The ufs-weather-model was [updated](https://github.com/ufs-community/ufs-weather-model/pull/1221) to use intel 2021/2022. Should the base set of modules be updated to use this version or should a set of modules for the forecast be added?

[...]

PR #747, opened 2022-04-20
Consolidate UFS build

### LINKED ISSUE #745: Consolidate UFS build
**Description**
The UFS build currently requires a user to determine whether to build for atmosphere-only (ATMW) or coupled (S2SW). This is no longer necessary, as the coupled executable can be used for atmosphere-only runs. The two options build with different CCPP suite lists, so these lists need to be combined

**Requirements**
Removal of special switches for build_ufs and build_all.

A UFS executable that is usable for all global-workflow applications.

**Dependencies**
None

WalterKolczynski-NOAA:
**Description**
Consolidates UFS build into a single executable. UFS will now always build with coupled components (S2SW app) and aerosols. Previous -a and -c options to build with coupled have been removed (-a was already defunct). The previous CCPP suite lists have been combined so all previously available suites will be available in the executable. This list likely needs to be trimmed to remove no longer supported options.

Fixes #745

**Type of change**

**How Has This Been Tested?**
- [x] Build on Hera
 
**Checklist**


WalterKolczynski-NOAA:
While this is mostly a maintenance issue, it is also much more convenient when testing to not have to deal with two different executables. I've actually been doing this manually when testing for a while.

JessicaMeixner-NOAA:
Two questions about this PR: 
1. What about the lack of being able to run with 32bit=Y for ATM only?  Isn't this a significant increase in resources for those running atm-only at this time? 
2. There is now the APP=S2SWA in ufs-weather-model.  Should we be using this in the build in the workflow now?

WalterKolczynski-NOAA:
> Two questions about this PR:
> 
> 1. What about the lack of being able to run with 32bit=Y for ATM only?  Isn't this a significant increase in resources for those running atm-only at this time?

As near as I can tell, we've been compiling 64-bit no matter what anyway. Might as well just condense it until mixed-mode FMS is ready.

> 2. There is now the APP=S2SWA in ufs-weather-model.  Should we be using this in the build in the workflow now?

Once we update to a UFS hash that allows this, yes. Unless we already are? In which case I can update this PR.

Also: what happened to putting A and W in alphabetical order?

junwang-noaa:
If you are using the Prototype-p8b tag, the APP S2SWA is the configuration
with all the components. The S2SWA is the app name, changing the order of W
and A won't work.

XianwuXue-NOAA (commented):
I like this changes. If you can improve the link_workflow.sh to only link the building components in *.cfg, for example: the exes in gsi is not needed in coupled mode, however, the current link_workflow.sh still link them to the exec folder but with broken links.

JessicaMeixner-NOAA:
@WalterKolczynski-NOAA the work required for : https://github.com/NOAA-EMC/global-workflow/issues/736 will mean that we need different builds for ATMW (which is currently not working in the workflow) and for S2SWA

KateFriedman-NOAA (approved):
Changes look good, thanks @WalterKolczynski-NOAA !

XianwuXue-NOAA (approved):

WalterKolczynski-NOAA:
We're going to have to keep different builds to support ATMW (though that is currently broken) as well as some future configurations, so killing this PR.

PR #755, opened 2022-04-25
Update WCOSS2 env file cpu-bind flags: core to depth

KateFriedman-NOAA:
**Description**

Based on conversations with NCO and EMC developers, we are now changing the `cpu-bind` launcher flag argument from `core` to `depth`. Change `--cpu-bind core` to `--cpu-bind depth`. This is to assist with runtime variability with threaded codes.

Refs: #399

Change also already made in the GSI `calcanl_gfs.py` script by @MichaelLueken-NOAA:

```
sorc/gsi.fd/ush/calcanl_gfs.py:        ExecCMDMPI1_host = 'mpiexec -l -n 1 --cpu-bind depth --depth '+str(NThreads)
sorc/gsi.fd/ush/calcanl_gfs.py:        ExecCMDMPI10_host = 'mpiexec -l -n 10 --cpu-bind depth --depth '+str(NThreads)
```

@GeorgeGayno-NOAA FYI, I found one other instance of `--cpu-bind core` outside of the GSI and global-workflow that you'll likely want to change for yourself (not used in ops):

```
sorc/ufs_utils.fd/reg_tests/global_cycle/driver.wcoss2.sh:export APRUNCY="mpiexec -n 6 -ppn 6 --cpu-bind core --depth ${OMP_NUM_THREADS_CY}"
```

### DIFF
--- a/env/WCOSS2.env
+++ b/env/WCOSS2.env
@@ -41,7 +41,7 @@ elif [ $step = "anal" ]; then
     export FI_OFI_RXM_SAR_LIMIT=3145728
 
     export NTHREADS_GSI=$nth_anal
-    export APRUN_GSI="$launcher -n ${npe_gsi:-${npe_anal:-$PBS_NP}} -ppn $npe_node_anal --cpu-bind core --depth $NTHREADS_GSI"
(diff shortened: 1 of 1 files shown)


aerorahul (approved):
looks good.

Issue #665, opened 2022-02-23
Support GFSv16.2.x (WCOSS2 port) on supported tier-1 platforms


KateFriedman-NOAA:
**Description**

This issue documents work to add support for the new GFSv16.2.x (WCOSS2 port version) on supported R&D platforms (Hera & Orion). Additional updates will also be committed to wrap up remaining WCOSS2 dev updates not required for operations (e.g. obsproc, GDA, verification).

Changes will be committed to the `dev_v16.2.x` branch.

**Requirements**

The `dev_v16.2.x` branch should support developers running the GFSv16.2.x package on WCOSS2, Hera, and Orion, in both forecast-only and cycled mode.

**Acceptance Criteria (Definition of Done)**

A user can install the `dev_v16.2.x` branch on WCOSS2, Hera, and/or Orion (out-of-the-box) and run either forecast-only or cycled parallels to support development for future GFS versions.

**Dependencies**

The following related issues:
- Issue #399 
- Issue #639

KateFriedman-NOAA:
### Testing on WCOSS2 in support of developers looking to run GFSv16.3 tests in coming months

#### Test: Cycled test with obsproc package integrated into `dev_v16` branch

Successfully completed unattended cycled test on Dogwood (no verification pieces yet): 2022032718 - 2022032912

```
Kate.Friedman@dlogin03:/lfs/h2/emc/global/noscrub/Kate.Friedman/expdir/devv16a> rocotostat -d devv16a.db -w devv16a.xml -s
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202203271800        Done    Mar 28 2022 18:03:08    Mar 29 2022 14:05:15
202203280000        Done    Mar 28 2022 18:03:08    Mar 29 2022 13:55:14
202203280600        Done    Mar 28 2022 18:03:08    Mar 29 2022 13:55:14
202203281200        Done    Mar 29 2022 13:13:14    Mar 29 2022 18:35:13
202203281800        Done    Mar 29 2022 14:00:19    Mar 29 2022 19:50:13
202203290000        Done    Mar 29 2022 14:00:19    Mar 29 2022 21:15:11
202203290600        Done    Mar 29 2022 14:10:12    Mar 29 2022 23:05:08
202203291200        Done    Mar 29 2022 18:40:17    Mar 30 2022 00:55:12
```

Used WCOSS1 ops GDA data pushed over to Dogwood from WCOSS-Dell.

[...]

PR #779, opened 2022-05-12
Low resolution Orion testing updates for dev_v16

KateFriedman-NOAA:
**Description**

This PR includes preliminary updates from testing the `dev_v16` branch at low resolutions on Orion. Additional tests on Orion will need to be performed, tests still need to happen on Hera, and updates from WCOSS2 testing will come in via separate PRs as well.

Changes included for now:

1. update verif_global tag to `verif_global_v2.9.0` in `Externals.cfg` (already in `checkout.sh`)
2. add `export launcher_PREP` settings to env files (`srun` on Hera/Orion and `mpiexec` on WCOSS2), needed for running obsproc/prepobs packages updated to support development GFS runs
3. add `export USE_CFP="NO"` for GLDAS on Orion (likely make a similar change for Hera when those tests are performed)
4. comment out p-dump if-block in `config.base.emc.dyn`, will evaluate deleting entirely in future, leaving for record for now
5. retirement of WCOSS_C and WCOSS_DELL_P3 references in numerous configs
6. updates to `config.fv3.emc.dyn` for supporting jobs that rely on node size, set C96 `DELTIM=450`, and pull in some WCOSS2 values while still supporting Orion
7. updates to `config.resources.emc.dyn` from a) `develop` enhancements (e.g. eobs), b) WCOSS2 testing, and c) testing on Orion; includes several checks of `npe_node_$step` values against `npe_node_max` to accommodate tested values on WCOSS2 and if larger than `npe_node_max` set `npe_node_$step` to `npe_node_max` (e,g, `if [[ "$npe_node_esfc" -gt "$npe_node_max" ]]; then export npe_node_esfc=$npe_node_max`)
8. add `export cdate10=${PDY}${cyc}` to `config.prep` for interfacing with obsproc package which now includes an override for the `cdate10` variable for GFS entry
9. update python_ver on Orion in `versions/orion.ver` to `export python_ver=3.7.5` based on testing

There will be a follow-up PR from `dev_v16` for updates from testing high res cycled on WCOSS2.

**Type of change**

- [x] Port/maintenance updates

**How Has This Been Tested?**

- [x] Cycled low res tests on Orion
- [x] Cycled high res tests on WCOSS2 (using nco resource configs though)

Refs #665


parm/config/config.base.emc.dyn
```diff
@@ -100,9 +100,9 @@ export EXPDIR="@EXPDIR@/$PSLOT"
 export ROTDIR="@ROTDIR@/$PSLOT"
 export ROTDIR_DUMP="YES"                #Note: A value of "NO" does not currently work
 export DUMP_SUFFIX=""
-if [[ "$CDATE" -ge "2019092100" && "$CDATE" -le "2019110700" ]]; then
-    export DUMP_SUFFIX="p"              # Use dumps from NCO GFS v15.3 parallel
-fi
+#if [[ "$CDATE" -ge "2019092100" && "$CDATE" -le "2019110700" ]]; then
```
WalterKolczynski-NOAA: Remove dead code?

parm/config/config.fv3.emc.dyn
```diff
@@ -22,10 +22,6 @@ echo "BEGIN: config.fv3"
 
 if [[ "$machine" = "WCOSS2" ]]; then
    export npe_node_max=128
-elif [[ "$machine" = "WCOSS_DELL_P3" ]]; then
```
WalterKolczynski-NOAA, line 25: I'd rather keep the Dell settings until it is actually retired. Cray is fine to remove, since it already doesn't work.

WalterKolczynski-NOAA (changes requested):

parm/config/config.base.emc.dyn
KateFriedman-NOAA: Yeah, I guess I'm not too attached to this block. We won't have these p-dumps on WCOSS2 anyway so no need. Will remove.

KateFriedman-NOAA: Removed!

parm/config/config.fv3.emc.dyn
KateFriedman-NOAA: I have put the WCOSS_DELL_P3 references back in. Will remove after official retirement.

parm/config/config.prep
```diff
@@ -9,6 +9,7 @@ echo "BEGIN: config.prep"
 . $EXPDIR/config.resources prep
 
 export DO_MAKEPREPBUFR="YES"   # if NO, will copy prepbufr from globaldump
+export cdate10=${PDY}${cyc}
```
aerorahul, line 12: Is PDY and cyc set at the job level or in `prep.sh` with `setPDY`?

KateFriedman-NOAA: When run via rocoto both `PDY` and `cyc` are exported at submission time. They are two of the known environment variables for each job at runtime. Example from the rocoto xml (specifically the prep task block):

```
        <envar><name>PDY</name><value><cyclestr>@Y@m@d</cyclestr></value></envar>
        <envar><name>cyc</name><value><cyclestr>@H</cyclestr></value></envar>
```

I considered using `CDATE` as well but I know we will be moving away from `CDATE` in the near future and replacing all `CDATE` instances with `$PDY$cyc` anyway.

In operations via ecflow setpdy.sh is used to set `PDY` and I see `cyc` set at submission, although in ops `cdate10` is set from `ncepdate` and the override I added into the obsproc scripts won't be used.

Right now setpdy.sh doesn't work outside of the operational environment since it uses a file that only exists in production. Here is what happens when setpdy.sh is invoked on Orion with `export COMROOT=/work/noaa/global/glopara/com` set from config.base (which I set for other uses and to avoid errors at runtime by not having `COMROOT` set):
```
0 + setpdy.sh
sed: can't read /work/noaa/global/glopara/com/date/t18z: No such file or directory
```

I considered and tried using setpdy.sh in the prep jobs on Orion but it would require setting up something similar to the `com/date/tCCz` files like is used in ops. For our development needs it's much simpler to set `cdate10` ourselves and have it used by the `cdate10` variable override in the obsproc script.

The setpdy.sh issue is related to general issues with `prod_util` outside of the operational environment and the various utilities within it being used elsewhere.

@lgannoaa When running the GFS in ecflow in the dev environment, how do you set `PDY` and `cyc` at runtime? Thanks!

lgannoaa: @KateFriedman-NOAA ecflow take PDY from include level envir-p1.h. CYC is set from the ecflow def files.

KateFriedman-NOAA: @lgannoaa Ok, thanks for that information!

@aerorahul We have `PDY` and `cyc` set at runtime from above in both the rocoto and ecflow setups.

parm/config/config.resources.emc.dyn
```diff
@@ -317,25 +333,33 @@ elif [ $step = "eupd" ]; then
...
-      export nth_eupd=14
+      export npe_eupd=480
+      export nth_eupd=6
       if [[ "$machine" = "WCOSS_DELL_P3" ]]; then
 	export npe_eupd=960
         export nth_eupd=7
       fi
+      if [[ "$machine" = "HERA" ]]; then
+        export npe_eupd=150
+        export nth_eupd=40
```
aerorahul, line 344: what? 40 threads?

KateFriedman-NOAA: This is what is already in feature/ops-orion for this job on Hera. There was a need to do this and it would need to be retested to confirm why and if still needed.
https://github.com/NOAA-EMC/global-workflow/blob/feature/ops-orion/parm/config/config.resources.emc.dyn#L321:L324
Didn't need to do this on Orion though.

WalterKolczynski-NOAA (approved):

parm/config/config.resources.emc.dyn
aerorahul: Ok. Thanks for providing history.

aerorahul (approved):
Thanks for addressing the comments. 
Looks good. 
We need this in sooner before Hera disrupts the work of our teams.

PR #794, opened 2022-05-17
Refactor some build scripts

### LINKED ISSUE #745: Consolidate UFS build

### LINKED ISSUE #751: Build DataAtmosphere configuration of the UFS-weather-model
The global-workflow currently build the "coupled" model.  It is noted that the coupled model has the ability to run in atmosphere-only mode.

This issue is to extend that to enable building the model in data atmosphere configuration in support of ocean DA capabilities.

Explore if the data-atmosphere configuration in the ufs-weather-model needs to be built differently.
If so, enable switches to allow building this configuration.

WalterKolczynski-NOAA:
**Description**
Refactors the build_ufs script to allow the build of any UFS app to support future expansion of global-workflow capability. The default should be sufficient for most users, as S2SWA can be used for ATM, ATMA, S2S, and S2SW as well. The new format of the command is now:
```
build_ufs.sh [-a UFS_app][-v]
  -a UFS_app:
    Specify the UFS application to build. The default if none is provided is S2SWA
  -v:
    Turn on verbose mode (BUILD_VERBOSE)
```

build_all.sh is similarly refactored to include the same options. Also, a new `-c build_config` option is added to specify an alternative list of programs to build.
```
build_all.sh [-a UFS_app][-c build_config][-h][-v]
  -a UFS_app:
    Build a specific UFS app instead of the default
  -c build_config:
    Selectively build based on the provided config instead of the default config
  -h:
    Print usage message and exit
  -v:
    Run all scripts in verbose mode
```

partial_build.sh is also updated to take the new `-c` option to specify a build config.
```
partial_build.sh [-c config_file][-h][-v]
  -c config_file:
    Selectively build based on the provided config. The default if none is specified is gfs_build.cfg
  -h:
    Print usage message and exit
  -v:
    Run in verbose mode
```

In addition to the above, build_all and partial_build had their indentation redone and their usage/help statement updated/added. Also, the build configuration file was renamed to from fv3gfs_build.cfg to gfs_build.cfg.

Fixes #745
Fixes #751

**Type of change**

**How Has This Been Tested?**

**Checklist**


JessicaMeixner-NOAA:
@WalterKolczynski-NOAA this is a really nice feature !  Have you thought about having optional inputs to add options to the ufs build such as debug mode ( -DDEBUG=ON), 32 bit builds (-D32BIT=ON), or other similar compile time options (-DMULTI_GASES=ON, -DMOVING_NEST=ON)?

WalterKolczynski-NOAA:
Debug is definitely a future possibility (for all build scripts), but we also don't want to give too many options up front either.

JessicaMeixner-NOAA:
The 32bit build is going to be a need fairly soon, so it seems smart to plan ahead for it while you're making these similar types of changes.

WalterKolczynski-NOAA:
@JessicaMeixner-NOAA with the removal of the command line option to specify CCPP suites, are there different suites we need to build with if we pass ATMW instead of the default list?

aerorahul:
@JessicaMeixner-NOAA @WalterKolczynski-NOAA 
I have mixed feelings about giving the user too much flexibility in configuring the model build options from inside the workflow.
The workflow does not support all the different combinations of the model.  If the user wants to do debugging of the model, they should be able to build the executable from the tag **outside** of the workflow and drop the executable in the `exec` folder, rather than calling the `build_all.sh` script and going through the hoops of the global-workflow.
My 2c.

JessicaMeixner-NOAA:
@aerorahul   I understand not wanting to give too much flexibility within the workflow.  However, I disagree with building the model outside of the workflow with the debug option. I'd just hard code it and use the script to keep things consistent.  Using a completely different build script, etc introduces possibility for unintentional divergence in options, etc.  If debug is needed, it can be manually hard-coded and instead of being added as a user-defined option. 

@WalterKolczynski-NOAA given: 
```
APP="S2SWA"
CCPP_SUITES="FV3_GFS_v16,FV3_GFS_v16_RRTMGP,FV3_GFS_v16_ugwpv1,FV3_GFS_v17_p8,FV3_GFS_v16_coupled_nsstNoahmpUGWPv1,FV3_GFS_v16_coupled_p7_rrtmgp,FV3_GFS_v17_coupled_p8"
```

I'd remove the "coupled" suites for ATMW and have: 
```
APP="ATMW"
CCPP_SUITES="FV3_GFS_v16,FV3_GFS_v16_RRTMGP,FV3_GFS_v16_ugwpv1,FV3_GFS_v17_p8"
```

WalterKolczynski-NOAA:
Okay, if ATMW uses the same suites, I don't see a need to change it.

I would love to trim the list of suites down though, if you can suggest ones from earlier in prototype testing we no longer need.

JessicaMeixner-NOAA:
I'd propose: 
CCPP_SUITES="FV3_GFS_v16,FV3_GFS_v17_p8,FV3_GFS_v16_coupled_nsstNoahmpUGWPv1,FV3_GFS_v17_coupled_p8"
since we need the P8 suite files which are FV3_GFS_v17_coupled_p8/FV3_GFS_v17_p8, and then we need something without Thompson MP for DA, so I'd recommend FV3_GFS_v16_coupled_nsstNoahmpUGWPv1 which is the p7 suite. The physics group might have additional suite files or requirements for their testing//development.

sorc/build_all.sh
```diff
@@ -1,4 +1,5 @@
-#!/bin/sh
+#! /usr/bin/env bash
+
 set +x
```
aerorahul, line 3: Does `set -eu` work? If so, please add.

JessicaMeixner-NOAA:
Looking at the latest [suite files](https://github.com/NOAA-EMC/fv3atm/tree/1e9c3da5e0cc9efcc434cb3a2ea08a9f521939d2/ccpp/suites), the following should be removed: 
FV3_GFS_v16_RRTMGP and FV3_GFS_v16_coupled_p7_rrtmgp 

Also, the README should probably also be updated with these new options when they're finalized.

aerorahul (approved):
looks good.

aerorahul:
> @aerorahul I understand not wanting to give too much flexibility within the workflow. However, I disagree with building the model outside of the workflow with the debug option. I'd just hard code it and use the script to keep things consistent. Using a completely different build script, etc introduces possibility for unintentional divergence in options, etc. If debug is needed, it can be manually hard-coded and instead of being added as a user-defined option.
> 
> @WalterKolczynski-NOAA given:
> 
> ```
> APP="S2SWA"
> CCPP_SUITES="FV3_GFS_v16,FV3_GFS_v16_RRTMGP,FV3_GFS_v16_ugwpv1,FV3_GFS_v17_p8,FV3_GFS_v16_coupled_nsstNoahmpUGWPv1,FV3_GFS_v16_coupled_p7_rrtmgp,FV3_GFS_v17_coupled_p8"
> ```
> 
> I'd remove the "coupled" suites for ATMW and have:
> 
> ```
> APP="ATMW"
> CCPP_SUITES="FV3_GFS_v16,FV3_GFS_v16_RRTMGP,FV3_GFS_v16_ugwpv1,FV3_GFS_v17_p8"
> ```

Ok.  As `Debug` is a very special case, it is for the user to edit `build_ufs.sh` and add the appropriate `DEBUG` flag to the `cmake -DAPP=${APP} -DCCPP_SUITES=${CCPP_SUITES} ..` line.

Out of curiosity, how often does one "debug" the model in the workflow rather than in the model regression tests controlled environment? And how many users do we have that need this level of convenience.

JessicaMeixner-NOAA:
@aerorahul to capture what we just discussed offline: I think it's fine to not put in the debug option and just keep the flexibility to specifying an APP. Debug is not used very often when trying to debug the model for the workflow.

JessicaMeixner-NOAA (approved):

binli2337 (approved):
Looks good. Thanks!

sorc/build_all.sh
WalterKolczynski-NOAA: Sorry, missed this before I merged. But adding `set -eu` to all scripts is part of my long-term plans (see Issue #397).

WalterKolczynski-NOAA:
> Looking at the latest [suite files](https://github.com/NOAA-EMC/fv3atm/tree/1e9c3da5e0cc9efcc434cb3a2ea08a9f521939d2/ccpp/suites), the following should be removed: FV3_GFS_v16_RRTMGP and FV3_GFS_v16_coupled_p7_rrtmgp
> 
> Also, the README should probably also be updated with these new options when they're finalized.

Looks like #795 is going to remove the dead suites, so I left it for there.

PR #801, opened 2022-05-20
Fix build script options

### LINKED ISSUE #800: Scripts called by build_all receive invalid options if none are set
**Current behavior**
If [build_all.sh](https://github.com/NOAA-EMC/global-workflow/blob/develop/sorc/build_all.sh) is called with command-line options, but none are passed to a downstream script, the script will error because `$@` is not reset.

For instance, if [build_all.sh](https://github.com/NOAA-EMC/global-workflow/blob/develop/sorc/build_all.sh) is called with only the `-a` option, [partial_build.sh](https://github.com/NOAA-EMC/global-workflow/blob/develop/sorc/partial_build.sh) should receive no arguments, but instead `$@` contains the `-a` option which causes partial_build.sh to fail because it doesn't expect it:

**Expected behavior**
Scripts should all work properly for any assortment of valid options

**Machines affected**
All

**To Reproduce**
```
hfe02 sorc/ > ./build_all.sh -a ATM
./build_all.sh: illegal option -- a
Unrecognized option: ?
Usage: ./partial_build.sh [-c config_file][-h][-v]
  -h:
    Print this help message and exit
  -v:
    Turn on verbose mode
  -c config_file:
    Override default config file to determine whether to build each program [default: gfs_build.cfg]
```
(getopts erroneously reports the error as being in build_all.sh, but it is in partial_build.sh)

**Context**
First reported by @CatherineThomas-NOAA

**Possible Implementation**
Consume `$@` using shift when processing options in build_all.sh, so it isn't accidentally passed to other scripts when called without arguments.

WalterKolczynski-NOAA:
**Description**
There was an issue where if build_all.sh was called with only options that are not used by partial_build.sh, partial_build.sh would see those options because `$@` was not overwritten. The getopts sections have been updated to consume those options so they are no longer present if `$@` isn't overwritten in the new script call.

Also fixed the getopts statement in partial_build.sh so we handle errors instead of bash (bash would report the wrong script if the script is sourced)

Fixes #800

**Type of change**

**How Has This Been Tested?**
  
**Checklist**

### DIFF
--- a/sorc/build_all.sh
+++ b/sorc/build_all.sh
@@ -36,17 +36,29 @@ _verbose_opt=""
 OPTIND=1
 while getopts ":a:c:hv" option; do
 	case "${option}" in
-		a) _build_ufs_opt+="-a ${OPTARG} ";;
-		c) _partial_opt+="-c ${OPTARG} ";;
-		h) _usage;;
+		a)
+			_build_ufs_opt+="-a ${OPTARG} "
+			shift 2
+			;;
+		c)
+			_partial_opt+="-c ${OPTARG} "
+			shift 2
+			;;
+		h)
+			_usage
+			shift 1
+			;;
 		# s) _build_ufs_opt+="-s ${OPTARG} ";;
-		v) _verbose_opt="-v";;
+		v)
+			_verbose_opt="-v"
+			shift 1
+			;;
 		\?)
-			echo "Unrecognized option: ${option}"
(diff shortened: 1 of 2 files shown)


CatherineThomas-NOAA (approved):
The branch builds successfully on Orion and Dell (@WalterKolczynski-NOAA already tested Hera).  Looks good to me.  Thanks for your quick action @WalterKolczynski-NOAA!

KateFriedman-NOAA (approved):

Issue #812, opened 2022-05-25
Update hpc-stack and compiler versions


WalterKolczynski-NOAA:
**Description**
HPCs are moving away from the Intel 2018 compiler, so we should update to one of the newer compiler versions. We should also take the opportunity to move to the hpc-stack/1.2.0

**Requirements**
Programs should compile and run using Intel 2021 or 2022 compilers

**Acceptance Criteria (Definition of Done)**
- All workflow-controlled modulefiles updated to use hpc-stack/1.2.0
- All workflow-controlled programs build using Intel 2021 or 2022
- All workflow jobs run loading Intel 2021 or 2022 at runtime
(These changes may also require other library updates, depending on what is available.)

**Dependencies**
Workflow-controlled programs have no dependencies, but those controlled by other components will need similar changes to move the entire system to a newer compiler. These may also be needed to prevent mismatches at runtime.

ulmononian:
Hi @WalterKolczynski-NOAA. I encountered an issue with the load of met/9.1 on Orion in S2SW (coupled, forecast-only) and ATM (uncoupled, forecast only) tests using the develop branch. The same error is produced in the gfsfcst task of the ATM case and gfswaveinit task of the S2SW case:
<img width="880" alt="Screen Shot 2022-05-23 at 11 42 45 AM" src="https://user-images.githubusercontent.com/43379611/170303137-81cdebbe-a838-4eab-a6d3-1d886703c4e4.png"> <img width="872" alt="Screen Shot 2022-05-23 at 11 55 25 AM" src="https://user-images.githubusercontent.com/43379611/170303191-a1bdb8ef-97a2-493f-a942-74514672c112.png">

I was able to bypass the issue by removing line 55 from module_base.orion.lua (load(pathJoin("met", "9.1"))). When I check the modules loaded by load_fv3gfs_modules.sh when testing outside of runtime, met/9.1 seems to still load despite removing this line:

<img width="672" alt="Screen Shot 2022-05-25 at 8 48 56 AM" src="https://user-images.githubusercontent.com/43379611/170304454-e3c804f1-802c-4b13-a250-775cbb4257fb.png">



I did not know if this was worth its own issue, and thought perhaps it was related to this issue. Thank you, Cameron.

WalterKolczynski-NOAA:
> I did not know if this was worth its own issue, and thought perhaps it was related to this issue. Thank you, Cameron.

This looks like an unrelated issue. Please open a new one.

[...]

PR #831, opened 2022-06-01
updates for wave mesh cap in workflow

### LINKED ISSUE #736: Update wave related settings and code for P8c

JessicaMeixner-NOAA:
**Description**

Updates to work with the new WW3 mesh cap, including pointing to the new ufs-weather-model tag for p8c. 
The log files were removed for the non-pdgen wave script and ush files. 

Fixes #736 

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)

**How Has This Been Tested?**

- [x] Clone and Build tests on WCOSS Dell P3
- [x] Coupled with aero forecast test on Orion
- [x] Coupled no-aero forecast  test on Hera

**Checklist**


modulefiles/module_base.hera.lua
```diff
@@ -46,12 +46,7 @@ load(pathJoin("cdo", "1.9.5"))
 load(pathJoin("R", "3.5.0"))
 
 -- Temporary until official hpc-stack is updated
-prepend_path("MODULEPATH", "/scratch2/NCEPDEV/ensemble/save/Walter.Kolczynski/hpc-stack/modulefiles/stack")
-load(pathJoin("hpc", "1.2.0"))
-load(pathJoin("hpc-intel", "18.0.5.274"))
 load(pathJoin("hpc-miniconda3", "4.6.14"))
 load(pathJoin("ufswm", "1.0.0"))
-load(pathJoin("met", "9.1"))
-load(pathJoin("metplus", "3.1"))
```
WalterKolczynski-NOAA: Why was all this deleted? We still need all this to run verification.

parm/config/config.defaults.s2sw
```diff
@@ -17,25 +17,32 @@ use_cice_alb=".true."
 
 # config.fv3
 DELTIM=300
-layout_x_gfs=24
-layout_y_gfs=16
+#layout_x_gfs=24
+#layout_y_gfs=16
```
WalterKolczynski-NOAA: Please remove the dead code

sorc/build_ufs.sh
```diff
@@ -8,7 +8,6 @@ CCPP_SUITES="FV3_GFS_v16,FV3_GFS_v16_ugwpv1,FV3_GFS_v17_p8,FV3_GFS_v16_coupled_n
 while getopts "a:s:v" option; do
   case "${option}" in
     a) APP="${OPTARG}" ;;
-    # s) CCPP_SUITES="${OPTARG}";;
```
WalterKolczynski-NOAA, line 11: We can remove the 's' from getopts too.

sorc/checkout.sh
```diff
@@ -29,9 +29,11 @@ mkdir -p ${logdir}
 
 echo ufs-weather-model checkout ...
 if [[ ! -d ufs_model.fd ]] ; then
-    git clone https://github.com/ufs-community/ufs-weather-model ufs_model.fd >> ${logdir}/checkout-ufs_model.log 2>&1
+    #git clone https://github.com/ufs-community/ufs-weather-model ufs_model.fd >> ${logdir}/checkout-ufs_model.log 2>&1
+    git clone https://github.com/DeniseWorthen/ufs-weather-model ufs_model.fd >> ${logdir}/checkout-ufs_model.log 2>&1
     cd ufs_model.fd
-    git checkout ${ufs_model_hash:-5c2d1a9}
+    #git checkout ${ufs_model_hash:-5c2d1a9}
+    git checkout feature/cmake_meshcapbuild
```
WalterKolczynski-NOAA: This needs to be a hash or tag, not a branch

sorc/checkout.sh
```diff
@@ -29,9 +29,11 @@ mkdir -p ${logdir}
 
 echo ufs-weather-model checkout ...
 if [[ ! -d ufs_model.fd ]] ; then
-    git clone https://github.com/ufs-community/ufs-weather-model ufs_model.fd >> ${logdir}/checkout-ufs_model.log 2>&1
+    #git clone https://github.com/ufs-community/ufs-weather-model ufs_model.fd >> ${logdir}/checkout-ufs_model.log 2>&1
+    git clone https://github.com/DeniseWorthen/ufs-weather-model ufs_model.fd >> ${logdir}/checkout-ufs_model.log 2>&1
```
WalterKolczynski-NOAA: Is this going to be merged into UFS soon?

Also, this whole bit will need to be redone after the checkout refactor that just went in. Should be simple though.

ush/forecast_postdet.sh
```diff
@@ -614,13 +614,18 @@ WW3_postdet() {
   COMPONENTwave=${COMPONENTwave:-${RUN}wave}
 
   #Link mod_def files for wave grids
-  array=($WAVECUR_FID $WAVEICE_FID $WAVEWND_FID $waveuoutpGRD $waveGRD $waveesmfGRD $wavesbsGRD $wavepostGRD $waveinterpGRD)
-  echo "Wave Grids: $WAVECUR_FID $WAVEICE_FID $WAVEWND_FID $waveuoutpGRD $waveGRD $waveesmfGRD $wavesbsGRD $wavepostGRD $waveinterpGRD"
-  grdALL=$(printf "%s\n" "${array[@]}" | sort -u | tr '\n' ' ')
+  if [ $waveMULTIGRID = ".true." ]; then
```
WalterKolczynski-NOAA, line 617: Instead of setting a new variable beforehand, why not just determine multi-grid from grdALL?

ush/forecast_postdet.sh
```diff
@@ -614,13 +614,18 @@ WW3_postdet() {
...
 
-  for wavGRD in ${grdALL}; do
-    $NCP $ROTDIR/${CDUMP}.${PDY}/${cyc}/wave/rundata/${COMPONENTwave}.mod_def.$wavGRD $DATA/mod_def.$wavGRD
-  done
+    for wavGRD in ${grdALL}; do
+      $NCP $ROTDIR/${CDUMP}.${PDY}/${cyc}/wave/rundata/${COMPONENTwave}.mod_def.$wavGRD $DATA/mod_def.$wavGRD
+    done
+  else 
+    #if shel, only 1 waveGRD which is linked to mod_def.ww3 
+    $NCP $ROTDIR/${CDUMP}.${PDY}/${cyc}/wave/rundata/${COMPONENTwave}.mod_def.$waveGRD $DATA/mod_def.ww3
```
WalterKolczynski-NOAA, line 627: Did WW3 change this to a more clumsy way of using a different suffix when there is only one grid? Or did we just always assume multi-grid before? Either way I'd love some consistency from WW3 here.

ush/forecast_postdet.sh
```diff
@@ -645,13 +650,21 @@ WW3_postdet() {
...
+      fi
     fi
   done  
 
-  for wavGRD in $waveGRD ; do
-    eval $NLN $datwave/${wavprfx}.log.${wavGRD}.${PDY}${cyc} log.${wavGRD}
-  done 
+  if [ $waveMULTIGRID = ".true." ]; then
+    for wavGRD in $waveGRD ; do
+      eval $NLN $datwave/${wavprfx}.log.${wavGRD}.${PDY}${cyc} log.${wavGRD}
```
WalterKolczynski-NOAA: Can we do all these without using eval? I know that's how it was done before, but I'd like to reduce the number of evals in our scripts, not increase them.

ush/forecast_postdet.sh
```diff
@@ -709,11 +733,18 @@ WW3_nml() {
   echo "SUB ${FUNCNAME[0]}: Copying input files for WW3"
   WAV_MOD_TAG=${CDUMP}wave${waveMEMB}
   if [ "${USE_WAV_RMP:-YES}" = "YES" ]; then
-    for file in $(ls $COMINwave/rundata/rmp_src_to_dst_conserv_*) ; do
-      $NLN $file $DATA/
-    done
-  fi 
-  $NLN $COMINwave/rundata/ww3_multi.${CDUMPwave}${WAV_MEMBER}.${cycle}.inp $DATA/ww3_multi.inp
+    if ls $FIXwave/rmp_src_to_dst_conserv_* 2> /dev/null
```
WalterKolczynski-NOAA: I'd rather this be something more readable, like:
`if (( $( ls -1 $FIXwave/rmp_src_to_dst_conserv_* > /dev/null | wc -l) > 0 )); then`

It does the same thing, but it is clearer we are counting the number of files and checking against zero.

modulefiles/module_base.hera.lua
```diff
@@ -5,37 +5,37 @@ Load environment to run GFS on Hera
 prepend_path("MODULEPATH", "/scratch2/NCEPDEV/nwprod/hpc-stack/libs/hpc-stack/modulefiles/stack")
 
 load(pathJoin("hpc", "1.1.0"))
-load(pathJoin("hpc-intel", "18.0.5.274"))
-load(pathJoin("hpc-impi", "2018.0.4"))
```
WalterKolczynski-NOAA: If we're going to update the compiler and module versions, we should do it for Orion too (Dell doesn't have the newer compiler and will be decommissioned soon).

sorc/build_ww3prepost.sh
```diff
@@ -1,6 +1,20 @@
 #!/bin/sh
 set -x
 
+# Determine which switch to use 
+ww3switch=model/esmf/switch
+while getopts "c" option; do
+  case "${option}" in
+    c)
```
WalterKolczynski-NOAA: When does -c get used? There is also no hook in build_all to turn it on. Does it need to be coordinated with build_ufs somehow?

WalterKolczynski-NOAA (changes requested):
With the compiler and module changes, we'll need to test this uncoupled as well.

modulefiles/module_base.hera.lua
JessicaMeixner-NOAA: Actually the verification job ran? 

And there is no met/metplus for the updated intel and I thought I understood from you that this shouldn't hold up the PR? Maybe I'm mistaken?

sorc/checkout.sh
WalterKolczynski-NOAA: Looks like you took care of this while I was reviewing.

sorc/checkout.sh
WalterKolczynski-NOAA: Looks like you took care of this while I was reviewing.

JessicaMeixner-NOAA: looks like the most recent code change changing this to a tag didn't get updated, it should be fixed with the merged in code now.

modulefiles/module_base.hera.lua
WalterKolczynski-NOAA: Looks like you took care of this while I was reviewing.

sorc/checkout.sh
JessicaMeixner-NOAA: Yeah, somehow my last change reverting this didn't get in, it's changed to be pointing to the p8c tag now.

ush/forecast_postdet.sh
JessicaMeixner-NOAA: We always used multi-grid before.  Shel uses 'ww3' as the extension.  Even if WW3 changes, it's not going to be changed in time for this PR.

ush/forecast_postdet.sh
JessicaMeixner-NOAA: I will try to see if this works with out the eval and report back.  As you mentioned, this is how it was done before and not something new.

modulefiles/module_base.hera.lua
JessicaMeixner-NOAA: Orion has now been updated, an oversight.  Working on the testing there now.

ush/forecast_postdet.sh
JessicaMeixner-NOAA: It cannot be simply determined from grdALL and therefore requires a new variable.

modulefiles/module_base.hera.lua
WalterKolczynski-NOAA: I think verification "runs" but is just skipped for coupled. I think it may just be several cycles before it kicks in.

I definitely don't want to re-break verification right after I fixed it. I'm trying to push hpc-stack to get the PR in to add MET and MET+ to the official installations, but if you have to use the updated compiler and loading the old one at the end unloads the stuff you need, I can try to build MET and MET+ myself. Right now I'm pointing to a hacked version of the system one in a custom stack location.

ush/forecast_postdet.sh
WalterKolczynski-NOAA: I knew WW3 wouldn't change in time for this PR, just annoying we have to have all these ifs because the filename pattern changes. Could we instead have a variable like wave_suffix that equals wavGRD if multi-grid and 'ww3' otherwise? Then we can just always loop over that and eliminate the rest of these new ifs.

sorc/build_ww3prepost.sh
JessicaMeixner-NOAA: @WalterKolczynski-NOAA thanks for finding this... it was passed in before at one point, I guess it got messed up in a bad merge when the -c was removed from the build_all.  I'll have to re-think this now.

modulefiles/module_base.hera.lua
JessicaMeixner-NOAA: I don't want to break verification either, but it seems ran and loading the old intel is not going to work. Please let me know what the best path forward is.

sorc/build_ww3prepost.sh
WalterKolczynski-NOAA: Maybe use `-m` for 'meshcap' in `build_all.sh` instead, since `-c` is taken? Can easily turn -m for build_all into -c for build_ww3prepost.

WalterKolczynski-NOAA: Nevermind, the way you changed it to use APP is better.

JessicaMeixner-NOAA: It's based on the APP, so ATMW (pervious -c) versus S2S*, so I've updated this to mimic to ufs build options.

sorc/build_ww3prepost.sh
```diff
@@ -1,6 +1,29 @@
...
+      ;;
+  esac
+done
+
+
+# Determine which switch to use 
+if [ APP == ATMW ]; then 
+  ww3switch=model/esmf/switch 
+else 
+  ww3switch=model/bin/switch_meshcap
```
WalterKolczynski-NOAA, line 23: Does `$SWITCHFILE` down below need to be updated to use this variable?

WalterKolczynski-NOAA (changes requested):

sorc/build_ww3prepost.sh
JessicaMeixner-NOAA: yes! thank you for catching this too!!

modulefiles/module_base.hera.lua
WalterKolczynski-NOAA: Okay, let me see what I can do about building MET and MET+.

JessicaMeixner-NOAA:
@WalterKolczynski-NOAA tests with the updated code from review are ongoing.   I believe there is only one major unresolved issue which is the Met/MetPlus modules.  On orion there also was not the ufswm module for the hpc-miniconda3.  I'm not sure if that's also a Met/MetPlus verification requirement or something else?

WalterKolczynski-NOAA:
> @WalterKolczynski-NOAA tests with the updated code from review are ongoing. I believe there is only one major unresolved issue which is the Met/MetPlus modules. On orion there also was not the ufswm module for the hpc-miniconda3. I'm not sure if that's also a Met/MetPlus verification requirement or something else?

That's why it was using the one I built. It's a requirement for aerosol init, and is what started the whole issue with verification in the first place.

JessicaMeixner-NOAA:
Can we create a second set of modules for the aerosol init job and verification jobs to use the older version of intel with your hpc-stack modules?  Or has anyone started on the issue https://github.com/NOAA-EMC/global-workflow/issues/812 which is more/less what some of this is requiring/doing? 

The orion test is currently running (it was in the queue from yesterday until early this morning) 

And I ran both coupled with and without aerosol on hera: 
/scratch1/NCEPDEV/climate/Jessica.Meixner/p8c/reviews/wave01
/scratch1/NCEPDEV/climate/Jessica.Meixner/p8c/reviews/aero01

@WalterKolczynski-NOAA are there other tests I should be running? Should I be reaching out to the hpc-stack team that we have ufswm on hera but not on orion?  Metplus?

JessicaMeixner-NOAA:
I have commented on the hpc-stack issue for the ufswm module on orion https://github.com/NOAA-EMC/hpc-stack/issues/449 
and for the PR for the MetPlus https://github.com/NOAA-EMC/hpc-stack/pull/324 about the outstanding module issues for this PR.

JessicaMeixner-NOAA:
The experiment for rolling back the modules has started and it's output (com, etc) can be found here: 
/scratch1/NCEPDEV/climate/Jessica.Meixner/p8c/rollbackmods/roll01

JessicaMeixner-NOAA:
The roll back test succeeded on hera, so I committed the changes to the branch here.

WalterKolczynski-NOAA (approved):
Approved pending testing.

JessicaMeixner-NOAA:
The test on orion has finished: /work/noaa/marine/jmeixner/p8c/aero01 
the coupled_ic job failed because the wave ICs need to be resync'd after the permissions issues were fixed on Hera. I manually copied over the IC for the test.

PR #849, opened 2022-06-10
Update experiment setup script for free-forecast ensemble

### LINKED ISSUE #789: Update experiment setup script for free-forecast ensemble
**Description**
As we move towards the merger of GEFS into the GFS system, we need to be able to generate experiments with an arbitrary number of free-forecast ensemble members in addition to the number of EnKF members.

**Requirements**
Ability to generate experiments (using `setup_expt.py`) with a number of free-forecast ensemble members determined via a command-line switch similar to how the number of EnKF members is determined. The default should be 0, indicating no perturbed members (control only). This will require `config.base` to be updated to provide a corresponding new variable which will be used by all the job scripts later.

A `setup_expt.py` script that takes in a number of member for the free-forecast and sets an appropriate variable in `config.base`.

**Dependencies**
None

XianwuXue-NOAA:
**Description**

As we move towards the merger of GEFS into the GFS system, we need to be able to generate experiments using setup_expt.py with an arbitrary number of free-forecast ensemble members in addition to the number of EnKF members.

**Type of change**

- [x] This change requires a documentation update (add "gefs" to setup_expt.py)

**How Has This Been Tested?**

Tested on Hera for ATM only GEFS forecast-only

  
**Checklist**


ush/rocoto/setup_expt.py
```diff
@@ -30,7 +30,8 @@ def fill_COMROT(host, inputs):
 
     fill_modes = {
         'cycled': fill_COMROT_cycled,
-        'forecast-only': fill_COMROT_forecasts
+        'forecast-only': fill_COMROT_forecasts,
+        'gefs': fill_COMROT_gefs
```
WalterKolczynski-NOAA: Instead of adding a new mode, there should be a new command line option to specify the number of ensemble members (with a default of 0). While we are starting with forecast-only, at some point in the future we will want to run full cycling and also have a forecast ensemble.

ush/rocoto/setup_expt.py
```diff
@@ -231,18 +241,21 @@ def input_args():
                           choices=['warm', 'cold'], required=False, default='cold')
 
     # cycled mode additional arguments
-    cycled.add_argument('--resens', help='resolution of the ensemble model forecast',
-                        type=int, required=False, default=192)
-    cycled.add_argument('--nens', help='number of ensemble members',
-                        type=int, required=False, default=20)
+    for subp in [cycled, gefs]:
```
WalterKolczynski-NOAA: New options need to be created for the forecast ensemble rather than overloading the options that were created for EnKF. Existing options should be renamed `--nens-enkf` and `--res-enkf` to avoid confusion.

ush/rocoto/setup_expt.py
```diff
@@ -147,12 +155,12 @@ def edit_baseconfig(host, inputs):
         "@APP@": inputs.app,
     }
 
-    if inputs.mode in ['cycled']:
+    if inputs.mode in ['cycled', 'gefs']:
         extend_dict = {
             "@CASEENS@": f'C{inputs.resens}',
```
WalterKolczynski-NOAA: This variable name will need to be changed in config.base and everywhere downstream to something like `$CASE_ENKF` to avoid confusion with new resolution variable for GEFS.

WalterKolczynski-NOAA (changes requested):
This needs to be reworked to be more inline with what we discussed previously (unless I'm forgetting or missed a conversation).

Also, the PR title and description need a *lot* of work, as they are practically non-existent right now.

ush/rocoto/setup_expt.py
WalterKolczynski-NOAA: Actually, we might want to use -1 as the default (meaning high-res only). For now assume GEFS-resolution only if the number of members >= 0, though we'll probably need a new switch in the future to turn off the high-res instead, since we'll want the option to everything (GFS, GEFS, GDAS, EnKF) together at some point, but we don't always want to run high-res when we run GEFS.

ush/rocoto/setup_expt.py
XianwuXue-NOAA: OK. I will add --nens-gefs, --res-gefs to setup_expt.py and also make some changes in config.gefs.

ush/rocoto/setup_expt.py
XianwuXue-NOAA: How about I change to, however, I keep inputs.resens and inputs.nens, then users don't need to change command options

    if inputs.mode in ['cycled']:
        extend_dict = {
            "@CASEENS@": f'C{inputs.resens}',
            "@NMEM_ENKF@": inputs.nens,
        }
    elif inputs.mode in ['gefs']:
        extend_dict = {
            "@CASEENS_GEFS@": f'C{inputs.resens}',
            "@NMEM_GEFS@": inputs.nens,
        }
    elif inputs.mode in ['forecast-only', 'gefs']:
        extend_dict = {
            "@DO_AERO@": inputs.aerosols,
        }

ush/rocoto/setup_expt.py
XianwuXue-NOAA: How about the changes below:

    if inputs.mode in ['cycled']:
        extend_dict = {
            "@CASEENS@": f'C{inputs.resens}',
            "@NMEM_ENKF@": inputs.nens,
        }
    elif inputs.mode in ['gefs']:
        extend_dict = {
            "@CASEENS_GEFS@": f'C{inputs.resens}',
            "@NMEM_GEFS@": inputs.nens,
        }
    elif inputs.mode in ['forecast-only', 'gefs']:
        extend_dict = {
            "@DO_AERO@": inputs.aerosols,
        }

ush/rocoto/setup_expt.py
WalterKolczynski-NOAA: This suggested code won't even work properly, and I still want to update the options so it is clear they belong to EnKF. As always when we change a script signature, we will make an announcement when the change is merged so everyone is aware.

WalterKolczynski-NOAA: Plus, I still don't want to add gefs as a new mode.

XianwuXue-NOAA: If you don't want to add gefs as a new mode, do you have any suggestions?

> This suggested code won't even work properly, and I still want to update the options so it is clear they belong to EnKF. As always when we change a script signature, we will make an announcement when the change is merged so everyone is aware.

Yes, I should also change the configure file. If you agree these changes, I will work on it.

XianwuXue-NOAA: > Plus, I still don't want to add gefs as a new mode.

I think adding gefs as a new mode is clear. Any suggestions?

WalterKolczynski-NOAA: Yes, it's in the first change request (along with why)

XianwuXue-NOAA: Originally, I want to reuse the existing variables. However, based on this conversation, I changed my mind to add new variables. If I misunderstood, please let me know

XianwuXue-NOAA:
@WalterKolczynski-NOAA I have modified the variables and also added new options for GEFS.

.gitignore
```diff
@@ -122,7 +122,7 @@ sorc/wafs_gcip.fd
 sorc/wafs_grib2_0p25.fd
 sorc/wafs_makewafs.fd
 sorc/wafs_setmissing.fd
-
+sorc/upp.fd
```
aerorahul: how is this related to this PR?

parm/config/config.gefs
```diff
@@ -0,0 +1,11 @@
+#!/bin/ksh -x
```
aerorahul, line 1: Which job is sourcing this file?
It may be needed in subsequent PR's.
Consider adding when it is used.

ush/rocoto/workflow_utils.py
```diff
@@ -272,6 +272,9 @@ def source_configs(configs, tasks):
         elif 'wave' in task:
             files.append(find_config(f'config.wave', configs))
             files.append(find_config(f'config.{task}', configs))
+        elif 'gefs' in task:
+            files.append(find_config(f'config.gefs', configs))
+            files.append(find_config(f'config.{task}', configs))
```
aerorahul: There are no gefs tasks currently, so this is a placeholder.
Since there is work ongoing on refactoring setup_workflow, this could be held till actual tasks are being added.

parm/config/config.base.emc.dyn
```diff
@@ -224,6 +225,8 @@ if [[ "$CDUMP" == "gdas" ]] ; then
    export FTSFS=10
 elif [[ "$CDUMP" == "gfs" ]] ; then
    export FHCYC=24
+elif [[ "$CDUMP" == "gefs" ]] ; then
+   export FHCYC=24
```
aerorahul: May be replace lines 223-230 with `case ${CDUMP} in` statements and condense `gfs` and `gefs`

parm/config/config.base.emc.dyn
```diff
@@ -207,7 +207,8 @@ export OPS_RES="C768" # Do not change
 # Resolution specific parameters
 export LEVS=128
 export CASE="@CASECTL@"
-export CASE_ENKF="@CASEENS@"
+export CASE_ENKF="@CASEENS_ENKF@"
+export CASE_GEFS="@CASEENS_GEFS@"
```
aerorahul: I am still not convinced we will have a use case for running GEFS that is a resolution different than the cycled ensemble resolution.

aerorahul (changes requested):
please see comments.

.gitignore
XianwuXue-NOAA: No relate to this PR, so I will delete it.

parm/config/config.gefs
XianwuXue-NOAA: Currently, this config file is just for placeholder. If there is no need for now until we need this file, I can delete this file.

ush/rocoto/workflow_utils.py
XianwuXue-NOAA: Currently, I use the existing getic, init and fcst tasks for GEFS, later, I will add recenter task for GEFS. OK, I can delete these lines until actual tasks are being added

parm/config/config.base.emc.dyn
XianwuXue-NOAA: OK. I will do it.

parm/config/config.base.emc.dyn
XianwuXue-NOAA: @aerorahul Any suggestions? Should I change it back to 'export CASE_ENKF="@CASEENS@"' and the corresponding changes for the options or keep this changes?

PR #862, opened 2022-06-15
Resource adjustments and updated obsproc/prepobs packages for tcvitals bug fix

KateFriedman-NOAA:
**Description**

This PR includes memory updates and new obsproc/prepobs package versions:

- Memory increases for some non-exclusive jobs that were getting the `Cgroup mem limit exceeded` warning message on WCOSS2.
- Also translated a few more high res WCOSS2 resource settings from `config.resources.nco.static` to `config.resources.emc.dyn` and from `config.fv3.nco.static` to `config.fv3.emc.dyn`.
- New obsproc R&D v1.0.2, includes WCOSS2 ops updates and tcvitals bug fix. Updated `obsproc_run_ver` in `versions/*.ver` files for WCOSS2, Hera, Orion.
- New prepobs R&D v1.0.1, includes WCOSS2 ops updates and tcvitals bug fix. Updated `prepobs_run_ver` in `versions/*.ver` files for WCOSS2, Hera, Orion.

New obsproc/prepobs packages have been installed on WCOSS2, Hera, and Orion.

Will merge these changes into the `release/gfs.v16.3.0` branch next.

**Type of change**

**How Has This Been Tested?**

- [x] Clone and build tests on WCOSS2, Hera, Orion (obsproc/prepobs packages)
- [x] Cycled test on WCOSS2
  
Refs #665, #744

### DIFF
--- a/parm/config/config.resources.emc.dyn
+++ b/parm/config/config.resources.emc.dyn
@@ -64,7 +64,8 @@ elif [ $step = "waveprep" ]; then
     export nth_waveprep=1
     export npe_node_waveprep=$npe_waveprep
     export npe_node_waveprep_gfs=$npe_waveprep_gfs
-    export memory_waveprep="200GB"
+    export memory_waveprep="100GB"
+    export memory_waveprep_gfs="200GB"
     export NTASKS=$npe_waveprep
     export NTASKS_gfs=$npe_waveprep_gfs
 
@@ -110,7 +111,7 @@ elif [ $step = "wavegempak" ]; then
     export nth_wavegempak=1
     export npe_node_wavegempak=$npe_wavegempak
     export NTASKS=$npe_wavegempak
-    export memory_wavegempak="1GB"
+    export memory_wavegempak="10GB"
 
 elif [ $step = "waveawipsbulls" ]; then
 
--- a/parm/config/config.resources.emc.dyn
+++ b/parm/config/config.resources.emc.dyn
@@ -43,7 +43,9 @@ if [ $step = "prep" -o $step = "prepbufr" ]; then
     eval "export npe_$step=4"
     eval "export npe_node_$step=2"
     eval "export nth_$step=1"
-    eval "export memory_$step=40G"
+    if [[ "$machine" != "WCOSS2" ]]; then
+      eval "export memory_$step=40G"
+    fi
 
 elif [ $step = "waveinit" ]; then
 
@@ -62,7 +64,7 @@ elif [ $step = "waveprep" ]; then
     export nth_waveprep=1
     export npe_node_waveprep=$npe_waveprep
     export npe_node_waveprep_gfs=$npe_waveprep_gfs
-    export memory_waveprep="150GB"
+    export memory_waveprep="200GB"
     export NTASKS=$npe_waveprep
     export NTASKS_gfs=$npe_waveprep_gfs
 
@@ -74,7 +76,7 @@ elif [ $step = "wavepostsbs" ]; then
     export nth_wavepostsbs=1
     export npe_node_wavepostsbs=$npe_wavepostsbs
     export memory_wavepostsbs="10GB"
-    export memory_wavepostsbs_gfs="10GB"
+    export memory_wavepostsbs_gfs="40GB"
     export NTASKS=$npe_wavepostsbs
 
 elif [ $step = "wavepostbndpnt" ]; then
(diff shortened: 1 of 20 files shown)


WalterKolczynski-NOAA (approved):

parm/config/config.resources.emc.dyn
```diff
@@ -43,7 +43,9 @@ if [ $step = "prep" -o $step = "prepbufr" ]; then
     eval "export npe_$step=4"
     eval "export npe_node_$step=2"
     eval "export nth_$step=1"
-    eval "export memory_$step=40G"
+    if [[ "$machine" != "WCOSS2" ]]; then
+      eval "export memory_$step=40G"
+    fi
```
aerorahul, line 48: What value for memory is being given on WCOSS2?
Not sure if this warrants an `if-block`

parm/config/config.resources.emc.dyn
```diff
@@ -192,15 +195,16 @@ elif [ $step = "fcst" ]; then
     export npe_node_fcst=$(echo "$npe_node_max / $nth_fcst" | bc)
     export npe_node_fcst_gfs=$(echo "$npe_node_max / $nth_fcst_gfs" | bc)
     if [[ "$machine" == "WCOSS2" ]]; then
+       export wtime_fcst_gfs="02:30:00"
```
aerorahul: Is this really necessary?
Or is WCOSS2 really that fast, that `6:00:00` on line 190 a "safe bet"?

parm/config/config.resources.emc.dyn
```diff
@@ -192,15 +195,16 @@ elif [ $step = "fcst" ]; then
...
        export npe_node_fcst=32
-       export npe_node_fcst_gfs=42
+       export npe_node_fcst_gfs=24
```
aerorahul, line 199: How do the values on L200 and L199 differ from the calculations on L195 and L196?

aerorahul (commented):
Nothing major, but a few questions.

parm/config/config.resources.emc.dyn
KateFriedman-NOAA: The prep job runs exclusively on one node on WCOSS2 so we don't want a memory setting for it in here (setup scripts only set exclhost in the xml if no memory is set for job). The prep job fails with 40G on WCOSS2 when tested, 40G is the R&D value that has been tested.

parm/config/config.resources.emc.dyn
KateFriedman-NOAA: L195 = 128/3 = 42.6, which rounds up to 43 in the xml
L196 = 128/5 = 25.6, which rounds up to 26 in the xml

Neither value lays nicely across the WCOSS2 nodes.

George V and I did a lot of testing to get the values on L199 and L200. We want users using these values for C768 on WCOSS2. May also want them for lower resolutions too but haven't fully vetted those resolutions on WCOSS2 yet.

This reminded me I forgot to adjust the C768 block in `config.fv3.emc.dyn` with the tested values for `layout_x_gfs` and `WRTTASK_PER_GROUP_GFS` from `config.fv3.nco.static`. Please see that change in this PR now too.

parm/config/config.resources.emc.dyn
KateFriedman-NOAA: The 6hr gfs runtime covers the job for all resolutions on all machines, so yeah, line 190 is the "safe bet". It's not super necessary (beyond the possibility of jobs waiting longer on WCOSS2 down the road when the machine gets busy) so I have removed the line. Can review runtimes again later.

parm/config/config.resources.emc.dyn
aerorahul: Ok. thanks.  
In a subsequent PR, we might want to know what the memory footprint is on WCOSS2

parm/config/config.resources.emc.dyn
aerorahul: Ok. So these values are specific to C768.
May be then please add a note to that effect.

aerorahul (approved):
Looks good.  Just add a note about C768 for those calculations and that they will be revised for other resolutions when we get to it.

parm/config/config.resources.emc.dyn
KateFriedman-NOAA: Will do, thanks!

parm/config/config.resources.emc.dyn
KateFriedman-NOAA: Done. I added a check for CASE=768 to make sure those values aren't used outside of high res on WCOSS2.

Issue #415, opened 2021-08-17
Rework firstcyc job


KateFriedman-NOAA:
Cycled mode currently has a `firstcyc` job in the first half cycle. It's dependency is currently on the existence of the rocoto log 4 cycles later. This dependency often holds up the forward progress of the 4th cycle when the cycle throttle of 3 is met because firstcyc isn't done yet and therefore the first half cycle can't complete and become Inactive (freeing up a cycle in cyclethrottle).

Easy workaround is to increase cycle throttle to 4 in live experiments but better yet, a review of the need/use of `firstcyc` job should happen. Review how the `firstcyc` job is used and, if needed, update what its dependency should be. Remove if not truly required.

KateFriedman-NOAA:
Work complete, thanks @JianKuang-UMD !

PR #66, opened 2020-05-12
Gefswave gempak rstrt

RobertoPadilla-NOAA:
Restart for wave_gempak 
Fixing filenames for wave_stat if SENDDBN=YES

PR #871, opened 2022-06-21
Add GDASapp (first wave of JEDI changes)

### LINKED ISSUE #521: Extend global workflow to include JEDI based atmospheric DA
This issue is opened to document the addition of a JEDI based atmospheric DA component to the global workflow.   Please see [Prototype 3DVar Cycling Capability](https://github.com/NOAA-EMC/JEDI-Eval/projects/2) for additional information.

RussTreadon-NOAA:
**Description**
This PR is opened to merge changes associated with g-w issue #521 from g-w branch feature/ufsda_gdasapp into develop.   feature/ufsda_gdasapp contains g-w extensions for JEDI based atmospheric DA.   

Specifically, this PR adds the option to add JEDI based variational and/or ensemble DA jobs to replace GSI based DA jobs.   The toggling on/off of JEDI_VAR and JEDI_ENS jobs is controlled via two new variables added to `config.base.emc.dyn` and `config.base.nco.static`
```
# DA engine
export DO_JEDIVAR="NO"
export DO_JEDIENS="NO"
```
When both variables are `NO`, the global workflow uses GSI based DA jobs.   Thus, the PR does not alter the default behavior of the develop global workflow.

When `DO_JEDIVAR=YES`, GSI jobs `anal` and `analdiag` are replaced by JEDI_VAR jobs `atmanalprep`, `atmanalrun`, and `atmanalpost`.  When `DO_JEDIENS=YES`, GSI jobs `eobs`, `ediag`, and `eupd` are replaced by JEDI_ENS jobs `atmensanalprep`, `atmensanalrun`, and `atmensanalpost`.   

`checkout.sh`, `build_all.sh`, and `link_workflow.sh` are updated to clone, build, and install the GDASapp in the global workflow. 
  Local directory `sorc/gdas.cd` contains the GDASApp superstructure plus the relevant components of JEDI needed to run GDASApp.

Closes #521

**Type of change**

**How Has This Been Tested?**
Cycled parallels have been run on Orion using a C96C48L127 testbed over the period 21-31 December 2021.   GSI and JEDI based combinations have been tested and found to successfully run to completion.    Cycled tests have been limited to Orion because the JEDI database needed to run the test period only exists on Orion.   A clone and build test was been completed on Hera.  It is not possible to build JEDI apps on WCOSS2 since required libraries have not been installed.
  
**Checklist**


RussTreadon-NOAA:
The following PRs in other NOAA-EMC repositories are associated with this g-w PR

- [NOAA EMC/GSI PR #415](https://github.com/NOAA-EMC/GSI/pull/415)
  - This GSI PR removes global_cycle from DA jobs and scripts.   
  - This g-w PR includes the addition of global_cycle jobs and scripts.
- [NOAA EMC/GDASApp PR #66](https://github.com/NOAA-EMC/GDASApp/pull/66)
  - This GDASApp PR adds JEDI cycling components to GDASApp. 
  - This g-w PR adds g-w hooks to exercise JEDI atmospheric DA components in cycling mode.

jobs/rocoto/atmanalprep.sh
```diff
@@ -0,0 +1,16 @@
...
+###############################################################
+# Source FV3GFS workflow modules
+. $HOMEgfs/ush/load_fv3gfs_modules.sh
+status=$?
+[[ $status -ne 0 ]] && exit $status
+
+###############################################################
+# set JJOB variables
+export configs="base atmanal atmanalprep"
+export EXSCRIPT=${GDASPREPPY:-$HOMEgfs/sorc/gdas.cd/scripts/exgdas_global_atmos_analysis_prep.py}
```
aerorahul: These variables should not be in the rocoto job script. 
Thinking ahead for the `ecflow` equivalent, would they be in the `.ecf`?
`base` adn `atmanal` are common to the 3 `atmanalXXXX.sh` jobs.  The last one `prep` or `post` or `run` need to be passed through.
I can think on this a bit more, but wanted to put this down as a comment.

parm/config/config.atmanal
```diff
@@ -0,0 +1,24 @@
+#!/bin/ksh -x
```
aerorahul: might as well use `bash` here and all `config` files.

parm/config/config.postanl
```diff
@@ -0,0 +1,46 @@
...
+if [ $machine = "WCOSS_DELL_P3" ]; then
+    export npe_dwn=28
+else
+    export npe_dwn=24
+fi
+
+export GRIBVERSION='grib2'
+export SENDCOM="YES"
+
+echo "END: config.post"
```
aerorahul:
```suggestion
echo "END: config.postanl"
```

scripts/exglobal_atmos_sfcanl.sh
```diff
@@ -0,0 +1,300 @@
...
+##############################################################
+# Output files
+if [ $DONST = "YES" ]; then
+   $NLN $DTFANL dtfanl
+fi
+
+
+##############################################################
+# Update surface fields in the FV3 restart's using global_cycle
+if [ $DOGCYCLE = "YES" ]; then
```
aerorahul: Is there a need for this `if` block?
If `DOGCYCLE` = `NO`, we should probably not call this job at all, right?
May be I am missing something.

sorc/build_gdas.sh
```diff
@@ -0,0 +1,33 @@
...
+
+source ./machine-setup.sh > /dev/null 2>&1
+cwd=`pwd`
+
+USE_PREINST_LIBS=${USE_PREINST_LIBS:-"true"}
+if [ $USE_PREINST_LIBS = true ]; then
+  export MOD_PATH=/scratch3/NCEPDEV/nwprod/lib/modulefiles
+else
+  export MOD_PATH=${cwd}/lib/modulefiles
+fi
```
aerorahul: I think we removed this `MOD_PATH` business.  We can check w/ @WalterKolczynski-NOAA when he is back.

ush/rocoto/setup_workflow.py
```diff
@@ -42,7 +42,7 @@ def main():
         print(f'input arg:     --expdir = {repr(args.expdir)}')
         sys.exit(1)
 
-    gfs_steps = ['prep', 'anal', 'analdiag', 'analcalc', 'gldas', 'fcst', 'postsnd', 'post', 'vrfy', 'arch']
+    gfs_steps = ['prep', 'analcalc', 'gldas', 'fcst', 'postsnd', 'postanl', 'post', 'vrfy', 'arch']
```
aerorahul: I will have to understand the changes in this script with the changes in the PR that consolidates the XML generation.  I can give feedback on this part separately.

jobs/rocoto/atmanalprep.sh
aerorahul: The `EXSCRIPT` variable should be defined in `config.atmanlprep`.
I am still thinking more on the `configs` and what to do here.

jobs/rocoto/atmanalrun.sh
```diff
@@ -0,0 +1,16 @@
...
+###############################################################
+# Source FV3GFS workflow modules
+. $HOMEgfs/ush/load_fv3gfs_modules.sh
+status=$?
+[[ $status -ne 0 ]] && exit $status
+
+###############################################################
+# set JJOB variables
+export configs="base atmanal atmanalrun"
+export EXSCRIPT=${GDASRUNSH:-$HOMEgfs/sorc/gdas.cd/scripts/exgdas_global_atmos_analysis_run.sh}
```
aerorahul: Should move `EXSCRIPT` to `config.atmanalrun`.

This script (`atmanalrun.sh`) is just a wrapper for rocoto similar to the `.ecf` files for ecflow.

jobs/rocoto/atmensanalpost.sh
```diff
@@ -0,0 +1,16 @@
...
+###############################################################
+# Source FV3GFS workflow modules
+. $HOMEgfs/ush/load_fv3gfs_modules.sh
+status=$?
+[[ $status -ne 0 ]] && exit $status
+
+###############################################################
+# set JJOB variables
+export configs="base atmensanal atmensanalpost"
+export EXSCRIPT=${GDASPOSTPY:-$HOMEgfs/sorc/gdas.cd/scripts/exgdas_global_atmos_ensanal_post.py}
```
aerorahul: same comment here for `EXSCRIPT`.

jobs/rocoto/atmensanalprep.sh
```diff
@@ -0,0 +1,16 @@
...
+###############################################################
+# Source FV3GFS workflow modules
+. $HOMEgfs/ush/load_fv3gfs_modules.sh
+status=$?
+[[ $status -ne 0 ]] && exit $status
+
+###############################################################
+# set JJOB variables
+export configs="base atmensanal atmensanalprep"
+export EXSCRIPT=${GDASPREPPY:-$HOMEgfs/sorc/gdas.cd/scripts/exgdas_global_atmos_analysis_prep.py}
```
aerorahul: here as well.

jobs/rocoto/atmensanalrun.sh
```diff
@@ -0,0 +1,16 @@
...
+###############################################################
+# Source FV3GFS workflow modules
+. $HOMEgfs/ush/load_fv3gfs_modules.sh
+status=$?
+[[ $status -ne 0 ]] && exit $status
+
+###############################################################
+# set JJOB variables
+export configs="base atmensanal atmensanalrun"
+export EXSCRIPT=${GDASRUNSH:-$HOMEgfs/sorc/gdas.cd/scripts/exgdas_global_atmos_ensanal_run.sh}
```
aerorahul: and here.

jobs/rocoto/postanl.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash -x
```
aerorahul: Curious on how this differs from `post.sh` other than the `exit 99` for `FHRGRP` that is for the forecast hours?

aerorahul: https://github.com/NOAA-EMC/global-workflow/blob/develop/jobs/rocoto/post.sh

.gitignore
```diff
@@ -76,6 +76,7 @@ sorc/gfs_post.fd
 sorc/gfs_wafs.fd
 sorc/gldas.fd
 sorc/gsi.fd
+sorc/gdas.cd
```
guillaumevernieres, line 85: typo?

CoryMartin-NOAA: Per NCO implementation standards `.fd` indicates a directory with mainly fortran source, `.cd` with mainly C source. I made the executive decision that JEDI is 'more C than F'. :-)

sorc/checkout.sh
```diff
@@ -131,7 +131,8 @@ mkdir -p ${logdir}
 # The checkout version should always be a speciifc commit (hash or tag), not a branch
 errs=0
 checkout "ufs_model.fd"    "https://github.com/ufs-community/ufs-weather-model" "${ufs_model_hash:-Prototype-P8c}"; errs=$((errs + $?))
-checkout "gsi.fd"          "https://github.com/NOAA-EMC/GSI.git"                "a62dec6"                         ; errs=$((errs + $?))
+checkout "gsi.fd"          "https://github.com/NOAA-EMC/GSI.git"                "develop"                         ; errs=$((errs + $?))
```
WalterKolczynski-NOAA: As noted in the comment above, checkout versions should always be a *specific* version (hash or tag), not a branch.

jobs/JGLOBAL_ATMOS_SFCANL
```diff
@@ -0,0 +1,179 @@
...
+
+
+export ATMGES="$COMIN_GES/${GPREFIX}atmf006${GSUFFIX}"
+if [ ! -f $ATMGES ]; then
+    echo "FATAL ERROR: FILE MISSING: ATMGES = $ATMGES"
+    exit 1
+fi
+
+
+# Get LEVS
```
WalterKolczynski-NOAA: Should this be done at the script level instead? I know NCO doesn't like calculations in the j-job.

jobs/JGLOBAL_ATMOS_SFCANL
```diff
@@ -0,0 +1,179 @@
...
+    export LEVS=$($NCLEN $ATMGES pfull) 
+    status=$?
+else
+    export LEVS=$($NEMSIOGET $ATMGES dimz | awk '{print $2}')
+    status=$?
+fi
+[[ $status -ne 0 ]] && exit $status
+
+
+##if [ $DOHYBVAR = "YES" ]; then
```
WalterKolczynski-NOAA: Please remove inactive code

jobs/JGLOBAL_ATMOS_SFCANL
```diff
@@ -0,0 +1,179 @@
...
+##    export ATMGES_ENSMEAN="$COMIN_GES_ENS/${GPREFIX}atmf006.ensmean$GSUFFIX"
+##    if [ ! -f $ATMGES_ENSMEAN ]; then
+##        echo "FATAL ERROR: FILE MISSING: ATMGES_ENSMEAN = $ATMGES_ENSMEAN"
+##        exit 2
+##    fi
+##fi
+
+
+# Link observational data
```
WalterKolczynski-NOAA: As above, would this be better at the script level?

jobs/JGLOBAL_ATMOS_SFCANL
```diff
@@ -0,0 +1,179 @@
...
+
+
+# Update surface fields with global_cycle
+export DOGCYCLE=${DOGCYCLE:-"YES"}
+
+
+###############################################################
+# Run relevant script
+env
+msg="HAS BEGUN on `hostname`"
```
WalterKolczynski-NOAA: We've removed all of the backticks from g-w. Please use `$(hostname)`.

parm/config/config.atmensanal
```diff
@@ -0,0 +1,27 @@
...
+
+echo "BEGIN: config.atmensanal"
+
+export OBS_YAML_DIR=$HOMEgfs/sorc/gdas.cd/parm/atm/obs/config/
+export OBS_LIST=$HOMEgfs/sorc/gdas.cd/parm/atm/obs/lists/lgetkf_prototype.yaml
+export BERROR_YAML=$HOMEgfs/sorc/gdas.cd/parm/atm/berror/hybvar_bump.yaml
+export ATMENSYAML=$HOMEgfs/sorc/gdas.cd/parm/atm/lgetkf/lgetkf.yaml
+##export FV3JEDI_FIX=/work2/noaa/da/cmartin/GDASApp/fix
+##export FV3JEDI_FIX=/work2/noaa/da/fhan/GDASApp/fix
+export FV3JEDI_FIX=/work2/noaa/da/rtreadon/GDASApp/fix
```
WalterKolczynski-NOAA: To match the way other fix directories are handled, this needs to be relative to the g-w clone and `link_workflow.sh` should be updated to link the directory in. Ideally, the fix directory is added into the master g-w fix location so it can be properly sync'd on all HPC, but we might be able to use the personal locations on an interim basis. Thoughts @KateFriedman-NOAA ?

parm/config/config.atmensanal
```diff
@@ -0,0 +1,27 @@
...
+export R2D2_OBS_DB='ufsda_test'
+export R2D2_OBS_DUMP='oper_gdas'
+export R2D2_OBS_SRC='ncdiag'
+export R2D2_BC_SRC='gsi'
+##export R2D2_BC_DUMP='oper_gdas'
```
WalterKolczynski-NOAA, line 17: Please remove inactive code.

parm/config/config.atmensanalrun
```diff
@@ -0,0 +1,16 @@
...
+# Atm LETKFs specific
+
+echo "BEGIN: config.atmensanalrun"
+
+# Get task specific resources
+. $EXPDIR/config.resources atmensanalrun
+
+# Task specific variables
+export JEDIENSEXE=$HOMEgfs/exec/fv3jedi_letkf.x
+##export JEDIENSEXE="/work/noaa/da/Russ.Treadon/git/global_workflow/ufsda_gdasapp/sorc/gdas.cd_pr1730/build/bin/fv3jedi_letkf.x"
```
WalterKolczynski-NOAA: Please remove inactive code.

parm/config/config.postanl
```diff
@@ -0,0 +1,46 @@
...
+export POSTJJOBSH="$HOMEpost/jobs/JGLOBAL_NCEPPOST"
+export GFSDOWNSH="$HOMEpost/ush/fv3gfs_downstream_nems.sh"
+export GFSDWNSH="$HOMEpost/ush/fv3gfs_dwn_nems.sh"
+
+export POSTGPSH="$HOMEpost/ush/gfs_nceppost.sh"
+export POSTGPEXEC="$HOMEpost/exec/gfs_ncep_post"
+export GOESF=NO                              # goes image
+export FLXF=YES                               # grib2 flux file written by post
+
+export npe_postgp=$npe_post
```
WalterKolczynski-NOAA: All these resource settings should be set in `config.resources`.

parm/config/config.resources
```diff
@@ -136,6 +138,31 @@ elif [ $step = "waveawipsgridded" ]; then
...
+    export layout_x=1
+    export layout_y=1
+
+    export wtime_atmanalprep="00:10:00"
+    export npe_atmanalprep=1
+    export nth_atmanalprep=1
+    export npe_node_atmanalprep=$(echo "$npe_node_max / $nth_atmanalprep" | bc)
+    export memory_atmanalprep="3072M"
+
+    export wtime_atmanalrun="00:30:00"
```
WalterKolczynski-NOAA, line 155: Is there a need for all three jobs to get the resources of the other two, or could these be separated into individual if-blocks?

parm/config/config.resources
```diff
@@ -418,6 +463,30 @@ elif [ $step = "coupled_ic" ]; then
...
+    # make below case dependent later
+    export layout_x=2
+    export layout_y=3
+
+    export wtime_atmensanalprep="00:10:00"
+    export npe_atmensanalprep=1
+    export nth_atmensanalprep=1
+    export npe_node_atmensanalprep=$(echo "$npe_node_max / $nth_atmensanalprep" | bc)
+
+    export wtime_atmensanalrun="00:30:00"
```
WalterKolczynski-NOAA, line 473: As above, do these need to be together or can it be separated into three individual blocks?

parm/config/config.atmanal
WalterKolczynski-NOAA: Not just config, but all new scripts. I'm slowly trying to convert all shell scripts to bash, so it'd be better if more non-bash weren't added.

scripts/exglobal_atmos_sfcanl.sh
```diff
@@ -0,0 +1,300 @@
...
+export gesenvir=${gesenvir:-$envir}
+
+# Observations
+OPREFIX=${OPREFIX:-""}
+OSUFFIX=${OSUFFIX:-""}
+
+# Guess files
+GPREFIX=${GPREFIX:-""}
+GSUFFIX=${GSUFFIX:-$SUFFIX}
+ATMG03=${ATMG03:-${COMIN_GES}/${GPREFIX}atmf003${GSUFFIX}}
```
WalterKolczynski-NOAA: Can we condense this with a loop instead of hand-coding multiple lines doing the same thing?

scripts/exglobal_atmos_sfcanl.sh
```diff
@@ -0,0 +1,300 @@
...
+fi
+
+cd $DATA || exit 99
+
+[[ $DONST = "YES" ]] && $NLN $NSSTBF nsstbufr
+
+
+##############################################################
+# Required model guess files
+$NLN $SFCG03 sfcf03
```
WalterKolczynski-NOAA: As above, can we use loops to create these links instead of a bunch of hand-coded lines? I'd also love for these two sections to be combined instead of separated by 80 other lines of code so it is easier to follow what is going on.

sorc/build_all.sh
```diff
@@ -132,6 +132,20 @@ $Build_gsi && {
 	((err+=$rc))
 }
 
+#------------------------------------
+# build gdas
+#------------------------------------
+$Build_gdas && {
```
WalterKolczynski-NOAA, line 153: Please use proper indentation for this block

sorc/build_gdas.sh
WalterKolczynski-NOAA: Fairly sure we did.

sorc/build_gdas.sh
```diff
@@ -0,0 +1,33 @@
+#! /usr/bin/env bash
+set -eux
+
+source ./machine-setup.sh > /dev/null 2>&1
+cwd=`pwd`
```
WalterKolczynski-NOAA: Please replace backticks with `$(pwd)`

ush/rocoto/setup_workflow.py
```diff
@@ -241,20 +249,30 @@ def get_gdasgfs_resources(dict_configs, cdump='gdas'):
     do_gldas = base.get('DO_GLDAS', 'NO').upper()
     do_wave = base.get('DO_WAVE', 'NO').upper()
     do_wave_cdump = base.get('WAVE_CDUMP', 'BOTH').upper()
+    do_jedivar = base.get('DO_JEDIVAR', 'NO').upper()
     reservation = base.get('RESERVATION', 'NONE').upper()
 
-    #tasks = ['prep', 'anal', 'fcst', 'post', 'vrfy', 'arch']
-    tasks = ['prep', 'anal', 'analcalc']
+    if do_jedivar in ['Y', 'YES']:
```
WalterKolczynski-NOAA: Worried about how these extensive changes are going to conflict with @aerorahul's refactor of the setup_workflow scripts.

WalterKolczynski-NOAA (changes requested):

parm/config/config.atmensanal
KateFriedman-NOAA: New fix files should go into the main fix sets. @RussTreadon-NOAA please provide the best pickup location for the new JEDI fix files and let me know the preferred subfolder name for them (`fix_jedi`?). I can pull them into the main set and then rsync to the other platform copies. Go-live on WCOSS2 is tomorrow so this may take me a day or so to get to. Thanks!

CoryMartin-NOAA: Since this is still very much a 'fluid' system (and we don't yet have a good solution for background error fix files), I suggest we stick with the interim basis personal location of FV3JEDI_FIX for now, but will defer to you and @aerorahul on this matter, it will just lead to more work for EIB folks over the next year or two :-)

WalterKolczynski-NOAA: Even if we go that route, I'd like all that to be handled in `link_workflow` and FV3JEDI_FIX to just be a path in the g-w tree like the others, so we keep the machine-specific paths contained.

CoryMartin-NOAA: yeah that makes sense

RussTreadon-NOAA:
I'll start working through comments tomorrow.   Thanks for the feedback.  This is the start of the long g-w JEDI transition.  g-w and DA will need to maintain two systems, GSI-DA and JEDI-DA, for an extended period until JEDI-DA replaces all GSI-DA in the GFS.

RussTreadon-NOAA:
**Repository question**
The new job `JGLOBAL_ATMOS_SFCANL` and new script `exglobal_atmos_sfcanl.sh` execute `global_cycle` and `global_cycle.sh`.  Source code for this executable and the script reside in ufs_utils.   

Moving ahead do we want to maintain the surface analysis j-job and script in the [g-w](https://github.com/NOAA-EMC/global-workflow.git) repo or the [ufs_utils](https://github.com/ufs-community/UFS_UTILS.git) repo?

For the time being I will make suggested changes to the sfcanl j-job and ex-script in `feature/ufsda-gdasapp` but please discuss & decide upon the long term home for this j-job and ex-script.   The outcome of this discussion has bearing on other apps executed by the workflow.

Adding @GeorgeGayno-NOAA for his awareness.

sorc/checkout.sh
RussTreadon-NOAA: The current head of NOAA-EMC/GSI develop is [c10644e](https://github.com/NOAA-EMC/GSI/commit/c10644ee443cb7a99c4300b9037d705a49501023).   My working copy of `sorc/checkout.sh` now specifies this hash instead of develop.   I'll update this hash again if NOAA-EMC/GSI develop is updated before this PR is merged into g-w develop.

jobs/JGLOBAL_ATMOS_SFCANL
RussTreadon-NOAA: Good point.   The `Get LEVS` section has been removed from `JGLOBAL_ATMOS_SFCANL`.   Script `exglobal_atmos_sfcanl.sh` already contains lines to determine `LEVS`

jobs/JGLOBAL_ATMOS_SFCANL
RussTreadon-NOAA: The comment is incorrect.  No files are being linked.  Furthermore, global_cycle does not use `PREPQC`, `PREPQCPF1, or `TCVITL`.   The `DONST` line has been moved to `exglobal_atmos_sfcanl.sh`.  This section of scripting has been removed from `JGLOBAL_ATMOS_SFCANL`.

jobs/JGLOBAL_ATMOS_SFCANL
RussTreadon-NOAA: Inactive lines removed.

jobs/JGLOBAL_ATMOS_SFCANL
RussTreadon-NOAA: Use `$(hostname)` to replace backticks

parm/config/config.atmensanal
RussTreadon-NOAA: Since this will be an iterative process, I can assume (for the time being) the responsibility for maintaining a `fix_jedi` directory in the main g-w fix on Orion, Hera, and WCOSS2.   Yes, we cannot yet run JEDI on WCOSS2 but I hope we can do so before we press to far into fall 2022.   I'll add `fix_jedi` to `sorc/link_workflow.sh`.   JEDI config files will then be updated to point at `fix_jedi` in the g-w clone.  Does this work for everyone?

WalterKolczynski-NOAA: @RussTreadon-NOAA Yes, Kate and I agree this is the best path for now. When fix_jedi becomes more stable, we can add it to the master fix location.

RussTreadon-NOAA: @WalterKolczynski-NOAA , my preference is to add `fix_jedi` to the Hera and Orion `FIX_DIR` specified in `link_workflow.sh.   
```
elif [ $machine = "hera" ]; then
    FIX_DIR="/scratch1/NCEPDEV/global/glopara/fix_NEW"
elif [ $machine = "orion" ]; then
    FIX_DIR="/work/noaa/global/glopara/fix_NEW"
```

parm/config/config.atmensanal
RussTreadon-NOAA: Inactive code removed.

parm/config/config.atmensanalrun
RussTreadon-NOAA: Inactive code removed.

parm/config/config.postanl
RussTreadon-NOAA: Resource settings moved to `config.resources`.

parm/config/config.resources
RussTreadon-NOAA: This section has been refactored into three if-blocks.

parm/config/config.resources
RussTreadon-NOAA: This section has be refactored into three if-blocks.

scripts/exglobal_atmos_sfcanl.sh
RussTreadon-NOAA: Examination of job log file shows that `ATM*` and `SFC*` variables are not used.   These variables plus other unused variables have been removed.

scripts/exglobal_atmos_sfcanl.sh
RussTreadon-NOAA: As noted above, `SFC*` variables not used and, therefore, removed.

sorc/build_all.sh
RussTreadon-NOAA: Indent to be consistent with other sections.

sorc/build_gdas.sh
RussTreadon-NOAA: `USE_PREINST_LIBS` section removed.  backticks around `pwd` replaced with `$(pwd)`

ush/rocoto/setup_workflow.py
RussTreadon-NOAA: Yes, these changes likely will conflict with @aerorahul 's refactor of the setup_workflow scripts.  His refactoring of the setup_workflow scripts has priority.   I'll work with him to add JEDI-DA pieces into the refactored scripts.

parm/config/config.atmanal
RussTreadon-NOAA: New config files have been converted to `bash`

parm/config/config.postanl
RussTreadon-NOAA: `config.post` changed to `config.postanl`

scripts/exglobal_atmos_sfcanl.sh
RussTreadon-NOAA: You are correct.   The `DOGCYCLE` `if` block has been removed and indentation adjusted.

sorc/build_gdas.sh
RussTreadon-NOAA: `MOD_PATH` removed

ush/rocoto/setup_workflow.py
RussTreadon-NOAA: Agreed.  Your refactoring of the XML generation take priority.  I can work with you add JEDI pieces into the refactored XML setup scripts.

jobs/rocoto/atmanalprep.sh
RussTreadon-NOAA: You are correct.  This approach does not work with `ecflow`.    

Currently the GDASApp has a single j-job, `JGDAS_GLOBAL_ATMOS_ANALYSIS`, for atmospheric JEDI-VAR DA.   This single j-job is used for prep, run, and post with toggles placed in rocoto jobs.   

One way to refactor this is to have separate GDASApp j-jobs.  That is, we replace `JGDAS_GLOBAL_ATMOS_ANALYSIS` with

- `JGDAS_GLOBAL_ATMOS_ANALYSIS_PREP`
- `JGDAS_GLOBAL_ATMOS_ANALYSIS_RUN`
- `JGDAS_GLOBAL_ATMOS_ANALYSIS_POST`

The toggles currently in `atmanalprep.sh`, `atmanalrun.sh`, and `atmanalpost.sh` would then move into their respective j-jobs.
   
This conversation also applies to JEDI-ENS DA.   The GDASApp currently has a single j-job for JEDI-ENS DA prep, run, and post.   It, too, could be refactored as suggested above.

Thoughts, comments? @aerorahul , @CoryMartin-NOAA , @WalterKolczynski-NOAA

RussTreadon-NOAA: Agreed.  See previous comment for possible resolution of this issue.

jobs/rocoto/atmanalrun.sh
RussTreadon-NOAA: Agreed.  See suggested resolution above.

jobs/rocoto/atmensanalpost.sh
RussTreadon-NOAA: Agreed.  See suggested resolution above.

jobs/rocoto/atmensanalprep.sh
RussTreadon-NOAA: Agreed.  See suggested resolution above.

jobs/rocoto/atmensanalrun.sh
RussTreadon-NOAA: Agreed.  See suggested resolution above.

jobs/rocoto/postanl.sh
RussTreadon-NOAA: You are correct.  The only difference is the lines you identify.   The UPP team should take the lead on this change.   The UPP application for the analysis differs from the forecast.   

We run executable `gfs_ncep_post -> ../sorc/upp.fd/exec/upp.x` when processing the `atmanl` and `sfcanl` files.   `upp.x` output is then used to generate analysis pgrb files.   The model runs the inline UPP.   `post.sh` does not execute `upp.x`  Instead, it processes inline UPP output to create pgrb files.  When `post.sh` includes the analysis files, the post job which processes analysis files and possibly initial forecast hours has different dependencies than subsequent post jobs which only process inline UPP output from the forecast model.   'postanl.sh` was added in recognition of these distinctions.   

Initial cycling tests with JEDI-VAR benefited from moving the analysis UPP into its own job.   Subsequent local changes to JEDI-VAR _may_ lessen the need for this.   This would need to be tested to confirm.   

That said, the above distinctions between analysis and forecast UPP applications remain.  If the UPP team were to adopt this approach, you are right in that `postanl.sh` could and should be simplified.

WalterKolczynski-NOAA: If the scripts are identical, we should just use the same script. I definitely don't think it is worth it to maintain all this duplicate code for a single variable bounds check. If we really think it is necessary (I don't), we can put it in the config file.

parm/config/config.atmensanal
WalterKolczynski-NOAA: @RussTreadon-NOAA Okay, coordinate that with @KateFriedman-NOAA then

scripts/exglobal_atmos_sfcanl.sh
WalterKolczynski-NOAA: Even better!

jobs/JGLOBAL_ATMOS_SFCANL
```diff
@@ -62,7 +62,7 @@ export COMPONENT=${COMPONENT:-atmos}
 if [ $RUN_ENVIR = "nco" ]; then
     export ROTDIR=${COMROOT:?}/$NET/$envir
 fi
-export DO_CALC_ANALYSIS=${DO_CALC_ANALYSIS:-"YES"} 
+##export DO_CALC_ANALYSIS=${DO_CALC_ANALYSIS:-"YES"}
```
WalterKolczynski-NOAA: Please remove this now-inactive code

parm/config/config.resources
```diff
@@ -330,8 +334,12 @@ elif [ $step = "postanl" ]; then
     export npe_postanl=112
     export nth_postanl=1
     export npe_node_postanl=40
+    export npe_dwn=24
     export npe_node_dwn=$npe_node_max
-    if [[ "$machine" = "WCOSS_DELL_P3" ]]; then export npe_node_postanl=28 ; fi
+    if [[ "$machine" = "WCOSS_DELL_P3" ]]; then
+	export npe_dwn=28
```
WalterKolczynski-NOAA: This if-block should be indented

WalterKolczynski-NOAA (changes requested):
Just a couple new minor ones have popped up, plus the outstanding questions about postanal.sh.

jobs/rocoto/postanl.sh
RussTreadon-NOAA: 'config.postanl' and `postanl.sh` have been removed from a working copy of `feature/ufsda-gdasapp` on Hera.   References to `postanl` have been removed from `setup_workflow.py` and `config.resources`.  A four cycle GSI-based DA parallel is now running on Hera to confirm no change in the results.    A four cycle JEDI-based DA parallel will be run on Orion once the machine is returned to service.   Assuming both tests pass, `postanl` will be removed from the branch.

jobs/JGLOBAL_ATMOS_SFCANL
RussTreadon-NOAA: Inactive code removed.

parm/config/config.resources
RussTreadon-NOAA: It seems these lines are indented in the repo snapshot of [`parm/config/config.resources`](https://github.com/NOAA-EMC/global-workflow/blob/feature/ufsda-gdasapp/parm/config/config.resources#L339)
```
    export npe_node_dwn=$npe_node_max
	export npe_dwn=28
	export npe_node_postanl=28
    fi
    if [[ "$machine" == "WCOSS_C" ]]; then export memory_postanl="3072M"; fi
```

WalterKolczynski-NOAA: What you pasted isn't indented either. Here's what I see:

image

The two exports are at the same level as the if ... DELL

RussTreadon-NOAA: Interesting.  It's indented on my screen.   The scripting in question is in a `postanl` section.  Reviews have requested removal of `postanl`.  I'm in the process of removing `postatnl` and testing that removal does not break anything.    Assuming successful tests the script fragment in question will be removed and not in the repo.

WalterKolczynski-NOAA: If we can get by without the job too, great, but I was mostly concerned about the duplicate script.

As for this, maybe a space/tab conflict causing different indentation to be shown?

jobs/rocoto/atmanalprep.sh
RussTreadon-NOAA: [6674720](https://github.com/NOAA-EMC/global-workflow/commit/667472055b8db74e05467f0c7b02eb6a67bce9d1) split the multifunction `JGDAS_GLOBAL_ATMOS_ANALYSIS` and `JGDAS_GLOBAL_ATMOS_ENSANAL` j-jobs into function specific j-jobs as outlined above.

RussTreadon-NOAA: [6674720](https://github.com/NOAA-EMC/global-workflow/commit/667472055b8db74e05467f0c7b02eb6a67bce9d1) removed the `configs` and `EXSCRIPT` lines from `atmanalprep.sh`.  These lines are now in the function specific j-job, `JGDAS_GLOBAL_ATMOS_ANALYSIS_PREP`.

jobs/rocoto/atmanalrun.sh
RussTreadon-NOAA: Addressed via [6674720](https://github.com/NOAA-EMC/global-workflow/commit/667472055b8db74e05467f0c7b02eb6a67bce9d1)

jobs/rocoto/atmensanalpost.sh
RussTreadon-NOAA: Addressed via [6674720](https://github.com/NOAA-EMC/global-workflow/commit/667472055b8db74e05467f0c7b02eb6a67bce9d1)

jobs/rocoto/atmensanalprep.sh
RussTreadon-NOAA: Addressed via [6674720](https://github.com/NOAA-EMC/global-workflow/commit/667472055b8db74e05467f0c7b02eb6a67bce9d1)

jobs/rocoto/atmensanalrun.sh
RussTreadon-NOAA: Addressed via [6674720](https://github.com/NOAA-EMC/global-workflow/commit/667472055b8db74e05467f0c7b02eb6a67bce9d1)

jobs/rocoto/postanl.sh
RussTreadon-NOAA: [6674720](https://github.com/NOAA-EMC/global-workflow/commit/667472055b8db74e05467f0c7b02eb6a67bce9d1) removes all references to postanl.   This includes `jobs/rocoto/postanl.sh`, `parm/config/config.postanl`, `postanl` sections in `config.resources`, and `postanl` sections in `ush/rocoto/setup_workflow.py`

RussTreadon-NOAA:
xml generation has changed in `develop`.   As such, JEDI-VAR and JEDI-ENS xml generation must be added to the new python scripts in `develop/workflow`.  This work is underway in `feature/ufsda-gdasapp`.  This PR will be updated once this development is complete.   Then it's time to re-review this PR and hopefully merge it into `develop`.

RussTreadon-NOAA:
The commit to `feature/ufsda-gdasapp` at fbbc582c adds JEDI_VAR and JED_ENS scripting to `workflow/applications.py` and `workflow/rocoto/workflow_tasks.py`.   When `workflow/setup_xml.py` is executed from `feature/ufsda-gdasapp` without JEDI_VAR and JEDI_ENS, it reproduces the xml generated by the `develop` branch.

I will run two C96/C48L127 parallels on Orion:
- use GSI based DA to ensure `feature/ufsda-gdasapp` reproduces `develop`
- use JEDI based DA to ensure `feature/ufsda-gdasapp` reproduces previous JEDI DA results

workflow/applications.py
```diff
@@ -170,15 +172,27 @@ def _cycled_configs(self):
         Returns the config_files that are involved in the cycled app
         """
 
-        configs = ['prep',
-                   'anal', 'sfcanl', 'analdiag', 'analcalc',
-                   'fcst', 'post', 'vrfy', 'arch']
+        configs = ['prep']
+##                   'sfcanl', 'analcalc',
+##                   'fcst', 'post', 'vrfy', 'arch']
```
WalterKolczynski-NOAA: Please remove the inactive lines

workflow/applications.py
```diff
@@ -331,12 +352,19 @@ def _get_cycled_task_names(self):
...
+                hybrid_gdas_or_gfs_tasks += ['ediag'] if self.lobsdiag_forenkf else ['eomg']
             hybrid_gdas_tasks += ['ecen', 'esfc', 'efcs', 'epos', 'earc']
 
         # Collect all "gdas" cycle tasks
-        gdas_tasks = gdas_gfs_common_tasks_before_fcst + ['analdiag']
+        gdas_tasks = []
+        if self.do_jedivar:
+            gdas_tasks += gdas_gfs_common_tasks_before_fcst
+        else:
+            gdas_tasks += gdas_gfs_common_tasks_before_fcst + ['analdiag']
```
WalterKolczynski-NOAA: Could this more simply be:
```
gdas_tasks = gdas_gfs_common_tasks_before_fcst
if not self.do_jedivar:
    gdas_tasks += ['analdiag']
```

workflow/rocoto/workflow_tasks.py
```diff
@@ -376,6 +384,65 @@ def analdiag(self):
...
+    def atmanalprep(self):
+
+        suffix = self._base["SUFFIX"]
+        dump_suffix = self._base["DUMP_SUFFIX"]
+        gfs_cyc = self._base["gfs_cyc"]
+        dmpdir = self._base["DMPDIR"]
+        gfs_enkf = True if self.app_config.do_hybvar and 'gfs' in self.app_config.eupd_cdumps else False
+
+        deps = []
+        dep_dict = {'type': 'metatask', 'name': f'{"gdas"}post', 'offset': '-06:00:00'}
```
WalterKolczynski-NOAA: This should be just `'name': 'gdaspost'`, as there's no need to use an f-string if you are just going to put a literal in the { }.

workflow/rocoto/workflow_tasks.py
```diff
@@ -376,6 +384,65 @@ def analdiag(self):
 
         return task
 
+    def atmanalprep(self):
+
+        suffix = self._base["SUFFIX"]
+        dump_suffix = self._base["DUMP_SUFFIX"]
+        gfs_cyc = self._base["gfs_cyc"]
+        dmpdir = self._base["DMPDIR"]
+        gfs_enkf = True if self.app_config.do_hybvar and 'gfs' in self.app_config.eupd_cdumps else False
```
WalterKolczynski-NOAA: Change this variable to `do_gfs_enkf` so it is clear if is a bool.

workflow/rocoto/workflow_tasks.py
```diff
@@ -376,6 +384,65 @@ def analdiag(self):
...
+                              cycledef=cycledef)
+        return task
+
+    def atmanalrun(self):
+
+        deps = []
+        dep_dict = {'type': 'task', 'name': f'{self.cdump}atmanalprep'}
+        deps.append(rocoto.add_dependency(dep_dict))
+        if self.app_config.do_hybvar:
+            dep_dict = {'type': 'metatask', 'name': f'{"gdas"}epmn', 'offset': '-06:00:00'}
```
WalterKolczynski-NOAA: Same as above.

workflow/rocoto/workflow_tasks.py
```diff
@@ -880,6 +947,61 @@ def eupd(self):
...
+    def atmensanalprep(self):
+
+        suffix = self._base["SUFFIX"]
+        dump_suffix = self._base["DUMP_SUFFIX"]
+        gfs_cyc = self._base["gfs_cyc"]
+        dmpdir = self._base["DMPDIR"]
+        gfs_enkf = True if self.app_config.do_hybvar and 'gfs' in self.app_config.eupd_cdumps else False
+
+        deps = []
+        dep_dict = {'type': 'metatask', 'name': f'{"gdas"}post', 'offset': '-06:00:00'}
```
WalterKolczynski-NOAA: Same as above

workflow/rocoto/workflow_tasks.py
```diff
@@ -880,6 +947,61 @@ def eupd(self):
...
+                              cycledef=cycledef)
+
+        return task
+
+    def atmensanalrun(self):
+
+        deps = []
+        dep_dict = {'type': 'task', 'name': f'{self.cdump}atmensanalprep'}
+        deps.append(rocoto.add_dependency(dep_dict))
+        dep_dict = {'type': 'metatask', 'name': f'{"gdas"}epmn', 'offset': '-06:00:00'}
```
WalterKolczynski-NOAA: Same as above.

WalterKolczynski-NOAA (changes requested):
Mostly minor

parm/config/config.resources
RussTreadon-NOAA: Indentation issue resolved.

workflow/applications.py
RussTreadon-NOAA: Sorry about that.   The inactive lines have been removed.  Done at [594f68d](https://github.com/NOAA-EMC/global-workflow/commit/594f68db15d744f195b33a11170cce4fb4173ef6)

workflow/rocoto/workflow_tasks.py
RussTreadon-NOAA: Thanks.  Recommended change made at [594f68d](https://github.com/NOAA-EMC/global-workflow/commit/594f68db15d744f195b33a11170cce4fb4173ef6)

workflow/rocoto/workflow_tasks.py
RussTreadon-NOAA: Recommended change made at [594f68d](https://github.com/NOAA-EMC/global-workflow/commit/594f68db15d744f195b33a11170cce4fb4173ef6)

workflow/rocoto/workflow_tasks.py
RussTreadon-NOAA: Recommended change made at [594f68d](https://github.com/NOAA-EMC/global-workflow/commit/594f68db15d744f195b33a11170cce4fb4173ef6)

workflow/rocoto/workflow_tasks.py
RussTreadon-NOAA: Recommended change made at [594f68d](https://github.com/NOAA-EMC/global-workflow/commit/594f68db15d744f195b33a11170cce4fb4173ef6)

workflow/rocoto/workflow_tasks.py
RussTreadon-NOAA: Recommended change made at [39fe881](https://github.com/NOAA-EMC/global-workflow/commit/39fe8813f544713b3205e0a669a04b2060fd8b96)

workflow/applications.py
RussTreadon-NOAA: @WalterKolczynski-NOAA , when I make the suggested change, `setup_xml.py` fails.  Here's an example using config files for GSI-based DA.

```
Orion-login-2[86] rtreadon$ ./setup_xml.py /work/noaa/da/Russ.Treadon/para_gfs/prdev3
sourcing config.prep
sourcing config.anal
sourcing config.analdiag
sourcing config.sfcanl
sourcing config.analcalc
sourcing config.fcst
sourcing config.post
sourcing config.vrfy
sourcing config.arch
sourcing config.gldas
sourcing config.eobs
sourcing config.eomg
sourcing config.ediag
sourcing config.eupd
sourcing config.ecen
sourcing config.esfc
sourcing config.efcs
sourcing config.echgres
sourcing config.epos
sourcing config.earc
sourcing config.metp
  File "./setup_xml.py", line 57, in <module>
    xml = RocotoXML(app_config)
  File "/work/noaa/da/rtreadon/git/global_workflow/update/workflow/rocoto/workflow_xml.py", line 24, in __init__
    task_list = get_wf_tasks(app_config)
  File "/work/noaa/da/rtreadon/git/global_workflow/update/workflow/rocoto/workflow_tasks.py", line 1227, in get_wf_tasks
    tasks.append(task_obj.get_task(task_name))
  File "/work/noaa/da/rtreadon/git/global_workflow/update/workflow/rocoto/workflow_tasks.py", line 129, in get_task
    return getattr(self, task_name, *args, **kwargs)()
  File "/work/noaa/da/rtreadon/git/global_workflow/update/workflow/rocoto/workflow_tasks.py", line 1107, in echgres
    self._is_this_a_gdas_task(self.cdump, 'echgres')
  File "/work/noaa/da/rtreadon/git/global_workflow/update/workflow/rocoto/workflow_tasks.py", line 64, in _is_this_a_gdas_task
    raise TypeError(f'{task_name} must be part of the "gdas" cycle and not {cdump}')
TypeError: echgres must be part of the "gdas" cycle and not gfs
Orion-login-2[87] rtreadon$
```
The line `gdas_tasks = gdas_gfs_common_tasks_before_fcst` results in `gdas_gfs_common_tasks_before_fcst` being updated when `gdas_tasks` is updated.   I confirmed this by adding `print` statements for `gdas_gfs_common_tasks_before_fcst` after each line in which `gdas_tasks` was updated.

As a result, by the time we get to 
```
        # Collect "gfs" cycle tasks
        gfs_tasks = gdas_gfs_common_tasks_before_fcst
```
`gdas_gfs_common_tasks_before_fcst` contains common gdas_gfs tasks plus gdas only tasks.

I am not famiiliar with python.  Does the assignment `gdas_tasks = gdas_gfs_common_tasks_before_fcst`
mean that these two lists reference the same memory location?  Therefore, updates to `gdas_tasks` also update `gdas_gfs_common_tasks_before_fcst`.

CoryMartin-NOAA: "two lists reference the same memory location? " @RussTreadon-NOAA yes almost always it is a 'shallow copy'.

RussTreadon-NOAA: Thanks, @CoryMartin-NOAA .   Given this, can we simplify

[workflow/applications.py](https://github.com/NOAA-EMC/global-workflow/pull/871/files/abb15aba906e6774265441a9ad854517347bb0a2#diff-d20f9a08d10f570a2399255f8699544a2232eb355efa36a64e094cb7cd790765)
```
        gdas_tasks = []
        if self.do_jedivar:
            gdas_tasks += gdas_gfs_common_tasks_before_fcst
        else:
            gdas_tasks += gdas_gfs_common_tasks_before_fcst + ['analdiag']
```
I thought
```
        # Collect all "gdas" cycle tasks
        gdas_tasks += gdas_gfs_common_tasks_before_fcst
        if not self.do_jedivar:
            gdas_tasks += gdas_gfs_common_tasks_before_fcst + ['analdiag']
```
might work, but python flags this with `UnboundLocalError: local variable 'gdas_tasks' referenced before assignment` ...which kinda makes sense.

CoryMartin-NOAA:
```
import copy
gdas_tasks = copy.deepcopy(gdas_gfs_common_tasks_before_fcst)
```
should do the trick

aerorahul: Does this not work?
```
gdas_tasks = gdas_gfs_common_tasks_before_fcst.copy()
```

CoryMartin-NOAA: Oh it probably does yeah

RussTreadon-NOAA: Thank you, @aerorahul and @CoryMartin-NOAA , for your suggestions.  I tried both suggestions.   The resulting xml file winds up with two `task_name` entries for jobs `gdasanal`, `gdassfcanl`, `gdasanalcalc`, and `gdasprep`

RussTreadon-NOAA: @aerorahul and @CoryMartin-NOAA .  I found and corrected the mistake which resulted in double entries for the cited jobs.   The following works

```
        # Collect all "gdas" cycle tasks
        gdas_tasks = gdas_gfs_common_tasks_before_fcst.copy()
        if not self.do_jedivar:
            gdas_tasks += ['analdiag']
```

RussTreadon-NOAA: Update as suggested at [866467e](https://github.com/NOAA-EMC/global-workflow/commit/866467e49b97cb7036f1366193adeac1985fe39c)

RussTreadon-NOAA:
The GSI and JEDI based DA parallels have completed two cycles on Orion.  Results are consistent with previous runs.

workflow/applications.py
WalterKolczynski-NOAA: Yay teamwork!

WalterKolczynski-NOAA (approved):

WalterKolczynski-NOAA:
@RussTreadon-NOAA if you can add a bit more to the PR description about what changed and usage, I can merge this tonight.

WalterKolczynski-NOAA:
> @RussTreadon-NOAA if you can add a bit more to the PR description about what changed and usage, I can merge this tonight.

Usage being how to invoke it, not necessarily how to actually use it in a research sense.

RussTreadon-NOAA:
@WalterKolczynski-NOAA , the description has been updated.  Is this sufficient?

WalterKolczynski-NOAA:
> @WalterKolczynski-NOAA , the description has been updated. Is this sufficient?

Yes, excellent. Thanks Russ. This will be merged shortly.

WalterKolczynski-NOAA:
@RussTreadon-NOAA oh, one last thing. Does this actually close #521?

RussTreadon-NOAA:
@WalterKolczynski-NOAA .   Yes, please close #521 with this PR.    Issue #521 focuses on atmospheric JEDI DA.   This PR adds the infrastructure for prototype atmospheric JEDI DA to the global workflow.   Once the machinery is in place we can move onto the science.

PR #874, opened 2022-06-24
for ops: Update calculation of starting time of rerun based on if wave restarts exist 

JessicaMeixner-NOAA:
**Description**

When WCOSS2 had some degraded file system performance, the wave model stalled for ~20 minutes meaning that the atmospheric model advanced significantly more than 12 hours ahead in forecast time from the wave model.  Upon restart, the wave model ICs were not there and the wave post jobs failed due to missing information.   

This PR updates the calculation for what the starting time is for restarts by also checking if wave restart files exist.  In addition, the model will fatally exit if wave ICs do not exist for RERUN=YES. 

**Type of change**

**How Has This Been Tested?**

- [x] Tested forecast script with canned data from ops to check re-run date calculation
  
**Checklist**

### DIFF
--- a/scripts/exglobal_forecast.sh
+++ b/scripts/exglobal_forecast.sh
@@ -449,10 +449,7 @@ if [ $cplwav = ".true." ]; then
         echo "WARNING: Wave ICs are absent, will start from rest"
       fi
     else
-      if [ ! -f ${RSTDIR_WAVE}/${PDYT}.${cyct}0000.restart.${wavGRD} ]; then
-        echo "ERROR: wave IC is missing for RERUN, exiting"
-        exit 1
-      else
+      if [ -f ${RSTDIR_WAVE}/${PDYT}.${cyct}0000.restart.${wavGRD} ]; then
         $NLN ${RSTDIR_WAVE}/${PDYT}.${cyct}0000.restart.${wavGRD} $DATA/restart.${wavGRD}
       fi
     fi
--- a/scripts/exglobal_forecast.sh
+++ b/scripts/exglobal_forecast.sh
@@ -443,10 +443,10 @@ if [ $cplwav = ".true." ]; then
 
   for wavGRD in $waveGRD ; do
     if [ $RERUN = "NO" ]; then
-      if [ ! -f ${WRDIR}/${sPDY}.${scyc}0000.restart.${wavGRD} ]; then 
-        echo "WARNING: NON-FATAL ERROR wave IC is missing, will start from rest"
-      else
+      if [ -f ${WRDIR}/${sPDY}.${scyc}0000.restart.${wavGRD} ]; then
         $NLN ${WRDIR}/${sPDY}.${scyc}0000.restart.${wavGRD} $DATA/restart.${wavGRD}
+      else
+        echo "WARNING: Wave ICs are absent, will start from rest"
       fi
     else
       if [ ! -f ${RSTDIR_WAVE}/${PDYT}.${cyct}0000.restart.${wavGRD} ]; then
--- a/scripts/exglobal_forecast.sh
+++ b/scripts/exglobal_forecast.sh
@@ -181,7 +181,18 @@ if [ $CDUMP = "gfs" -a $rst_invt1 -gt 0 -a $FHMAX -gt $rst_invt1 -a $filecount -
         cycs=$(echo $SDATE | cut -c9-10)
         flag1=$RSTDIR_ATM/${PDYS}.${cycs}0000.coupler.res
         flag2=$RSTDIR_ATM/coupler.res
-        if [ -s $flag1 ]; then
+        #make sure that the wave restart files also exist if cplwav=true
+        waverstok=".true."
+        if [ $cplwav = ".true." ]; then
+          for wavGRD in $waveGRD ; do
+            if [ ! -f ${RSTDIR_WAVE}/${PDYS}.${cycs}0000.restart.${wavGRD} ]; then
+              waverstok=".false."
+            fi
+          done
+        fi
+
+        if [ -s $flag1 -a $waverstok = ".true." ]; then
+	#if [ -s $flag1 ]; then
             CDATE_RST=$SDATE          
             [[ $RERUN = "YES" ]] && break
             mv $flag1 ${flag1}.old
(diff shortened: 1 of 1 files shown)


aerorahul (approved):
I have reviewed the changes in this PR and the `RERUN=YES` will produce a fatal error if appropriate Wave IC's are not found.

scripts/exglobal_forecast.sh
```diff
@@ -434,13 +445,16 @@ if [ $cplwav = ".true." ]; then
...
       if [ ! -f ${WRDIR}/${sPDY}.${scyc}0000.restart.${wavGRD} ]; then 
         echo "WARNING: NON-FATAL ERROR wave IC is missing, will start from rest"
+      else
+        $NLN ${WRDIR}/${sPDY}.${scyc}0000.restart.${wavGRD} $DATA/restart.${wavGRD}
       fi
-      $NLN ${WRDIR}/${sPDY}.${scyc}0000.restart.${wavGRD} $DATA/restart.${wavGRD}
     else
       if [ ! -f ${RSTDIR_WAVE}/${PDYT}.${cyct}0000.restart.${wavGRD} ]; then
-        echo "WARNING: NON-FATAL ERROR wave IC is missing, will start from rest"
+        echo "ERROR: wave IC is missing for RERUN, exiting"
```
WeiWei-NCO: Why it's a fatal error now?

JessicaMeixner-NOAA: @aerorahul and I discussed this.  We believe that the model should have failed in a situation that we had last week.  In a RERUN, we should have a wave IC and not start from rest.  We should never get to this "failure" now because of the first chunk of code, but think it's best to have this failure alerted in an odd chance we do.

WeiWei-NCO: If it (the new fatal) ever fails, what should we do about it? Should we apply any special procedures to recover? If so, can the script be smarter to do it for us?

If it's a rerun, there is a failure. I think it would be better to just start from rest to minimize the delay and a warning message would be enough.

JessicaMeixner-NOAA: We should restart from FHR 0 if it fails here.  Starting from rest gets you results faster, but the quality of those results would not be good for the wave model.

WeiWei-NCO: Can we restart from FHR 0 if it fails instead of exiting?

JessicaMeixner-NOAA: The logic here: https://github.com/JessicaMeixner-NOAA/global-workflow/blob/bug/waverstops/scripts/exglobal_forecast.sh#L171-L204 means that if there isn't a valid WW3 rerun restart, RERUN=NO and we would start at 0, meaning you should never get to this chunk of code.

WeiWei-NCO: Sorry I am having a hard time understanding why there would be a fatal that we don't expect at all and we would have no idea how to recover from it.

JessicaMeixner-NOAA: I guess my question to that would be, why would you not want to fail if for some reason you got to this point?  In terms of recovering, the solution would be to run again, but again I don't know how you get to this point after the above logic.   If @aerorahul agrees, we can revert back to a warning, but I'd prefer to keep the change for a fatal error.

WeiWei-NCO: Okay let me ask this....What would happen if we had such changes in production last week? Would it have proceeded with RERUN=NO or failed with RERUN=YES? Proceeded with NO I assume, and that's all we need.

Now to the details...Why do we want to check the same files twice? [It](https://github.com/JessicaMeixner-NOAA/global-workflow/blob/bug/waverstops/scripts/exglobal_forecast.sh#L452-L454) was used to check availability, but with the newly added logic, we won't get to this point. If we won't, why it's still in the script? I am fine with a fatal, but I still feel it's unnecessary in this case. 

Maybe we can set up a 15-30 minutes meeting next week to discuss this.

Thanks!

JessicaMeixner-NOAA: If this happened in production last week, the first chunk of code in this PR https://github.com/JessicaMeixner-NOAA/global-workflow/blob/bug/waverstops/scripts/exglobal_forecast.sh#L171-L204
would have ensured that we went back to a start time for the restart that we also had a wave restart for.  So RERUN=YES, but it would have been an earlier restart time.  This error statement would never be invoked with the first section of code. 

I'm happy to meet next week to discuss further.

aerorahul: We could re-work this logic between lines L452-457 and replace them as:
```
if [ -f ${RSTDIR_WAVE}/${PDYT}.${cyct}0000.restart.${wavGRD} ]; then
   $NLN ${RSTDIR_WAVE}/${PDYT}.${cyct}0000.restart.${wavGRD} $DATA/restart.${wavGRD}
fi
```
There is no condition for this file to not be present given the checks above.

scripts/exglobal_forecast.sh
```diff
@@ -434,13 +445,16 @@ if [ $cplwav = ".true." ]; then
     if [ $RERUN = "NO" ]; then
       if [ ! -f ${WRDIR}/${sPDY}.${scyc}0000.restart.${wavGRD} ]; then 
         echo "WARNING: NON-FATAL ERROR wave IC is missing, will start from rest"
+      else
+        $NLN ${WRDIR}/${sPDY}.${scyc}0000.restart.${wavGRD} $DATA/restart.${wavGRD}
       fi
```
aerorahul:
```suggestion
      if [ -f ${WRDIR}/${sPDY}.${scyc}0000.restart.${wavGRD} ]; then
        $NLN ${WRDIR}/${sPDY}.${scyc}0000.restart.${wavGRD} $DATA/restart.${wavGRD}
      else
        echo "WARNING: Wave ICs are absent, will start from rest"
      fi
```

scripts/exglobal_forecast.sh
JessicaMeixner-NOAA: That works for me.  @WeiWei-NCO if that is also okay with you I can update this branch accordingly.

WeiWei-NCO: Works for me too. Thanks, Jessica and Rahul!

JessicaMeixner-NOAA:
@WeiWei-NCO @aerorahul exglobal_forecast has been updated for the linking of the wave restart files since we will never get to the "error" part.

WeiWei-NCO (approved):
Looks good! Thank you, @JessicaMeixner-NOAA and @aerorahul!

aerorahul (approved):
looks good.

PR #881, opened 2022-06-29
Update buoy locations

aliabdolali:
**Description**
This PR updates wave_gfs.buoys* files.

Update the coordinate of NDBC buoys
Sync changes in gfs.v16.3 to develop
No dependencies are required for this change.

- [na] Bug fix (non-breaking change which fixes an issue)
- [na] New feature (non-breaking change which adds functionality)
- [na] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [na] This change requires a documentation update


WalterKolczynski-NOAA (approved):
Approved, though I have questions before I merge:

Why are the locations being updated? Were they incorrect before, or are they floaters, or maybe just moved slightly? Are these only being used to for the buoy output? My main concern is if we need to redesign to have some temporal component to these locations if they are moving over time.

aliabdolali:
> Approved, though I have questions before I merge:
> 
> Why are the locations being updated? Were they incorrect before, or are they floaters, or maybe just moved slightly? Are these only being used to for the buoy output? My main concern is if we need to redesign to have some temporal component to these locations if they are moving over time.

Thanks, Walter for approving it.
A couple of locations were defined in the wrong hemisphere, so it was wrong from the beginning. Also, some of the buoys were dislocated after maintenance or drifted slightly, so we need to check them every few years to make sure the coordinates are correct. 
My suggestion is to read the coordinates from a live server, but not sure if this is doable within the global workflow.
Note that these changes are not often, but something that is better to be accurate.

Issue #427, opened 2021-08-23
Test/fix cycled mode in feature/coupled_sprint


KateFriedman-NOAA:
Began testing feature/coupled_sprint in cycled mode on supported platforms as part of pre-commit testing for PR #418. Will commit needed updates/fixes to feature/coupled_sprint.

Testing C192C96L127 on WCOSS-Dell, Hera, Orion.

KateFriedman-NOAA:
@CoryMartin-NOAA I am testing cycled mode on Mars (C192C96L127) using the new feature/coupled_sprint branch, which is the develop branch plus changes to support the coupled system (coming in piece-by-piece from feature/coupled_crow). We haven't touched much outside of the forecast jobs yet and this is my first try at cycled mode after we updated the forecast jobs. My eobs job is not running normally, it doesn't fail but it also doesn't produce several needed output files (cnvstat.ensmean, oznstat.ensmean, radstat.ensmean) so the eupd job fails:

```
3.926 + mpirun -n 1 cfp /gpfs/dell3/stmp/Kate.Friedman/RUNDIRS/blah/2020090200/gdas/blah_gdaseupd_00.70417192/mp_untar.sh
tar: /gpfs/dell3/ptmp/Kate.Friedman/comrot/blah/enkfgdas.20200902/00/atmos/gdas.t00z.cnvstat.ensmean: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
tar: /gpfs/dell3/ptmp/Kate.Friedman/comrot/blah/enkfgdas.20200902/00/atmos/gdas.t00z.oznstat.ensmean: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
tar: /gpfs/dell3/ptmp/Kate.Friedman/comrot/blah/enkfgdas.20200902/00/atmos/gdas.t00z.radstat.ensmean: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
```
log: /gpfs/dell3/ptmp/Kate.Friedman/comrot/blah/logs/2020090200/gdaseupd.log

I notice the following in the log (`INTERPOLATION GRID NOT MONOTONIC IN SIMPIN1`):

```
 4782  create_ges_tendencies: successfully complete
 4783  create_ges_derivatives: successfully complete
 4784   INTERPOLATION GRID NOT MONOTONIC IN SIMPIN1
 4785   i,x1grid,dx=           1  -4.65113432491888      -1.630027530344513E-002
[...]

PR #889, opened 2022-07-05
build GSI utilities from GSI-utils repo

aerorahul:
**Description**
The GSI Utilities have been moved to https://github.com/NOAA-EMC/GSI-utils

This PR:
- checks out GSI utilities from the above repo
- builds the utilities from the above repo
- does not build the utilities from the GSI repo.  In a subsequent PR when the GSI removes the utilities, that option will become redundant.
- updates gsi-monitor hash and uses `build.sh` from that repository
- creates links appropriately.

**Type of change**

**How Has This Been Tested?**

- [X] Clone and Build tests on Hera
- [ ] Cycled test on Orion
- [N/A] Forecast-only test on Hera
  
**Checklist**


sorc/link_workflow.sh
```diff
@@ -278,36 +286,39 @@ done
 #--link source code directories
 #------------------------------
 
+    [[ -d ncdiag.fd ]] && rm -rf ncdiag.fd
```
RussTreadon-NOAA, line 292: **Two comments.**
1) We are still in `exec/` when this section is executed.  We should move line 298, namely, 

    `cd ${pwd}/../sorc   ||   exit 8`

    to just before line 289.   Otherwise the ndiag, global_gsi and global_enkf source directories are created in the wrong location.

2)  I think operational Implementation standards require the source code and executable name to be consistent.  For example, if the executable is `gsi.x`, the source code directory should be `gsi.fd`.   If this is true, we need to change the source code directory names in this section to be consistent with the executable name.

aerorahul: Thanks @RussTreadon-NOAA for catching the linking of the directories.
I have retained the names of the executables to be consistent with the source code.  `ncdiag.fd` will become moot soon.

RussTreadon-NOAA: It dawned on me that we already have `gsi.fd` from `checkout.sh`.   Thus, the following section of `link_workflow.sh` is not needed
```
    [[ -d gsi.fd ]] && rm -rf gsi.fd
    $SLINK gsi.fd/src/gsi     gsi.fd
```
In fact, retaining the above erroneously removes the clone of NOAA-EMC/GSI in `gsi.fd`.

RussTreadon-NOAA (commented):
Changes look good apart from removal of `gsi.fd` link from `link_workflow.sh`.   My previous comment prompted this change.  I was wrong and corrected myself in a subsequent comment.  Assuming `link_workflow.sh` is fixed, this set of changes is ready for merger into g-w develop.  

Once GSI PR [#427](https://github.com/NOAA-EMC/GSI/pull/427) is merged into GSI develop this g-w PR or a subsequent g-w PR needs to  
1) clean up the _TODO_ section of `build_gsi.sh`
2) update the hash for GSI develop in `checkout.sh` to the GSI PR [#427](https://github.com/NOAA-EMC/GSI/pull/427) hash or one thereafter.

.gitignore
```diff
@@ -77,6 +77,7 @@ sorc/gfs_post.fd
 sorc/gfs_wafs.fd
 sorc/gldas.fd
 sorc/gsi.fd
+sorc/gsi_utils.fd
```
RussTreadon-NOAA, line 85: Lines 105 and 106 reference `global_enkf.fd` and `global_gsi.fd`.   These directories are not created by the updated `link_workflow.sh`.   Instead we have `gsi.fd` and `enkf.fd`.     We can remove `global_gsi.fd` and `global_enkf.fd`.   `gsi.fd` is already in `.gitignore`.  We can add `enkf.fd`.   

For the time being `link_workflow.sh` creates `ncdiag.fd`.   For completeness we could replace `ncdiag_cat.fd` in `.gitignore` with `ncdiag.fd`

aerorahul: That raises an interesting question.
`gsi.fd` from `checkout.sh` is the "umbrella" GSI containing the source codes for `GSI`, `EnKF` and soon to be removed `ncdiag`.
Do (Should) we need to "link" `GSI/src/gsi` and `GSI/src/enkf` to `sorc`?
EE2 standards say yes.
What if we call the `gsi.fd` from `checkout.sh` as `GSIApp.fd` and then retain the symlinks?
Just a thought for now and can be worked later.
I will pin this for discussion w/ NCO.

aerorahul: @RussTreadon-NOAA 
See if you agree with [5457fe0](https://github.com/NOAA-EMC/global-workflow/pull/889/commits/5457fe052524fbe65c87f90ccdf5aea54a70b36d) for the time being.
`gsi_enkf.fd` is accurate to what contains in that directory.

RussTreadon-NOAA: Cloning NOAA-EMC/GSI into `GSIApp.fd` and updating links would work.  If we take this approach we may want to similarly clone NOAA-EMC/GDASApp into `GDASApp.fd` and update JEDI links accordingly.   

I agree.  We don't need to answer all these questions in this particular PR.   We simply need this PR to build GSI utilities from the GSI-utils repo and update links accordingly.

sorc/build_gsi.sh
```diff
@@ -1,36 +1,34 @@
...
-export REGRESSION_TESTS="NO"
-./build.sh "$cwd/gsi.fd"
+#TODO - remove UTIL_OPTS line after GSI removes all utilities from its repo.
+BUILD_TYPE=${BUILD_TYPE:-"Release"} \
+BUILD_VERBOSE=${BUILD_VERBOSE:-"YES"} \
+GSI_MODE=GFS \
+ENKF_MODE=GFS \
+REGRESSION_TESTS=NO \
+UTIL_OPTS="-DBUILD_UTIL=OFF" \
+./gsi.fd/ush/build.sh
```
RussTreadon-NOAA: Given changes to `checkout.sh` need to replace `gsi.fd` with `gsi_enkf.fd`.  For clarity we should rename `build_gsi.sh` as `build_gsi_enkf.sh`

sorc/build_all.sh
```diff
@@ -93,145 +92,159 @@ ERRSCRIPT=${ERRSCRIPT:-'eval [[ $err = 0 ]]'}
...
+  ./build_gsi.sh $_ops_opt $_verbose_opt > $logs_dir/build_gsi.log 2>&1
+  rc=$?
+  if [[ $rc -ne 0 ]] ; then
+    echo "Fatal error in building gsi."
+    echo "The log file is in $logs_dir/build_gsi.log"
+  fi
+  ((err+=$rc))
+}
+
+#------------------------------------
```
RussTreadon-NOAA, line 136: build_gsi section above is now more accurately described as build_gsi_enkf. If we rename `build_gsi.sh` as `build_gsi_enkf.sh`, corresponding changes are needed above.

RussTreadon-NOAA:
After updating to [0cc159d](https://github.com/NOAA-EMC/global-workflow/pull/889/commits/0cc159d3cb2bf4a0aacd73588a59d810b74c933e) found it necessary to update  `gfs_build.cfg` as follows
```
- Building gsi (gsi) .................................... yes
+ Building gsi_enkf (gsi_enkf) .......................... yes
```
Absent this change `build_all.sh` aborts with
```
Orion-login-1[48] rtreadon$ ./build_all.sh
*** Unrecognized line in config file "gfs_build.cfg":
Building gsi (gsi) .................................... yes
```

aerorahul:
> After updating to [0cc159d](https://github.com/NOAA-EMC/global-workflow/pull/889/commits/0cc159d3cb2bf4a0aacd73588a59d810b74c933e) found it necessary to update `gfs_build.cfg` as follows
> 
> ```
> - Building gsi (gsi) .................................... yes
> + Building gsi_enkf (gsi_enkf) .......................... yes
> ```
> 
> Absent this change `build_all.sh` aborts with
> 
> ```
> Orion-login-1[48] rtreadon$ ./build_all.sh
> *** Unrecognized line in config file "gfs_build.cfg":
> Building gsi (gsi) .................................... yes
> ```

Ok.
I see whats happening here.
This is incredibly complicated set of logic and jumping through hoops for something that shouldn't be this convoluted.

RussTreadon-NOAA:
Sorry, a bit more hoop jumping.

Execution of `./build_all.sh` revealed a problem with gsi_monitor.

```
Orion-login-1[55] rtreadon$ ./build_all.sh
 .... Building WW3 pre and post execs ....
 .... Building forecast model ....
 .... Building gsi and enkf ....
 .... Building gsi utilities ....
 .... Building gsi monitor ....
Fatal error in building gsi monitor.
The log file is in /work/noaa/da/Russ.Treadon/git/global_workflow/pr889/sorc/logs/build_gsi_monitor.log
```

We need to replace `gsi.fd` with `gsi_enkf.fd` in `build_gsi_monitor.sh`

```
-GSI_INSTALL_PREFIX="${cwd}/gsi.fd/install"
+GSI_INSTALL_PREFIX="${cwd}/gsi_enkf.fd/install"
```

Also updated error message in `build_gsi_monitor.sh`
 ```
else
   echo << EOF
 FATAL: ${GSI_INSTALL_PREFIX} does not exist
-       Have you built GSI yet?
+       Have you built GSI and EnKF yet?
 EOF
   exit 2
```

**Comment by aerorahul** [comment]
@RussTreadon-NOAA 
Let me make sure this all works before you review again.
I don't want you to troubleshoot errors and mistakes.
I can close the PR, if that helps.

**Comment by RussTreadon-NOAA** [comment]
@aerorahul , I think we're at the end of the build issues for this PR.   `build_all.sh` just finished in my working directory.   It ran to completion.    

The only potential issue for `link_workfow.sh` is whether or not we want to add `sorc/enkf.fd`, `sorc/gsi.fd`, and `sorc/ncdiag.fd` to `.gitignore`.   _git status_ returns these three directories as _Untracked files_.    `sorc/ncdiag.fd` will be removed in the future.   `enkf.fd` and `gsi.fd` will remian.

**Comment by aerorahul** [comment]
I have now, done all the steps:
`checkout.sh`,
`build_all.sh`, 
`link_workflow.sh`
and I don't see errors or stale links.

File: sorc/build_gsi_enkf.sh [hunk 5]
```diff
@@ -0,0 +1,34 @@
...
+      echo "[$BASH_SOURCE]: ${option} requires an argument"
+      usage
+      ;;
+  esac
+done
+shift $((OPTIND-1))
+
+#TODO - remove UTIL_OPTS line after GSI removes all utilities from its repo.
+BUILD_TYPE=${BUILD_TYPE:-"Release"} \
+BUILD_VERBOSE=${BUILD_VERBOSE:-"YES"} \
```
WalterKolczynski-NOAA: Default should be NO so the -v option actually does something, but I think having it at all is unnecessary since the default if not set is no anyway.

sorc/build_all.sh
```diff
@@ -12,22 +12,21 @@ set +x
...
-		  -c build_config:
-		    Selectively build based on the provided config instead of the default config
-		  -h:
-		    print this help message and exit
-		  -v:
-		    Execute all build scripts with -v option to turn on verbose where supported
-
-	EOF
-	exit 1
+  cat << EOF
```
WalterKolczynski-NOAA, line 15: Still don't approve of this change that breaks the nice formatting.

sorc/build_gsi_monitor.sh
```diff
@@ -1,80 +1,40 @@
 #! /usr/bin/env bash
 set -eux
 
-source ./machine-setup.sh > /dev/null 2>&1
```
WalterKolczynski-NOAA, line 4: I think this may need to stay to support partial build? If not, we should take it out of all the build scripts.

RussTreadon-NOAA (approved):
Clone `feature/gsi-utils` at 4db8b89d.  Run `checkout.sh`, `build_all.sh`, and `link_workflow.sh`.   All three scripts ran without error.  A quick check of g-w directories does not show any incomplete links.  _git status_ does not return any untracked files.

aerorahul:
Waiting on @WalterKolczynski-NOAA to review.

sorc/build_gsi_monitor.sh
```diff
@@ -1,80 +1,40 @@
...
-set -x
-cmake $CMAKE_OPTS $DIR_ROOT
-make -j ${BUILD_JOBS:-8} VERBOSE=${BUILD_VERBOSE:-}
-make install
-set +x
-
-# Clean up build directory
-rm -Rf "${BUILD_DIR}"
+BUILD_TYPE=${BUILD_TYPE:-"Release"} \
+BUILD_VERBOSE=${BUILD_VERBOSE:-"YES"} \
```
WalterKolczynski-NOAA: Default should be "NO"

sorc/build_gsi_monitor.sh
```diff
@@ -1,80 +1,40 @@
...
+CMAKE_OPTS=${CMAKE_OPTS} \
```
WalterKolczynski-NOAA, line 37: Does this line do anything?

sorc/build_gsi_utils.sh
```diff
@@ -0,0 +1,29 @@
...
+    :)
+      echo "[$BASH_SOURCE]: ${option} requires an argument"
+      usage
+      ;;
+  esac
+done
+shift $((OPTIND-1))
+
+BUILD_TYPE=${BUILD_TYPE:-"Release"} \
+BUILD_VERBOSE=${BUILD_VERBOSE:-"YES"} \
```
WalterKolczynski-NOAA: As before

WalterKolczynski-NOAA (changes requested):

sorc/build_gsi_monitor.sh
aerorahul: Can you verify it is needed for partial build?
I tried with a partial build for the GSI, GSI-utils and GSI-monitor and had no impact.
I think it is necessary for builds that loads modules in the build scripts in this repo.
It is not necessary for GSI, GSI-utils or GSI-monitor that uses the `build.sh` from their repos.

sorc/build_gsi_monitor.sh
aerorahul: Yes.  It passes `-Dncdiag_ROOT=${GSI_INSTALL_PREFIX}` to the gsi-monitor cmake command.

sorc/build_all.sh
aerorahul: formatting is preserved.

sorc/build_gsi_monitor.sh
WalterKolczynski-NOAA: Got it

WalterKolczynski-NOAA (commented):
I still protest the conversion to spaces that breaks the heredoc indentation, but will let it through if I must. I want to wait until after the cycling PR gets merged, since that is imminent, more important, and modifies the same area. This one will have to be slightly modified to bring in those changes.

WalterKolczynski-NOAA (approved):

PR #929, opened 2022-07-23
Add preamble, convert to bash, and remove env

WalterKolczynski-NOAA:
**Description**

Add preamble, convert to bash, and remove env

This is the first in a wave of commits to improve and standardize the scripts within global workflow. In this commit, all scripts run during execution are converted to bash and a preamble is added to every script that is not sourced by another script.

Every script executed during a forecast cycle is converted to bash. This was mostly straightforward, though there were a couple Korne-shell conventions (primarily using `typeset` to format strings) that had to be replaced with bash-compatable alternatives like `printf`. This in turn required a few modification to prevent zero-padded numbers from being treated as octals (other may have been pre-existing bugs).

The preamble contains a number of feature to standardize code and improve script flow and debugging.

- First, it uses two variables, `$STRICT` and `$TRACE` to control the behavior of `set`. When `$STRICT` is `"YES"`, error on undefined variables (`set -u`) and exit on non-zero return (`set -e`) are turned on. When `$TRACE` is `"YES"`, command trace (`set -x`) is turned on and a useful string is set to `$PS4` that gives the name and line number of the script. Both `$STRICT` and `$TRACE` default to `"YES"`. They also set up commands, `$ERR_EXIT_ON` and `$TRACE_ON`, that will restore the setting of each in the event a script needs to temporarily turn them off.

- Second, the preamble sets up primative timing of scripts using Posix `date`.

- Third, it echos the script is beginning and at what time.

- Finally, it also establishes a postamble function and sets it to run as a trap of EXIT. The postamble will use the end time to calculate the run time of the script, then print that the script has ended at what time, how long has elapsed, and the exit code. By setting this up as a trap instead of just calling it at the end of the script, it ensures the postamble is called even if the script exits early because there is an error.

- In response to this standardization, parts of scripts that performed these preamble functions (announcing start/end, `set -x`, etc) have been deleted. For some scripts where temporarily turning off `-x` or `-e` is needed, they now use `$ERR_EXIT_ON` and `$TRACE_ON` to return to the correct state afterwards, instead of blindly turning the setting back on.

- Additionally, some modifications were needed to comply with `set -eu`. Mostly taking care of undefined variables, but also a couple instances where a non-zero return code had to be dealt with.

If users wish to use their own preamble script instead, the default script can be overriden by setting `$PREAMBLE_SCRIPT` before the run begins.

Instance where scripts would print the full list of environment variables have been removed. These can be spot added back in to debug as necessary. Alternatively, a future PR will add them back in in a standardized way.

`rstprod.sh` is added to the link list from gsi_monitor.fd, as it is needed for the radmon scripts.

The placeholders for AWIPS and GEMPAK in the Hera and Orion environment scripts were replaced with the correct definitions. There were also other modifications to AWIPS and GEMPAK scripts to get it working for development (AWIPS still isn't and will be fixed in the future).

GSI scripts that were brought in recently had all of their backticks replaced with `$(  )` as was done with all other script previously.

Refs: #397

**Type of change**
- [x] Script maintenance

**How Has This Been Tested?**
- [x] Cycled test on Hera with BUFRSND, GEMPAK, AWIPS, WAFS
- [x] Coupled test on Hera
  
**Checklist**


WalterKolczynski-NOAA:
Cycled and coupled confirmed bitwise identical to develop on Hera.

AWIPS fails, but gets farther than develop. More effort will be needed to get that working.

env/HERA.env
```diff
@@ -261,9 +261,17 @@ elif [ $step = "postsnd" ]; then
...
+    export APRUN_AWIPSCFP="$launcher -n $npe_awips --multi-prog"
 
 elif [ $step = "gempak" ]; then
 
-     echo "WARNING: $step is not enabled on $machine!"
+    nth_max=$(($npe_node_max / $npe_node_gempak))
+
+    export NTHREADS_GEMPAK=${nth_gempak:-1}
+    [[ $NTHREADS_GEMPAK -gt $nth_max ]] && export NTHREADS_GEMPAK=$nth_max
+    export APRUN="$launcher -n $npe_gempak --multi-prog"
```
aerorahul, line 276: Is `gempak` supported on Hera, Orion?

jobs/rocoto/anal.sh
```diff
@@ -1,4 +1,9 @@
-#!/bin/bash -x
+#! /usr/bin/env bash
+
+PREAMBLE_SCRIPT="${PREAMBLE_SCRIPT:-$HOMEgfs/ush/preamble.sh}"
+if [ -f "${PREAMBLE_SCRIPT}" ]; then
+  source $PREAMBLE_SCRIPT
+fi
```
aerorahul: What is the point of sourcing `PREAMPLE_SCRIPT` in `anal.sh` when it is sourced in the J-JOB.

I can see you need to source it in jobs that don't have a j-job e.g. `aerosol_init.sh` above.

jobs/rocoto/awips.sh
```diff
@@ -73,7 +78,7 @@ for fhr in $fhrlst; do
     fhmax=84
     if [ $fhr -ge $fhmin -a $fhr -le $fhmax ] ; then
 	if [[ $(expr $fhr % 3) -eq 0 ]]; then
-            fhr3=$(printf %03i $fhr)
+            fhr3=$(printf %03d $((10#$fhr)))
```
aerorahul, line 78: why is the one on the left wrong?

jobs/rocoto/earc.sh
```diff
@@ -134,20 +139,26 @@ if [ $ENSGRP -eq 0 ]; then
...
+        ${ERR_EXIT_ON:-set -eu}
     fi
 
     #-- Archive online for verification and diagnostics
     [[ ! -d $ARCDIR ]] && mkdir -p $ARCDIR
     cd $ARCDIR
 
-    $NCP $ROTDIR/enkf${CDUMP}.$PDY/$cyc/$COMPONENT/${CDUMP}.t${cyc}z.enkfstat         enkfstat.${CDUMP}.$CDATE
-    $NCP $ROTDIR/enkf${CDUMP}.$PDY/$cyc/$COMPONENT/${CDUMP}.t${cyc}z.gsistat.ensmean  gsistat.${CDUMP}.${CDATE}.ensmean
+    if [[ -f $ROTDIR/enkf${CDUMP}.$PDY/$cyc/$COMPONENT/${CDUMP}.t${cyc}z.enkfstat ]]; then
```
aerorahul: why not use `safe_copy` here?

jobs/rocoto/waveprep.sh
```diff
@@ -1,4 +1,9 @@
-#!/bin/bash -x
+#! /usr/bin/env bash
+
+PREAMBLE_SCRIPT="${PREAMBLE_SCRIPT:-$HOMEgfs/ush/preamble.sh}"
+if [ -f "${PREAMBLE_SCRIPT}" ]; then
+  source $PREAMBLE_SCRIPT
+fi
```
aerorahul: capping back at the last rocoto job wrapper, why is this necessary here, when the j-job does it first thing.

scripts/exgdas_atmos_chgres_forenkf.sh
```diff
@@ -18,11 +18,9 @@
...
 ################################################################################
 
-#  Set environment.
-export VERBOSE=${VERBOSE:-"YES"}
-if [ $VERBOSE = "YES" ]; then
-   echo $(date) EXECUTING $0 $* >&2
-   set -x
+PREAMBLE_SCRIPT="${PREAMBLE_SCRIPT:-$HOMEgfs/ush/preamble.sh}"
+if [ -f "${PREAMBLE_SCRIPT}" ]; then
+  source $PREAMBLE_SCRIPT
```
aerorahul: Why is this needed in the exscript when the JJob has already set the necessary tracing, error handling etc.
This seems like a lot of duplication.

scripts/exgfs_atmos_awips_20km_1p0deg.sh
```diff
@@ -75,7 +77,7 @@ echo "#######################################"
 echo " Process GRIB AWIP GRIB2 PRODUCTS      "
 echo "#######################################"
 echo " "
-set -x
+${TRACE_ON:-set -x}
```
aerorahul, line 77: What is the use of the preamble if every script will set/reset this.

scripts/exgfs_wave_post_gridded_sbs.sh
```diff
@@ -55,7 +55,7 @@
   echo "Starting at : $(date)"
   echo '-------------'
   echo ' '
-  [[ "$LOUD" = YES ]] && set -x
+  ${TRACE_ON:-set -x}
```
aerorahul, line 54: Technically this is correct, but the entire `set +x` before and echo and `set -x` after the echo is useless when `TRACE_ON` is set.
What real difference does it make to do `set +x` for an echo statement when the default project wide is `set -x`?

aerorahul (changes requested):
Several questions.  See inline comments.

env/HERA.env
WalterKolczynski-NOAA: There is a gempak module available on both machines.

jobs/rocoto/awips.sh
WalterKolczynski-NOAA: bash has an annoying tendency to treat zero-padded numbers as octals. The 10# forces it to be treated as a decimal (which is then zero-padded again; it's annoying). Without this is will complain about numbers like '018', because 8 is an invalid octal digit.

jobs/rocoto/earc.sh
WalterKolczynski-NOAA: Probably just missed it. Will update. I think we need to have someone review this and identify which files we should fatally fail on and which we can continue on without.

jobs/rocoto/anal.sh
WalterKolczynski-NOAA: Otherwise you don't have it while the rocoto pre-job runs, and there are sometimes issues (especially undefined variables) there too.

scripts/exgfs_atmos_awips_20km_1p0deg.sh
WalterKolczynski-NOAA: This is only used if the script is using `set +x` before this point, usually to echo a bunch of lines (the wave scripts do a lot of this). This makes sure after that section has completed, it returns to its prior state instead of blinding turning `set -x` back on. This way if you turn `TRACE=NO`, then it won't turn the trace on here.

scripts/exgfs_wave_post_gridded_sbs.sh
WalterKolczynski-NOAA: I didn't want to go through all the (mostly wave) scripts and change this overall paradigm just yet (this PR is big enough). When I move on to giving these scripts closer attention, I plan on revisiting this. Simplest solution then will be to leave trace on (if it is on) and echo a heredoc instead of individual echo lines (which is how all multiline echos should be handled).

jobs/rocoto/awips.sh
aerorahul: What is the incoming value of `fhr`?  Does it have the `0` padded?
If it doesn't, you would not get the error.

jobs/rocoto/awips.sh
```diff
@@ -1,4 +1,9 @@
-#!/bin/bash -x
+#! /usr/bin/env bash
```
aerorahul, line 1: I also don't understand this job.
It doesn't have a typical j-job?
How is this being called in ops?

jobs/rocoto/awips.sh
WalterKolczynski-NOAA: Probably not? Something to evaluate later.

jobs/rocoto/awips.sh
WalterKolczynski-NOAA: The jjobs are coded in `AWIPS20SH` and `AWIPSG2SH`, which are set in [config.awips](https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/config/config.awips). Those are then called by this script.

Yeah, this arrangement doesn't make sense to me either. Still really want to get rid of all these `jobs/rocoto/` scripts anyway and call j-jobs directly.

scripts/exgdas_atmos_chgres_forenkf.sh
WalterKolczynski-NOAA: The settings aren't handed down to daughter scripts (unless sourced, of course).

WalterKolczynski-NOAA: Also, even if the `set` built-in were, the way I have the postamble set up with the trap wouldn't work.

jobs/rocoto/awips.sh
aerorahul: I see whats going with `awips.sh`.
This is being treated like `arch.sh` or any other without a j-job.
The j-job does not follow the convention of the other global-workflow j-jobs.
The configs are sourced in this rocoto wrapper job and probably not in the j-job.

jobs/rocoto/earc.sh
WalterKolczynski-NOAA: Oh, it's not that I missed it. I fixed this bit; then later I realized it was going to keep being an issue and made a function instead, but never replaced the first edit with the function. Fixed now.

We could still use some feedback on which of these (if any) should be fatal. These changes are just ignoring missing files so it completes without regard to whether we _should_ or not.

WalterKolczynski-NOAA: Wait, safe_copy isn't even in this script, it's in the other archive script. Should I copy that function here, even though it is only needed a few times in this script?

aerorahul: perhaps put that in a `ush/workflow_utils.sh` and source it.  You could call it `file_utils.sh` if you wish to use other file/directory manipulations.

jobs/rocoto/anal.sh
```diff
@@ -10,4 +12,6 @@ status=$?
 # Execute the JJOB
 $HOMEgfs/jobs/JGLOBAL_ATMOS_ANALYSIS
 status=$?
-exit $status
+
+
+exit ${status}
```
KateFriedman-NOAA, line 17: Why {} for $status here but not in all of the other jobs/rocoto scripts?

jobs/rocoto/vrfy.sh
```diff
@@ -56,7 +58,7 @@ echo
 echo "=============== START TO GENERATE QUARTER DEGREE GRIB1 FILES ==============="
 if [ $MKPGB4PRCP = "YES" -a $CDUMP = "gfs" ]; then
     if [ ! -d $ARCDIR ]; then mkdir $ARCDIR ; fi
-    nthreads_env=${OMP_NUM_THREADS:-1} # get threads set in env
+    nthreads_env=${OMP_NUM_THREADS:-1} # get threads set in env
```
KateFriedman-NOAA: Remove extra whitespace at end of line 61.

jobs/rocoto/earc.sh
WalterKolczynski-NOAA: Done. I'm not going to go through every script (yet) and replace all the file checking with safe_copy(), but did apply it to earc.

WalterKolczynski-NOAA: Maybe I should rename it non_blocking_copy?

WalterKolczynski-NOAA: Changed to nb_copy

ush/preamble.sh
```diff
@@ -0,0 +1,86 @@
...
+	end_time=$(date +%s)
+	elapsed_sec=$((end_time - start_time))
+	elapsed=$(date -d@${elapsed_sec} -u +%H:%M:%S)
+
+	# Announce the script has ended, then pass the error code up
+	echo "End ${script} at $(date -u) with error code ${rc:-0} (time elapsed: ${elapsed})"
+	exit ${rc}
+}
+
+# Place the postable in a trap so it is always called no matter how the script exits
```
KateFriedman-NOAA: Spelling mistake, missing the "m" in "postable"?

KateFriedman-NOAA (approved):
Left a few minor comments. Looks fantastic! Approve pending resolution of other reviewer comments and any remaining testing. Thank you for these updates @WalterKolczynski-NOAA !

jobs/rocoto/anal.sh
WalterKolczynski-NOAA: This was the one I happened to change at some point for some reason. Eventually I will likely make sure `{ }` is always used.

ush/preamble.sh
WalterKolczynski-NOAA: done

jobs/rocoto/vrfy.sh
WalterKolczynski-NOAA: done

jobs/rocoto/anal.sh
KateFriedman-NOAA: Roger that!

aerorahul (approved):
Looks good. 
Hopefully now the stdout will look nice and make it easier to trace failures

jobs/rocoto/awips.sh
aerorahul: Ok.

jobs/rocoto/anal.sh
aerorahul: Something to look for to avoid repetition later.

scripts/exgdas_atmos_chgres_forenkf.sh
aerorahul: to be evaluated later.

scripts/exgfs_wave_post_gridded_sbs.sh
aerorahul: Understood.

NeilBarton-NOAA:
Was this tested with DA? I'm getting unbound variable errors when running

WalterKolczynski-NOAA:
> Was this tested with DA? I'm getting unbound variable errors when running

Yes, except the new JEDI-based stuff.

PR #931, opened 2022-07-26
Optimize DA clone, build, and link

### LINKED ISSUE #930: Optimize DA clone & build
**Description**
This issue is opened to document changes to optimize the checkout and build for DA components.   This development is done in g-w branch `feature/clone`.

**Requirements**
This issue adds flexibility to the g-w clone and build via the following changes:
- add option to select DA repositories to clone
- modify build so that it only builds cloned DA components.   

_The default checkout behavior is to **not build** any DA components._

The work in this issue is complete when
- default behavior is as outlined above
- added options yield the desired behavior
- PR is opened, reviewed, approved, and merged

**Dependencies**
This issue has no dependencies on other issues or PRs.

RussTreadon-NOAA:
**Description**
The PR contains changes to optimize the DA clone, build, and link.   Changes are made to `checkout.sh`, `build_all.sh`, and `link_workflow.sh` in the g-w `sorc/` directory.  These changes are in g-w branch `feature/clone`

Two arguments are added to `checkout.sh` to allow the user to specify which DA package to build the global workflow with.  These options are 
- `-g`: clone from the [GSI](https://github.com/NOAA-EMC/GSI) repo and build the g-w for GSI-based DA
- `-u`:  clone from the [GDASApp](https://github.com/NOAA-EMC/GDASApp) repo and build the g-w for UFS-based DA

If no option is specified, `checkout.sh` does not clone any DA and DA related repos.  This is the default behavior of `checkout.sh`.   (_DA related_ repos include [GLDAS](https://github.com/NOAA-EMC/GLDAS), [GSI-utils](https://github.com/NOAA-EMC/GSI-utils), and [GSI-Monitor](https://github.com/NOAA-EMC/GSI-Monitor).)

`build_all.sh` is modified to detect which repos and have been cloned and to build accordingly.   `link_workflow.sh` is modified to detect which directories are present and link/copy accordingly.
 
**Type of change**

Merger of this PR into develop is a _breaking change_ in that the default behavior of `checkout.sh` is altered.   Currently, `checkout.sh` with no options clones the GSI repo.   This behavior changes when this PR is merged into develop.   Thereafter, `checkout.sh` with no option results in a local g-w without DA and DA-related repos.   If developers want to cycle with GSI-based DA, they **must** include execute `checkout.sh -g`.   If UFS-based DA is desired, the command to execute is `checkout.sh -u`

**How Has This Been Tested?**
g-w branch `feature/clone` has been cloned on Orion and various options have been tested with `checkout.sh`.   The intended behavior is observed for each case.  Issue #930 documents these tests and their results.

**Checklist**


sorc/checkout.sh
```diff
@@ -96,13 +102,23 @@ function checkout() {
...
 }
 
-while getopts ":chm:o" option; do
+while getopts ":chgum:o" option; do
   case $option in
     c)
       echo "Recieved -c flag, will delete any existing directories and start clean"
       export CLEAN="YES"
       ;;
+    g)
```
aerorahul, line 118: What happens if one does:
```
./checkout.sh -g -u
```

sorc/link_workflow.sh
```diff
@@ -28,6 +28,7 @@ SLINK="ln -fs"
 pwd=$(pwd -P)
 
 # Link post
+[[ -d upp.fd ]] && rm -rf upp.fd
```
aerorahul, line 31: Shouldn't this be `-L` or `-h`  since `upp.fd` at this level is a symlink?

sorc/link_workflow.sh
```diff
@@ -85,13 +85,17 @@ fi
...
+  cd ${pwd}/../jobs               ||exit 8
     $LINK ../sorc/gldas.fd/jobs/JGDAS_ATMOS_GLDAS            .
+fi
 cd ${pwd}/../parm               ||exit 8
     # [[ -d post ]] && rm -rf post
     # $LINK ../sorc/upp.fd/parm                           post
-    [[ -d gldas ]] && rm -rf gldas
-    $LINK ../sorc/gldas.fd/parm                              gldas
+    if [ -d ../sorc/gldas.fd ]; then
+      [[ -d gldas ]] && rm -rf gldas
```
aerorahul, line 96: Would this also not be a link?

aerorahul (commented):
Changes look good.
I do have a few comments and question.
1. `./checkout.sh -g -u`, what happens here?
2. `-d` for symlinks.  I realize you are simply following what is in place, so this is more of a comment for the code-managers and maintainers to use proper test.

sorc/link_workflow.sh
CoryMartin-NOAA: But since it links to a dir it should return `0` still

sorc/checkout.sh
CoryMartin-NOAA, line 118: Similar to Rahul's question. Can we just keep `CHECKOUT_GSI="YES"` for `-g` and `CHECKOUT_GDAS="YES"` for `-u` so that `-gu` would allow for both to be built? I could see a use case where we might want both analyses to run on the same model fcst (but only one increment) for testing purposes.

RussTreadon-NOAA: Good question.  `checkout.sh -g -u` results in the GDASApp and DA related repos (gldas, gsi_utils, gsi_monitor) being cloned.   Arguments are processed sequentially so the `-g` settings are overridden by `-u`.    `checkout.sh -u -g` resulted in the GSI repo being cloned along with DA related repos.   

What scripting should be added to `checkout.sh` if we want to flag `checkout.sh -g -u` as an error?  I don't know how to trap this type of error cleanly and efficiently.

sorc/checkout.sh
```diff
@@ -96,13 +102,23 @@ function checkout() {
...
+      echo "Receieved -g flag for optional checkout of GSI-based DA"
+      CHECKOUT_GSI="YES"
+      CHECKOUT_GDAS="NO"
```
aerorahul:
```suggestion
```

sorc/checkout.sh
```diff
@@ -96,13 +102,23 @@ function checkout() {
...
+      ;;
     h)  usage;;
+    u)
+      echo "Received -u flag for optional checkout of UFS-based DA"
+      CHECKOUT_GSI="NO"
```
aerorahul:
```suggestion
```

sorc/checkout.sh
```diff
@@ -131,14 +147,23 @@ mkdir -p ${logdir}
...
 checkout "ufs_model.fd"    "https://github.com/ufs-community/ufs-weather-model" "${ufs_model_hash:-b97375c}"      ; errs=$((errs + $?))
-checkout "gsi_enkf.fd"     "https://github.com/NOAA-EMC/GSI.git"                "67f5ab4"                         ; errs=$((errs + $?))
-checkout "gsi_utils.fd"    "https://github.com/NOAA-EMC/GSI-Utils.git"          "322cc7b"                         ; errs=$((errs + $?))
-checkout "gsi_monitor.fd"  "https://github.com/NOAA-EMC/GSI-Monitor.git"        "acf8870"                         ; errs=$((errs + $?))
-checkout "gdas.cd"         "https://github.com/NOAA-EMC/GDASApp.git"            "5952c9d"                         ; errs=$((errs + $?))
-checkout "gldas.fd"        "https://github.com/NOAA-EMC/GLDAS.git"              "fd8ba62"                         ; errs=$((errs + $?))
 checkout "ufs_utils.fd"    "https://github.com/ufs-community/UFS_UTILS.git"     "a2b0817"                         ; errs=$((errs + $?))
 checkout "verif-global.fd" "https://github.com/NOAA-EMC/EMC_verif-global.git"   "c267780"                         ; errs=$((errs + $?))
 
+if [[ "${CHECKOUT_GSI:-NO}" == "YES" ]]; then
```
aerorahul: Define `CHECKOUT_GSI=NO` and `CHECKOUT_GDAS=NO` at the top of `checkout.sh` before the arguments are parsed.

aerorahul (commented):
Suggested changes would do the trick, I think.

sorc/checkout.sh
aerorahul: `-g -u` should checkout the GSI and the GDAS apps, so that would allow "intercomparison" exercises to proceed.

sorc/link_workflow.sh
RussTreadon-NOAA: You are right.  I was just following the pattern already in `link_workflow.sh`.   If we replace `-d` with `-L` or `-h`, it should be done consistently throughout `link_workflow.sh`.   Repo code managers should review and update the sections of `link_workflow.sh` as directed by the g-w team.   What's the g-w team's preference:  `-L` or `-h` for symlinks?

sorc/link_workflow.sh
RussTreadon-NOAA: Same reply as above.  The GLDAS code manager should replace `-d` with the check the g-w team prefers.

sorc/checkout.sh
RussTreadon-NOAA: This works for me.  I agree with @CoryMartin-NOAA .   I can see developers wanting to run both GSI and UFS based DA from the same g-w installation.  This type of testing will frequently occur as we cross check between the two DA engines.   Performing these tests from the same `HOMEgfs` makes sense.    

I confirmed that @CoryMartin-NOAA 's suggestion works.   After making the suggested changes in a working copy, `checkout.sh -g -u` results in both the GSI and GDASApp repos being cloned.   Is this approach acceptable to the g-w team?

sorc/checkout.sh
RussTreadon-NOAA: _at the top of checkout.sh_  - Is this the literal top

```
#!/bin/sh
set +x
set -u

CHECKOUT_GSI=NO
CHECKOUT_GDAS=NO

function usage() {
```

or after the function definitions

```
  fi
  echo
  return 0
}

CHECKOUT_GSI=NO
CHECKOUT_GDAS=NO

while getopts ":chgum:o" option; do
  case $option in
    c)
```

If we set `CHECKOUT_GSI` and `CHECKOUT_GDAS` in either of these locations, the default definitions in if tests are no longer necessary

```
if [[ "${CHECKOUT_GSI:-NO}" == "YES" ]]; then
  checkout "gsi_enkf.fd" "https://github.com/NOAA-EMC/GSI.git" "67f5ab4"; errs=$((errs + $?))
fi

if [[ "${CHECKOUT_GDAS:-NO}" == "YES" ]]; then
  checkout "gdas.cd" "https://github.com/NOAA-EMC/GDASApp.git" "5952c9d"; errs=$((errs + $?))
fi
```

If we adopt this approach, shall we apply it to other variables:  `CLEAN`, `checkout_gtg`, `checkout_wafs`, `ufs_model_hash`?

sorc/checkout.sh
RussTreadon-NOAA: Commit @CoryMartin-NOAA 's suggestion at [ff2192d](https://github.com/NOAA-EMC/global-workflow/commit/ff2192dc6fa005ee52b8c8493271a27a8ef569c2)

sorc/checkout.sh
RussTreadon-NOAA: Commit @CoryMartin-NOAA 's suggestion at [ff2192d](https://github.com/NOAA-EMC/global-workflow/commit/ff2192dc6fa005ee52b8c8493271a27a8ef569c2)

aerorahul: @RussTreadon-NOAA 
The latter is what I had in mind; i.e. after the function definitions.
And yes, if you would do the same for the other argument options, that would be great.  Thanks~

sorc/link_workflow.sh
aerorahul: You can leave it with `-d` for now.  I will make a mental note of it and save it for later.

CoryMartin-NOAA (approved):

sorc/link_workflow.sh
RussTreadon-NOAA: I tested `-L` and it works as intended.   This test reminded me that NCO's execution of `link_workflow.sh` sets `LINK="cp -rp"`.    Thus, for NCO, 

```
# Link post
[[ -d upp.fd ]] && rm -rf upp.fd
$LINK ufs_model.fd/FV3/upp upp.fd
```
results in a physical directory named `upp.fd`   In this case `-d` is the proper test.   If we use `-h` or `-L`, the test is `false` and the directory is not removed.   This may not matter so much since the next line does a `cp -rp`.   For clarity and consistency, it would be good for the directory to be removed before `ufs_model.fd/FV3/upp` is copied.

I'm not sure how we want to handle the `ln -fs` versus `cp -rp` differences in `LINK` depending on the `RUN_ENVIR`.

One idea.   `link_workflow.sh` includes both `LINK` and `SLINK`.    `-h` or `-L` checks are appropriate for `SLINK`.   `-d` works for `LINK`.   We could review the use of `LINK` and `SLINK` in `link_workflow.sh` and update as appropriate.   Once this is done, a second sweep through `link_workflow.sh` could ensure `-d` is used for `LINK` directories and `-h` or `_L` is used for `SLINK` directories.   This or related development should be the subject of a new issue and subsequent PR.

sorc/checkout.sh
RussTreadon-NOAA: Suggested change committed to `feature/clone` at [ced1632](https://github.com/NOAA-EMC/global-workflow/commit/ced16325b3de7e7fd013363e2c59478a9babe173)

aerorahul (approved):
Thank you for making GDASApp and GSI optional to speed up checkout and builds for applications not using these components.
Also, thanks for improving the overall script.

sorc/link_workflow.sh
WalterKolczynski-NOAA: Keep in mind when NCO runs these, they make copies instead of symlinks.

sorc/checkout.sh
```diff
@@ -30,6 +34,8 @@ function checkout() {
   #   topdir [default: $(pwd)]: parent directory to your checkout
   #   logdir [default: $(pwd)]: where you want logfiles written
   #   CLEAN [default: NO]:      whether to delete existing directories and create a fresh clone
+  #   CHECKOUT_GSI [default: NO]:  whether to clone GSI repository (GSI-based DA)
+  #   CHECKOUT_GDAS [default: NO]: whether to clone GDASApp repository (UFS-based DA)
```
WalterKolczynski-NOAA: This is the docblock for the function, not the whole script. `CHECKOUT_GSI` and `CHECKOUT_GDAS` are not used by the function.

sorc/link_workflow.sh
```diff
@@ -137,29 +145,43 @@ fi
...
+#------------------------------
+if [ -d ../sorc/gdas.cd ]; then
+  cd ${pwd}/../fix                ||exit 8
+    [[ -d fix_gdas ]] && rm -rf fix_gdas
+    $LINK $FIX_DIR/fix_gdas .
+fi
 
 #------------------------------
-#--add GDASApp files
+#--add GDASApp files
```
WalterKolczynski-NOAA: Extra space was added at the end of this line

WalterKolczynski-NOAA (changes requested):

sorc/checkout.sh
RussTreadon-NOAA: `CHECKOUT_GSI` and `CHECKOUT_GDAS` lines removed.   Done at [53befc3](https://github.com/NOAA-EMC/global-workflow/commit/53befc3d11a5b6e8ff5404e8eed2ecd48a6bc0bc)

WalterKolczynski-NOAA (approved):

sorc/link_workflow.sh
RussTreadon-NOAA: Extra space removed at [53befc3](https://github.com/NOAA-EMC/global-workflow/commit/53befc3d11a5b6e8ff5404e8eed2ecd48a6bc0bc)

PR #932, opened 2022-07-26
Changes to resources and upstream/downstream jobs from low res testing on supported platforms (dev_v16)

KateFriedman-NOAA:
**Description**

This PR includes a varied set of changes from testing `dev_v16` (GFSv16.2.1) at lower resolutions on WCOSS2, Hera, and Orion. It also includes changes to support `verif-global.v2.9.5` and a change to how the prepobs module loads.

Changes

1. Slight hack to support `verif-global.v2.9.5` on Hera/Orion: unset several version variables in `jobs/rocoto/metp.sh` when on Hera or Orion. These workflow level versions conflict with the versions that the METplus package needs to run on those platforms. This is only for the R&D machines in the v16 package, the `develop` branch will be using a better solution. This does not impact other jobs.
2. Move the prepobs module load from `jobs/rocoto/prep.sh` into the module_base files. Remove `HOMEprepobs` from `config.base.emc.dyn`.
3. Add `COMINtcvital` to `config.base.emc.dyn`; was missing and is needed in dev runs.
4. Set `OUTPUT_FILETYPES` to serial netcdf when C384 on WCOSS2. Related to similar change for C384 EnKF forecast jobs on WCOSS2. Needs additional investigations in `develop` branch work.
5. Update machine-setup.sh to use `module reset` on Hera/Orion; includes necessary hack to set `LMOD_SYSTEM_DEFAULT_MODULES=contrib` first to make `module reset` work. Remove `module purge` on Hera, not needed if doing `module reset`.
6. Update `workflow_utils.py` to add `debug=true` to all job submissions on WCOSS2.
7. Set `export MPICH_MPIIO_HINTS="*:romio_cb_write=disable"` in analcalc block of `config.resources.emc.dyn` and `config.resources.nco.static`: needed for C384 analcalc jobs on WCOSS2.
8. Resource updates to `config.fv3.emc.dyn`: 
* Increase `npe_wav*` values as needed for WCOSS2 (waves disabled elsewhere).
* Force C192 fcst thread value to 1 on WCOSS2, fails with anything larger (investigate further in `develop` branch).
* Update C384 block: 2 threads on R&Ds, 2 WRITE_GROUPS for gfs on R&Ds, and on WCOSS2 use 5 threads and 8 WRITE_GROUPS (to support running serial netcdf)
9. Resource updates to `config.resources.emc.dyn`
* Update WCOSS2 `npe_node_anal` check to only set to 15 when C768.
* Add C384 to WCOSS2 `npe_node_fcst*` check.
* Remove resolution check for eobs values when on WCOSS2, use the ops resources for all resolutions (optimize further in `develop` branch).
* When on WCOSS2 use ops eupd resources for all resolutions (investigate further in `develop` branch).
* Increase `nth_eupd` when C384 on Hera from 4 to 8 to get more memory for job there.

**Type of change**

**How Has This Been Tested?**

- [x] Clone and build tests on WCOSS2, Hera, Orion
- [x] Cycled tests on WCOSS2, Hera, Orion: C384C192L127 & C192C96L127
- [x] Cycled test on WCOSS2 @ C768C384L127

Confirmed fit2obs and METplus jobs work and produce output. Had @malloryprow check on METplus jobs and stat files to confirm version unsets hack works.
  
Refs #665


jobs/rocoto/metp.sh
```diff
@@ -51,6 +51,13 @@ export DATAROOT="$RUNDIR/$CDATE/$CDUMP/metp.${jobid}"
 [[ -d $DATAROOT ]] && rm -rf $DATAROOT
 mkdir -p $DATAROOT
 
+if [ $machine = "HERA" -o $machine = "ORION" ]; then
+  unset prod_util_ver
+  unset grib_util_ver
+  unset nco_ver
+  unset intel_ver
+  unset anaconda_ver
```
WalterKolczynski-NOAA, line 59: Reminder to myself and note to everyone else: Kate and I discussed this and decided this was the best way to handle things for dev 16, which uses an older version of the metp script that loads its own modules with specific versions that are different than we are using elsewhere. In develop, these will be set properly as we have a verif branch that no longer loads its own modules.

WalterKolczynski-NOAA (approved):

WalterKolczynski-NOAA (approved):

parm/config/config.resources.nco.static
```diff
@@ -127,6 +127,8 @@ elif [ $step = "anal" ]; then
 
 elif [ $step = "analcalc" ]; then
 
+    export MPICH_MPIIO_HINTS="*:romio_cb_write=disable"
```
aerorahul, line 130: remind me again why this shouldn't be set site-wide?

ush/rocoto/workflow_utils.py
```diff
@@ -339,6 +339,10 @@ def get_resources(machine, cfg, task, reservation, cdump='gdas'):
             natstr = "-l place=vscatter"
             if memory is None:
                natstr += ":exclhost"
+            natstr += ",debug=true"
+
+        if machine in ['WCOSS2'] and task in ['arch', 'earc', 'getic']:
+            natstr = "-l debug=true"
```
aerorahul: You could move these 2 lines to 337 and drop the `and task in ['arch' ...]:`
Then change the line 339 `=` to `+=`
You won't need 342.

aerorahul (commented):
a question and a suggestion.

parm/config/config.resources.nco.static
KateFriedman-NOAA: It slows down the analysis jobs significantly when set in the anal block of WCOSS2.env (the GFSv16 GSI analcalc JJOB sources the "anal" block of WCOSS2.env). Setting this in the analcalc block of config.resources* or config.analcalc were the only ways I could see to set it only for analcalc.

I can reexamine this setting for other jobs in the WCOSS2 `develop` port and set it more appropriately since we own those GSI scripts now. I wanted to be cautious with `dev_v16` since these updates will feed into the `release/gfs.v16.3.0` branch after this PR goes in.

aerorahul: Ok. Its fine for high impact work, but we should reexamine they this job needs this and not any other.

ush/rocoto/workflow_utils.py
KateFriedman-NOAA: I ingested your suggestion and further reduced that section from:
```
 if machine in ['WCOSS2'] and task not in ['arch', 'earc', 'getic']:
            natstr = "-l place=vscatter"
            if memory is None:
               natstr += ":exclhost"
            natstr += ",debug=true"

        if machine in ['WCOSS2'] and task in ['arch', 'earc', 'getic']:
            natstr = "-l debug=true"
```
to:
```
if machine in ['WCOSS2']:
    natstr = "-l debug=true"

    if task not in ['arch', 'earc', 'getic']:
       natstr += ",place=vscatter"
       if memory is None:
           natstr += ":exclhost"
```

The NATIVE entity in the rocoto xmls on WCOSS2 used to be (exclusive job example):
`-l place=vscatter:exclhost,debug=true`
...and are now:
`-l debug=true,place=vscatter:exclhost`
I have been running with this update on WCOSS2 all afternoon and it works. No observed submission errors. I have committed the adjustment to `ush/rocoto/workflow_utils.py` @ 0a21614.

parm/config/config.resources.nco.static
KateFriedman-NOAA: The [efcs jobs need this setting](https://github.com/NOAA-EMC/global-workflow/blob/dev_v16/env/WCOSS2.env#L127) but that's the only other job that I've come across that needs it on WCOSS2. Will reexamine this setting in the port of `develop` to WCOSS2.

aerorahul (approved):
looks good.

PR #942, opened 2022-07-28
Remove MPICH_MPIIO_HINTS analcalc setting

KateFriedman-NOAA:
**Description**

Remove the MPICH_MPIIO_HINTS analcalc setting from `config.resources.nco.static`. Leaving it in the `config.resources.emc.dyn` version.

Refs: #665


aerorahul (approved):
this does not impact NCO.

PR #975, opened 2022-08-16
Bring ecflow script changes from production that are not in release/gfs.v16.3.0.

lgannoaa:
It bring ecflow script changes from production that are not in release/gfs.v16.3.0.

### DIFF
--- a/ecf/scripts/enkfgdas/analysis/recenter/jenkfgdas_sfc.ecf
+++ b/ecf/scripts/enkfgdas/analysis/recenter/jenkfgdas_sfc.ecf
@@ -4,7 +4,7 @@
 #PBS -q %QUEUE%
 #PBS -A %PROJ%-%PROJENVIR%
 #PBS -l walltime=00:06:00
-#PBS -l select=1:mpiprocs=80:ompthreads=1:ncpus=80:mem=80GB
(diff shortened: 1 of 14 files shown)


aerorahul:
@WeiWei-NCO 
As we prepare for code handoff, we would like to bring these uncommitted changes (I assume were made by the NCO team) from current production into the release branch that is in preparation for GFS v16.3.0.
A review from you to validate these changes made by NCO would be greatly appreciated.
Also tagging @StevenEarle-NCO for any additional guidance.

aerorahul:
Just a quick note before I forget.
The resource updates in the `.ecf` should be consistent w/ `config.resources`.   In this case `config.resources.nco.static` and may be `config.resources.emc.dyn`.  
If they are not brought in, there won't be a catastrophic failure, however, we will deviate as most of the resource updates are in memory allocation within the `ecf`
However, there are a couple of `wafs` ecf jobs in this PR that is altering the task count.  That will have an impact.  WAFS jobs do not source `config` files to obtain their resource allocation.  They do their own thing and likely these are hard-wired in the  exscripts.  
The development team should ensure they are consistent in v16.3.0 release branch.

WalterKolczynski-NOAA (approved):

PR #1007, opened 2022-08-31
Add shell linter Github action for pull requests

WalterKolczynski-NOAA:
**Description**
Adds a Github action to apply a shell linter to all pull requests. The linter will identify any deviations from the [Google style guide](https://google.github.io/styleguide/shellguide.html) and flag them during the code review.

To begin, only the ush/preamble.sh file is checked. This PR also includes updates to that script to make it compliant with G-Style.

Refs: #397

**Type of change**

**How Has This Been Tested?**
- [x] Unit test of preamble script
- [x] Test PR incorporating the action
  
**Checklist**

### DIFF
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -2,7 +2,6 @@
 name: linters
 on:
   pull_request:
-    types: [opened, reopened, edited]
 
 permissions:
   contents: read
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -1,17 +1,32 @@
+#
 name: linters
 on:
   pull_request:
     types: [opened, reopened, edited]
+
+permissions:
+  contents: read
+
 defaults:
   run:
     shell: bash -o pipefail {0}
+
 jobs:
   lint:
     runs-on: ubuntu-latest
+
+    permissions:
+      security-events: write
+      pull-requests: write
+
     steps:
       - name: Checkout code
         uses: actions/checkout@v1
-      - name: Run Shellcheck
-        uses: azohra/shell-linter@latest
         with:
-          exclude-paths: "sorc/*.fd"
+          fetch-depth: 0
+
+      - name: Lint shell scripts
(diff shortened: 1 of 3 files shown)


aerorahul (approved):
Looks good to me.
And we now have standards.

.github/workflows/linters.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v1
+        with:
+          fetch-depth: 0
+
+      - name: Lint shell scripts
+        uses: redhat-plumbers-in-action/differential-shellcheck@latest
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          shell-scripts: ush/preamble.sh
```
jamacku, line 31: Unfortunately, this will not work as you expect. [`shell-scripts`](https://github.com/redhat-plumbers-in-action/differential-shellcheck#shell-scripts) option takes the path to a text file that contains a list of file paths that will be ignored. I know that it is a bit clumsy, and it would deserve to rework on the linter side.

But I think you could be fine by running a linter with default settings. It performs differential scans, so only new changes introduced via PR will be reported.

I'm glad you are using my action. If you have any issues or requests, feel free to reach out to me or create issue on [`differential-shellcheck`](https://github.com/redhat-plumbers-in-action/differential-shellcheck).

WalterKolczynski-NOAA: Seemed to work when I tested it, but thanks for the information. It was probably just accidental that it did what I expected. I don't know that using a separate file is really all that clumsy at all.

Actually really like your action, thank you for making it available. My only request would be to make a new one that does the same for python linter ;). I'm having trouble finding one quite as well integrated into GitHub actions, to the point I am considering adapting yours for a different linter.

jamacku: I'm glad you like it. :+1: 

We are planning to create also other linters, including `differential-cppcheck` and `differential-pylint`. I can't give you any exact release date, but they are coming. They will be available at: [@redhat-plumbers-in-action](https://github.com/redhat-plumbers-in-action)

PR #1013, opened 2022-09-06
Correct shell linter config

WalterKolczynski-NOAA:
**Description**
Per comment from the developer on PR #1008 after it was merged, the `shell-scripts` option for the shell linter does not do the globbing we expected. However, the linter action also only annotates when syntax issues are *added* (at least, this is my understanding). It will not flag files with existing errors, even if they are otherwise changed. So, immediately processing all shell files shouldn't be an issue (fingers crossed).

Refs: #397

**Type of change**

**How Has This Been Tested?**
N/A?

**Checklist**

### DIFF
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -28,4 +28,3 @@ jobs:
         uses: redhat-plumbers-in-action/differential-shellcheck@latest
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


KateFriedman-NOAA (approved):

AndrewEichmann-NOAA:
@WalterKolczynski-NOAA Does this impact PR https://github.com/NOAA-EMC/global-workflow/pull/674 ?

PR #1016, opened 2022-09-07
Correct diagnostic file handling in ush/ozn_xtrct.sh

### LINKED ISSUE #1004: Correct diagnostic file handling in ush/ozn_xtrct.sh
Testing for PR #674 revealed a problem in ush/ozn_xtrct.sh.  

**Expected behavior**
vrfy should run without error.

**Current behavior**

Diagnostic file names in the oznstat file are not being parsed correctly.  The filename parsing logic is insufficiently robust and fails on orion only.  The problem manifests when the oznmon_time.x executable attempts to open a diagnostic file that doesn't exist, causing the executable to halt.   This is reported in the gdasvrfy.log.

**Machines affected**
Orion

In addition to fixing the file parsing, a check will be added to verify the diagnostic file exists and, if it doesn't, avoid running the executable.

EdwardSafford-NOAA:
**Description**
The OznMon's ush/ozn_xtrct.sh script creates a list of available sat/instrument sources for a given cycle in order to report missing files.  The logic for extracting the list of sources worked on wcoss2 and hera but failed on orion.  The logic used a listing of the diagnostic files which included user name.  That created a parsing problem if the user name was not the expected length.  The solution is to simply do a single line listing (ls -1) and parse the results.  

Additionally the processing logic was condensed and now avoids running the extraction executables in the case of a missing diag file.

Fixes #1004 .

**Type of change**

**How Has This Been Tested?**

1. Cloned global-workflow from fork.
2. Ran checkout.sh, build_all.sh, link_workflow.sh to set up cycled experiment.
3. Set up `$EXPDIR` for rocoto-based experiment.
4. Populated `$ROTDIR` with files needed to run OznMon (oznstat files).
5. Reduced vrfy.sh to only run the oznmon, and reduced workflow to only run gdasvrfy job.
6. Ran on orion and wcoss2.  Expected files were produced on both platforms; no errors occurred on orion.
  
**Checklist**

### DIFF
--- a/ush/ozn_xtrct.sh
+++ b/ush/ozn_xtrct.sh
@@ -232,10 +232,10 @@ EOF
             $NCP ${type}.${ptype}.${PDATE}.ieee_d.${Z} ${TANKverf_ozn}/horiz/
       
 
-            echo "finished processing ptype, type:  $ptype, $type"
+            echo "finished processing ptype, type:  ${ptype}, ${type}"
 
          else
-            echo "diag file for $type.$ptype not found"
+            echo "diag file for ${type}.${ptype} not found"
          fi
 
       done  # type in satype
--- a/ush/ozn_xtrct.sh
+++ b/ush/ozn_xtrct.sh
@@ -102,7 +102,7 @@ ozn_ptype=${ozn_ptype:-"ges anl"}
 #  a problem, reported by an iret value of 2
 #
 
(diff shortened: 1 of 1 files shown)


WalterKolczynski-NOAA:
Please update the title so it actually says what this PR is doing. Compare with the other open PRs.

WalterKolczynski-NOAA (approved):

EdwardSafford-NOAA:
@WalterKolczynski-NOAA Done.

PR #1029, opened 2022-09-22
initial commit for incoming yaml work

aerorahul:
**Description**
The global-workflow in preparation for GFSv17, and GEFSv13 is adding JEDI and coupled pre-processing/model/DA/post-processing tasks. 

This PR:
- adds basic ability to read and process yaml files.  
- adds ability to include and parse nested yaml files
- adds ability to parse arbitrary templates e.g. `ufs-weather-model` regression tests templates, jinja2 templates, environment variables, etc.

A linter and logger will be added shortly when these tools are exercised.

**Type of change**

**How Has This Been Tested?**
- [X] No tasks use this at the moment.  This is a placeholder for work to come.
  
**Checklist**


ush/python/setup.cfg
```diff
@@ -0,0 +1,62 @@
+[metadata]
+name = pygw
+version = 0.0.1
+description = Global applications specific workflow related tools
+long_description = file: README.md
+long_description_content_type = text/markdown
+author = "NOAA/NCEP/EMC"
```
CoryMartin-NOAA: don't forget the National Weather Service!

CoryMartin-NOAA (commented):
One small comment, but looks great overall! Looking forward to changing GDASApp to use this.

ush/python/setup.cfg
aerorahul: done.

CoryMartin-NOAA (approved):

ush/python/pygw/src/pygw/attrdict.py
```diff
@@ -0,0 +1,169 @@
...
+            else:
+                for key, val in iter(arg):
+                    __self[key] = __self._hook(val)
+
+        for key, val in kwargs.items():
+            __self[key] = __self._hook(val)
+
+    def __setattr__(self, name, value):
+        if hasattr(self.__class__, name):
+            raise AttributeError("'AttrDict' object attribute "
```
WalterKolczynski-NOAA, line 37: Use f-string?

ush/python/pygw/src/pygw/attrdict.py
```diff
@@ -0,0 +1,169 @@
...
+            object.__delattr__(self, '__parent')
+            object.__delattr__(self, '__key')
+
+    def __add__(self, other):
+        if not self.keys():
+            return other
+        else:
+            self_type = type(self).__name__
+            other_type = type(other).__name__
+            msg = "unsupported operand type(s) for +: '{}' and '{}'"
```
WalterKolczynski-NOAA, line 65: f-string?

ush/python/pygw/src/pygw/fsutils.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@contextlib.contextmanager
+def chdir(path):
+    cwd = os.getcwd()
+    try:
+        os.chdir(path)
+        yield
+    finally:
+        os.chdir(cwd)
```
WalterKolczynski-NOAA, line 37: Is this necessary to recover from the failed `os.chdir(path)`? Otherwise not seeing the point to changing the directory to the one we are in.

Also, I'm not sure about silently failing to change the directory like this. I think a lot of the time we'd want to throw an exception. But maybe I'm misunderstanding how this would be used.

ush/python/pygw/src/pygw/template.py
```diff
@@ -0,0 +1,191 @@
...
+    try:
+        len(s)
+    except TypeError:
+        return True
+    else:
+        return False
+
+
+def is_single_type_or_string(s):
+    if isinstance(s, str):
```
WalterKolczynski-NOAA, line 184: Could this function more simply just be
`return isinstance(s,str) or is_single_type(s)`

ush/python/pygw/src/pygw/timetools.py
```diff
@@ -0,0 +1,111 @@
+import re
+import datetime
+
+
+__all__ = ["to_datetime", "to_timedelta",
+           "to_YMDH", "to_YMD",
+           "strftime", "strptime"]
```
WalterKolczynski-NOAA, line 8: I had added a function to CROW that also converted a timedelta back to a string for use in rocoto cyclestr offsets. Not sure if would be useful to add here as well:
```
def td_to_HMS(td):
    hours, remainder = divmod(int(td.total_seconds()), 3600)
    minutes, seconds = divmod(remainder, 60)
    return f"{hours:02d}:{minutes:02d}:{seconds:02d}"
```

ush/python/pygw/src/pygw/yaml_file.py
```diff
@@ -0,0 +1,144 @@
...
+    Reads a YAML file as an AttrDict and recursively converts
+    nested dictionaries into AttrDict.
+    This is the entry point for all YAML files.
+    """
+
+    def __init__(self, path=None, data=None):
+        super().__init__()
+        if path:
+            config = parse_yaml(path=path)
+        elif data:
```
WalterKolczynski-NOAA, line 24: Do we want to say something if both are defined (which means the data setting is ignored)?

WalterKolczynski-NOAA (changes requested):
A couple minor things, but also needs [numpy-style](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) docstrings throughout.

ush/python/pygw/src/pygw/attrdict.py
aerorahul: I copied this `attrdict.py` from an external github repo and note it in the documentation up front.
I can update this to f-string, but I am inclined to stay true to the upstream source as much as possible.

ush/python/pygw/src/pygw/fsutils.py
aerorahul: we should throw an error if it cannot `chdir` to the directory.

ush/python/pygw/src/pygw/timetools.py
aerorahul: yes.  I can add that method.  Seems useful.

ush/python/pygw/src/pygw/yaml_file.py
aerorahul: good point.

ush/python/pygw/src/pygw/template.py
aerorahul: trying to stay true to the upstream `template.py` to manage changes easily.

ush/python/pygw/src/pygw/fsutils.py
aerorahul: On second thought, a `WARNING` for now.  We can raise an exception in the future.

WalterKolczynski-NOAA (approved):

WalterKolczynski-NOAA:
Looks good now except documentation

Issue #1028, opened 2022-09-22
Radiance verification fails to find diag files


DavidHuber-NOAA:
**Expected behavior**

The ges diag files should be found and used when performing radiance verification.

**Current behavior**

The lines
`if [[ -f "diag_${type}_ges.${PDATE}*.${Z}" ]]; then`
and
`if [[ -f "diag_${type}_anl.${PDATE}*.${Z}" ]]; then`
in [exgdas_atmos_verfrad.sh](https://github.com/NOAA-EMC/global-workflow/blob/develop/scripts/exgdas_atmos_verfrad.sh#L147-L155) fail to find anything as the use of wildcards in a `-f` conditional are not expanded.

**Machines affected**

All

**To Reproduce**

Run a gdasvrfy job in a cycle in which gdasanal also ran.

**Possible Implementation**

One possible fix is to use ls instead of -f:
```
if ls "diag_${type}_ges.${PDATE}*.${Z}" 1> /dev/null 2>&1; then
  mv diag_${type}_ges.${PDATE}*.${Z} ${type}.${Z}
  ${UNCOMPRESS} ./${type}.${Z}
else
  echo "WARNING: diag_${type}_ges.${PDATE}*.${Z} not available, skipping"
fi
```

WalterKolczynski-NOAA:
Prefer `if (( ls "diag_${type}_ges.${PDATE}*.${Z}" 2> /dev/null | wc -l > 0 )); then` as a solution, as it is more explicit in what it is doing.

PR #1031, opened 2022-09-23
Fix radiance verification failing to find diag files

EdwardSafford-NOAA:
**Description**

Correct exgdas_atmos_verfrad's failure to find diagnostic files.  Additionally, fixed two other problems in child scripts which caused halts and/or failure to move resulting data files to the correct location.  

Fixes issue #1028.

**Type of change**

**How Has This Been Tested?**

1. Set up the necessary data in comrot (radstat and abias files).
2. Modified the workflow xml file to only run the vrfy step.
3. Reduced the vrfy step to only run the radiance monitor.
4. The radiance monitor now correctly locates the diagnostic files.  The 4 child scripts run to completion and produce the expected output in the expected location.

- [x] Cycled test on Orion
  
**Checklist**

### DIFF
--- a/scripts/exgdas_atmos_verfrad.sh
+++ b/scripts/exgdas_atmos_verfrad.sh
@@ -144,7 +144,7 @@ if [[ -s ${radstat} && -s ${biascr} ]]; then
          netcdf=1
       fi
      
-      if [[ -f "diag_${type}_ges.${PDATE}*.${Z}" ]]; then
+      if (( `ls diag_${type}_ges.${PDATE}*.${Z} 2> /dev/null | wc -l` > 0 )); then
         mv diag_${type}_ges.${PDATE}*.${Z} ${type}.${Z}
         ${UNCOMPRESS} ./${type}.${Z}
       else
@@ -152,7 +152,7 @@ if [[ -s ${radstat} && -s ${biascr} ]]; then
       fi
      
       if [[ $USE_ANL -eq 1 ]]; then
-        if [[ -f "diag_${type}_anl.${PDATE}*.${Z}" ]]; then
+        if (( `ls diag_${type}_anl.${PDATE}*.${Z} 2> /dev/null | wc -l` > 0 )); then
           mv diag_${type}_anl.${PDATE}*.${Z} ${type}_anl.${Z}
           ${UNCOMPRESS} ./${type}_anl.${Z}
         else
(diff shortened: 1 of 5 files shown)


scripts/exgdas_atmos_verfrad.sh
```diff
@@ -173,17 +173,17 @@ if [[ -s ${radstat} && -s ${biascr} ]]; then
...
-    ${USHradmon}/radmon_verf_time.sh ${PDATE}
+    ${USHradmon}/radmon_verf_time.sh "${PDATE}"
     rc_time=$?
 
     #--------------------------------------
     #  optionally run clean_tankdir script
     #
     if [[ ${CLEAN_TANKVERF:-0} -eq 1 ]]; then
-       ${USHradmon}/clean_tankdir.sh glb 60
+       "${USHradmon}/clean_tankdir.sh glb 60"
```
WalterKolczynski-NOAA: Only the first argument should be in quotations

scripts/exgdas_atmos_verfrad.sh
```diff
@@ -214,7 +214,7 @@ export CHGRP_CMD=${CHGRP_CMD:-"chgrp ${group_name:-rstprod}"}
 rlist="saphir"
 for rtype in $rlist; do
   if compgen -G "$TANKverf_rad/*${rtype}*" > /dev/null; then
-    ${CHGRP_CMD} $TANKverf_rad/*${rtype}*
+    ${CHGRP_CMD} "$TANKverf_rad/*${rtype}*"
```
WalterKolczynski-NOAA: Globs won't work inside of quotations. Try
`"${TANKverf_rad}/"*${rtype}*`

If the linter still complains at that, use
`"${TANKverf_rad}/"*"${rtype}"*`

The `compgen` commands are different, because they want it as a single argument that it will expand itself (and won't work without them). Same for `find`. But regular shell commands like chmod want the argument expanded into multiple arguments first.

WalterKolczynski-NOAA (changes requested):

scripts/exgdas_atmos_verfrad.sh
EdwardSafford-NOAA: That makes sense, thanks.  Change made.

scripts/exgdas_atmos_verfrad.sh
EdwardSafford-NOAA: The first solution does pass the linter.

DavidHuber-NOAA (approved):
Looks good.  I gave this a test run this morning and it works well for me.  Thanks for the quick turnaround @EdwardSafford-NOAA!

WalterKolczynski-NOAA (approved):

PR #1038, opened 2022-09-29
Rename post scripts

WalterKolczynski-NOAA:
**Description**
Rename post scripts to remove the unnecessary "ncep" qualifier. Also changes the executable name for upp to be the one produced by upp (`upp.x`).

The ush/global_nceppost.sh script is removed as it is not used by any job. We may want to rename the gfs_post.sh script (the one that is used) to global_post.sh to be more reflective of its usage.

**Type of change**
- [x] Maintenance (alteration that does not change functionality)

**How Has This Been Tested?**
- [x] Forecast-only test on Hera
- [x] Cycled test on Hera
  
**Checklist**


KateFriedman-NOAA (approved):
Changes look good! Thanks @WalterKolczynski-NOAA ! Approve pending resolution of linter warnings.

jobs/JGLOBAL_ATMOS_POST
```diff
@@ -1,46 +1,51 @@
...
-for config in $configs; do
-    . $config_path/config.$config
+export EXPDIR=${EXPDIR:-${HOMEgfs}/parm/config}
+config_path=${EXPDIR:-${PACKAGEROOT}/gfs.${gfs_ver}/parm/config}
+for config in ${configs}; do
+    # shellcheck source-path=parm/config
+    . "${config_path}/config.${config}"
     status=$?
-    [[ $status -ne 0 ]] && exit $status
+    (( status != 0 )) && exit "${status}"
```
aerorahul, line 16: Are you sure `(( )) && ` or `(( )) || ` works?
My understanding is that `(( ))` is used when performing arithmetic.  The result of the arithmetic will set the exit code (0 for true).

jobs/JGLOBAL_ATMOS_POST
```diff
@@ -1,46 +1,51 @@
...
 
 ##########################################
 # Source machine runtime environment
 ##########################################
-. $HOMEgfs/env/${machine}.env post
+# shellcheck source-path=env
+. ${HOMEgfs}/env/${machine}.env post
 status=$?
-[[ $status -ne 0 ]] && exit $status
+(( status != 0 )) && exit "${status}"
```
aerorahul, line 25: same question as above.

jobs/JGLOBAL_ATMOS_POST
```diff
@@ -1,46 +1,51 @@
...
-    [[ $status -ne 0 ]] && exit $status
+    (( status != 0 )) && exit "${status}"
 done
 
 
 ##########################################
 # Source machine runtime environment
 ##########################################
-. $HOMEgfs/env/${machine}.env post
+# shellcheck source-path=env
```
aerorahul: are we adding this now to every file?
Can this be set in higher level env?

jobs/JGLOBAL_ATMOS_POST
```diff
@@ -1,46 +1,51 @@
...
+cd "${DATA}" || exit 1
 
 
 ##############################################
 # Run setpdy and initialize PDY variables
 ##############################################
-export cycle="t${cyc}z"
+export cycle="t${cyc:?}z"
 setpdy.sh
+# shellcheck disable=SC1091
```
aerorahul: I am not sure littering the code w/ this type of stuff is a good idea.
Its distracting for one, and its hard to know what are we disabling and why specifically here.

jobs/JGLOBAL_ATMOS_POST
```diff
@@ -68,30 +73,32 @@ export COMPONENT=${COMPONENT:-atmos}
...
 export RERUN=${RERUN:-NO}
 export HOMECRTM=${HOMECRTM:-${PACKAGEROOT:?}/lib/crtm/${crtm_ver:?}}
 export FIXCRTM=${CRTM_FIX:-${HOMECRTM}/fix}
-export PARMpost=${PARMpost:-$HOMEgfs/parm/post}
+export PARMpost=${PARMpost:-${HOMEgfs}/parm/post}
 export INLINE_POST=${WRITE_DOPOST:-".false."}
 
-export COMIN=${COMIN:-$ROTDIR/$RUN.$PDY/$cyc/$COMPONENT}
-export COMOUT=${COMOUT:-$ROTDIR/$RUN.$PDY/$cyc/$COMPONENT}
+export COMIN=${COMIN:-${ROTDIR}/${RUN}.${PDY}/${cyc}/${COMPONENT}}
```
aerorahul, line 78: Is there ever a reason for `COMPONENT` in this script to not be `atmos`?
Line 70 should be:
`export COMPONENT="atmos"`

jobs/JGLOBAL_ATMOS_POST
```diff
@@ -117,9 +124,9 @@ export SLEEP_INT=5
 ###############################################################
 # Run relevant exglobal script
 
-$SCRgfs/ex${RUN}_atmos_nceppost.sh
+"${SCRgfs}/ex${RUN}_atmos_post.sh"
```
aerorahul, line 122: Are the `exgfs_atmos_post.sh` and `exgdas_atmos_post.sh` that different?
Just curious here.

aerorahul (commented):
Several comments:
- Use of `(( ))` at status checks.  See specific comment
- Use of shellcheck statements.  See specific comment
- Some other, inline.

jobs/JGLOBAL_ATMOS_POST
WalterKolczynski-NOAA: Yes. (( )) evaluates in arithmetic context and can handle inequalities: https://mywiki.wooledge.org/ArithmeticExpression

Quick test:
```
hfe02 rename_post/ > if ((  2 > 0 )); then echo "hi"; fi
hi
hfe02 rename_post/ > if ((  0 > 2 )); then echo "hi"; fi                                                                                                  
hfe02 rename_post/ > (( 2 > 0 )) || echo "hi"
hfe02 rename_post/ > (( 0 > 2 )) || echo "hi"                                                                                                             
hi
```

jobs/JGLOBAL_ATMOS_POST
WalterKolczynski-NOAA: No, but we may be able to do without it. I was hoping to get fewer other complaints if shellcheck can source the right files, and I'm not sure it is even helping (especially this one, which I don't think is followed).

jobs/JGLOBAL_ATMOS_POST
WalterKolczynski-NOAA: The latter issue can be solved by explaining in comment, but I was trying to make them minimally intrusive. So far, I've only needed shellcheck directives in a couple places per file. We could turn certain warnings off all the time, but, well, then they'd be off all the time.

jobs/JGLOBAL_ATMOS_POST
WalterKolczynski-NOAA: Not that I know of. I was just updating for the linter.

jobs/JGLOBAL_ATMOS_POST
WalterKolczynski-NOAA: Not too different. Comparing more carefully and merging them is part of the next phase. This PR is supposed to just be renaming the scripts (and then all the linter complaints).

jobs/JGLOBAL_ATMOS_POST
aerorahul: Are we testing here if the arithmetic works or evaluating the expression?
```
(( $status -ne 0 ))
```
returns the status of the arithmetic evaluation of the expression.  So even if the arithmetic was wrong, but evaluated correctly, the return status would be True.
```
[[ $status -ne 0 ]]
```
evaluates the expression and returns a True or False.

In other words, just because it works, is this the right test?

WalterKolczynski-NOAA: Guess I'm not sure I see the difference unless there is a string where a number should be (since a string would be treated as 0 in an arithmetic context).

aerorahul: I suppose it doesn't make a difference in the end.  Just be consistent and make it a rule.
```
(( 2*0 )) && echo "hi"
(( 2*1 )) && echo "hi"
hi
```
Not sure why the first is `False`.  It is a perfectly fine arithmetic expression.

WalterKolczynski-NOAA: I would only use this form as a logical unless it includes a comparison operator like the examples I gave. Really, we would never use `(( var = a + b ))` for math either, as it will return a non-zero exit code most of the time (use `var=$(( a + b ))` instead).

PR #1070, opened 2022-10-13
Resource updates to support WCOSS2

KateFriedman-NOAA:
**Description**

This PR primarily brings back resource updates from testing on WCOSS2. These updates have been tested on Orion and further adjusted as needed. Testing on Hera continues but is ok so far. There are also changes to add `BASE_CPLIC` for WCOSS2.

Changes:

1. pull in some GFSv16.3 adjustments (mainly from NCO reverting memory increases)
2. add WCOSS2 `BASE_CPLIC` to `config.coupled_ic`
3. add `make_nsstbufr` and `make_nsstbufr` to `wcoss2.yaml` hosts file to match S4 updates to other hosts files
4. remove errant `)` in `wcoss2.yaml` hosts file
5. updates to `config.fv3`
* add WCOSS2 `npe_node_max` 
* change `WRTTASK_PER_GROUP` from `$npe_node_max` to 64 for every resolution (half of node size on WCOSS2)
* add check for `WRTTASK_PER_GROUP` to set it back to `$npe_node_max` if 64 is greater than $npe_node_max on machine (Hera/Orion)
* change C192 `nth_fv3` from 2 to 1 (2 fails on WCOSS2 but 1 works everywhere, further examine this on WCOSS2 during optimization work)
* change C384 `WRITE_GROUP` from 1 to 2 (from WCOSS2 GFSv16 port work, works everywhere)
* move C384 from `netcdf_parallel` to `netcdf` (serial) case check at bottom; need to run C384 with serial netcdf on WCOSS2 (needs further investigation)
6. updates to `config.resources`
* add WCOSS2 `npe_node_max`
* add {} around variables
* replace `--exclusive` with `is_exclusive=True`
* update prep job to run exclusively on WCOSS2 but set 40GB memory elsewhere
* pull in WCOSS2 GFSv16 port resource updates:
  * add `_gfs` versions of some job resource variables where needed to define gdas and gfs suite versions of jobs differently
  * adjust some walltimes
  * add `npe_node_cycle` where needed for WCOSS2 `APRUN_CYCLE` launcher commands
  * add memory settings for non-exclusive jobs (setting `is_exclusive=True` otherwise)
  * update some `npe_node_*` values to either be `npe_node_max / thread #` or if greater than `npe_node_max` set it to `npe_node_max` (for R&Ds)
  * reduce `npe_wavepostbndpnt` from 280 to 240
  * increase `npe_wavepostbndpntbll` from 280 to 448
  * reduce `npe_wavepostpnt` from 280 to 200
  * change `npe_wavegempak` from `npe_node_max` to 1
  * change `npe_waveawipsbulls` from `npe_node_max` to 1
  * change `npe_waveawipsgridded` from `npe_node_max` to 1
  * adjust C768 and C384 analysis job resources
  * reduce `npe_analdiag` from 112 to 96 and add note `# Should be at least twice npe_ediag`
  * increase `npe_gldas` from 96 to 112 and change `npe_node_gaussian` to be `npe_node_max / nth_gaussian`
  * increase `npe_post` from 112 to 126 and set `npe_node_post*` values to `npe_post`
  * change `npe_wafsgrib2` from 1 to 18
  * change `npe_wafsgrib20p25` from 1 to 11
  * set `memory_echgres` to 200GB on WCOSS2
  * reduce `npe_ediag` from 56 to 48
  * add WCOSS2 blocks to eupd section, use ops resources for C768 and C384; optimization is needed; also update Hera settings to stop using 40 threads (npe_node_max)
  * reduce `nth_ecen` from 6 to 4 on WCOSS2 and Orion
  * reduce `nth_epos` from 6 to 4 on WCOSS2 and Orion
  * update postsnd resources everywhere
  * reduce awips resources to 1 npe, 1 npe_node, and 1 thread
  * update gempak resources

Resource changes in generated xmls on R&Ds:

* all
  * post job nodes reduced from 10 to 4
  * some ppn values reduced from 40 to a smaller non-npe_node_max value (no node # changes for these jobs though)
  * gdasfcst nodes reduced from 10 to 5
  * gdasepos nodes reduced from 14 to 8 on Orion
* coupled C384
  * gfswavepostpnt nodes reduced from 7 to 4 

Notes: 

1. non-atmos-only and non-coupled jobs were not tested on WCOSS2, downstream ops-only jobs (e.g. gempak/awips) were also not tested for this PR
2. work is still needed for METplus and fit2obs on WCOSS2
3. S2SWA is not supported on WCOSS2 (waiting for mapl)
4. ocnpost does not work on WCOSS2

**Type of change**

Updates for supporting WCOSS2.

**How Has This Been Tested?**

- [x] Clone and Build tests on WCOSS2, Hera, and Orion
- [x] Cycled atmos-only C768C384L127 tests on WCOSS2
- [x] Cycled atmos-only C384C192L127 and C192C96L127 tests on WCOSS2, Hera, Orion
- [x] Forecast-only S2SW on WCOSS2 and Orion

Some continued testing is happening on Hera, Orion, and WCOSS2.
  
Refs #419


parm/config/config.resources
```diff
@@ -67,72 +73,88 @@ elif [ ${step} = "waveinit" ]; then
     export wtime_waveinit="00:10:00"
     export npe_waveinit=12
     export nth_waveinit=1
-    export npe_node_waveinit=$(echo "${npe_node_max} / ${nth_waveinit}" | bc)
+    export npe_node_waveinit=$npe_waveinit
```
WalterKolczynski-NOAA: Won't some of these change if the number of threads is increased later?

WalterKolczynski-NOAA (changes requested):
Confused about these `npe_node_*` changes in general. Values should never be larger than the core size, otherwise you might put multiple ranks on the same CPU.

parm/config/config.resources
```diff
@@ -67,72 +73,88 @@ elif [ ${step} = "waveinit" ]; then
...
     export nth_waveawipsbulls=1
     export npe_node_waveawipsbulls=$(echo "${npe_node_max} / ${nth_waveawipsbulls}" | bc)
     export NTASKS=${npe_waveawipsbulls}
+    export is_exclusive=True
 
 elif [ ${step} = "waveawipsgridded" ]; then
 
-    export wtime_waveawipsgridded="00:30:00"
-    export npe_waveawipsgridded=${npe_node_max}
+    export wtime_waveawipsgridded="02:00:00"
```
aerorahul, line 152: that is a huge increase.

parm/config/config.resources
```diff
@@ -206,13 +229,21 @@ elif [ ${step} = "aeroanlfinal" ]; then
 
 elif [ ${step} = "anal" ]; then
 
-    export wtime_anal="01:00:00"
-    export npe_anal=1000
+    export wtime_anal="00:50:00"
+    export wtime_anal_gfs="00:40:00"
```
aerorahul, line 233: I understand the granularity, but 10m is hardly any thing. So would keeping `gfs` the same as `gdas` have meaningful impact?

aerorahul (commented):
most of the changes look good.
a couple of questions.

parm/config/config.resources
KateFriedman-NOAA: Some of these were brought in from the GFSv16 WCOSS2 port. I reviewed the `npe_node_*` variables and made adjustments. These adjustments did not change the xml resources for a high-res cycled test on WCOSS2 so that's good. The wave jobs aren't run in testing for this PR so we'll need to check them when we turn waves back on in the future. Similar note for the wafs jobs.

I also checked that the xml resources don't change on Hera with these updates. They don't.

I also pulled in recent updates from GFSv16.3 (ecf scripts and memory reductions).

Please review the `config.resources` changes.

parm/config/config.resources
KateFriedman-NOAA: This was a change from GFSv16 ops: https://github.com/NOAA-EMC/global-workflow/blob/operations/ecf/scripts/gfs/wave/post/jgfs_wave_prdgen_gridded.ecf#L6

parm/config/config.resources
KateFriedman-NOAA: In development, no. We'll just need to revisit these walltimes for GFSv17 ops (these are the current ops walltimes). Want me to change the `gfs` value to be 50mins?

WalterKolczynski-NOAA (approved):

parm/config/config.fv3
```diff
@@ -52,9 +54,11 @@ case $case_in in
         export nth_fv3_gfs=2
         export cdmbgwd="0.071,2.1,1.0,1.0"  # mountain blocking, ogwd, cgwd, cgwd src scaling
         export WRITE_GROUP=1
-        export WRTTASK_PER_GROUP=$npe_node_max
+        export WRTTASK_PER_GROUP=64
```
aerorahul, line 57: Can you explain why 64 is hard-wired here?  Isn't `npe_node_max=40` on Orion, in which case this is going to be an issue. No?
I think you are trying to set this so that on WCOSS it won't be requesting 128 cores. Correct?

parm/config/config.fv3
```diff
@@ -79,28 +85,32 @@ case $case_in in
         export layout_y=6
         export layout_x_gfs=4
         export layout_y_gfs=6
-        export nth_fv3=2
+        export nth_fv3=1
         export nth_fv3_gfs=2
         export cdmbgwd="0.23,1.5,1.0,1.0"  # mountain blocking, ogwd, cgwd, cgwd src scaling
         export WRITE_GROUP=1
-        export WRTTASK_PER_GROUP=$npe_node_max
+        export WRTTASK_PER_GROUP=64
```
aerorahul, line 92: I think I must be confused about this.

aerorahul (commented):
I think I need more explanations on the hardwiring of 64 across for all platforms.

parm/config/config.fv3
KateFriedman-NOAA: During the port of GFSv16 to WCOSS2 George V and I found that WRTTASK_PER_GROUP=64 was good for the resolutions I tested. I set `WRTTASK_PER_GROUP=64` for all resolutions as the default but also include a check in the lines after each of these to set it back to `$npe_node_max` (what it was before) when `WRTTASK_PER_GROUP > npe_node_max` (on the other platforms). Example:
```
if [[ "$WRTTASK_PER_GROUP" -gt "$npe_node_max" ]]; then export WRTTASK_PER_GROUP=$npe_node_max ; fi
```
So the values for the R&Ds is preserved and WCOSS2 is considered. The same thing was done in the `dev_v16.2.x` branch:
https://github.com/NOAA-EMC/global-workflow/blob/dev_v16.2.x/parm/config/config.fv3.emc.dyn

parm/config/config.fv3
KateFriedman-NOAA: See comment reply above.

WalterKolczynski-NOAA (approved):

PR #1073, opened 2022-10-17
Update UFS to Sept 9 version

WalterKolczynski-NOAA:
**Description**
Updates the UFS version. This captures the conversion of UFS module from TCL to lua. A couple of the CICE namelist variables are no longer valid in this version, so they are removed.

Due to memory limitations on Hera and the increased memory requirements of GOCART, the number of threads is increased there when running the forecast with aerosols.

Also added a temporary block to delete any existing gocart output files. The ability to clobber files was deactivated a while ago and I got fed up with forecast jobs failing on retry.

**Type of change**
- [x] Maintenance

**How Has This Been Tested?**
- [x] Cycled test on Orion
- [x] Coupled test on Orion
- [x] ATMA test on Hera
- [x] Coupled test on Hera
  
**Checklist**

### DIFF
--- a/modulefiles/modulefile.ww3.orion.lua
+++ b/modulefiles/modulefile.ww3.orion.lua
@@ -22,5 +22,4 @@ load(pathJoin("g2", "3.4.5"))
 load(pathJoin("hdf5", "1.10.6"))
 load(pathJoin("netcdf", "4.7.4"))
 
-load(pathJoin("w3nco", "2.4.1"))
 load(pathJoin("w3emc", "2.9.2"))
--- a/modulefiles/modulefile.ww3.orion.lua
+++ b/modulefiles/modulefile.ww3.orion.lua
@@ -23,3 +23,4 @@ load(pathJoin("hdf5", "1.10.6"))
 load(pathJoin("netcdf", "4.7.4"))
 
 load(pathJoin("w3nco", "2.4.1"))
+load(pathJoin("w3emc", "2.9.2"))
--- a/ush/parsing_namelists_CICE.sh
+++ b/ush/parsing_namelists_CICE.sh
@@ -80,8 +80,6 @@ cat > ice_in <<eof
    restart_FY   = .false.
    tr_lvl       = .true.
    restart_lvl  = .false.
-   tr_pond_cesm = .false.
-   restart_pond_cesm = .false.
    tr_pond_topo = .false.
    restart_pond_topo = .false.
    tr_pond_lvl  = $tr_pond_lvl
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -1,7 +1,7 @@
 # External sub-modules of global-workflow
 
 [UFS]
-tag = Prototype-P8
+tag = 3c3548d
 local_path = sorc/ufs_model.fd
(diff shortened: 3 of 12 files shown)


modulefiles/modulefile.ww3.orion.lua
```diff
@@ -23,3 +23,4 @@ load(pathJoin("hdf5", "1.10.6"))
 load(pathJoin("netcdf", "4.7.4"))
 
 load(pathJoin("w3nco", "2.4.1"))
```
JessicaMeixner-NOAA: I think you can also remove w3nco.  Do you have paths to the output of your coupled model run by any chance I could look at?

WalterKolczynski-NOAA: Pretty sure it wouldn't build without it, which is why I added it. It was probably a month ago at this point, so memory is fuzzy.

`/scratch2/NCEPDEV/ensemble/noscrub/Walter.Kolczynski/global-workflow/update_ufs_cpl`

JessicaMeixner-NOAA: We don't run with it in the standalone WW3 regression tests, which is why I'm pretty sure you can remove it.  Maybe I can run a quick test to see.  It's not a big deal to carry around the extra module but this module list has been trimmed down to not have any extras, which is why I pointed it out.  Thanks for the path!

aerorahul: I think so too.
`w3nco` is no longer required in any library/executable, and if it is, we need to snuff it out.
I know it is being automatically loaded when `nemsio/2.5.2` is loaded.  It was removed from `nemsio/2.5.3+`

If necessary, we can hold this change for a follow up PR since testing for this PR has completed.

ush/parsing_namelists_CICE.sh
```diff
@@ -80,8 +80,6 @@ cat > ice_in <<eof
    restart_FY   = .false.
    tr_lvl       = .true.
    restart_lvl  = .false.
-   tr_pond_cesm = .false.
```
aerorahul, line 83: can you add a comment in the PR description, so this does not seem like trying to sneak in?

aerorahul (approved):
A couple of comments.
As long as the other reviewers are satisfied, this PR can be merged.

ush/parsing_namelists_CICE.sh
WalterKolczynski-NOAA: Done

modulefiles/modulefile.ww3.orion.lua
WalterKolczynski-NOAA: Sorry, my wires were crossed. I was thinking of the thing I added, but this is about removing the thing it basically replaced. Doing a test build just to be sure, but should work.

WalterKolczynski-NOAA: Done. Also changed the modules on wcoss2 and s4 over to w3emc, as those machines weren't in the repo when I first made these changes.

JessicaMeixner-NOAA (approved):
The updates match what I see here: https://github.com/ufs-community/ufs-weather-model/compare/Prototype-P8...3c3548d

PR #1075, opened 2022-10-17
Updates from NCO for GFSv16.3.0

KateFriedman-NOAA:
**Description**

This PR brings back changes from NCO/Carlos which were made in the para/prod installs of `gfs.v16.3.0` on WCOSS2 leading up to the 30-day pre-implementation parallel.

Changes:

1. Add line to `jobs/JGLOBAL_ATMOS_EMCSFC_SFC_PREP` to check if `COMOUT` exists and if not create it.
2. Updates to `run.ver`: a) remove `bufr_ver` (not needed in `run.ver`) and b) update `bufr_dump_ver` to 1.1.2.
3. Revert some memory increases based on a request from NCO (updated release notes as well):

> Hi Everyone,
>We're going to stick with the memory requests from 16.2 with few exceptions; we know this won't get rid of cgroup warning messages and that's OK. Since Kate and Wei worked on the transition many months ago, there has been new information from GDIT, which is essentially to ignore these warnings until they can figure out what's going on with them. They are clearly inaccurate. For example, you've requested jgfs_atmos_gempak_meta to go from 2GB to 200GB; this is logically insane, even though I totally understand where it's coming from -- I'm sure you raised it until the cgroup warnings were gone. But this job shouldn't use more than a couple hundred MB so requesting 200GB is not a path we should go down. We'll fill the system very fast if we get rid of all these warnings by inflating memory allocations to unreasonable levels.
> The reality is that we don't have reliable memory usage statistics on WCOSS2 and we should not tune around bad data. I also realize this leaves us susceptible to failure because we don't know actual memory used from the system.
> Steven Earle

**How Has This Been Tested?**

NCO using changes in pre-implementation testing.


aerorahul (approved):
looks good.

WalterKolczynski-NOAA (approved):

KateFriedman-NOAA:
@aerorahul @WalterKolczynski-NOAA I committed the memory adjustments after getting confirmation from NCO (see statement from Steven Earle in PR description). Please reconfirm your approval of this PR. Thanks!

WalterKolczynski-NOAA:
@KateFriedman-NOAA I would recommend running a full-res test on WCOSS2 with these settings if you haven't already and check the files carefully. @lgannoaa reported during the dev meeting yesterday that the cgroup errors can cause corrupted data when EMC runs it even if the job reports as successful. However, when NCO runs it presumably the additional memory is just provided because the data is fine. (Lin, please correct if this is a misinterpretation of what you said.)

We probably need to merge this into the release branch as-is anyway since this is what NCO is running with, but it needs to be evaluated before merger into the dev_16.x branch. Approving on that basis.

WalterKolczynski-NOAA (approved):

KateFriedman-NOAA:
@WalterKolczynski-NOAA thanks for letting me know what @lgannoaa reported. @lgannoaa can you provide more details on where you see the corruption (e.g. which jobs or output files)? Could you send us an email with those details? Thanks!

> We probably need to merge this into the release branch as-is anyway since this is what NCO is running with, but it needs to be evaluated before merger into the dev_16.x branch.

Agreed! I will merge this into the release branch for now since it's what NCO is using but test before merging it back into the developer-supporting branch.

PR #1078, opened 2022-10-17
Use more cycledefs for task control

### LINKED ISSUE #1003: Archive jobs fail for fourth full cycle when starting with a half-cycle
**Expected behavior**
Archive jobs should complete normally

**Current behavior**
When running cold start, the archive job fails for the fourth full cycle (24 h after the half-cycle).

**Machines affected**
All

**To Reproduce**
Set up a cycled cold-start experiment that runs for more than a day.

**Detailed Description**

Failure is caused near the end of the archive script when it attempts to delete data no longer needed. It is checking the rocoto log(!) to see if the cycle 24 h previous completed successfully. However, the key string that is being searched for, `This cycle is complete: Success` is not written for the half-cycle. This is likely caused by not all jobs running in the half-cycle and no task assigned as the "final" task, so the cycle is never "complete" from a rocoto perspective.

**Possible Implementation**
There are a lot of things going on here that can be fixed. First, there are prerequisites that can be used so the job only runs if the cycle is complete (if it exists) rather than parsing the rocoto log file. Even if that approach is used, it will be unsuccessful unless the half-cycle is marked as complete. This can be done by setting `final="true"` in the task definition, or preferably by specifying a new cycledef for the half-cycle that includes only the jobs to be run (or both for good measure).

WalterKolczynski-NOAA:
**Description**
Splits the existing rocoto cycle definitions up to offer better job control. This means that only the jobs that are due to run will appear in a cycle's job list from rocotostat/rocotoviewer. It also allows for the removal of some of the cycleexist dependencies that were there solely to prevent the job from running in the half cycle. A side effect of this change is that the half-cycle will be recognized as a completed cycle, fixing the bug with archive jobs starting in the fourth cycle (#1003).

The gdas cycledef has been split into a `gdas_half` for the first half- cycle and `gdas` for the other GDAS cycles. Tasks that run during that first half-cycle therefore run on two cycledefs.

For gfs, instead of slicing perpindicular to time, a new cycledef `gfs_cont` (continuity) was created in parallel to the existing gfs cycledef that omits the first cycle. This was done since only one job (`aerosol_init`) currently skips the first cycle, and this prevents the need to provide two cycledefs for every gfs task but one.

Since some time math is now being done on sdate in workflow_xml.py, we now keep those as datetime objects and only convert to string when writing the cycledef strings.

In order to access the pygw utilities in the workflow directory, a symlink is created in `workflow` pointing to the pygw location in `ush`. A better solution may be found in the future.

Fixes #1003

**Type of change**

**How Has This Been Tested?**
- [x] 4½-cycle cold-start C96/48 cycled test on Orion
- [x] 4½-cycle warm-start C384/192 cycled test on Orion
- [x] Coupled prototype test on Orion
- [x] 3-cycle forecast-only C96 ATMA test on Orion (test aerosol_init)
- [ ] GDASapp
  
**Checklist**

### DIFF
--- a/workflow/rocoto/workflow_tasks.py
+++ b/workflow/rocoto/workflow_tasks.py
@@ -310,10 +310,6 @@ class Tasks:
             interval = self._base['INTERVAL']
         offset = f'-{interval}'
 
-        # Previous cycle
-        dep_dict = {'type': 'cycleexist', 'offset': offset}
-        deps.append(rocoto.add_dependency(dep_dict))
-
         # Files from previous cycle
         files = [f'@Y@m@d.@H0000.fv_core.res.nc'] + \
                 [f'@Y@m@d.@H0000.fv_core.res.tile{tile}.nc' for tile in range(1, self.n_tiles + 1)] + \
@@ -326,8 +322,10 @@ class Tasks:
 
         dependencies = rocoto.create_dependency(dep_condition='and', dep=deps)
 
+        cycledef = 'gfs_cont'
         resources = self.get_resource('aerosol_init')
-        task = create_wf_task('aerosol_init', resources, cdump=self.cdump, envar=self.envars, dependency=dependencies)
+        task = create_wf_task('aerosol_init', resources, cdump=self.cdump, envar=self.envars, dependency=dependencies,
+                              cycledef=cycledef)
 
         return task
 
@@ -387,8 +385,6 @@ class Tasks:
         deps = []
         dep_dict = {'type': 'task', 'name': f'{self.cdump}anal'}
         deps.append(rocoto.add_dependency(dep_dict))
-        dep_dict = {'type': 'cycleexist', 'offset': '-06:00:00'}
-        deps.append(rocoto.add_dependency(dep_dict))
         dependencies = rocoto.create_dependency(dep_condition='and', dep=deps)
 
         resources = self.get_resource('analdiag')
@@ -446,8 +442,6 @@ class Tasks:
         deps = []
         dep_dict = {'type': 'task', 'name': f'{self.cdump}atmanalrun'}
         deps.append(rocoto.add_dependency(dep_dict))
-        dep_dict = {'type': 'cycleexist', 'offset': '-06:00:00'}
-        deps.append(rocoto.add_dependency(dep_dict))
         dependencies = rocoto.create_dependency(dep_condition='and', dep=deps)
 
         resources = self.get_resource('atmanalpost')
@@ -459,7 +453,6 @@ class Tasks:
 
         suffix = self._base["SUFFIX"]
         dump_suffix = self._base["DUMP_SUFFIX"]
-        gfs_cyc = self._base["gfs_cyc"]
         dmpdir = self._base["DMPDIR"]
 
         deps = []
@@ -494,8 +487,6 @@ class Tasks:
         deps = []
         dep_dict = {'type': 'task', 'name': f'{self.cdump}aeroanlrun'}
         deps.append(rocoto.add_dependency(dep_dict))
-        dep_dict = {'type': 'cycleexist', 'offset': '-06:00:00'}
-        deps.append(rocoto.add_dependency(dep_dict))
         dependencies = rocoto.create_dependency(dep_condition='and', dep=deps)
 
         resources = self.get_resource('aeroanlfinal')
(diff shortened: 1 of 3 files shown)


WalterKolczynski-NOAA:
Will need some help confirming that the GDASapp job flow is still correct, since I don't know how to test that app myself.

I'd also love some confirmation testing on all the other modes. I tried to hit everything that might matter for these changes.

workflow/rocoto/workflow_tasks.py
```diff
@@ -326,8 +322,10 @@ def aerosol_init(self):
 
         dependencies = rocoto.create_dependency(dep_condition='and', dep=deps)
 
+        cycledef = 'gfs_cont'
```
CoryMartin-NOAA: suggest `ctrl` if this is meaning 'gfs control

aerorahul (approved):
The `gfs_cont` is meant to represent "continuous aerosols".   In other words, it is no longer a "parallel" forecasts, but is now "sequential" since every subsequent CDATE depends on the aerosol backgrounds (restarts) from the forecast valid at the CDATE initialized from the previous CDATE.

In that frame, the continuous aerosols operates like the "cycled" system which is sequential.

CoryMartin-NOAA:
> The `gfs_cont` is meant to represent "continuous aerosols". In other words, it is no longer a "parallel" forecasts, but is now "sequential" since every subsequent CDATE depends on the aerosol backgrounds (restarts) from the forecast valid at the CDATE initialized from the previous CDATE.
> 
> In that frame, the continuous aerosols operates like the "cycled" system which is sequential.

Yeah I guess that is a confusing name then... This is really a temporary thing though so I won't fight it, as the 'aerosol init' will be replaced by aerosol DA soon.

aerorahul:
> > The `gfs_cont` is meant to represent "continuous aerosols". In other words, it is no longer a "parallel" forecasts, but is now "sequential" since every subsequent CDATE depends on the aerosol backgrounds (restarts) from the forecast valid at the CDATE initialized from the previous CDATE.
> > In that frame, the continuous aerosols operates like the "cycled" system which is sequential.
> 
> Yeah I guess that is a confusing name then... This is really a temporary thing though so I won't fight it, as the 'aerosol init' will be replaced by aerosol DA soon.

@CoryMartin-NOAA 
There will still be those who wish to run continuous aerosols without DA.  I can't say why they would want to (perhaps cost savings from not having to run DA), but there will be some who would argue for such capability.  In essence, continuous aerosols is "Identity DA".

CoryMartin-NOAA:
@aerorahul not disagreeing there. I guess my point is, why is it a separate 'GFS' and not just an option to turn on/off?

CoryMartin-NOAA (approved):
`gfs_cont` is less than ideal but otherwise seems fine to me

workflow/rocoto/workflow_tasks.py
WalterKolczynski-NOAA: Changed to `gfs_seq`.

WalterKolczynski-NOAA:
@CoryMartin-NOAA did you test to make sure I didn't break the GDASapp?

CoryMartin-NOAA:
@WalterKolczynski-NOAA I have not yet. To be honest, it is only used by like 2-3 people at the moment so I wouldn't worry too much about breaking GDASApp compatibility at this time, since all of our stuff is definitely still in flux. In my opinion, don't let us hold you up (unless @guillaumevernieres or @RussTreadon-NOAA want to make sure it works before merging)

guillaumevernieres:
> @WalterKolczynski-NOAA I have not yet. To be honest, it is only used by like 2-3 people at the moment so I wouldn't worry too much about breaking GDASApp compatibility at this time, since all of our stuff is definitely still in flux. In my opinion, don't let us hold you up (unless @guillaumevernieres or @RussTreadon-NOAA want to make sure it works before merging)

I'm with @CoryMartin-NOAA , you don't need to worry about the GDASApp stuff.

WalterKolczynski-NOAA:
> > @WalterKolczynski-NOAA I have not yet. To be honest, it is only used by like 2-3 people at the moment so I wouldn't worry too much about breaking GDASApp compatibility at this time, since all of our stuff is definitely still in flux. In my opinion, don't let us hold you up (unless @guillaumevernieres or @RussTreadon-NOAA want to make sure it works before merging)
> 
> I'm with @CoryMartin-NOAA , you don't need to worry about the GDASApp stuff.

Okay. It's a simple change later if anything is wrong, I just hate breaking things (I do enough of that unknowingly).

PR #1080, opened 2022-10-18
First step towards making j-jobs consistent in use from ecflow and rocoto

aerorahul:
**Description**

Several jobs run in development mode do not adhere to EE2 standards.  Sometimes these development jobs do get used for operations at which time there is a rush in getting them to be standardized.
These jobs are also used to run parallel retrospectives and near-real time parallels.

This work is performed with GFSv17 and GEFSv13 as the target implementation.

This PR attempts to standardize the use of several variables as defined in the NCO EE2 standards document.
Specifically;
The following are defined in the job-card:
`DATAROOT`
`cyc`
`HOMEgfs`
`job`
`jobid=$job/$$` or `jobid=$PBS_JOBNAME/$PBS_JOBID` depending on Rocoto/Ops

The following are defined in the j-job:
`pid`
`DATA=DATAROOT/jobid`
`cycle`

This PR also fixes a bug in `exgfs_atmos_post.sh` that is only encountered when `WAFSF=YES` and in the `gdaspostanl` job.  An overzealous linter added extra white-space and put quotes around an argument for `wgrib2`.

**Type of change**

It is not intended to break any frequently used existing features, however testing of several features e.g. WAFS jobs, BUFRSND jobs are untested in `develop`.  It is unlikely these jobs run in `develop` to begin with and fixing them will be done in a future PR.

Several of the development jobs e.g. `vrfy`, `metp`, `prep`, `ocnpost`, `arch`, `earc` do not follow EE2 standards.
Testing was done to ensure these jobs run without failure, however extended testing is necessary to validate these tests as some of these jobs require extended data to be generated.

Future additions of jobs/scripts will be checked for EE2 compliance, regardless of if it is used in development or operations.

**How Has This Been Tested?**
All tests were performed on Orion.
- [X] Cycled test for two and half cycles
- [X] Single prototype S2SW forecast-only test case
  
**Checklist**

- [?] New and existing tests pass with my changes


scripts/exgfs_atmos_post.sh
```diff
@@ -206,7 +206,7 @@
           wafsgrid="latlon 0:1440:0.25 90:721:-0.25"
           ${WGRIB2} "${PGBOUT}" -set_grib_type same -new_grid_winds earth \
           -new_grid_interpolation bilinear -set_bitmap 1 \
-          -new_grid "${wafsgrid}" "${PGBOUT}.tmp"
+          -new_grid ${wafsgrid} "${PGBOUT}.tmp"
```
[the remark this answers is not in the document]
aerorahul: This is fixing a linter recommendation that caused the failure.
Mental note to self: Not all linter warnings are accurate.

jobs/JGDAS_ATMOS_ANALYSIS_DIAG
```diff
@@ -2,7 +2,29 @@
...
+. ./PDY
+
+
+##############################################
+# Determine Job Output Name on System
+##############################################
+export pid=${pid:-$$}
+export pgmout="OUTPUT.${pid}"
+export pgmerr=errfile
+
```
WalterKolczynski-NOAA, line 27: As discussed in the call, I think we can move all this to a jjob header script that is sourced to have one for all j-jobs.

scripts/exgfs_atmos_post.sh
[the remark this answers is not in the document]
WalterKolczynski-NOAA: Add a `# shellcheck disable=SC2086` before this line and a `shellcheck disable=` after to suppress the warning permanently.

scripts/exglobal_atmos_analysis.sh
```diff
@@ -317,7 +317,7 @@ else
 fi
 
 # GSI Fix files
-RTMFIX=${RTMFIX:-${CRTM_FIX}}
+RTMFIX=${CRTM_FIX}
```
WalterKolczynski-NOAA, line 320: If we are going to do this, we should remove the line from `config.base` that sets `${RTMFIX}`, since changing it won't do anything.

workflow/applications.py
```diff
@@ -103,7 +103,7 @@ def __init__(self, configuration: Configuration) -> None:
         self.do_bufrsnd = _base.get('DO_BUFRSND', False)
         self.do_gempak = _base.get('DO_GEMPAK', False)
         self.do_awips = _base.get('DO_AWIPS', False)
-        self.do_wafs = _base.get('DO_WAFS', False)
+        self.do_wafs = _base.get('WAFSF', False)
```
WalterKolczynski-NOAA, line 106: Is there a todo to change this to `DO_WAFS` wherever it is being used so we can change it back to match the other settings?

WalterKolczynski-NOAA (changes requested):
Only big thing is the j-job header idea we discussed on the call. Maybe we can have a footer too, though I think right now it would just be for deleting `${DATA}`.

aerorahul:
will reopen from my fork.
closing.

PR #1131, opened 2022-11-22
Update C48 forecast to run with one thread

### LINKED ISSUE #1129: C48 fails on WCOSS2
**Expected behavior**
`C48` should run on all platforms.

**Current behavior**
`C48` is currently the ensemble resolution test case and fails on WCOSS2.  (All) members successfully complete integration on Hera and Orion.

**Machines affected**
WCOSS2

**To Reproduce**
Try and run a cycled experiment at C96/C48 on WCOSS2.

**Detailed Description**
It is found that the number of threads is set to 2 for C48 and that seems to be causing the failures on WCOSS2.

**Possible Implementation**
Reduce number of threads for C48 to 1 for all platforms unless someone from the modeling teams can address the threading issue.  Running w/ 1 thread should allow development and testing of the global-workflow.

aerorahul:
**Description**
C48 forecasts currently fail on WCOSS2 when running with two threads. Two threads are likely not really needed, so the forecast job is reduced to single-threaded on all platforms, at least until the problem can be corrected.

Also moves the block capping the write block size to the node size outside the resolution-specific block, since that is needed for all resolutions.

Fixes #1129 

**Type of change**

This PR also removes the repeated common logic from each of the case statements.

**How Has This Been Tested?**

Tested as part of other ongoing PR's 
  
**Checklist**

### DIFF
--- a/parm/config/config.fv3
+++ b/parm/config/config.fv3
@@ -50,15 +50,13 @@ case ${case_in} in
         export layout_y=2
         export layout_x_gfs=3
         export layout_y_gfs=2
-        export nth_fv3=2
-        export nth_fv3_gfs=2
+        export nth_fv3=1
+        export nth_fv3_gfs=1
         export cdmbgwd="0.071,2.1,1.0,1.0"  # mountain blocking, ogwd, cgwd, cgwd src scaling
         export WRITE_GROUP=1
         export WRTTASK_PER_GROUP=64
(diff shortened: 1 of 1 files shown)


XianwuXue-NOAA (approved):

WalterKolczynski-NOAA (approved):

WalterKolczynski-NOAA:
Updated the PR title and description to actually have one.

PR #1146, opened 2022-12-05
Allow customizations based on user/configuration

aerorahul:
**Description**

The current configuration is rigid in the global-workflow.  This will remain so for the foreseeable future.
However, some minor tweaks can be made to allow for configuring the applications supported by global workflow to create custom experiment configuration.

This PR:
- allows a user to create and specify a yaml file containing the defaults for their experiments.  Future upgrades could allow for nested inclusions.
- removes hard-wired variables and pulls them up to a yaml.  Some user input variables remain in place.

This PR is driven by the need of the GDASApp developers to allow custom configurations for testing.  The yaml file containing the customizations can live anywhere (even in GDASApp), but should follow and update the `defaults.yaml` introduced in this PR. 

**Type of change**

**How Has This Been Tested?**

`setup_expt.py` has been used to create the experiment directories from `develop` and this branch.
The experiment directories were directly compared and found to be identical.
  
**Checklist**


aerorahul:
`config.aeroanl` is used to demonstrate how this works in providing an updated/customized configuration file from a yaml file containing those intended changes.

workflow/setup_expt.py
```diff
@@ -108,47 +111,51 @@ def fill_EXPDIR(inputs):
     return
 
 
+def update_configs(host, inputs):
+
+    # First update config.base
+    edit_baseconfig(host, inputs)
+
+    yaml_path = inputs.yaml
```
WalterKolczynski-NOAA, line 119: Instead of looking for a specific filename, can the filename be an optional keyword argument to setup_expt.py?

WalterKolczynski-NOAA (approved):
Would like to move away from at-square to jinja at least for our internal stuff, but that would be a different PR. This looks like a good launching point.

Otherwise just one request, but it can be deferred to a future PR.

WalterKolczynski-NOAA:
The [config.defaults.s2sw](https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/config/config.defaults.s2sw) file should be quickly converted to one of these input files after the PR goes through.

workflow/setup_expt.py
aerorahul: ha! `inputs.yaml` is the variable.  The value is `HOMEgfs/parm/config/yaml/defaults.yaml` or anything that the user provides at the command-line via the `--yaml` argument.

WalterKolczynski-NOAA: Duh, there's not quotation marks making this a string. And I somehow missed the new argument, despite looking for it. My brain must still be on leave.

guillaumevernieres (approved):
Code looks good but I did not test.

CoryMartin-NOAA (approved):
Like @guillaumevernieres I did not test but this looks like a great first step! What are people's thoughts on replacing:
```
--resdet 768 --resens 384 --gfs_cyc 4
```
for example with a YAML:
```
base:
  resdet: 768
  resens: 384
  gfs_cyc: 4
```
Will this be the plan or do we not want to get rid of all of those command line args?

aerorahul:
> Like @guillaumevernieres I did not test but this looks like a great first step! What are people's thoughts on replacing:
> 
> ```
> --resdet 768 --resens 384 --gfs_cyc 4
> ```
> 
> for example with a YAML:
> 
> ```
> base:
>   resdet: 768
>   resens: 384
>   gfs_cyc: 4
> ```
> 
> Will this be the plan or do we not want to get rid of all of those command line args?

In principle I like the idea of getting rid of all command-line arguments.
Some thoughts:
- it is convenient to have some e.g. `pslot` as a command-line argument rather than; opening a yaml file, editing out `pslot`, save and then run `setup_expt.py`.   Or we can have an automatic `pslot` generator e.g. `goobly-ga`, `yaba-daba-doo`.  Point is, CLA provides flexibility and ease of use.  However, that can be made optional and the user can provide something from the yaml and will be over-written if the user also provides something from CLA. 

In other words, I don't know and would like to hear user perspectives before deciding.  This is also not something IMO, we should spend time on when there are other pressing matters.

PR #1165, opened 2022-12-09
Remove existing directory if it exists when getic runs

### LINKED ISSUE #1156: Error with getic task if ROTDIR directory not cleaned up
**Expected behavior**
User should not have to clean up $ROTDIR before running getic again after it fails

**Current behavior**
After I ran the getic task and it failed (see issue #1155), the job was resubmitted. The job gave an error when running the following statement.

` mv ${EXTRACT_DIR}/${CDUMP}.${yy}${mm}${dd}/${hh}/* ${ROTDIR}/${CDUMP}.${yy}${mm}${dd}/${hh}
`

mv: cannot move <EXTRACT_DIR>/gfs.20221115/00/* to <ROT_DIR>/gfs.20221115/00/atmos': Directory not empty

**Machines affected**
All

**To Reproduce**
run getic.sh before fixing it for issue #1155 and then resubmit job

lgannoaa:
**Description**
If getic is run with an existing `$ROTDIR` (possibly because it failed the first time), it would fail because the directory already exists. Now any existing directory is deleted before attempting to move the ICs.

Fixes #1156

**Type of change**

### DIFF
--- a/jobs/rocoto/getic.sh
+++ b/jobs/rocoto/getic.sh
@@ -101,7 +101,15 @@ else # Pull chgres cube inputs for cold start IC generation
 fi
 
 # Move extracted data to ROTDIR
-if [ ! -d ${ROTDIR}/${CDUMP}.${yy}${mm}${dd}/${hh}/${COMPONENT} ]; then mkdir -p ${ROTDIR}/${CDUMP}.${yy}${mm}${dd}/${hh}/${COMPONENT} ; fi
+if [ ! -d ${ROTDIR}/${CDUMP}.${yy}${mm}${dd}/${hh}/${COMPONENT} ]; then
+  mkdir -p ${ROTDIR}/${CDUMP}.${yy}${mm}${dd}/${hh}/${COMPONENT}
+else
+  # Exception handling on case if ICS already exist
+  echo "The ICS target directory is not empty. Please backup or remove ${ROTDIR}/${CDUMP}.${yy}${mm}${dd}/${hh}/${COMPONENT}"
+  echo "This job will now set to fail"
+  exit 7
+fi
+
 if [ $gfs_ver = v16 -a $RETRO = "YES" ]; then
   mv ${EXTRACT_DIR}/${CDUMP}.${yy}${mm}${dd}/${hh}/* ${ROTDIR}/${CDUMP}.${yy}${mm}${dd}/${hh}/${COMPONENT}
 else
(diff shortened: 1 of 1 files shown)


aerorahul (approved):

WalterKolczynski-NOAA:
I don't think we want the job to fail if it exists, I think we just want it to overwrite. It already errors now (thanks to `set -e`).

Alternatively, just skip if all the files exists.

@KateFriedman-NOAA thoughts?

KateFriedman-NOAA:
Agreed, don't want this job to fail if the files already exist. I could see either non-failing result though: overwrite or skip. For now, I vote for skip and to have the job exit with success. I believe either or both the init and fcst jobs will fire off when the same files are available (data dependency) anyway.

jobs/rocoto/getic.sh
```diff
@@ -101,7 +101,14 @@ else # Pull chgres cube inputs for cold start IC generation
...
 
 # Move extracted data to ROTDIR
-if [ ! -d ${ROTDIR}/${CDUMP}.${yy}${mm}${dd}/${hh}/${COMPONENT} ]; then mkdir -p ${ROTDIR}/${CDUMP}.${yy}${mm}${dd}/${hh}/${COMPONENT} ; fi
+if [ ! -d ${ROTDIR}/${CDUMP}.${yy}${mm}${dd}/${hh}/${COMPONENT} ]; then
+  mkdir -p ${ROTDIR}/${CDUMP}.${yy}${mm}${dd}/${hh}/${COMPONENT}
+else
+  # Exception handling on case if ICS already exist
+  echo "The ICS target directory is not empty. Exiting job"
+  exit 0
+fi
```
KateFriedman-NOAA: This is good if we only wanna check that the $COMPONENT folder exists and then exit. Thinking about this a bit more, I'm wondering if we should also check for the various IC files within before exiting. Thoughts @WalterKolczynski-NOAA ?

aerorahul:
Can we please keep the discussion in this PR confined to resolving the issue that was opened as a bug and resolve this ASAP?
If further enhancements are desired, feel free to open a issue and discuss it there.

KateFriedman-NOAA (approved):

KateFriedman-NOAA:
Noted @aerorahul . I approve this bug fix. Will discuss further enhancements elsewhere.

jobs/rocoto/getic.sh
aerorahul: I 
```suggestion
if [ -d ${ROTDIR}/${CDUMP}.${yy}${mm}${dd}/${hh}/${COMPONENT} ]; then
    rm -rf ${ROTDIR}/${CDUMP}.${yy}${mm}${dd}/${hh}/${COMPONENT}
fi
mkdir -p ${ROTDIR}/${CDUMP}.${yy}${mm}${dd}/${hh}/${COMPONENT}
```

WalterKolczynski-NOAA (approved):

PR #1168, opened 2022-12-10
Enable checking for python norms and fix violating code

aerorahul:
**Description**

This PR:
- enables checking coding norms on python scripts using `pycodestyle` via github actions
- adds the rules of the norms in `.pycodestyle`
- fixes python scripts that violate the norms using `autopep8`.

If a python script is failing during norm check, it can be fixed with the following command:
```
$> autopep8 -i --max-line-length 160 --ignore E402,W504 /path/to/offending_script.py
```

**Type of change**

**How Has This Been Tested?**

Tested in github actions itself.
  
**Checklist**

### DIFF
--- a/.github/workflows/pynorms.yaml
+++ b/.github/workflows/pynorms.yaml
@@ -1,0 +1,24 @@
+name: Check Norms
+on: [push, pull_request]
+
+jobs:
+  check_norms:
+    runs-on: ubuntu-latest
+    name: Check Python coding norms with pycodestyle
+
+    steps:
+
+    - name: Install dependencies
+      run: |
+        pip install --upgrade pip
+        pip install pycodestyle
+
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        path: global-workflow
+
+    - name: Run pycodestyle
+      run: |
+        cd $GITHUB_WORKSPACE/global-workflow
+        pycodestyle -v --config ./.pycodestyle .
--- a/.github/workflows/pynorms.yaml
+++ b/.github/workflows/pynorms.yaml
@@ -14,7 +14,7 @@ jobs:
         pip install pycodestyle
 
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         path: global-workflow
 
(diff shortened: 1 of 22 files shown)


.pycodestyle
```diff
@@ -0,0 +1,6 @@
+[pycodestyle]
+count = False
+ignore = E226,E401,E402,W504
```
WalterKolczynski-NOAA: Any reason why we are ignoring these four things?

E226 - Whitespace around operator
E401 - Multiple imports on one line
E402 - Import not at top of file
W504 - Line break after binary operator instead of before

.github/workflows/pynorms.yaml
```diff
@@ -0,0 +1,24 @@
+name: Check Norms
+on: [push, pull_request]
```
WalterKolczynski-NOAA, line 2: Should we add `workflow_dispatch` so we can run it manually in GitHub? This didn't make sense for the differential shellcheck, since it requires a PR to diff, but here it would be sensible.

WalterKolczynski-NOAA (changes requested):

.pycodestyle
aerorahul: Not sure.  I just copied these from `emcpy`.
Seems like we should not ignore these (atleast the first 3) as they improve code readability.
I am indifferent to `W504`.
Will fix today.

.github/workflows/pynorms.yaml
aerorahul: Not sure how a manual dispatch would produce a different result from a push or pull_request, as a commit needs to be made to fix any errors.
But sure.

.pycodestyle
aerorahul: fixed.
@CoryMartin-NOAA 
The exgdas scripts are loading `ufsda` and not at the top of the code.  For now we are retaining `E402`, but I suggest we remove that soon.

CoryMartin-NOAA: I hope to remove that sooner rather than later and wrap it into `pygw`.

WalterKolczynski-NOAA (approved):

.github/workflows/pynorms.yaml
WalterKolczynski-NOAA: Deferred to a future PR.

.pycodestyle
WalterKolczynski-NOAA: E402 and W504 deferred to the future.

PR #1179, opened 2022-12-14
Stop checking for restarts on non-GFS CDUMPs

### LINKED ISSUE #1140: Unbound variable (RSTDIR_ATM) in forecast_det.sh (minor bug)
Your bug may already be reported!
Please search on the [Issue tracker](https://github.com/NOAA-EMC/global-workflow/issues) before creating one.

**Expected behavior**

FV3_GFS_det should run without error when called during gdasfcst jobs.  `RSTDIR_ATM` is only a required variable for gfsfcst jobs (set in [forecast_predet.sh:212](https://github.com/NOAA-EMC/global-workflow/blob/e2c29749ffe4050ce4e0273ef4802b8458bbc59c/ush/forecast_predet.sh#L212)) and should not be referenced.

**Current behavior**

`RSTDIR_ATM` is referenced during both gdas and gfs fcst steps, resulting in an unbound error at [forecast_det.sh:43](https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/forecast_det.sh#L43) during gdasfcst.  It appears that this unbound error only affects the `find` command within the command substitution and so flow continues with `filecount=0`, which does not affect on the remainder of the function.  I'm not sure if this is the usual/constant behavior when `set -u` is toggled.

**Machines affected**

Found on Google Cloud.

**To Reproduce**

Run a gdasfcst job and check the resulting log.

**Possible Implementation**

Replace
`filecount=$(find $RSTDIR_ATM -type f | wc -l)`
with
`[[ $CDUMP = "gfs" ]] && filecount=$(find $RSTDIR_ATM -type f | wc -l)`
or
remove `$CDUMP = "gfs"` from the proceeding `if` and move the `filecount` assignment and the proceeding `if` into a new `if [ $CDUMP = "gfs" ]; then` block.

lgannoaa:
**Description**
The forecast was attempting to check `RSTDIR_ATM` for restart files, but that variable is not defined for gdas, resulting in an unbound variable error. Because the error occurred in a subshell, the error had no impact on the execution, just an unwanted error message in the log. Now that directory is only checked when the `$CDUMP` is 'gfs'.

Also fixes some linter warnings.

Fixes #1140 

- [X] Forecast-only test on Hera

### DIFF
--- a/ush/forecast_det.sh
+++ b/ush/forecast_det.sh
@@ -19,55 +19,55 @@ FV3_GFS_det(){
   res_latlon_dynamics="''"
 
   # Determine if this is a warm start or cold start
-  if [ -f $gmemdir/RESTART/${sPDY}.${scyc}0000.coupler.res ]; then
+  if [[ -f "${gmemdir}"/RESTART/"${sPDY}"."${scyc}"0000.coupler.res ]]; then
     export warm_start=".true."
   fi
 
   # turn IAU off for cold start
   DOIAU_coldstart=${DOIAU_coldstart:-"NO"}
-  if [ $DOIAU = "YES" -a $warm_start = ".false." ] || [ $DOIAU_coldstart = "YES" -a $warm_start = ".true." ]; then
+  if [[ "${DOIAU}" = "YES" ]] && [[ ${warm_start} = ".false." ]] || [[ "${DOIAU_coldstart}" = "YES" ]] && [[ ${warm_start} = ".true." ]]; then
     export DOIAU="NO"
-    echo "turning off IAU since warm_start = $warm_start"
+    echo "turning off IAU since warm_start = ${warm_start}"
     DOIAU_coldstart="YES"
     IAU_OFFSET=0
-    sCDATE=$CDATE
-    sPDY=$PDY
-    scyc=$cyc
-    tPDY=$sPDY
-    tcyc=$cyc
+    sCDATE=${CDATE}
+    sPDY=${PDY}
+    scyc=${cyc}
+    tPDY=${sPDY}
+    tcyc=${cyc}
   fi
 
   #-------------------------------------------------------
   # determine if restart IC exists to continue from a previous forecast
   RERUN="NO"
-  filecount=$(find $RSTDIR_ATM -type f | wc -l)
-  if [ $CDUMP = "gfs" -a $rst_invt1 -gt 0 -a $FHMAX -gt $rst_invt1 -a $filecount -gt 10 ]; then
+  [[ ${CDUMP} = "gfs" ]] && filecount=$(find "${RSTDIR_ATM}" -type f | wc -l)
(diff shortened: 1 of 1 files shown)


ush/forecast_det.sh
```diff
@@ -19,55 +19,55 @@ FV3_GFS_det(){
   res_latlon_dynamics="''"
 
   # Determine if this is a warm start or cold start
-  if [ -f $gmemdir/RESTART/${sPDY}.${scyc}0000.coupler.res ]; then
+  if [[ -f "${gmemdir}"/RESTART/"${sPDY}"."${scyc}"0000.coupler.res ]]; then
```
WalterKolczynski-NOAA: Please quote the entire path and not each individual variable.

ush/forecast_det.sh
```diff
@@ -19,55 +19,55 @@ FV3_GFS_det(){
...
+      flag2=${RSTDIR_ATM}/coupler.res
 
       #make sure that the wave restart files also exist if cplwav=true
       waverstok=".true."
-      if [ $cplwav = ".true." ]; then
-        for wavGRD in $waveGRD ; do
-          if [ ! -f ${RSTDIR_WAVE}/${PDYS}.${cycs}0000.restart.${wavGRD} ]; then
+      if [[ "${cplwav}" = ".true." ]]; then
+        for wavGRD in ${waveGRD} ; do
+          if [[ ! -f "${RSTDIR_WAVE}"/"${PDYS}"."${cycs}"0000.restart."${wavGRD}" ]]; then
```
WalterKolczynski-NOAA: Same here. Please quote the entire path.

ush/forecast_det.sh
```diff
@@ -19,55 +19,55 @@ FV3_GFS_det(){
...
 
-      if [ -s $flag1 -a $waverstok = ".true." ]; then
-        CDATE_RST=$SDATE
-        [[ $RERUN = "YES" ]] && break
-        mv $flag1 ${flag1}.old
-        if [ -s $flag2 ]; then mv $flag2 ${flag2}.old ;fi
+      if [[ -s "${flag1}" ]] && [[ ${waverstok} = ".true." ]]; then
+        CDATE_RST=${SDATE}
+        [[ ${RERUN} = "YES" ]] && break
+        mv "${flag1}" "${flag1}".old
```
WalterKolczynski-NOAA: And here. Should be `mv "${flag1}" "${flag1}.old"`

ush/forecast_det.sh
```diff
@@ -19,55 +19,55 @@ FV3_GFS_det(){
...
+        if [[ -s "${flag2}" ]]; then mv "${flag2}" "${flag2}".old ;fi
```
WalterKolczynski-NOAA: And here

WalterKolczynski-NOAA (changes requested):
I know this is how shellcheck "fixes" missing quote strings automatically, but the results are ugly and more difficult to read. It is suggested you fix these warnings manually by quoting entire strings ("canonical" quoting). Just make sure you don't accidentally include any spaces within the quotes that would unintentionally combine multiple arguments into one. We have a feature request in with shellcheck to provide the option of using canonical quotes, but in the meantime we need to do it manually.

WalterKolczynski-NOAA (approved):

PR #1186, opened 2022-12-16
Fix checking for restart files

### LINKED ISSUE #1140: Unbound variable (RSTDIR_ATM) in forecast_det.sh (minor bug)

### LINKED ISSUE #1185: Fix a bug in  ush/forecast_det.sh
**Expected behavior**

The workflow can run forecast jobs for gdas

**Current behavior**

When I run gdasfcst and gdasefmn jobs, they will fail because filecount: unbound variable in ush/forecast_det.sh: line 44

**Machines affected**

WCOSS2

**To Reproduce**

Run gdas or early enkf using the latest develop branch

**Detailed Description**

lfs/h2/emc/ens/noscrub/xianwu.xue/GEFS_v13_da/gw_gefs_da/scripts/../ush/forecast_det.sh: line 44: filecount: unbound variable

XianwuXue-NOAA:
**Description**
Undoes the portion of PR #1179 that caused a new bug while attempting to fix #1140, without removing the linter fixes. Instead `/dev/null` is 'searched' if `${RSTDIR_ATM}` is not defined. That situation will always result in zero files found, ensuring a rerun is not triggered.

Fixes #1140 
Fixes #1185
Moots #1190 

**Type of change**

**How Has This Been Tested?**
- [x] Cycled with early EnKF on Orion 
- [x] Full coupled on Orion
- [x] Forecast-only on Orion
  
**Checklist**

### DIFF
--- a/ush/forecast_det.sh
+++ b/ush/forecast_det.sh
@@ -40,7 +40,7 @@ FV3_GFS_det(){
   #-------------------------------------------------------
   # determine if restart IC exists to continue from a previous forecast
   RERUN="NO"
-  filecount=$(find "${RSTDIR_ATM}" -type f | wc -l)
+  filecount=$(find "${RSTDIR_ATM:-/dev/null}" -type f | wc -l)
   if [ ${CDUMP} = "gfs" -a ${rst_invt1} -gt 0 -a ${FHMAX} -gt ${rst_invt1} -a ${filecount} -gt 10 ]; then
     reverse=$(echo "${restart_interval[@]} " | tac -s ' ')
     for xfh in ${reverse} ; do
--- a/ush/forecast_det.sh
+++ b/ush/forecast_det.sh
@@ -40,7 +40,7 @@ FV3_GFS_det(){
   #-------------------------------------------------------
   # determine if restart IC exists to continue from a previous forecast
   RERUN="NO"
-  [[ ${CDUMP} = "gfs" ]] && filecount=$(find "${RSTDIR_ATM}" -type f | wc -l)
+  filecount=$(find "${RSTDIR_ATM}" -type f | wc -l)
   if [ ${CDUMP} = "gfs" -a ${rst_invt1} -gt 0 -a ${FHMAX} -gt ${rst_invt1} -a ${filecount} -gt 10 ]; then
     reverse=$(echo "${restart_interval[@]} " | tac -s ' ')
     for xfh in ${reverse} ; do


WalterKolczynski-NOAA:
This undoes a fix I just merged this morning: #1179 

I suspect this is only a problem with early-cycle enkf, which (I think) uses gfs as the `$CDUMP`. That line should short-circuit false before reaching `filecount` for gdas.

What we need is a more direct solution to the original issue this morning's PR was supposed to fix (`RSTDIR_ATM` being unbound). Use this:

`filecount=$(find "${RSTDIR_ATM:-/dev/null}" -type f | wc -l)`

XianwuXue-NOAA:
> This undoes a fix I just merged this morning: #1179
> 
> I suspect this is only a problem with early-cycle enkf, which (I think) uses gfs as the `$CDUMP`. That line should short-circuit false before reaching `filecount` for gdas.
> 
> What we need is a more direct solution to the original issue this morning's PR was supposed to fix (`RSTDIR_ATM` being unbound). Use this:
> 
> `filecount=$(find "${RSTDIR_ATM:-/dev/null}" -type f | wc -l)`

Good method. In fact, it will not affect  not only the early enkf, but also the late enkf. I just push a new comment based on your suggestions.

WalterKolczynski-NOAA (approved):
Looks good once all the testing is complete

XianwuXue-NOAA:
@WalterKolczynski-NOAA Tested and confirmed the changes

CoryMartin-NOAA:
This is not just a problem with early cycle EnKF, I just found a similar issue with the half cycle gdasfcst run.

XianwuXue-NOAA:
> This is not just a problem with early cycle EnKF, I just found a similar issue with the half cycle gdasfcst run.

@CoryMartin-NOAA Thanks a lot for your checking. Yes, same in my test. So this PR should fix the issue.

aerorahul:
The offending line here was added as part of the PR #1179 
In the PR description, it says that a forecast-only test was performed.
This section of the script is used also by the cycled mode.
Please revert PR #1179  and re-open issue #1140 
The PR #1179 may be resolving the particular issue in #1140 but is causing other issues.

lgannoaa:
The test with #1140 was done with forecast-only. However, the filecount variable in question was not initialized. That caused problem when running gdas and enkf. I have a fix and is in testing now.
Will update issue 1140 and submit PR as needed.

WalterKolczynski-NOAA:
I'm able to run cycled (with early cycle), fo, and coupled all successfully using this PR, with no mention of any unbound variable in the forecast logs.

XianwuXue-NOAA:
> I'm able to run cycled (with early cycle), fo, and coupled all successfully using this PR, with no mention of any unbound variable in the forecast logs.

Thanks. I ran 1.5 cycles with early enkf C768/C384 successfully

Issue #1187, opened 2022-12-16
Fix 'DEBUG' option in build_ufs.sh


mdtoyNOAA:
**Expected behavior**

When compiling the ufs_model with sorc/build_ufs.sh and the "-d" option, the model should be compiled with "--DEBUG=ON".

**Current behavior**

Line 16 of the code assigns "BUILD_TYPE" to "Debug" (mixed case) when option "-d" passed in, however, the logic test on line 31 tests for BUILD_TYPE = "DEBUG" (all caps), so the debug mode can never be invoked.

**Machines affected**

All machines.

**To Reproduce**

This is a trivial problem.  Easy to reproduce.

**Detailed Description**

The fix is trivial and it is to change "Debug" to "DEBUG" in line 16 of sorc/build_ufs.sh.

**Note**
I will submit a PR regarding this fix.

PR #1188, opened 2022-12-16
Fixed 'DEBUG' option in build_ufs.sh

mdtoyNOAA:
**Description**
When compiling the ufs_model with sorc/build_ufs.sh and the "-d" option, the model should be compiled with "--DEBUG=ON".   Line 16 of the code assigns "BUILD_TYPE" to "Debug" (mixed case) when option "-d" passed in, however, the logic test on line 31 tests for BUILD_TYPE = "DEBUG" (all caps), so the debug mode can never be invoked.

Fixes Issue [#1187](https://github.com/NOAA-EMC/global-workflow/issues/1187) 

**Type of change**

**How Has This Been Tested?**
I tested the change to the script on Hera.  Code now compiles in DEBUG mode using "-d" option.
  
**Checklist**

### DIFF
--- a/sorc/build_ufs.sh
+++ b/sorc/build_ufs.sh
@@ -13,7 +13,7 @@ source "${cwd}/ufs_model.fd/tests/module-setup.sh"
 
 while getopts ":da:v" option; do
   case "${option}" in
-    d) BUILD_TYPE="Debug";;
+    d) BUILD_TYPE="DEBUG";;
     a) APP="${OPTARG}" ;;
     v) export BUILD_VERBOSE="YES";;
     :)


sorc/build_ufs.sh
```diff
@@ -13,7 +13,7 @@ source "${cwd}/ufs_model.fd/tests/module-setup.sh"
 
 while getopts ":da:v" option; do
   case "${option}" in
-    d) BUILD_TYPE="Debug";;
+    d) BUILD_TYPE="DEBUG";;
```
aerorahul, line 16: I think this should be kept `Debug` (as that is the recognized `CMAKE_BUILD_TYPE` and an edit be made on line 31 as:
```
[[ ${BUILD_TYPE:-"Release"} = "Debug" ]] && MAKE_OPT+=" -DDEBUG=ON"
```

mdtoyNOAA: > I think this should be kept `Debug` (as that is the recognized `CMAKE_BUILD_TYPE` and an edit be made on line 31 as:
> 
> ```
> [[ ${BUILD_TYPE:-"Release"} = "Debug" ]] && MAKE_OPT+=" -DDEBUG=ON"
> ```

Whichever way is better works for me.  I had just noticed that in the other "sorc/build*.sh" scripts that the equivalent to line 16 always has "DEBUG" in all caps, i.e.,  "d) export BUILD_TYPE="DEBUG";;"

I'll stand by until I hear from the other reviewers about which way to go.

aerorahul: Thanks for checking.
In the interest of getting this in, we can do a consistency review at a later time.

aerorahul (approved):

sorc/build_ufs.sh
mdtoyNOAA: It seems that this case only effects the internals of the "build_ufs.sh" script, and wouldn't have an effect anywhere else.  But that's based on my limited knowledge of the workflow as a whole.

PR #1106, opened 2022-11-02
Add in initial 3DVar aerosol DA cycling capability

CoryMartin-NOAA:
**Description**
This PR adds an initial 3DVar aerosol DA cycling capability to support scientific development and testing towards an operationally viable candidate system for aerosol DA for GFSv17/GEFSv13.

This PR includes the following:

- Three new j-jobs and ex-scripts for aeroanlinit, aeroanlrun, and aeroanlfinal
- modifies the rocoto scripts that call the above j-jobs to actually call them
- makes modification to the config.resources and config.aeroanl files
- Introduces an object-oriented python structure to initialize/finalize the aerosol analysis with the intention of eventually using the top-level classes for other analysis jobs (soca, atm, land)

This (at least for now) closes #982 

**Type of change**

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

**How Has This Been Tested?**

Use app `ATMA`, and ensure that `nens=0`, and `resdet==resens==96` when setting up your experiment. I used a start date of `2019061400` through `2019061500` because I have sample obs staged for those cycles.

ICs are in `/scratch1/NCEPDEV/da/Cory.R.Martin/GEFS-Aero/ics_20190614/00/atmos/INPUT/` for C96

Until the fix files are updated, one will have to do two things:
- point to a new `fix/gdas/fv3jedi` symbolic link (see `/scratch1/NCEPDEV/da/Cory.R.Martin/GEFS-Aero/fix_dev/fv3jedi` on Hera)
- The `config.aeroanl` already points to @andytangborn 's C96 BUMP files on Orion, this path will need to be modified to `/scratch1/NCEPDEV/da/Andrew.Tangborn/JEDI/fv3-bundle_aero_mzp2/work_c96/Data/staticb_aero` for Hera, this will move to the default fix directory before merged in but will require coordination with @KateFriedman-NOAA for this and the above bullet

There are a few minor things that have to be changed to get this to work fully after running the setup_exp script.
- Change $DMPDIR in config.base (`/scratch1/NCEPDEV/da/Cory.R.Martin/GEFS-Aero/glopara_dump` on Hera, `/work2/noaa/da/cmartin/GEFS-Aero/glopara_dump/` on Orion)
- Until g-w is updated, change `config.resources` to use less than 96 tasks for the post jobs
- `export DOHYBVAR="NO"`, `export l4densvar=".false." `, `export lwrite4danl=".false."` in config.base
[...]

PR #1192, opened 2022-12-18
Add tests for configuration.py

aerorahul:
**Description**
This PR:
- extracts out the casting of strings to known datatypes into a method
- adds several tests for the public methods in `configuration.py`.

A note on the two disabled tests:
Two tests `test_parse_config1` and `test_parse_config2` are disabled.
These tests fail in the github runner.
The code in `configuration.py`, calls `_get_shell_env` which executes in a subprocess.  For some reason, it is executed in `/bin/sh` instead of `/bin/bash`.  `source` is only available in `bash`, and not in `sh`.  If `source` is replaced w/ `.`, this code will succeed in the runner.
On the localhost, these tests run as expected even with `source`, so I think this is something peculiar w/ the runner.
`bash` is available, but `subprocess` is opening a `sh` subprocess.

This was identified in the work when passing `os.environ` to the task e.g. PR #1106 e.g. line https://github.com/NOAA-EMC/global-workflow/blob/bae934f307d9565271f875c31484bc9e759e4666/scripts/exgdas_global_aero_analysis_initialize.py#L12 Currently, all values of the keys passed by `os.environ` will be strings.  Casting them as known datatypes can significantly improve the user experience.

Note:
2 tests for `parse_config` are failing in GH runner, but they pass on localhost.  Currently just marking them to skip till the issue with the runner is identified.

**Type of change**

**How Has This Been Tested?**
- [x] Generated experiment directories and XML for various combinations of `cycled`+`ATM` and `forecast-only`+`ATM, S2SWA`.  The resulting experiment directories and XML's were identical.
  
**Checklist**

### DIFF
--- a/ush/python/pygw/src/pygw/configuration.py
+++ b/ush/python/pygw/src/pygw/configuration.py
@@ -2,15 +2,14 @@ import glob
 import os
 import random
 import subprocess
-from datetime import datetime
 from pathlib import Path
 from pprint import pprint
 from typing import Union, List, Dict, Any
-from dateutil.parser import parse, ParserError
 
 from pygw.attrdict import AttrDict
+from pygw.timetools import to_datetime
 
-__all__ = ['Configuration', 'cast_as_known_dtype']
+__all__ = ['Configuration', 'cast_as_dtype', 'cast_strdict_as_dtypedict']
 
 
 class ShellScriptException(Exception):
@@ -80,11 +79,7 @@ class Configuration:
         if isinstance(files, (str, bytes)):
             files = [files]
         files = [self.find_config(file) for file in files]
-        varbles = AttrDict()
-        for key, value in self._get_script_env(files).items():
-            varbles[key] = cast_as_known_dtype(value)
-
-        return varbles
+        return cast_strdict_as_dtypedict(self._get_script_env(files))
 
     def print_config(self, files: Union[str, bytes, list]) -> None:
         """
@@ -127,7 +122,26 @@ class Configuration:
         return varbls
 
 
-def cast_as_known_dtype(string: str):
+def cast_strdict_as_dtypedict(ctx: Dict[str, str]) -> Dict[str, Any]:
+    """
+    Environment variables are typically stored as str
+    This method attempts to translate those into datatypes
+    Parameters
+    ----------
+    ctx : dict
+          dictionary with values as str
+    Returns
+    -------
+    varbles : dict
+              dictionary with values as datatypes
+    """
+    varbles = AttrDict()
+    for key, value in ctx.items():
+        varbles[key] = cast_as_dtype(value)
+    return varbles
+
+
+def cast_as_dtype(string: str) -> Union[str, int, float, bool, Any]:
     """
     Cast a value into known datatype
     Parameters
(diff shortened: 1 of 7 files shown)


ush/python/pygw/src/pygw/configuration.py
```diff
@@ -137,16 +121,59 @@ def _get_shell_env(scripts: List) -> Dict[str, Any]:
...
         except ValueError:
-            return value
+            return string
 
-    @staticmethod
-    def _true_or_not(value):
+    def _true_or_not(string: str):
         try:
-            return value.lower() in Configuration.TRUTHS
+            return string.lower() in TRUTHS
```
WalterKolczynski-NOAA, line 167: We're kinda treating the lists inconsistently. `TRUTHS` gets a single case format and all inputs are converted to lower for comparison, but `BOOLS` gets a number of allowed capitalizations (upper, lower, title) and has to match one of them. Personally prefer the former, so typos like `TRue` would match, but fine with either choice.

ush/python/pygw/src/pygw/timetools.py
```diff
@@ -9,12 +9,15 @@
...
 _DATETIME_RE = re.compile(
-    r"(?P<year>\d{4})(-)?(?P<month>\d{2})(-)?(?P<day>\d{2})(T)?(?P<hour>\d{2})?(:)?(?P<minute>\d{2})?(:)?(?P<second>\d{2})?(Z)?")
+    r"(?P<year>\d{4})(-)?(?P<month>\d{2})(-)?(?P<day>\d{2})"
+    r"(T)?(?P<hour>\d{2})?(:)?(?P<minute>\d{2})?(:)?(?P<second>\d{2})?(Z)?")
 
 _TIMEDELTA_HOURS_RE = re.compile(
-    r"(?P<sign>[+-])?((?P<days>\d+)[d])?(T)?((?P<hours>\d+)[H])?((?P<minutes>\d+)[M])?((?P<seconds>\d+)[S])?(Z)?")
+    r"(?P<sign>[+-])?"
+    r"((?P<days>\d+)[d])?(T)?((?P<hours>\d+)[H])?((?P<minutes>\d+)[M])?((?P<seconds>\d+)[S])?(Z)?")
 _TIMEDELTA_TIME_RE = re.compile(
```
WalterKolczynski-NOAA, line 18: These aren't new, but while I'm noticing it: these two constants are named `HOURS` and `TIME`, but they both take in hours, minutes, and seconds (in different formats). Are there better names to use (or maybe a list of regex under one variable)?

WalterKolczynski-NOAA (commented):
I know this is still draft, but a couple notes.

ush/python/pygw/src/pygw/timetools.py
aerorahul: I don't see why not. 
Do you mind if I hold this for another PR where I add some tests for `timetools.py` and then consolidate this?

ush/python/pygw/src/pygw/configuration.py
aerorahul: I understand that, but trying to account and anticipate for "user-error" and correct it is not usually considered safe practice (Don't remember where I read that).  Instead, I would be inclined to add an exception stating that `FAlse` or `trUe` is not an acceptable value.

Also note how these (bools) are cast.
The `if` statement checks if the `string` is in `BOOLS` (Upper and lowercase, as well as the Titlecase).  If the condition is satisfied, the `string` is converted to `lower` to match if it is `True` if the `string` is indeed `True`.

To extend `BOOLS` to account for the typos, would mean to extend `BOOLS` to:
```
BOOLS +=  ['yEs', 'yES', 'yeS', 'YeS']
```
And many more for `True` and `False`, `.TRUE.', `.FALSE'.

ush/python/pygw/src/pygw/timetools.py
WalterKolczynski-NOAA: No objection here.

ush/python/pygw/src/pygw/configuration.py
WalterKolczynski-NOAA: Still feels weird that if you call `_true_or_not()` directly on `"TRue"`, you'll get `True`, but if you check whether it is a boolean by looking for it in `BOOLS`, it isn't. Even if the current implementation only calls `_true_or_not()` after first checking `BOOLS`.

aerorahul: Note I have just moved the methods out of the `class Configuration` to make use of them outside.
Feel free to open a PR to improve/expand the use of typo-ed booleans.

aerorahul:
@CoryMartin-NOAA 
You should be able to use the method from `configuration.py`, `cast_strdict_as_dtypedict` in your PR #1106 as follows:

```
pygfs.task.aero_analysis.AerosolAnalysis(cast_strdict_as_dtypedict(os.environ))
```

It should greatly simply the use and access of the `config` in your scripts.

WalterKolczynski-NOAA (approved):
When you get a chance, can you add instructions somewhere in the wiki on how to run all the tests manually?

aerorahul:
> When you get a chance, can you add instructions somewhere in the wiki on how to run all the tests manually?

I have updated the `README.md` in the `pygw` to add instructions on how to run the tests manually.
I have tested them on Orion by loading the module `python/3.7.5`.
I don't think I can update the wiki.  Feel free to add these instructions to the appropriate part of the wiki.

Issue #1147, opened 2022-12-06
Add GEFS functions in fcst ush scripts


XianwuXue-NOAA:
**Description**

There are some functions for GEFS, however, some of the functions are missing and the others are empty. This issue is to add the missed functions for GEFS

**Requirements**

**Acceptance Criteria (Definition of Done)**

**Dependencies**

XianwuXue-NOAA:
Duplicated with the #1203

PR #1203, opened 2023-01-03
Make needed updates to run forecast from GEFS

XianwuXue-NOAA:
**Description**

Since some of the GEFS functions are empty or missing, this PR is to complete/add such functions and also make some variables to accept values from other scripts

Refs: #1147

**Type of change**

**How Has This Been Tested?**

Tested on WCOSS2 using GEFS workflow
  
**Checklist**


parm/config/config.resources
```diff
@@ -737,7 +737,7 @@ elif [ ${step} = "esfc" ]; then
 elif [ ${step} = "efcs" ]; then
 
     if [ ${CASE} = "C768" ]; then
-      export wtime_efcs="01:20:00"
+      export wtime_efcs="06:00:00"
```
WalterKolczynski-NOAA, line 740: This is a big change, and only for one resolution. Seems like there should be a `wtime_efcs_gfs` (and possibly other (`_gfs` settings) now to cover early cycle?

ush/forecast_postdet.sh
```diff
@@ -519,11 +520,19 @@ EOF
       logi=logf${FH3}
       pgbi=GFSPRS.GrbF${FH2}
       flxi=GFSFLX.GrbF${FH2}
-      atmo=$memdir/${CDUMP}.t${cyc}z.atmf${FH3}.$affix
-      sfco=$memdir/${CDUMP}.t${cyc}z.sfcf${FH3}.$affix
-      logo=$memdir/${CDUMP}.t${cyc}z.logf${FH3}.txt
-      pgbo=$memdir/${CDUMP}.t${cyc}z.master.grb2f${FH3}
-      flxo=$memdir/${CDUMP}.t${cyc}z.sfluxgrbf${FH3}.grib2
+      if [[ $RUN == "gefs" ]]; then
```
WalterKolczynski-NOAA: This should be temporary until GFS is updated to new COM structure.

ush/forecast_postdet.sh
```diff
@@ -534,11 +543,19 @@ EOF
     done
   else
     for n in $(seq 1 $ntiles); do
-      eval $NLN nggps2d.tile${n}.nc       $memdir/nggps2d.tile${n}.nc
-      eval $NLN nggps3d.tile${n}.nc       $memdir/nggps3d.tile${n}.nc
-      eval $NLN grid_spec.tile${n}.nc     $memdir/grid_spec.tile${n}.nc
-      eval $NLN atmos_static.tile${n}.nc  $memdir/atmos_static.tile${n}.nc
-      eval $NLN atmos_4xdaily.tile${n}.nc $memdir/atmos_4xdaily.tile${n}.nc
+      if [[ $RUN == "gefs" ]]; then
```
WalterKolczynski-NOAA: This too

ush/forecast_postdet.sh
```diff
@@ -556,6 +573,20 @@ FV3_GFS_nml(){
   echo SUB ${FUNCNAME[0]}: FV3 name lists and model configure file created
 }
 
+FV3_GEFS_nml(){
```
WalterKolczynski-NOAA: This is identical to FV3_GFS_nml, why is it needed?

ush/forecast_postdet.sh
```diff
@@ -599,14 +630,22 @@ data_out_GFS() {
           $NCP $file $memdir/RESTART/$file
         done
       fi
-    elif [ $CDUMP = "gfs" ]; then
-      $NCP $DATA/input.nml $ROTDIR/${CDUMP}.${PDY}/${cyc}/atmos/
+    elif [ $CDUMP = "gfs" ] || [ ${RUN} = "gefs" ]; then
```
WalterKolczynski-NOAA: Instead of the nested if, use an additional `elif`.

ush/forecast_predet.sh
```diff
@@ -241,13 +241,19 @@ FV3_GFS_predet(){
     rprefix=enkf$rCDUMP
     memchar=mem$(printf %03i $MEMBER)
   fi
-  memdir=$ROTDIR/${prefix}.$PDY/$cyc/atmos/$memchar
+  memdir=${memdir:-$ROTDIR/${prefix}.$PDY/$cyc/atmos/$memchar}
   if [ ! -d $memdir ]; then mkdir -p $memdir; fi
+  if [[ $RUN == "gefs" ]]; then
+    if [ ! -d ${memdir}/history ]; then mkdir -p ${memdir}/history; fi
```
WalterKolczynski-NOAA: With `mkdir -p`, you shouldn't need to check if the directory already exists first.

XianwuXue-NOAA:
@aerorahul @WalterKolczynski-NOAA Though there still have some warnings after the lint ShellCheck, I will not fix them because those warnings are not the modification in this PR. if you think I should fix the warnings, please let me know.

ush/forecast_postdet.sh
WalterKolczynski-NOAA: There are also a bunch of "GEFS" functions added that just call the GFS version. Is there a need to do that?

WalterKolczynski-NOAA (changes requested):

ush/forecast_postdet.sh
XianwuXue-NOAA: Yes.

ush/forecast_postdet.sh
XianwuXue-NOAA: yes

ush/forecast_postdet.sh
XianwuXue-NOAA: I added these functions is based on the exglobal_forecast.sh. Such as: Since the function should be similar with GFS functions, I just reuse the GFS functions. 

case $RUN in
  'data') DATM_nml;;
  'gfs') FV3_GFS_nml;;
  'gdas') FV3_GFS_nml;;
  'gefs') FV3_GEFS_nml;;
esac				#no namelist for data atmosphere

ush/forecast_postdet.sh
XianwuXue-NOAA: I think when the GFS new COM is ready, we don't need the "if [ ${RUN} = "gefs" ]; then" conditional. So this is only a temporary solution.

ush/forecast_postdet.sh
WalterKolczynski-NOAA: But `gfs` and `gdas` already both call `FV3_GFS_nml`, why can't `gefs`?

ush/forecast_predet.sh
XianwuXue-NOAA: I added the check to avoid run mkdir if it is already exists. Of course, it doesn't matter for "mkdir -p". I can delete the check if you want.

parm/config/config.resources
XianwuXue-NOAA: In fact, using "01:20:00" is only good for coarse resolution, if we want to run C768/C384, then the walltime is too short. Therefore, I increase the wall time. I agree that we may have an another variable to cover early cycle. However, in the early-enkf, we may not run gfsemn jobs since it will be run GEFSv13.

ush/forecast_postdet.sh
WalterKolczynski-NOAA: Still think it should be a separate elif. We'll recombine later if necessary (really, I think copying the `input.nml` can just be done for all).

ush/forecast_postdet.sh
XianwuXue-NOAA: Good insight. I can delete FV3_GEFS_nml to use FV3_GFS_nml

ush/forecast_postdet.sh
WalterKolczynski-NOAA: Maybe we should just make a `$HISTORY_DIR` and a `$MASTER_DIR` and set appropriately instead of all these if-blocks.

ush/forecast_postdet.sh
XianwuXue-NOAA: agree that copying the input.nml is done for all. If separate it, I should be in the else, because the if conditional use $CDUMP, but for GEFS, it should use $RUN to identify it is for GEFS. What do you think?

ush/forecast_postdet.sh
XianwuXue-NOAA: I agree, however, it will add more variables. If you agree, I can do this changes.

ush/forecast_postdet.sh
```diff
@@ -599,14 +600,18 @@ data_out_GFS() {
...
+      $NCP $DATA/input.nml ${ROTDIR}/${RUN}.${PDY}/$cyc/${mem:-""}/atmos/
```
XianwuXue-NOAA: In this change, I use ${RUN}.${PDY} instead of ${CDUMP}.${PDY} based on the NCO's standard. I also add the ${mem:-""} to deal with GEFS data.

XianwuXue-NOAA:
@WalterKolczynski-NOAA @aerorahul I just did the changes to remove the history and master folders. Please review my changes.

scripts/exgdas_enkf_fcst.sh
```diff
@@ -108,7 +108,9 @@ export LEVS=${LEVS_ENKF:-${LEVS:-64}}
 
 # nggps_diag_nml
 export FHOUT=${FHOUT_ENKF:-3}
-
+if [[ $CDUMP == "gfs" ]]; then
+    export FHOUT=${FHOUT_ENKF_GFS:-${FHOUT_ENKF:${FHOUT:-3}}}
+fi
```
aerorahul: Instead, how about this on the line above?
```suggestion
export FHOUT=${FHOUT_ENKF_${CDUMP}:-${FHOUT_ENKF:-${FHOUT:-3}}}
```

ush/forecast_predet.sh
```diff
@@ -266,6 +272,14 @@ FV3_GFS_predet(){
   echo "SUB ${FUNCNAME[0]}: pre-determination variables set"
 }
 
+FV3_GEFS_predet(){
+  echo "SUB ${FUNCNAME[0]}: Defining variables for FV3GFS"
```
aerorahul:
```suggestion
  echo "SUB ${FUNCNAME[0]}: Defining variables for FV3GEFS"
```

scripts/exgdas_enkf_fcst.sh
XianwuXue-NOAA: $CDUMP should be "gfs", not "GFS", should I change FHOUT_ENKF_GFS to FHOUT_ENKF_gfs?

ush/forecast_postdet.sh
WalterKolczynski-NOAA: You got rid of this one, but there are still a few others that weren't removed.

scripts/exgdas_enkf_fcst.sh
aerorahul: That makes sense.

ush/forecast_det.sh
```diff
@@ -41,7 +41,7 @@ FV3_GFS_det(){
   # determine if restart IC exists to continue from a previous forecast
   RERUN="NO"
   filecount=$(find "${RSTDIR_ATM:-/dev/null}" -type f | wc -l)
-  if [ ${CDUMP} = "gfs" -a ${rst_invt1} -gt 0 -a ${FHMAX} -gt ${rst_invt1} -a ${filecount} -gt 10 ]; then
+  if [[ ( ${CDUMP} = "gfs" || ( ${RUN} = "gefs" && ${CDATE_RST} = "" )) && ${rst_invt1} -gt 0 && ${FHMAX} -gt ${rst_invt1} && ${filecount} -gt 10 ]]; then
```
aerorahul, line 44: Is there a way to use `$RUN` or `$CDUMP` consistently.
I think replacing `$CDUMP` with `$RUN` would work here.
What is `$CDATE_RST`?

workflow/rocoto/workflow_tasks.py
```diff
@@ -1259,9 +1259,10 @@ def epos(self):
         def _get_eposgroups(epos):
             fhmin = epos['FHMIN_ENKF']
             fhmax = epos['FHMAX_ENKF']
+            fhout = epos['FHOUT_ENKF']
             if self.cdump == "gfs":
                 fhmax = epos['FHMAX_ENKF_GFS']
-            fhout = epos['FHOUT_ENKF']
+                fhout = epos['FHOUT_ENKF_GFS']
```
aerorahul, line 1265: If you make it `FHOUT_ENKF_gfs` in `config.base`, be use to make the same change here.

aerorahul (commented):
see some comments.

ush/forecast_det.sh
XianwuXue-NOAA: Because CDUMP can only be gfs or gdas in g-w, so I use RUN to determine this is for GEFS based on the exglobal_forecast.sh. 

For $CDATE_RST, is for RERUN=YES. See https://github.com/NOAA-EMC/global-workflow/blob/1a7f0746a78b04cf9e92ef49aca40a81d3d91f42/scripts/exglobal_fcst_nemsfv3gfs.sh#L158. I will do more research on it.

scripts/exgdas_enkf_fcst.sh
XianwuXue-NOAA: To be consistent, I think using FHOUT_ENKF_GFS is better than FHOUT_ENKF_gfs, because we don't have FHOUT_ENKF_GDAS variable. So I think I don't need to change it.

workflow/rocoto/workflow_tasks.py
XianwuXue-NOAA: Thanks. As mentioned in the previous answer, I think it is better to use FHOUT_ENKF_GFS since there is no variable FHOUT_ENKF_gdas. And also consistent with other variables, such as FHMAX_ENKF_GFS. How do you think?

ush/forecast_det.sh
XianwuXue-NOAA: CDATE_RST is to set the restart CDATE, especially for the  forecast_d16_35. Then forecast_d16_35 can run restart from FHMAXH instead of from FHMAXH-$restart_interval.

WalterKolczynski-NOAA (changes requested):

XianwuXue-NOAA:
@WalterKolczynski-NOAA @aerorahul I think I have modified the PR based on your comments. Please find time to review it. I think maybe I can add 'export rCDUMP="gdas"' to this PR to fix the bug if you agree.

ush/forecast_predet.sh
```diff
@@ -208,8 +208,8 @@ FV3_GFS_predet(){
   print_freq=${print_freq:-6}
 
   #-------------------------------------------------------
-  if [ $CDUMP = "gfs" ] && [ $rst_invt1 -gt 0 ] && [ $MEMBER -lt 0 ]; then
-    RSTDIR_ATM=${RSTDIR:-$ROTDIR}/${CDUMP}.${PDY}/${cyc}/atmos/RERUN_RESTART
+  if [[ $CDUMP = "gfs" || $RUN = "gefs" ]] && [ $rst_invt1 -gt 0 ]; then
+    RSTDIR_ATM=${RSTDIR_ATM:-${RSTDIR:-$ROTDIR}/${CDUMP}.${PDY}/${cyc}/atmos/RERUN_RESTART}
```
WalterKolczynski-NOAA: Do we need to have both `$RSTDIR` or `$RSTDIR_ATM`?

scripts/exgdas_enkf_fcst.sh
```diff
@@ -108,7 +108,9 @@ export LEVS=${LEVS_ENKF:-${LEVS:-64}}
 
 # nggps_diag_nml
 export FHOUT=${FHOUT_ENKF:-3}
-
+if [[ $CDUMP == "gfs" ]]; then
```
WalterKolczynski-NOAA: Any new code should comply with shellcheck (always use `{ }`).

ush/forecast_predet.sh
```diff
@@ -208,8 +208,8 @@ FV3_GFS_predet(){
   print_freq=${print_freq:-6}
 
   #-------------------------------------------------------
-  if [ $CDUMP = "gfs" ] && [ $rst_invt1 -gt 0 ] && [ $MEMBER -lt 0 ]; then
-    RSTDIR_ATM=${RSTDIR:-$ROTDIR}/${CDUMP}.${PDY}/${cyc}/atmos/RERUN_RESTART
+  if [[ $CDUMP = "gfs" || $RUN = "gefs" ]] && [ $rst_invt1 -gt 0 ]; then
```
WalterKolczynski-NOAA: More `${ }` needed

ush/forecast_predet.sh
```diff
@@ -232,7 +232,7 @@ FV3_GFS_predet(){
 
   #-------------------------------------------------------
   # member directory
-  if [ $MEMBER -lt 0 ]; then
+  if [[ $MEMBER -lt 0 || $RUN = "gefs" ]]; then
```
WalterKolczynski-NOAA: And here

ush/forecast_predet.sh
```diff
@@ -241,13 +241,13 @@ FV3_GFS_predet(){
     rprefix=enkf$rCDUMP
     memchar=mem$(printf %03i $MEMBER)
   fi
-  memdir=$ROTDIR/${prefix}.$PDY/$cyc/$memchar/atmos
+  memdir=${memdir:-$ROTDIR/${prefix}.$PDY/$cyc/$memchar/atmos}
```
WalterKolczynski-NOAA: And here

ush/forecast_predet.sh
```diff
@@ -241,13 +241,13 @@ FV3_GFS_predet(){
...
   if [ ! -d $memdir ]; then mkdir -p $memdir; fi
 
   GDATE=$($NDATE -$assim_freq $CDATE)
   gPDY=$(echo $GDATE | cut -c1-8)
   gcyc=$(echo $GDATE | cut -c9-10)
-  gmemdir=$ROTDIR/${rprefix}.$gPDY/$gcyc/$memchar/atmos
+  gmemdir=${gmemdir:-$ROTDIR/${rprefix}.$gPDY/$gcyc/$memchar/atmos}
```
WalterKolczynski-NOAA: And here

WalterKolczynski-NOAA (changes requested):

ush/forecast_predet.sh
XianwuXue-NOAA: I didn't find "RSTDIR=" in the g-w repo. However, I am not sure whether some others need it or not. So I keep it. If you think RSTDIR can be removed, then I will remove it.

scripts/exgdas_enkf_fcst.sh
XianwuXue-NOAA: Done

ush/forecast_predet.sh
XianwuXue-NOAA: Done

ush/forecast_predet.sh
XianwuXue-NOAA: Done

ush/forecast_predet.sh
XianwuXue-NOAA: Done

ush/forecast_predet.sh
XianwuXue-NOAA: Done.

scripts/exgdas_enkf_fcst.sh
```diff
@@ -108,7 +108,9 @@ export LEVS=${LEVS_ENKF:-${LEVS:-64}}
 
 # nggps_diag_nml
 export FHOUT=${FHOUT_ENKF:-3}
-
+if [[ ${CDUMP} == "gfs" ]]; then
```
XianwuXue-NOAA, line 111: This checking is for early cycle enkf.

XianwuXue-NOAA:
@WalterKolczynski-NOAA I did the changes by add {}. Thanks

XianwuXue-NOAA:
@WalterKolczynski-NOAA @aerorahul Do you have any comments about this PR?

ush/forecast_predet.sh
WalterKolczynski-NOAA: I don't see it either. Just remove it and only use RSTDIR_ATM

XianwuXue-NOAA: Done

WalterKolczynski-NOAA (approved):
Approved pending additional testing. Please confirm things still work for a cycled experiment using global-workflow.

XianwuXue-NOAA:
> Approved pending additional testing. Please confirm things still work for a cycled experiment using global-workflow.

I have tested 1.5 cycles of early cycle enkf and 1 cycle of GEFS workflow, everything works fine.

XianwuXue-NOAA:
@WalterKolczynski-NOAA @aerorahul I pushed one comment to avoid create invalid link for the final restart files.

ush/forecast_postdet.sh
```diff
@@ -134,7 +129,8 @@ EOF
     #.............................
 
   else ## cold start
-    for file in $(ls $memdir/INPUT/*.nc); do
+    ICSDIR=${ICSDIR:-${memdir}/INPUT}         # cold start initial conditions
```
WalterKolczynski-NOAA: I'm not sure if this is going to work or not. `config.base` sets `ICSDIR` with a value provided by `setup_exp.py`, but if one isn't provided, the value is `None`, to the fallback would never be used here. This will cause issues if people stage ICs manually. Plus, `setup_exp.py` should've already copied the ICs to the `INPUT` directory already.

WalterKolczynski-NOAA (changes requested):

ush/forecast_postdet.sh
XianwuXue-NOAA: The changes is to allow users can use different ICSDIR, otherwise, we mush put the ICS into the ${memdir}/INPUT.  Or I can use **ICSIDR_ATM** to be consistent with **RSTDIR_ATM**. How do you think?

XianwuXue-NOAA: If user does not provide the ICSDIR, then ICSDIR will use default value: ${memdir}/INPUT

XianwuXue-NOAA: I tested the forecast in GEFS repo to set ICSDIR, and it works well.

XianwuXue-NOAA: @aerorahul @WalterKolczynski-NOAA I just removed the ICSDIR.

WalterKolczynski-NOAA (approved):

aerorahul (approved):

PR #1218, opened 2023-01-06
Add missing AWIPS links

WalterKolczynski-NOAA:
**Description**
The mkgfsawps.x and overgridid.x executables were not being linked by the link script, but they are necessary to run AWIPS jobs.

This fix alone is not sufficient to restore AWIPS functionality.

Fixes #1217

**Type of change**
- [x] Bug fix (non-breaking change which fixes an issue)

**How Has This Been Tested?**
- [x] Clone, build, and link on Orion
  
**Checklist**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing tests pass with my changes

PR #1216, opened 2023-01-06
Increase wallclock for diag jobs

WalterKolczynski-NOAA:
**Description**
Diag jobs were failing due to insufficient wall clock, so the wall clock is increased until a more complete review of the resources can be completed.

This is part of a package of PRs that were tested together as part of a j-job refactor.
    
Refs #1215

**Type of change**
- [x] Bug fix (non-breaking change which fixes an issue)

**How Has This Been Tested?**
- [x] Cycled test on Orion
- [x] Cycled test on Hera
- [x] Cycled test on WCOSS

**Checklist**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing tests pass with my changes
- [ ] Any dependent changes have been merged and published

PR #1214, opened 2023-01-06
Use correct resources for GFS gempak

WalterKolczynski-NOAA:
**Description**
The GFS gempak job was using the GDAS settings instead of switching to the GFS settings for `npe` and `npe_node`.

This is part of a package of PRs that were tested together as part of a j-job refactor.

Fixes: #1213

**Type of change**
- [x] Bug fix (non-breaking change which fixes an issue)

**How Has This Been Tested?**
- [x] Cycled test on Orion
- [x] Cycled test on Hera
- [x] Cycled test on WCOSS2
  
**Checklist**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing tests pass with my changes
- [ ] Any dependent changes have been merged and published

** Dependencies **
This is part of a package of PRs that were tested together as part of a j-job refactor:
#1212

PR #1212, opened 2023-01-06
Fix post sounding job

WalterKolczynski-NOAA:
**Description**
A few minor errors were blocking the sounding job from completing successfully:
- CFP APRUN command was using incorrect number of PEs
- Updates were needed to run on Orion. Removed the machine-specific code for Hera and Jet so it runs the same on all machines, while refactoring it to use C-style loop.
- Bash base errors (see also #1195), which are now corrected, while refactoring to use C-style loop.

This is part of a package of PRs that were tested together as part of a j-job refactor.

Refs #1195
Fixes #1211

**Type of change**
- [x] Bug fix (non-breaking change which fixes an issue)

**How Has This Been Tested?**
- [x] Cycled test on Orion
- [x] Cycled test on Hera
- [x] Cycled test on WCOSS2

**Checklist**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing tests pass with my changes
- [ ] Any dependent changes have been merged and published

PR #1230, opened 2023-01-06
Abstract common j-job tasks

WalterKolczynski-NOAA:
**Description**
Takes all of the tasks that are common to all j-jobs and abstracts them out into a shared script that is sourced by each job:
- Set and create $DATA directory
- Call setpy and set $cycle
- Set pid, pgmout, and pgmerr
- Source config files
- Source machine environment file

The common j-job header script is called by passing the job name for the `${machine}.env` files using the `-e` option, and a list of config files to source with the `-c` option.
```
${HOMEgfs}/ush/jjob_header.sh -e [env_job] -c "[config1 [config2 [...]]]"
```
The job name argument (`-e`) is mandatory, and the config list is optional but recommend to always use as well.

Some pre j-job rocoto entry scripts (`jobs/rocoto/*`) are currently doing much more than they should be. These sometimes required extra finagling, usually pre-calling the jjob header in the rocoto script before it does something.

Refs: #1069 

**Type of change**

**How Has This Been Tested?**
- [x] Cycled test on Orion
- [x] Cycled test on Hera
- [x] Cycled test on WCOSS2
- [x] Coupled test on Orion
- [x] Coupled test on Hera
- [x] Couple test on WCOSS2

Post sounding (bufr) and gempak jobs complete successfully
AWIPS and WAFS failing for unrelated reasons
GDAS App jobs not tested
Some j-jobs are not called in the development workflow. TODOs an corresponding issues have been added to these.

**Checklist**

**Dependencies**
#1212
#1214
#1216
#1218

### DIFF
--- a/jobs/JGDAS_ATMOS_ANALYSIS_DIAG
+++ b/jobs/JGDAS_ATMOS_ANALYSIS_DIAG
@@ -1,49 +1,7 @@
 #! /usr/bin/env bash
 
 source "${HOMEgfs}/ush/preamble.sh"
-
-##############################################
-# make temp directory
-##############################################
-export DATA=${DATA:-${DATAROOT}/${jobid}}
-mkdir -p ${DATA}
-cd ${DATA}
-
-
-##############################################
-# Run setpdy and initialize PDY variables
-##############################################
-export cycle="t${cyc}z"
-setpdy.sh
-. ./PDY
-
-
-##############################################
-# Determine Job Output Name on System
-##############################################
-export pid=${pid:-$$}
-export pgmout="OUTPUT.${pid}"
-export pgmerr=errfile
-
-
-#############################
-# Source relevant config files
-#############################
-export EXPDIR=${EXPDIR:-${HOMEgfs}/parm/config}
-configs="base anal analdiag"
-for config in ${configs}; do
-    . ${EXPDIR}/config.${config}
-    status=$?
-    [[ ${status} -ne 0 ]] && exit ${status}
-done
-
-
-##########################################
-# Source machine runtime environment
-##########################################
-. ${HOMEgfs}/env/${machine}.env anal
-status=$?
-[[ ${status} -ne 0 ]] && exit ${status}
+source "${HOMEgfs}/ush/jjob_header.sh" base anal analdiag
 
 
 ##############################################
--- a/jobs/JGDAS_ATMOS_ANALYSIS_DIAG
+++ b/jobs/JGDAS_ATMOS_ANALYSIS_DIAG
@@ -1,7 +1,7 @@
 #! /usr/bin/env bash
 
 source "${HOMEgfs}/ush/preamble.sh"
-source "${HOMEgfs}/ush/jjob_header.sh" base anal analdiag
+source "${HOMEgfs}/ush/jjob_header.sh" -e "anal" -c "base anal analdiag"
 
 
 ##############################################
--- a/jobs/JGDAS_ATMOS_GLDAS
+++ b/jobs/JGDAS_ATMOS_GLDAS
@@ -1,7 +1,6 @@
 #! /usr/bin/env bash
 
 source "${HOMEgfs:?}/ush/preamble.sh"
-source "${HOMEgfs}/ush/jjob_header.sh" base gldas
 
 if [[ "${cyc:?}" -ne "${gldas_cyc:?}" ]]; then
     echo "GLDAS only runs for ${gldas_cyc} cycle; Skip GLDAS step for cycle ${cyc}"
@@ -16,6 +15,7 @@ if [[ "${CDATE}" -le "$(${NDATE:?} +"${xtime}" "${SDATE:?}")" ]]; then
     exit 0
 fi
 
+source "${HOMEgfs}/ush/jjob_header.sh" -e "gldas" -c "base gldas"
 
 ##############################################
 # Set variables used in the exglobal script
--- a/jobs/JGDAS_ATMOS_GLDAS
+++ b/jobs/JGDAS_ATMOS_GLDAS
@@ -1,9 +1,11 @@
 #! /usr/bin/env bash
 
 source "${HOMEgfs:?}/ush/preamble.sh"
+source "${HOMEgfs}/ush/jjob_header.sh" -e "gldas" -c "base gldas"
 
 if [[ "${cyc:?}" -ne "${gldas_cyc:?}" ]]; then
     echo "GLDAS only runs for ${gldas_cyc} cycle; Skip GLDAS step for cycle ${cyc}"
+    rm -Rf "${DATA}"
     exit 0
 fi
 
@@ -12,10 +14,10 @@ xtime=$((gldas_spinup_hours+12))
 if [[ "${CDATE}" -le "$(${NDATE:?} +"${xtime}" "${SDATE:?}")" ]]; then
     echo "GLDAS needs fluxes as forcing from cycles in previous ${xtime} hours"
     echo "starting from ${SDATE}. This gldas cycle is skipped"
+    rm -Rf "${DATA}"
     exit 0
 fi
 
-source "${HOMEgfs}/ush/jjob_header.sh" -e "gldas" -c "base gldas"
 
 ##############################################
 # Set variables used in the exglobal script
--- a/jobs/JGDAS_ATMOS_GLDAS
+++ b/jobs/JGDAS_ATMOS_GLDAS
@@ -1,42 +1,7 @@
 #! /usr/bin/env bash
 
 source "${HOMEgfs:?}/ush/preamble.sh"
-
-##############################################
-# Obtain unique process id (pid) and make temp directory
-##############################################
-export DATA=${DATA:-${DATAROOT}/${jobid}}
-mkdir -p "${DATA}"
-cd "${DATA}" || exit 1
-
-
-##############################################
-# Run setpdy and initialize PDY variables
-##############################################
-export cycle="t${cyc}z"
-setpdy.sh
-. ./PDY
-
-
-##############################################
-# Determine Job Output Name on System
-##############################################
-export pid=${pid:-$$}
-export pgmout="OUTPUT.${pid}"
-export pgmerr=errfile
-
-
-#############################
-# Source relevant config files
-#############################
-export EXPDIR=${EXPDIR:-${HOMEgfs}/parm/config}
-configs="base gldas"
-for config in ${configs}; do
-    . ${EXPDIR}/config.${config}
-    status=$?
-    [[ ${status} -ne 0 ]] && exit "${status}"
-done
-
+source "${HOMEgfs}/ush/jjob_header.sh" base gldas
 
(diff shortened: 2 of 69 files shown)


WalterKolczynski-NOAA:
First wave of testing passes on all machines. Pulled in develop changes and running a last test on Orion where I'm going to compare output against develop.

WalterKolczynski-NOAA:
Output confirmed identical to develop. Had to revert PR #1158 locally on both to get the analysis to run successfully, at least with the chosen date.

jobs/JGDAS_ATMOS_ANALYSIS_DIAG
```diff
@@ -1,49 +1,7 @@
...
-    . ${EXPDIR}/config.${config}
-    status=$?
-    [[ ${status} -ne 0 ]] && exit ${status}
-done
-
-
-##########################################
-# Source machine runtime environment
-##########################################
-. ${HOMEgfs}/env/${machine}.env anal
```
aerorahul, line 44: This job is passing `anal` as the argument for sourcing `$machine.env` file.  In the `jjob_header.sh`, it is going to use `analdiag` as that is set in the rocoto job `analdiag.sh`. There is no entry for `analdiag` in `$machine.env`, atleast not in HERA.env

jobs/JGDAS_ATMOS_GLDAS
```diff
@@ -1,42 +1,7 @@
...
-#############################
-export EXPDIR=${EXPDIR:-${HOMEgfs}/parm/config}
-configs="base gldas"
-for config in ${configs}; do
-    . ${EXPDIR}/config.${config}
-    status=$?
-    [[ ${status} -ne 0 ]] && exit "${status}"
-done
-
+source "${HOMEgfs}/ush/jjob_header.sh" base gldas
```
aerorahul: This is going to create the `DATA` directory.
If `gldas` is going to be skipped, we should remove the empty `DATA` directory as well.

jobs/JGDAS_ATMOS_VERFOZN
```diff
@@ -4,36 +4,12 @@
...
-. ./PDY
-
-
-####################################
-# Determine Job Output Name on System
-####################################
-export pid=$$
-export pgmout="OUTPUT.${pid}"
-export pgmerr=errfile
+source "${HOMEgfs}/ush/jjob_header.sh" base vrfy
```
aerorahul: This job does not source any configs.  We are now forcing it to source these files that may or may not be applicable for this job.
Likewise, it is going to source vrfy section in $machine.env

jobs/JGDAS_ATMOS_ANALYSIS_DIAG
WalterKolczynski-NOAA: Not in `ORION.env` either. Must not be important, since it completes and produces bit-identical results.

jobs/JGDAS_ENKF_DIAG
```diff
@@ -1,49 +1,7 @@
...
-done
-
-
-##########################################
-# Source machine runtime environment
-##########################################
-. ${HOMEgfs}/env/${machine}.env eobs
-status=$?
-[[ ${status} -ne 0 ]] && exit ${status}
+source "${HOMEgfs}/ush/jjob_header.sh" base anal eobs analdiag ediag
```
aerorahul: LHS uses `eobs` in $machine.env. 
RHS will be using ediag.

jobs/JGDAS_ATMOS_ANALYSIS_DIAG
WalterKolczynski-NOAA: (That's not a reason not to add an entry for analdiag, just maybe it should be empty?)

jobs/JGDAS_GLOBAL_OCEAN_ANALYSIS_POST
```diff
@@ -1,41 +1,18 @@
 #!/bin/bash
 export STRICT="NO"
 source "${HOMEgfs}/ush/preamble.sh"
+WIPE_DATA="NO"
+DATA="${DATA:-${DATAROOT}/${RUN}ocnanal_${cyc}}"
```
aerorahul: why can't this be:
DATA="${DATAROOT}/${RUN}ocnanal_${cyc}}"
There is no one upstream setting `DATA`

WalterKolczynski-NOAA: It can be. I just copied what was there before.

aerorahul (changes requested):
In spirit, this is a neat feature.
However, there are quite a few realizations where standardization likely hurts in the transparency of the jobs.

jobs/JGDAS_ATMOS_VERFOZN
WalterKolczynski-NOAA: It did, at least in development, but it was being done in the rocoto pre-job (`jobs/rocoto/vrfy.sh`) (and still is, for now). Added it here to be consistent with other j-jobs, since the job should be broken up so that only one j-job is called per rocoto task.

jobs/JGDAS_ATMOS_GLDAS
WalterKolczynski-NOAA: Moved to after the cycle checks.

jobs/JGDAS_ATMOS_ANALYSIS_DIAG
WalterKolczynski-NOAA: Refactored the jjob header to take in the job name to pass to the environment script.

jobs/JGDAS_ENKF_DIAG
WalterKolczynski-NOAA: Refactored the jjob header to take in the job name to pass to the environment script.

ush/jjob_header.sh
```diff
@@ -0,0 +1,114 @@
...
+        *)
+            echo "FATAL [${BASH_SOURCE[0]}]: Unrecognized option: ${option}"
+            exit 1
+            ;;
+    esac
+done
+shift $((OPTIND-1))
+
+if [[ -z ${env_job} ]]; then
+    echo "FATAL [${BASH_SOURCE[0]}]: Must specify a job name with -e"
```
aerorahul, line 62: I think an `exit 1` is missing after the echo.

jobs/JGDAS_ATMOS_GLDAS
WalterKolczynski-NOAA: Had to move it back because the script needs variables set in the gldas config files to determine whether to skip. Instead added removal of `$DATA` if aborting.

ush/jjob_header.sh
WalterKolczynski-NOAA: Fixed.

WalterKolczynski-NOAA:
Testing complete after addressing reviewer comments. Output still identical.

jobs/JGFS_ATMOS_AWIPS_20KM_1P0DEG
```diff
@@ -1,30 +1,11 @@
...
-#########################################################
-export DATA=${DATA:-${DATAROOT}/${jobid}}
-mkdir -p ${DATA}
-cd ${DATA}
 
-######################################
-# Set up the cycle variable
-######################################
-export cycle="t${cyc}z"
+export OMP_NUM_THREADS=${OMP_NUM_THREADS:-1}
```
aerorahul, line 7: shouldn't this be in the ex-script?

jobs/JGFS_ATMOS_AWIPS_G2
```diff
@@ -1,35 +1,17 @@
...
-export cycle="t${cyc}z"
+source "${HOMEgfs}/ush/preamble.sh"
+source "${HOMEgfs}/ush/jjob_header.sh" -e "awips" -c "base awips"
 
-###########################################
-# Run setpdy and initialize PDY variables
-###########################################
-setpdy.sh
-. ./PDY
+export OMP_NUM_THREADS=${OMP_NUM_THREADS:-1}
```
aerorahul, line 10: same here.

jobs/rocoto/awips.sh
```diff
@@ -94,14 +83,14 @@ for fhr in $fhrlst; do
             done
             
 	    export fcsthrs=$fhr3
-	    export job="jgfs_awips_f${fcsthrs}_20km_${cyc}"
-	    export DATA="${DATAROOT}/$job"
+	    # export job="jgfs_awips_f${fcsthrs}_20km_${cyc}"
+	    # export DATA="${DATAROOT}/$job"
```
aerorahul, line 87: are we sure this is not going to bomb?

jobs/rocoto/awips.sh
```diff
@@ -94,14 +83,14 @@ for fhr in $fhrlst; do
...
 	    $AWIPS20SH
 	fi
 	
 	if [[ $(expr $fhr % 6) -eq 0 ]]; then
-	    export job="jgfs_awips_f${fcsthrs}_${cyc}"
-	    export DATA="${DATAROOT}/$job"
+	    # export job="jgfs_awips_f${fcsthrs}_${cyc}"
```
aerorahul, line 92: same here

jobs/rocoto/awips.sh
```diff
@@ -131,12 +120,12 @@ for fhr in $fhrlst; do
             done
             
 	    export fcsthrs=$fhr3
-	    export job="jgfs_awips_f${fcsthrs}_20km_${cyc}"
-	    export DATA="${DATAROOT}/$job"
+	    # export job="jgfs_awips_f${fcsthrs}_20km_${cyc}"
```
aerorahul, line 123: and here.

aerorahul (approved):
Changes look great.
A lot of duplicate lines have been eliminated.
I like that the header takes in named arguments that improves readability.

A few inline comments could use some explaining.

jobs/JGFS_ATMOS_AWIPS_20KM_1P0DEG
WalterKolczynski-NOAA: Maybe, but this was here before (even though git diff is marking it as an addition; it was L5 before).

jobs/JGFS_ATMOS_AWIPS_G2
WalterKolczynski-NOAA: So was this. There is still much j-job cleaning to do.

aerorahul: indeed.

jobs/rocoto/awips.sh
WalterKolczynski-NOAA: No, but the AWIPS jobs aren't working right now anyway for unrelated reasons. It isn't causing any immediate issues up until that point. But that is why I left them as comments for now instead of just deleting them outright, so we can see how it was supposed to work before when we get a chance to fix the AWIPS job.

aerorahul: understood.  thanks for explaining.

PR #1234, opened 2023-01-10
Add utility to compare Fortran namelists

aerorahul:
**Description**
Often times it is necessary to compare Fortran namelists between a UFS-weather-model regression test and a global-workflow experiment, or in other example applications.

This PR adds a simple utility that loads two namelists and spits out the differences between them.  The differences are calculated as a departure from the first namelist.
This utility leverages `f90nml` (approved for use on WCOSS2)

The usage is as follows:
```
❯❯❯ python3 compare_f90nml.py -h
usage: compare_f90nml.py [-h] [-r] left_namelist right_namelist

Compare two Fortran namelists and display differences (left_namelist - right_namelist)

positional arguments:
  left_namelist   Left namelist to compare
  right_namelist  Right namelist to compare

options:
  -h, --help      show this help message and exit
  -r, --reverse   reverse diff (right_namelist - left_namelist) (default: False)
```

The comparison is done as follows:
- Both namelists are loaded
- We loop over the keys of `left_namelist`.  We look for the same key in the `right_namelist`.  If the key is found, the values are compared.  If the key is not found, a note is made that the key is undefined in `right_namelist`.
- Differences in the values are printed to screen.
- 
The `-r | --reverse` reverses the `namelists`.  This allows the user to use `right_namelist` as the reference.

If differences are found, they are shown as follows (examples of `input.nml` from the `control_p8` and `cpld_control_p8` regression tests of the ufs-weather-model)
```
❯❯❯ python3 compare_f90nml.py control_p8.nml cpld_control_p8.nml
comparing: control_p8.nml | cpld_control_p8.nml
-----------------------------------------------
atmos_model_nml:
  ccpp_suite : FV3_GFS_v17_p8 | FV3_GFS_v17_coupled_p8
fms_nml:
  domains_stack_size : 3000000 | 8000000
fv_core_nml:
  dnats : 0 | 2
gfs_physics_nml:
    min_seaice : 0.15 | 1e-06
  use_cice_alb : False | True
     nstf_name : [2, 1, 0, 0, 0] | [2, 0, 0, 0, 0]
        cplchm : False | True
        cplflx : False | True
        cplice : False | True
        cplwav : False | True
    cplwav2atm : False | True
```

**Type of change**

**How Has This Been Tested?**
As shown above in comparing 2 namelists from ufs-weather-model regression tests.
  
**Checklist**

### DIFF
--- a/ush/compare_f90nml.py
+++ b/ush/compare_f90nml.py
@@ -85,5 +85,5 @@ if __name__ == "__main__":
 
     msg = f"comparing: {nml1} | {nml2}"
     print(msg)
-    print("-"*len(msg))
+    print("-" * len(msg))
     compare_dicts(dict1, dict2)
--- a/ush/compare_f90nml.py
+++ b/ush/compare_f90nml.py
@@ -1,18 +1,23 @@
 #!/usr/bin/env python3
 
 import json
+import f90nml
 from typing import Dict
 from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter
-try:
-    import f90nml
-except (ImportError, ModuleNotFoundError):
-    raise ModuleNotFoundError(f"f90nml not found on this system, ABORT!")
 
 
 def get_dict_from_nml(filename: str) -> Dict:
     """
     Read a F90 namelist and convert to a dictionary.
     This method uses json to convert OrderedDictionary into regular dictionary
+    Parameters
+    ----------
+    filename: str
+              Name of the F90 namelist
+    Returns
+    -------
+    dictionary: Dict
+                F90 namelist returned as a dictionary
     """
     return json.loads(json.dumps(f90nml.read(filename).todict()))
 
@@ -26,6 +31,18 @@ def compare_dicts(dict1: Dict, dict2: Dict, path: str = "") -> None:
     If a matching key is not found, it is set to as UNDEFINED.
     Note: A reverse match is not performed in this method.  For reverse matching, use the -r option in the main driver.
     Note: This is a recursive method to handle nested dictionaries.
+    Parameters
+    ----------
+    dict1: Dict
+           First dictionary
+    dict2: Dict
+           Second dictionary
+    path:  str (optional)
+           default: ""
+           key (if nested dictionary)
+    Returns
+    -------
+    None
     """
 
     result = dict()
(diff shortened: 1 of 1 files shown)


ush/compare_f90nml.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python3
+
+import json
+from typing import Dict
+from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter
+try:
```
WalterKolczynski-NOAA: Is it necessary to put this load in a try/catch? If the module isn't found, it already throws the exception.

ush/compare_f90nml.py
```diff
@@ -0,0 +1,84 @@
...
+
+    _print_diffs(result)
+
+
+if __name__ == "__main__":
+
+    parser = ArgumentParser(
+        description=("Compare two Fortran namelists and display differences"),
+        formatter_class=ArgumentDefaultsHelpFormatter)
+    parser.add_argument('-n', '--namelists', help='name of two namelists to compare (namelist1 - namelist2)',
```
WalterKolczynski-NOAA: These are required (the script is kinda useless without namelists) and there are no other arguments. Why not make them positional arguments?

ush/compare_f90nml.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def compare_dicts(dict1: Dict, dict2: Dict, path: str = "") -> None:
+    """
+    Compare 2 dictionaries.
+    This is done by looping over keys in dictionary 1 and searching for them
+    in dictionary 2.
+    If a matching key is found, the values are compared.
+    If a matching key is not found, it is set to as UNDEFINED.
+    Note: A reverse match is not performed in this method.  For reverse matching, use the -r option in the main driver.
```
WalterKolczynski-NOAA, line 32: Might be a future improvement, but I think this would be even more useful if it inherently did both directions. As-is, there is a need to run it twice and mentally ignore duplicates.

Possible implementation:
Move all the diffing to another hidden function (like print), call it twice, and return the results dict. Then reverse all the lists in the second one and combine the dicts before sending it to the print function.

WalterKolczynski-NOAA (changes requested):
A few things, plus docstrings need arguments/returns (I like the descriptions).

ush/compare_f90nml.py
aerorahul: I generally like named arguments as they are more descriptive.
However, I also tried positional arguments and the usage looks as follows:
```
❯❯❯ python3 compare_f90nml.py -h
usage: compare_f90nml.py [-h] [-r] namelist namelist

Compare two Fortran namelists and display differences

positional arguments:
  namelist       name of two namelists to compare (namelist1 - namelist2)

options:
  -h, --help     show this help message and exit
  -r, --reverse  reverse diff (namelist2 - namelist1) (default: False)
```
It is not clear what is `namelist1` and `namelist2`.
I tried different variations for `metavar`, but did not find anything elegant.

ush/compare_f90nml.py
aerorahul: Sure.
Note however, that the `_print_diff` is called as one traverses down the dictionary.  The `compare_dicts` is not returning a difference dictionary.
There are off-the-shelf packages e.g. [deepdiff](https://deepdiff.readthedocs.io/en/latest/) that does this probably infinitely better.

ush/compare_f90nml.py
aerorahul: removed.
I was going for a friendly message with a traceback than an abrupt end.

ush/compare_f90nml.py
WalterKolczynski-NOAA: What would be wrong with this?
```
parser.add_argument('left_namelist', type=str, help="Left namelist to compare")
parser.add_argument('right_namelist', type=str, help="Right namelist to compare")
```

```
usage: test.py [-h] left_namelist right_namelist

positional arguments:
  left_namelist   Left namelist to compare
  right_namelist  Right namelist to compare

options:
  -h, --help      show this help message and exit
```

Left and right are clearer, but namelist1/namelist2 should work as well as the argument names.

aerorahul: done.

WalterKolczynski-NOAA (approved):

ush/compare_f90nml.py
WalterKolczynski-NOAA: Deferred to later.

Issue #269, opened 2021-02-17
Streamline module files


WalterKolczynski-NOAA:
With the move to hpc-stack, module lists are becoming uniform across machines and can be condensed into a single module for each build that covers all machines. The only machine-dependent portion is the use statement with the location of each stack and the compilation environment, which should be the same for all builds and loaded first. Furthermore, the module version can be determined via a versions file, so all modules are guaranteed to use the same version. These two steps will make porting and maintenance much easier, as changes will only need to be made in one location.

aerorahul:
This depends on the availability of spack-stack on RDHPCS

WalterKolczynski-NOAA:
OBE, we are moving to loading only the needed modules for each individual job (see #1521).

PR #1235, opened 2023-01-11
Add more user defined parameters for the marine DA

guillaumevernieres:
**Description**
This is the companion PR to [GDASApp PR#270](https://github.com/NOAA-EMC/GDASApp/pull/270) and needs to be merged before I can trigger the GDASApp `CI` (something that we will change eventually ...).
 
See [GDASApp issue #269](https://github.com/NOAA-EMC/GDASApp/issues/269)

**Type of change**

- [x] New feature (non-breaking change which adds functionality ... being a bit bold here, but sure)

**How Has This Been Tested?**

GDASApp `ctests`.
  
**Checklist**


parm/config/config.ocnanal
```diff
@@ -6,20 +6,27 @@
...
 export OBS_YAML_DIR=${HOMEgfs}/sorc/gdas.cd/parm/soca/obs/config
-export OBS_YAML=${HOMEgfs}/sorc/gdas.cd/parm/soca/obs/obs_list.yaml
-export OBS_LIST=${OBS_YAML_DIR}/obs_list.yaml
+export OBS_LIST=@SOCA_OBS_LIST@
+[[ -n "${OBS_LIST}" ]] || export OBS_LIST=${HOMEgfs}/sorc/gdas.cd/parm/soca/obs/obs_list.yaml
+export OBS_YAML=${OBS_LIST}
 export FV3JEDI_STAGE_YAML=${HOMEgfs}/sorc/gdas.cd/test/soca/testinput/dumy.yaml
 export SOCA_INPUT_FIX_DIR=@SOCA_INPUT_FIX_DIR@
 export SOCA_VARS=tocn,socn,ssh
+export SABER_BLOCKS_YAML=@SABER_BLOCKS_YAML@
```
RussTreadon-NOAA, line 15: Question:  What happens if `SOCA_INPUT_FIX_DIR` is not defined in the environment?    A similar question for `NICAS_RESOL` and `NCICAS_GRID_SIZE`?   `OBS_LIST` above provides a default in case `SOCA_OBS_LIST` is not defined (zero length).

RussTreadon-NOAA (commented):
Changes look good.   One minor question.

parm/config/config.ocnanal
guillaumevernieres: @RussTreadon-NOAA , they have to be defined in the `setup_expt.py` yaml of the g-w here `parm/config/yaml/defaults.yaml`. We host a default obs list yaml in parm, so pointing to that seems appropriate if nothing else is provided. At least that was my logic!

guillaumevernieres:
@WalterKolczynski-NOAA , @KateFriedman-NOAA , @aerorahul : Any idea why the `pytest` are failing? I couldn't make sense of the output, and I'm not sure if my changes are the reason for the failure. How do I run these tests locally?

parm/config/config.ocnanal
RussTreadon-NOAA: OK.   I was just wondering why we check to see if `OBS_LIST` is defined while we don't do so for other variables.   Seems the answer is that we have defaults in `parm/config/yaml/defults.yaml` and so it is not necessary to provide defaults in `parm/config/config.ocnanal`.     I don't know if the g-w team or NCO has preferences where / how we define variable defaults.

aerorahul:
> @WalterKolczynski-NOAA , @KateFriedman-NOAA , @aerorahul : Any idea why the `pytest` are failing? I couldn't make sense of the output, and I'm not sure if my changes are the reason for the failure. How do I run these tests locally?

I noticed them too and don't have a clue yet.  
Manually booting them works, but its, well manual, which kinda defeats the purpose of automated testing.
I'll take a look and see if there is an outage or existing issue in Github Actions world.

parm/config/config.ocnanal
guillaumevernieres: @RussTreadon-NOAA , that `OBS_LIST` is a temporary solution, we'll have to construct the list differently at run time since it certainly is going to be cycle dependent.

RussTreadon-NOAA: Got it.   No need for me to get hung up on this since it's likely to change in the future.

RussTreadon-NOAA (commented):
Have these changes been tested in a simple parallel to ensure they work the way intended?   Taking a step back can we cycle with `ocnanal`?

guillaumevernieres:
> Have these changes been tested in a simple parallel to ensure they work the way intended? Taking a step back can we cycle with `ocnanal`?

@RussTreadon-NOAA :
1 - We can't cycle the coupled model with `develop`, so no, you can't cycle `ocnanal` since it depends on ocean/ice bkg's. I have a branch that allows us to cycle the marine DA with mom6-cice-fv3 but I have not tested the changes proposed in this pr. I'm confident that our gdasapp jjobs tests are sufficient ... cough cough ... 
2 - The changes have been tested with the `g-w` `ctest` in the `GDASApp` and the recently merged stand alone application that runs a sequence of jjobs. Both of these cover the "code" changed in this branch.

RussTreadon-NOAA (approved):
OK.  You've tested what you can test.   Approve.

aerorahul (approved):
changes look fine to me.

WalterKolczynski-NOAA (approved):

PR #741, opened 2022-04-19
Merge GFSv16.1.7 updates into operations branch

KateFriedman-NOAA:
### GFSv16.1.7 - Tropical cyclone storm names for the 2022 season

**Description**

The GFSv16.1.7 was implemented into production during the 12z cycle on April 19th (2022041912). This PR merges the GFSv16.1.7 updates from the `release/gfs.v16.1.7` branch into the `operations` branch.

This implementation contains the following updates:

1. source code update: `sorc/syndat_getjtbul.fd/getjtbul.f` code by @Qingfu-Liu 
2. fix file update: `fix_am/syndat_stmnames` (external to system, updated in FIX_DIR `fix_nco_gfsv16` set on disk)
3. new release notes: `docs/Release_Notes.gfs.v16.1.7.txt`

**Type of change**

- New feature (non-breaking change which adds functionality)
- This change requires a documentation update

**How Has This Been Tested?**

The source code update was tested by @Qingfu-Liu on WCOSS-Dell prior to hand-off to NCO via the `EMC-v16.1.7` global-workflow tag.
  
Refs #698 
Closes #698

PR #1242, opened 2023-01-13
Use fracoro data for all new UFS applications 

### LINKED ISSUE #863: FIXfv3 not being set up correctly for atmos-only experiment in the latest workflow
Issue:  
An attempt to run  atmos-only experiment using the latest workflow (develop branch) and the model tag p8c revealed that FIXfv3 is not set up correctly for atmos-only runs if FRAC_GRID=.true..   The new VIIRS veg type datasets and new mask/oro datasets are only made available in the fix_fv3_fracoro directory. 

In the original [exglobal_forecast](https://github.com/yangfanglin/global-workflow/blob/feature/gfsv17/scripts/exglobal_forecast.sh).sh I have the following settings.
```
reso=025
[[ ${CASE} = "C192" ]] && reso=050
[[ ${CASE} = "C96" ]]  && reso=100
frac_grid=${frac_grid:-".true."}
if [ $frac_grid = ".false." ]; then
 FIXfv3=$FIX_DIR/fix_fv3_gmted2010/${CASE}
else
 FIXfv3=$FIX_DIR/fix_fv3_fracoro/${CASE}.mx${reso}_frac
fi
```
I think the latest fix, mask and oro datasets created by Shan/Mike/Helin should work for both fractional and non-fractional grid.   So the default FIXfv3 should be pointing to fix_fv3_fracoro instead of  fix_fv3_gmted2010 for all new UFS applications.   We need to use the new VIIRS veg type,  VIIRS based land-sea-lake mask and the updated oro dataset that was made to be consistent with MOM6, no matter if we are running coupled or atmos-only experiments.   

 fix_fv3_gmted2010 should only be used for running current operational model (GFS.v16)

----------------
Helin checked the model RT settings and found out the RTs are using the correct (new) datasets.

ChunxiZhang-NOAA:
**Description**

The new fracoro data should be used for all new UFS applications no matter if it uses frac_grid or not.

Most problems in Issue[#863](https://github.com/NOAA-EMC/global-workflow/issues/863) have been resolved. However, one problem remains, e.g., the latest fix, mask and oro datasets (fracoro) created by Shan/Mike/Helin should work for both fractional and non-fractional grid.

Note that it also requires the changes in UFS_UTILS. An corresponding PR[#741](https://github.com/ufs-community/UFS_UTILS/pull/741) in UFS_UTILS has been created.

Fixes: #863

Dependency: UFS_UTILS [PR#741](https://github.com/ufs-community/UFS_UTILS/pull/741)

**Type of change**

**How Has This Been Tested?**

The Forecast-only tests were successful on Orion. The tests are: Atmosphere-only C96 for FRAC_GRID on and off, respectively; Coupled c384 for GRAC_GRID on and off, respectively. However, the test for GLDAS is not done.


WalterKolczynski-NOAA:
@ChunxiZhang-NOAA is this still a draft? It also needs to be updated with current develop.

ChunxiZhang-NOAA:
@WalterKolczynski-NOAA Yes, it is still a draft. I will check with George tomorrow for the upstream PR.

WalterKolczynski-NOAA:
Today is @ChunxiZhang-NOAA's last day. Do we know who is going to take over this work? It looks like the UFS Utils PR it was waiting for was merged last week.

ChunxiZhang-NOAA:
@WalterKolczynski-NOAA I don't know who will take over this work, but Fanglin will arrange it.  This PR is basically ready, only needs to update the hash for UFS_UTILS to the latest one or 4e673bf. I can do it within the next few hours.

ChunxiZhang-NOAA:
@WalterKolczynski-NOAA  This PR is ready for review.

parm/config/config.init
```diff
@@ -13,9 +13,11 @@ echo "BEGIN: config.init"
 
 export UFS_DIR=${HOMEgfs}/sorc/ufs_utils.fd
 export GDASINIT_DIR=${UFS_DIR}/util/gdas_init
+export GDAS_INIT_DIR=${UFS_DIR}/util/gdas_init
```
WalterKolczynski-NOAA: Why are you introducing a second variable that does the exact same thing? I prefer this new form, but the old one would need to be removed and any existing references updated.

WalterKolczynski-NOAA (changes requested):

WalterKolczynski-NOAA:
@yangfanglin who is supposed to be taking this over?

yangfanglin:
@WalterKolczynski-NOAA Can we get this PR committed first since Chunxi said it was ready ?  We can come back to revisit the issue related to GDASINIT_DIR and GDAS_INIT_DIR.   I am afraid if we wait longer this PR will become more difficult to merge.   For now I do not have anyone can take over this PR.

parm/config/config.init
aerorahul:
```suggestion
```

aerorahul: We can commit this suggestion and merge the PR.
If this breaks, then someone will need to be assigned to fix it.

parm/config/config.init
```diff
@@ -16,6 +16,7 @@ export GDASINIT_DIR=${UFS_DIR}/util/gdas_init
 
 export CRES_HIRES=$CASE
 export CRES_ENKF=$CASE_ENKF
+export FRAC_ORO="yes"
```
aerorahul, line 19: What is `FRAC_ORO` and how does it differ from `FRAC_GRID=.true.`?

aerorahul: Do we need this?  It has been introduced here, but I can't see where it is used.

aerorahul: I see it is used in `ufs-utils`.  The question still begs asking, how does it differ from `FRAC_GRID`

WalterKolczynski-NOAA (approved):

PR #1264, opened 2023-01-21
Deploy documentation to RTD

aerorahul:
**Description**
This PR adds an Github action that will build and deploy documentation as artifacts for with every push to `develop`.
The artifacts can be downloaded from Actions -> Documentation -> Artifacts

**Type of change**

**How Has This Been Tested?**
To be tested with the github action.
  
**Checklist**


.github/workflows/docs.yaml
```diff
@@ -0,0 +1,54 @@
...
+    branches:
+      - develop
+      - feature/*
+      - main/*
+      - bugfix/*
+      - release/*
+    paths:
+      - docs/**
+  pull_request:
+    types: [opened, reopened, synchronize]
```
WalterKolczynski-NOAA, line 13: Not sure if all these should trigger the action. Shouldn't it only be on pushes to develop?

WalterKolczynski-NOAA (changes requested):

.github/workflows/docs.yaml
arunchawla-NOAA: If I understand this correctly this should create a documentation build for all the branches that have sphinx documentation and get updated. Where are the documentation files uploaded though?

arunchawla-NOAA (approved):
Is the build still deployed through RTD?

.github/workflows/docs.yaml
WalterKolczynski-NOAA: L38?

aerorahul: the uploads are as artifacts on github actions.  That is not RTD.  The artifacts will allow you to download the whole documentation.

To answer @WalterKolczynski-NOAA, I believe we want to build the latest documentation (`develop`), but we also want to build documentation for releases for sure.  Not sure about bugfixes and feature branches.

I know RTD does allow one to toggle between documentations between releases and latest.

WalterKolczynski-NOAA: Okay, if RTD can handle branches, I'm satisfied.

WalterKolczynski-NOAA (approved):

aerorahul:
> Is the build still deployed through RTD?

Yes.
Nothing at the RTD end has changed.
This PR simply automates the creation of sphinx documentation in the Github action and upload it as artifacts for a-posteriori inspection.

PR #1266, opened 2023-01-23
Update dev/gdasapp

guillaumevernieres:
That branch is behind develop and needs to be updated. I suspect this is the reason why the CI is not working on Orion? Thanks for pointing that out @RussTreadon-NOAA ! I can't figure out where in @CoryMartin-NOAA 's `ci` stuff we are cloning `dev/gdasapp` on Orion and `develop` on Hera ...

I'm not sure this branch is turning out to be amazingly useful. Maybe we should add an option to clone our g-w forks/branches instead of pointing to `dev/gdasapp` ... Food for thoughts for next week.


guillaumevernieres:
Damn ... the automated PR adds binaries from the docs. Will resubmit from a local merge of develop.

guillaumevernieres:
Well nope ... Somebody I won't name :) added binary files to this repo! 
Was [that](https://github.com/guillaumevernieres/global-workflow/blob/develop/docs/source/_static/GFS_v16_flowchart.png) intentional @WalterKolczynski-NOAA @KateFriedman-NOAA  @aerorahul?

aerorahul (approved):
approve (but not reviewed).

PR #1294, opened 2023-01-31
Deprecate `FDATE`

### LINKED ISSUE #1005: FDATE - overlooked or deprecated?
**Expected behavior**
All variables are filled out in `config.base`

**Current behavior**
After running the refactored setup scripts, when looking at the `config.base` file, there is a line:
`export FDATE=@FDATE@ `

Is this intentional? An oversight? What exactly is `$FDATE`? I see it being used here in `config.anal`:
https://github.com/NOAA-EMC/global-workflow/blob/392ca6fc71c390a75cf0c4d621daa491c721e08a/parm/config/config.anal#L15

When running a cycled experiment with older (pre GFSv16) ICs, the `gdasanal` job fails because it is trying to untar netCDF diagnostic files, but they are archived as binary diagnostic files. Removing the `FDATE` part of the above `if` causes the job to complete. 

My question is: is `$FDATE` used anywhere else? If not, can we modify `config.anal` to not key off of it?

**Machines affected**
all?

**To Reproduce**
Setup a cycled experiment and look at `config.base`.

aerorahul:
**Description**

There used to be a fringe functionality where the `gdas.tHHz.radstat` could be ignored in the first full cycle `gdasanal` task.
All cycled experiments that start with a cold-start or warm-start must have initial bias correction coefficients (the so called `gdas.tHHz.abias` files) and the `gdas.tHHz.radstat` file.

Experts may customize their configurations if they choose to setup without a `gdas.tHHz.radstat` file thereby initializing the radiance diagnostics.  They would anyways know what they are doing. 

There are no updates in documentation as this was never an advertised mode of starting an experiment.

Closes #1005 

**Type of change**

**How Has This Been Tested?**
- [X] Cycled test on Orion

A forecast test is not required for this change.
  
**Checklist**


WalterKolczynski-NOAA (approved):

CoryMartin-NOAA:
I just noticed this, does this fix the issue where the radstat files have binary format when it should be expecting netCDF diags?

aerorahul:
@CoryMartin-NOAA 
No.  We are no longer supporting "binary" radstat files going forward.
The users can download the "netcdf" radstat files from previous implementation parallels that go quite a few years back.
If a super user wishes to start an experiment from a date that does not have a "netcdf" radstat file, they can manually turn off radstat `USE_RADSTAT=NO` for the first analysis cycle and then remove that line going forward.

CoryMartin-NOAA:
> they can manually turn off radstat `USE_RADSTAT=NO` for the first analysis cycle and then remove that line going forward.

Ok that was my backup plan, thanks

PR #1306, opened 2023-02-10
Module base file reorder and reduction

### LINKED ISSUE #479: Module order and dependency check
Review modules being loaded on each platform (e.g. via module_base.*) to check they are loading in the proper order and dependencies are being met correctly. Check for module swaps and resolve as needed to maintain the correct module loads.

Module loads should also follow this general order/grouping:

1. compilers
2. mpi
3. standard 3rd party e..g jasper, etc.
4. hdf5
5. netcdf
6. nceplibs

KateFriedman-NOAA:
**Description**

This PR reduces and reorganizes the `module_base*` files. This work addresses issue #479 and the desire to reduce the modulefiles for runtime down to only modules needed for running, since components now all load the modules they need for buildtime themselves. Resulting contents of `module_base*` files will help with future work to move module loads into scripts and retire the `module_base*` files.

The following was done for this PR:

1. Address issue #479 needs to reorder modules: compilers, mpi, standard 3rd party, hdf5, netcdf, nceplibs
2. Address issue #479 needs to check dependencies for correct load order
3. Reduce/clean the `module_base` modulefiles to only load the modules that are actually needed/used at runtime and not load any unnecessary ones

For 1: Move module loads around to get them loading in the order prescribed in issue #479.
For 2: Check module dependencies of each module loaded in `module_base` to ensure no dependencies are missing and no reloads occur that aren't desired/expected.
For 3: Deleted modules that are not needed at runtime, generally ones that were carried through from older versions of these modulefiles and which were originally added to support codes building in global-workflow. Now that the `/sorc` folder contains no code, no longer need to support build modules in the `module_base*` files.

@DavidHuber-NOAA FYI on this PR. Please check the S4 and Jet `module_base` files for potential impacts on those platforms from my changes.

**Type of change**

Cleanup and organize

**Tests being done**

- [x] Clone and build tests on Orion
- [x] Cycled test on Orion (1.5 cycles), compared to control run = bit-wise identical results
- [x] Coupled forecast-only test on Orion, compared to control run = bit-wise identical results
- [x] Clone and build tests on Hera
- [x] Cycled test on Hera (1.5 cycles), compared to control run = bit-wise identical results
- [x] Coupled forecast-only test on Hera, compared to control run = bit-wise identical results
- [x] Clone and build tests on WCOSS2
- [x] Cycled test on WCOSS2 (1.5 cycles), compared to control run = bit-wise identical results
- [x] Coupled forecast-only test on WCOSS2, compared to control run = bit-wise identical results

Scanned logs in tests on Orion and Hera for new, unexpected, or silent errors. No new errors beyond the usual expected ones (e.g. "cannot remove file" ones).
  
**Checklist**

Resolves #479


modulefiles/module_base.jet.lua
```diff
@@ -8,26 +8,20 @@ load(pathJoin("hpc", "1.1.0"))
 load(pathJoin("hpc-intel", "18.0.5.274"))
 load(pathJoin("hpc-impi", "2018.4.274"))
 
+load(pathJoin("esmf", "8_2_0_beta_snapshot_14"))
+
 load(hpss)
```
DavidHuber-NOAA: This should be `load("hpss")`, with quotes around hpss.

DavidHuber-NOAA (approved):
The changes look good.  The current Jet module file has a bug in it where "hpss" is not quoted.  My Jet port will fix this, so I'm approving this whether that gets fixed or not.  Both module files loaded the correct modules on their respective systems (with the bug fix).

modulefiles/module_base.jet.lua
KateFriedman-NOAA: Noted, I've make this fix in my branch @ db77600f since you mentioned it. Thanks!

KateFriedman-NOAA:
Removing the `esmf` module load from the `module_base*` modulefiles and extending tests on Orion and Hera to check it's indeed not needed.

KateFriedman-NOAA:
All tests have now completed and shown reproducible results on all supported platforms.

aerorahul (approved):
testing shows reproducible results.
This is a step towards loading only necessary modules at runtime.

WalterKolczynski-NOAA (approved):

CoryMartin-NOAA:
@guillaumevernieres and I think this broke UFS DA related development, but not sure yet exactly why there are additional things like `nco` added here?

KateFriedman-NOAA:
> not sure yet exactly why there are additional things like nco added here?

I didn't add `nco` in this PR, I just moved its order of loading. Let me know if any modules that UFS DA needs were removed in this PR though and we can fix that.

CoryMartin-NOAA:
sorry @KateFriedman-NOAA you're right, it was just reordered. I don't know what is going on but there are all kinds of modulepath changes now that weren't there before... I'm still investigating.

CoryMartin-NOAA:
actually this change shouldn't have affected our stuff at all... sorry to have complained and rushed to assume this was the cause of the module issue... now I don't know what the problem is but it shouldn't be this.

KateFriedman-NOAA:
No worries! Let me know if you do happen to trace the issue back to this PR though. :)

Issue #1299, opened 2023-02-07
Only use CDUMP where appropriate


WalterKolczynski-NOAA:
**Description**
Much of the workflow uses `$CDUMP` where it should be using `$RUN` instead. `$RUN` represents the ecflow family of the current job and can be one of `gdas`, `gfs`, `enkfgdas`, and `enkfgfs`. `$CDUMP` refers to whether the job is part of the early cycle or late cycle (`gdas` or `gfs`). `$RUN` should always be used in defining output paths for the current job.

As part of this issue, places that still use afterward can remove workaround used to accommodate CDUMPs of `enkfgdas` and `enkfgfs` (see Issue #1298) can be removed.

Part of Epic #1069 

**Requirements**
`$CDUMP` used only where appropriate
Where `$CDUMP` remains, remove any workarounds used to strip `enkf` (i.e. `${CDUMP/enkf}` or `${CDUMP} =~ gdas`)

**Acceptance Criteria (Definition of Done)**
All workflow cases run properly using `$RUN`

**Dependencies**
#761 (this will partially resolve this issue)
#1298

Issue #761, opened 2022-04-26
Organize COM and refactor path definitions


WalterKolczynski-NOAA:
**Description**
The COM directory, particularly atmos, is currently very flat, with most files at the same level. This is already a bit unwieldy, a situation that would become much worse if ensembles were implemented without additional organization.

Original proposed organization: (**Please see PR #1421 for up-to-date reorg trees**)
```
$mem
|- atmos
  |- model_data
    |- analysis
    |- history
      |- (NetCDFs produced by model)
    |- input
    |- master
      |- (master files)
      |- (sfluxgrb files).
    |- restart
  |- products
    |- 0p25
    |- 1p00
    |- bufr [prod]
    |- gempak [prod]
    |- genesis_vital_$year [dev]
      |- all.vitals.gfs.gfso.$year
      |- genesis.vitals.gfso.$year
    |- goes_sim [prod]
      |- *.goessimpgrb2.*
    |- imagery [prod]
      |- *.gif
    |- minmon [prod]
    |- tracks [dev]
      |- storms.*
      |- trak.*
    |- wafs [prod]
      |- blended
      |- no_us
      |- unblended
    |- wmo [prod]
|- chem
  |- model_data
    |- history
|- ice
  |- model_data
    |- input
    |- history
    |- restart
  |- products
|- ocean
  |- model_data
    |- analysis
    |- history
    |- input
    |- restart
  |- products
    |- 0p25
    |- 0p50
|- wave
  |- model_data
    |- history
    |- prep
    |- restart
  |- products
    |- gridded
    |- station
```

**Requirements**
A sensible organization of output files into different directories, with each directory containing a reasonable number of files.

Ideally, this would be done using configurable filename templates to allow for easy changes in the future, assuming it can be done in bash in a sensible manner. Otherwise, it will have to wait until a possible future conversion to python.

**Acceptance Criteria (Definition of Done)**
End-to-end completion of workflow using the new COM structure.

**Dependencies**
Issues #679, #746, #769
With the understanding that any solution implemented now is only an interim solution until an agreement is reached with NCO on a final configuration.

Issue #1298

WalterKolczynski-NOAA:
After discussion with the global-workflow team, we've decided on a tentative structure overall and for atmos. Modifications to other component directories may be added as we go. I will keep the first post updated with the current proposed structure.

[...]

PR #1309, opened 2023-02-13
Change RUN and CDUMP for ensemble jobs

### LINKED ISSUE #1298: Change RUN for ensemble jobs
**Description**
Currently ensemble jobs utilize the same `$RUN` and `$CDUMP` as deterministic jobs. However, this violates NCO policy and is also causing complications in the COM refactor (#761).

Ideally we would like `CDUMP` to be separate from `RUN` and reference which dump is being used (likely the original intent, given the name), which would return it to its current options of 'gfs' or 'gdas' while allowing `RUN` to be 'enkfgfs' or 'enkfgdas'. However, the codebase uses `$CDUMP` in a lot of places where `$RUN` should be used (see #1298), so for the time being keep `CDUMP` and `RUN` identical. Reverting `CDUMP` will be part of #1298.

Part of Epic #1069 

**Requirements**
GDAS ensemble jobs that use `enkfgdas` as its `$RUN`
GFS ensemble jobs that use `enkfgfs` as its `$RUN`

Cycled forecast completes successfully using the new `$RUN` (and associated filename changes).

**Dependencies**
None

WalterKolczynski-NOAA:
**Description**
Updates `$RUN` and `$CDUMP` for ensemble jobs to include `enkf`.

Previously, the `$RUN` for EnKF jobs had been set to `gdas` or `gfs`. However, this violates NCO policy and was also complicating Issue #761 . Now the `$RUN` for EnKF jobs is either `enkfgdas` or `enkfgfs`.

Theoretically, `$CDUMP` shouldn't need to change. However, `$CDUMP` and `$RUN` are used interchangeably throughout much of the workflow (Issue #1299), so for now the `$CDUMP` is kept identical to `$RUN`. This will be corrected in a future PR.

This change **changes the name** of enkf output files. Files now begin `enkfgdas` (or `enkfgfs`) rather than `gdas`.

Closes #1298

**Type of change**

**How Has This Been Tested?**
- [x] Cycled test on Orion
  
**Checklist**


WalterKolczynski-NOAA:
Leaving draft for now until I do a bit more testing and figure out how we will address the GDAS App jobs. So far, I have not updated the GDAS App jobs, but they will be impacted by this change as well. Need to discuss how we want to handle those.

WalterKolczynski-NOAA:
@CoryMartin-NOAA @guillaumevernieres @HenryWinterbottom-NOAA
This change is going to require updates to UFSDA as well. Also note a couple other expected changes along the same lines so you're aware and don't think we are just repeatedly changing the same stuff without a plan:

- This `$RUN` change will be permanent, but `$CDUMP` will be reverted back once workflow has been updated to use it properly instead of as a stand-in for `$RUN`
- More fundamental changes are coming via the use of path templates as part of the #761. You can get an **early** sneak preview here: https://github.com/WalterKolczynski-NOAA/global-workflow/blob/feature/com_reorg/parm/config/config.com

WalterKolczynski-NOAA:
Outside of the filename changes, output for this PR confirmed to be bitwise-identical after 4½ cycles. A prototype coupled was also tested and found to be identical.

jobs/JGDAS_ATMOS_CHGRES_FORENKF
```diff
@@ -17,31 +17,19 @@ export DO_CALC_ANALYSIS=${DO_CALC_ANALYSIS:-"YES"}
...
-    export COMIN_GES_OBS="${ROTDIR}/${GDUMP}.${gPDY}/${gcyc}/obs"
+    export COMOUT="${ROTDIR}/${RUN/enkf}.${PDY}/${cyc}/atmos"
+    export COMOUT_ENS="${ROTDIR}/${RUN}.${PDY}/${cyc}"
 fi
 mkdir -m 775 -p ${COMOUT}
-# COMIN_GES and COMIN_GES_ENS are used in script
-export COMIN_GES="${ROTDIR}/${GDUMP}.${gPDY}/${gcyc}/atmos"
-export COMIN_GES_ENS="${ROTDIR}/enkfgdas.${gPDY}/${gcyc}"
+
+export ATMFCST_ENSRES="${COMOUT_ENS}/mem001/atmos/${APREFIX_ENS}atmf006.nc"
```
KateFriedman-NOAA, line 32: Curious about this line. This seems even more hardcoded than similar lines like this in other scripts. Example from `jobs/JGDAS_ATMOS_ANALYSIS_DIAG`:
```
export ATMGES_ENSMEAN="${COMIN_GES_ENS}/${GPREFIX_ENS}atmf006.ensmean.nc"
```
Is it not possible to set `ATMFCST_ENSRES` without hardcoding `mem001/atmos`? Or do we need this hardcoding?

scripts/exgdas_enkf_earc.sh
```diff
@@ -190,9 +190,9 @@ if [ "${GDATE}" -lt "${RDATE}" ]; then
     RDATE=${GDATE}
 fi
 rPDY=$(echo "${RDATE}" | cut -c1-8)
-clist="gdas gfs"
+clist="gdas gfs enkfgdas enkfgfs"
```
KateFriedman-NOAA: Why still list `gdas gfs` in `clist` if this script is only for enkf archival?

KateFriedman-NOAA (approved):
Couple small comments but approve!

jobs/JGDAS_ATMOS_CHGRES_FORENKF
WalterKolczynski-NOAA: I believe it is just needs an example of the file so it can extract the grid information. Since the number of members can vary, it is best to look for the first member, since that will be there no matter what the ensemble size.

KateFriedman-NOAA: Ok, gotcha, thanks!

scripts/exgdas_enkf_earc.sh
WalterKolczynski-NOAA: Good catch.

jobs/JGDAS_ENKF_DIAG
```diff
@@ -22,31 +22,33 @@ GDATE=$(${NDATE} -${assim_freq} ${CDATE})
 gPDY=$(echo ${GDATE} | cut -c1-8)
 gcyc=$(echo ${GDATE} | cut -c9-10)
 GDUMP=${GDUMP:-"gdas"}
+GDUMP_ENS=${GDUMP_ENS:-"enkf${GDUMP}"}
```
aerorahul, line 25: This is a j-job. There is no one above this job setting `GDUMP_ENS`.
```suggestion
GDUMP_ENS="enkf${GDUMP}"
```
Same on the line above for `GDUMP`.

jobs/JGDAS_ENKF_DIAG
```diff
@@ -22,31 +22,33 @@ GDATE=$(${NDATE} -${assim_freq} ${CDATE})
 gPDY=$(echo ${GDATE} | cut -c1-8)
 gcyc=$(echo ${GDATE} | cut -c9-10)
 GDUMP=${GDUMP:-"gdas"}
```
aerorahul, line 24:
```suggestion
GDUMP="gdas"
```

jobs/JGDAS_ENKF_DIAG
```diff
@@ -22,31 +22,33 @@ GDATE=$(${NDATE} -${assim_freq} ${CDATE})
...
+GDUMP_ENS=${GDUMP_ENS:-"enkf${GDUMP}"}
 
 export CASE=${CASE_ENKF}
-export CDUMP_OBS=${CDUMP_OBS:-${CDUMP}}
+export CDUMP_OBS=${CDUMP_OBS:-${CDUMP/enkf}}
+export CDUMP_CTL=${CDUMP_CTL:-${CDUMP/enkf}}
```
aerorahul, line 29: What is intent of introducing `CDUMP_CTL`?

aerorahul: Same for `CDUMP_OBS`

jobs/JGDAS_ENKF_ECEN
```diff
@@ -20,35 +20,23 @@ GDATE=$(${NDATE} -${assim_freq} ${CDATE})
 gPDY=$(echo ${GDATE} | cut -c1-8)
 gcyc=$(echo ${GDATE} | cut -c9-10)
 GDUMP=${GDUMP:-"gdas"}
+GDUMP_ENS=${GDUMP_ENS:-"enkf${GDUMP}"}
```
aerorahul, line 23: Similar comment as the one made above.

jobs/JGDAS_ENKF_ECEN
```diff
@@ -20,35 +20,23 @@ GDATE=$(${NDATE} -${assim_freq} ${CDATE})
...
 
 export CASE=${CASE_ENKF}
 
+export CDUMP_OBS=${CDUMP_OBS:-${CDUMP/enkf}}
```
aerorahul, line 27: `CDUMP_OBS` is needed for?

aerorahul (commented):
Can you explain the some basic choices on introducing variables that seem not necessary.
I will continue reviewing.

jobs/JGDAS_ENKF_DIAG
WalterKolczynski-NOAA: `CDUMP_OBS` was a pre-existing variable. I created `CDUMP_CTL` so as not to misuse `CDUMP_OBS` for the deterministic member (even though they will be the same for now). I think this all goes away in the COM refactor when I revert `CDUMP` and use it per our discussion for both of these.

jobs/JGDAS_ENKF_DIAG
WalterKolczynski-NOAA: This wasn't something I changed, but I've changed it now to what you suggested (I had wondered the same thing).

jobs/JGDAS_ENKF_ECEN
WalterKolczynski-NOAA: This was done because this is how other j-jobs did it (though I clearly wasn't consistent, since I didn't use a `CDUMP_CTL` and instead just did the substitution in-line). I didn't worry _too_ much about consistency since I know this is all about to be substantially refactored anyway.

aerorahul (approved):

aerorahul:
@CoryMartin-NOAA @guillaumevernieres 
This change will impact GDASapp.
Please be aware of these inbound changes, and that they will change shortly as this is work in progress to resolve several outstanding EE2 issues.

CoryMartin-NOAA:
@aerorahul I need to discuss with you how to handle additional obs now that `COMIN_OBS` has been changed

PR #1341, opened 2023-02-28
Run executables or scripts from within python.

### LINKED ISSUE #1337: Add capability to run executables from within python tasks
**Description**
Several new tasks added conform to `initialize`, `run` and `finalize` when setting up, executing and cleaning of a job function.
The `run` typically requires exporting environment variables (e.g. `OMP_NUM_THREADS`, etc.) before running the executable (e.g. ./exec.x, mpirun exec.x, etc.).
As of now, the `run` is a shell script. 

This feature will allow executing from within python through the use of `subprocess` or other available alternatives.

**Requirements**
Ability to consistently run executables from within python tasks.

There are some existing solutions; some developed at NOAA and some elsewhere in the community.
Rather than listing all of them here, some exploration will be done before implementing a few solutions.

aerorahul:
**Description**

This PR:
- adds the ability of running executable (binary or shell scripts) via the `subprocess.Popen` call
- allows setting env. variables for the subprocess without having to modify the calling environment.
- allows passing custom arguments to the executable
- allows capturing `stdout` and `stderr` as well as pass `stdin`

Fixes #1337 

The core of the code is adopted from `spack`.  Reference is provided in the code itself.

Documentation is added inline, however API documentation will need to wait till it is used in one of the tasks (Coming soon!)

Basic use will be as such:
```python
from pygw.executable import Executable

# Lets say we need to run a APRUN command e.g. "srun"
cmd = Executable('srun')

# Lets say we need to run the executable "fv3jedi_var.x"
cmd.add_default_arg('fv3jedi_var.x')

# Lets say we need to pass an argument yaml file to this executable
yaml_file = os.path.join(DATA, 'fv3jedi_var.yaml')
cmd.add_default_arg(yaml_file)

# Lets say we want to run w/ 4 threads
cmd.add_default_env('OMP_NUM_THREADS', 4)

# Run the command and capture the stdout and stderr in files named similarly.
cmd(output='stdout', error='stderr')
```

The `cmd` line above will translate to:
```bash
export OMP_NUM_THREADS=4
srun fv3jedi_var.x $DATA/fv3jedi_var.yaml 1>&stdout 2>&stderr
```

**Type of change**

**How Has This Been Tested?**

- [X] 2 pytests are added that expose the class `Executable` and `which` function
  
**Checklist**

### DIFF
--- a/ush/python/pygw/src/pygw/executable.py
+++ b/ush/python/pygw/src/pygw/executable.py
@@ -1,0 +1,278 @@
+#
+import os
+import shlex
+import subprocess
+import sys
+from typing import Any, Optional
+
+__all__ = ["Executable", "which", "ProcessError", "CommandNotFoundError"]
+
+
+class Executable:
+    """Class representing a program that can be run on the command line."""
+
+    def __init__(self, name: str):
+        """
+        Construct an executable object.
+
+        Parameters
+        ----------
+        name : str
+            name of the executable to run
+
+        Adopted from:
+        https://github.com/spack/spack/blob/develop/lib/spack/spack/util/executable.py
+        """
+        self.exe = shlex.split(str(name))
+        self.default_env = {}
+        self.returncode = None
+
+        if not self.exe:
+            raise ProcessError(f"Cannot construct executable for '{name}'")
+
+    def add_default_arg(self, arg: Any) -> None:
+        """Add a default argument to the command."""
+        self.exe.append(arg)
+
+    def add_default_env(self, key: str, value: Any) -> None:
+        """Set an environment variable when the command is run.
+
+        Parameters:
+            key: The environment variable to set
+            value: The value to set it to
+        """
+        self.default_env[key] = value
+
+    @property
+    def command(self) -> str:
+        """The command-line string.
+
+        Returns:
+            str: The executable and default arguments
+        """
+        return " ".join(self.exe)
+
+    @property
+    def name(self) -> str:
+        """The executable name.
+
+        Returns:
+            str: The basename of the executable
+        """
+        return os.path.basename(self.path)
+
+    @property
+    def path(self) -> str:
+        """The path to the executable.
+
+        Returns:
+            str: The path to the executable
+        """
+        return self.exe[0]
+
+    def __call__(self, *args, **kwargs):
+        """Run this executable in a subprocess.
+
+        Parameters:
+            *args (str): Command-line arguments to the executable to run
+
+        Keyword Arguments:
+            _dump_env (dict): Dict to be set to the environment actually
+                used (envisaged for testing purposes only)
+            env (dict): The environment with which to run the executable
+            fail_on_error (bool): Raise an exception if the subprocess returns
+                an error. Default is True. The return code is available as
+                ``exe.returncode``
+            ignore_errors (int or list): A list of error codes to ignore.
+                If these codes are returned, this process will not raise
+                an exception even if ``fail_on_error`` is set to ``True``
+            input: Where to read stdin from
+            output: Where to send stdout
+            error: Where to send stderr
+
+        Accepted values for input, output, and error:
+
+        * python streams, e.g. open Python file objects, or ``os.devnull``
+        * filenames, which will be automatically opened for writing
+        * ``str``, as in the Python string type. If you set these to ``str``,
+          output and error will be written to pipes and returned as a string.
+          If both ``output`` and ``error`` are set to ``str``, then one string
+          is returned containing output concatenated with error. Not valid
+          for ``input``
+        * ``str.split``, as in the ``split`` method of the Python string type.
+          Behaves the same as ``str``, except that value is also written to
+          ``stdout`` or ``stderr``.
+
+        By default, the subprocess inherits the parent's file descriptors.
+
+        """
+        # Environment
+        env_arg = kwargs.get("env", None)
+
+        # Setup default environment
+        env = os.environ.copy() if env_arg is None else {}
+        env.update(self.default_env)
+
+        # Apply env argument
+        if env_arg:
+            env.update(env_arg)
+
+        if "_dump_env" in kwargs:
+            kwargs["_dump_env"].clear()
+            kwargs["_dump_env"].update(env)
+
+        fail_on_error = kwargs.pop("fail_on_error", True)
+        ignore_errors = kwargs.pop("ignore_errors", ())
+
+        # If they just want to ignore one error code, make it a tuple.
+        if isinstance(ignore_errors, int):
+            ignore_errors = (ignore_errors,)
+
+        output = kwargs.pop("output", None)
+        error = kwargs.pop("error", None)
+        input = kwargs.pop("input", None)
+
+        if input is str:
+            raise ValueError("Cannot use `str` as input stream.")
+
+        def streamify(arg, mode):
+            if isinstance(arg, str):
+                return open(arg, mode), True
+            elif arg in (str, str.split):
+                return subprocess.PIPE, False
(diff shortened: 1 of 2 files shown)


ush/python/pygw/src/pygw/executable.py
```diff
@@ -0,0 +1,315 @@
...
+        self.returncode = None
+
+        if not self.exe:
+            raise ProcessError(f"Cannot construct executable for '{name}'")
+
+    def add_default_arg(self, arg: Any) -> None:
+        """Add a default argument to the command."""
+        self.exe.append(arg)
+
+    def add_default_env(self, key: str, value: Any) -> None:
```
WalterKolczynski-NOAA: Why `add_default_arg` and `add_default_env` instead of just `add_arg` and `add_env`? I don't understand what is "default" about these.

ush/python/pygw/src/pygw/executable.py
```diff
@@ -0,0 +1,315 @@
...
+            else:
+                return arg, False
+
+        istream, close_istream = streamify(input, "r")
+        ostream, close_ostream = streamify(output, "w")
+        estream, close_estream = streamify(error, "w")
+
+        cmd = self.exe + list(args)
+
+        escaped_cmd = ["'%s'" % arg.replace("'", "'\"'\"'") for arg in cmd]
```
WalterKolczynski-NOAA, line 195: I think this should be functionally identical:

```suggestion
        escaped_cmd = [f"'{arg.replace("'", "'\"'\"'")}'" for arg in cmd]
```

ush/python/pygw/src/tests/test_executable.py
```diff
@@ -0,0 +1,46 @@
...
+
+    cmd = Executable(str(test_file))
+    assert cmd.exe == [str(test_file)]
+
+    stdout_file = tmp_path / 'stdout'
+    stderr_file = tmp_path / 'stderr'
+    cmd(output=str(stdout_file), error=str(stderr_file))
+    with open(str(stdout_file)) as fh:
+        assert fh.read() == whoami + '\n'
+
```
WalterKolczynski-NOAA, line 35: Do we want additional tests of `Executable` where we set an environment variable, provide an input file, and stderr has non-null contents?

WalterKolczynski-NOAA (changes requested):
A couple things, plus the usual request for reST doc-blocks.

ush/python/pygw/src/tests/test_executable.py
aerorahul: More tests would definitely be great.  Feel free to conjure up some or we can add them in a future PR.

ush/python/pygw/src/pygw/executable.py
aerorahul: I think `default_env` makes sense because the `__call__` method takes optional keyword arguments `env` to over-ride or add to the default environment.  If `env` is passed, then the executable is run with that environment (`env`) + `default_env`.  If `env` is `None`, the executable is run with the `os.environ` + `default_env`

Similarly, one can add default args to the `exe` as well as pass command specific args via the `__call__` magic function.

ush/python/pygw/src/pygw/executable.py
aerorahul: If you are sure, I can commit.  I am a bit leary of all the `'` and `"`.  They play tricks with my eyes.

aerorahul:
> A couple things, plus the usual request for reST doc-blocks.

Can you please point out where the reST doc-blocks are missing?

WalterKolczynski-NOAA:
> > A couple things, plus the usual request for reST doc-blocks.
> 
> Can you please point out where the reST doc-blocks are missing?

I think most of it is there but misformatted. Parameters should be listed like this:
```
Parameters
----------
x : type
    Description of parameter `x`.
y
    Description of parameter `y` (with type not specified).
```
See https://numpydoc.readthedocs.io/en/latest/format.html#sections

These were also incomplete:
`add_default_arg`
`which_string` (yeah, it will be almost the same as `which`)
`test_executable`
`test_which`

Seems nit-picky, but will be important as we move to generating html documentation from the scripts.

ush/python/pygw/src/pygw/executable.py
WalterKolczynski-NOAA: I used the same quotation marks as were there, I just replaced `%s` with the substitution. Let me check real fast.

WalterKolczynski-NOAA: It does not work. Just leave as-is for now, as much as using `%` pains me.

ush/python/pygw/src/tests/test_executable.py
```diff
@@ -0,0 +1,52 @@
...
+
+script = """#!/bin/bash
+echo ${USER}
+"""
+
+
+def test_executable(tmp_path):
+    """
+    Tests the class `Executable`
```
WalterKolczynski-NOAA, line 14: Type and description of `tmp_path`? (Looks like a `Path`)

ush/python/pygw/src/tests/test_executable.py
```diff
@@ -0,0 +1,52 @@
...
+    cmd(output=str(stdout_file), error=str(stderr_file))
+    with open(str(stdout_file)) as fh:
+        assert fh.read() == whoami + '\n'
+
+
+def test_which(tmpdir):
+    """
+    Tests the `which()` function.
+    `which` should return `None` if the executable is not found
+    """
```
WalterKolczynski-NOAA, line 45: Type of `tmpdir`?

WalterKolczynski-NOAA (changes requested):

WalterKolczynski-NOAA (approved):

ush/python/pygw/src/pygw/executable.py
```diff
@@ -0,0 +1,354 @@
...
+    >>> from pygw.executable import Executable
+    >>> cmd = Executable('srun')  # Lets say we need to run command e.g. "srun"
+    >>> cmd.add_default_arg('my_exec.x')  # Lets say we need to run the executable "my_exec.x"
+    >>> cmd.add_default_arg('my_arg.yaml')  # Lets say we need to pass an argument to this executable e.g. "my_arg.yaml"
+    >>> cmd.add_default_env('OMP_NUM_THREADS', 4)  # Lets say we want to run w/ 4 threads in the environment
+    >>> cmd(output='stdout', error='stderr')  # Run the command and capture the stdout and stderr in files named similarly.
+
+    `cmd` line above will translate to:
+
+    $ export OMP_NUM_THREADS=4
```
CoryMartin-NOAA, line 26: devil's advocate here, this is great but how can we convince Joe User that this is better than a bash script for this sort of thing?

CoryMartin-NOAA (approved):

ush/python/pygw/src/pygw/executable.py
aerorahul: Because you don't have to step out of the Python interpreter and get back in to finish the work after the executable is done. 
Stepping back in incurs overhead in having to ensure that the environment is exactly the same as before stepping out.

aerorahul: I hope that is convincing enough for provenance of the process itself.

HenryRWinterbottom:
I thought I submitted this yesterday.

My main question is why this addition is necessary? This capability has already been tested and vetted across multiple workflows (i.e., NGGPS, UFS-RNR, and some I developed in the private sector).

The relevant module is here: https://github.com/HenryWinterbottom-NOAA/ufs_pyutils/blob/develop/execute/subprocess_interface.py

PR #1349, opened 2023-02-28
Make necessary bugfixes to get aerosol cycling going

CoryMartin-NOAA:
**Description**

This PR adds in a number of bugfixes that were overlooked in #1106 that now allows for 3DVar aerosol DA cycling on Hera/Orion.

This PR also updates the GDASApp hash to the most recent commit in develop.

**Type of change**

**How Has This Been Tested?**

- [x] Clone and Build tests on Hera and Orion
- [x] Cycled test on Hera and Orion
  
**Checklist**


ush/load_ufsda_modules.sh
```diff
@@ -53,6 +53,14 @@ elif [[ -d /scratch1 ]] ; then
   module load prod_util
 elif [[ -d /work ]] ; then
   # We are on MSU Orion
+  # prod_util stuff, find a better solution later...
```
RussTreadon-NOAA, line 56: Any estimate from the hpc-stack team when this problem will be resolved?   It would be nice to remove this hack sooner rather than later.

CoryMartin-NOAA: agreed, it's not likely to happen until the UFS spack-stack and the JEDI spack-stack are one and the same...

RussTreadon-NOAA (approved):
Given that 3DVar aerosol DA cycling tests have been run on Hera and Orion and the changes indeed restore the cycling capability, approve.

ush/load_ufsda_modules.sh
RussTreadon-NOAA: OK.  Thanks for the timing sequence.

andytangborn (approved):
Since it's cycling, then it's fine with me.

ush/load_ufsda_modules.sh
aerorahul: The stack unification will take a while till the GSI and GDASApp can be built and run w/ the same compiler version.

ush/load_ufsda_modules.sh
```diff
@@ -53,6 +53,14 @@ elif [[ -d /scratch1 ]] ; then
...
+  #module use /apps/contrib/NCEP/hpc-stack/libs/hpc-stack/modulefiles/compiler/intel/2022.1.2/
+  #module load prod_util
+  export UTILROOT=/work2/noaa/da/python/opt/intel-2022.1.2/prod_util/1.2.2
+  export MDATE=/work2/noaa/da/python/opt/intel-2022.1.2/prod_util/1.2.2/bin/mdate
+  export NDATE=/work2/noaa/da/python/opt/intel-2022.1.2/prod_util/1.2.2/bin/ndate
+  export NHOUR=/work2/noaa/da/python/opt/intel-2022.1.2/prod_util/1.2.2/bin/nhour
+  export FSYNC=/work2/noaa/da/python/opt/intel-2022.1.2/prod_util/1.2.2/bin/fsync_file
```
aerorahul, line 63: May I ask what is using this?

aerorahul (approved):
Changes look good.
I just have one question about those variables added in `load_ufsda_modules.sh`.
The new jobs should not be using those, so curious why they are introduced here.

ush/load_ufsda_modules.sh
CoryMartin-NOAA: Good question, there were things that were failing due to `set -e` because these weren't defined... perhaps that isn't true anymore?
It might not actually be anything besides this:
https://github.com/NOAA-EMC/global-workflow/blob/07fedaa649e4576a2bd3d2fca32bb6c144fa2a85/parm/config/config.base.emc.dyn#L148
and I just defined everyting in the prod_util module for completeness sake

CatherineThomas-NOAA (approved):
Looks good to me.  I assume the cycling tests were for ATMA, but I don't see how these changes could impact other non-aero configurations, so it's probably fine to not run other tests.

ush/load_ufsda_modules.sh
WalterKolczynski-NOAA: Why can't we just add prod_util to the UFSDA stack?

CoryMartin-NOAA: We could, but it is maintained by JCSDA (Dom) and probably isn't worth it? Would there be a need to include these tools in the workflow scripts before JEDI is on WCOSS2?

WalterKolczynski-NOAA (approved):

PR #1372, opened 2023-03-04
Refactor UFS-DA ATM var component to use python g-w

RussTreadon-NOAA:
**Description**
This PR contains UFS-DA ATM var changes in PR https://github.com/NOAA-EMC/global-workflow/pull/1354.   

Below is a list of changes in this PR
- rename UFS-DA ATM var jobs `atmanalprep`, `atmanalrun`, and `atmanalpost` as `atmanlinit`, `atmanlrun`, and `atmanlfinal`, respectively
- replace UFS-DA ATM var shell scripts with python scripts
- rename UFS-DA ATM var j-jobs consistent with initialize, run, and finalize functions.  Update j-jobs to execute python scripts instead of shell scripts
- rename UFS-DA ATM var rocoto jobs to be consistent with initialize, run, and finalize functions.  Update jobs to set python paths and execute renamed j-jobs
- update rocoto workflow generation to new names for UFS-DA ATM var jobs 
- update UFS-DA ATM var job names in machine dependent `env` files to new job names
- rename UFS-DA ATM var configuration files consistent with change in job names
- add UFS-DA ATM yaml files for UFS-DA fix files, crtm files, and increments
- add python class for UFS-DA ATM var analysis
- link UFS-DA python increment conversion script from GDASApp to g-w `ush` directory

The above changes are part of a larger g-w effort to transition from shell scripts to python.  UFS-DA Aerosol was the first GDASApp system to be converted.   This PR represents the second GDASApp system, UFS-DA atmospheric variational DA, to be converted.  PR #1373 contains changes to convert the UFS-DA ensemble DA to a python based approach.   

Fixes (in part) https://github.com/NOAA-EMC/global-workflow/issues/1313
Depends (in part) on PR #1370 and GDASApp [#388](https://github.com/NOAA-EMC/GDASApp/pull/388)

**Type of change**
- [x] Update feature (non-breaking change which refactors existing functionality)

**How Has This Been Tested?**
- [x] Clone and Build on Hera and Orion
- [x] Cycled tests on Hera and Orion
  
**Checklist**

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code

PR #1370, opened 2023-03-03
Stage bias files for UFSDA aerosols

RussTreadon-NOAA:
**Description**
Updates UFSDA aerosols to stage bias files. Also adds in the threads and aprun commands for the run portion of the UFDA aero jobs.

Fixes (in part) #1313.

**Type of change**
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Update existing feature

**How Has This Been Tested?**
- [x] Clone and Build on Hera and Orion
- [x] Cycled tests on Hera and Orion
  
**Checklist**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code

PR #1354, opened 2023-03-01
Refactor UFS-DA ATM var and ens components to use python g-w

RussTreadon-NOAA:
**Description**
g-w is moving toward a python based approach.   The UFS-DA Aerosol component uses the python g-w approach.  This PR  replaces the non-python based UFS-DA ATM var and ens components with python based equivalents.

Fixes #1313 

**Type of change**

- [x] Update feature (non-breaking change which refactors existing functionality)

**How Has This Been Tested?**

- [ ] Clone and Build tests on Hera and Orion
- [x] Cycled test on Hera

  
**Checklist**

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] New and existing tests pass with my changes
- [ ] Any dependent changes have been merged and published

PR #1373, opened 2023-03-04
Refactor UFS-DA ATM ens component to use python g-w

### LINKED ISSUE #1313: Refactor UFS-DA ATM to use python based approach of UFS_DA Aerosol
**Description**
Refactor UFS-DA ATM g-w components to use UFS-DA Aerosol python based approach implemented via PR [#1106](https://github.com/NOAA-EMC/global-workflow/pull/1106). 

Specifically the g-w components activated via `DO_JEDIVAR="YES"` and `DO_JEDIENS="YES"` will be examined and aligned with the python based approach used by UFS-DA Aerosol g-w components.

**Requirements**
Refactoring of UFS-DA ATM g-w components should not alter existing ATM analysis results.

This issue will be closed when it is possible to cycle UFS-DA ATM g-w components using the python based approach of UFS-DA Aerosol g-w components and the PR to fold these changes into g-w `develop` is approved and merged.

**Dependencies**
Changes to the UFS-DA ATM component may require changes in the GDASApp ATM component (e.g., yaml files, python scripts).   GDASApp dependencies, if they arise, will be added to this issue.   In turn, this issue will be listed as a dependency in relevant GDASApp issues.

RussTreadon-NOAA:
**Description**
This PR contains UFS-DA ATM ens changes in PR #1354.

Below is a list of changes in this PR
- rename UFS-DA ATM ens jobs atmensanalprep, atmensanalrun, and atmensanalpost as atmensanlinit, atmensanlrun, and atmensanlfinal, respectively
- replace UFS-DA ATM ens shell scripts with python scripts
- rename UFS-DA ATM ens j-jobs consistent with initialize, run, and finalize functions. Update j-jobs to execute python scripts instead of shell scripts
- rename UFS-DA ATM ens rocoto jobs to be consistent with initialize, run, and finalize functions. Update jobs to set python paths and execute renamed j-jobs
- update rocoto workflow generation to new names for UFS-DA ATM ens jobs
- update UFS-DA ATM ens job names in machine dependent env files to new job names
- rename UFS-DA ATM ens configuration files consistent with change in job names
- add python class for UFS-DA ATM ens analysis
- unify JEDIEXE link for UFS-DA Aerosol, ATM, and ENS
- properly set `cycledefs` for `gfsatmanlinit`
- remove unused `FV3JEDI_FIX` from atmanl and atmensanl config

The above changes are part of a larger g-w effort to transition from shell scripts to python. UFS-DA Aerosol was the first GDASApp system to be converted. PR #1372 converted UFS-DA atmospheric variational DA to the python based approach.  This PR converts converts UFS-DA atmospheric local ensemble DA to the python based approach.

Fixes #1313
Depends (in part) on #1370 and #1372 and GDASApp [#388](https://github.com/NOAA-EMC/GDASApp/pull/388)

**Type of change**
- [x] Update feature (non-breaking change which refactors existing functionality)

**How Has This Been Tested?**
- [x] Clone and Build tests on Hera and Orion
- [x] Cycled tests on Hera and Orion

  
**Checklist**


RussTreadon-NOAA:
`RussTreadon-NOAA:feature/refactor_jedivar` has been updated to the current head of `develop`, [7e661f4](https://github.com/RussTreadon-NOAA/global-workflow/commit/7e661f4cd00cf99b12e55574318bbd0eb1a5eaa0).   The changes in PR #1372 and #1373 have been merged into `RussTreadon-NOAA:feature/refactor_jediatm`.   The combined set of changes are being cycled in the prufsatm on Hera.   The prufsatm is correctly processing files.

The changes in this PR are ready for review.

RussTreadon-NOAA:
@aerorahul and @WalterKolczynski-NOAA :  The changes in this PR are ready for review.   
- `RussTreadon-NOAA:feature/refactor_jediens` is up to date with the current head ([b2ed864](https://github.com/NOAA-EMC/global-workflow/commit/b2ed8648f80946de85983a51664b120540854cc9)) of `develop`.    
- The prufsatm on Hera has been cycled with `feature/refactor_jediens` with acceptable behavior.

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -0,0 +1,332 @@
...
+        # loop over ensemble members
+        bkglist = []
+        for imem in range(1, task_config.NMEM_ENKF + 1):
+            memchar = f"mem{imem:03d}"
+
+            # make run directory for member restart files
+            bkgdir = [
+                os.path.join(task_config.DATA, 'bkg', memchar)
+            ]
+            FileHandler({'mkdir': bkgdir}).sync()
```
aerorahul: Instead of calling this for every member, consider creating a list and calling `FileHandler` once.

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -0,0 +1,332 @@
...
+        ----------
+        bkg_dict: Dict
+            a dictionary containing the list of model background files to copy for FileHandler
+        """
+        # NOTE for now this is FV3 RESTART files and just assumed to be fh006
+
+        bkgdir = [
+            os.path.join(task_config.DATA, 'bkg'),
+        ]
+        FileHandler({'mkdir': bkgdir}).sync()
```
aerorahul: This purpose of this method is to construct a dictionary.
It should not call `FileHandler` in it.

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -0,0 +1,332 @@
...
+                    bkglist.append([os.path.join(rst_dir, basename), os.path.join(run_dir, basename)])
+
+        bkg_dict = {
+            'copy': bkglist,
+        }
+
+        return bkg_dict
+
+    @logit(logger)
+    def jedi2fv3inc(self: Analysis) -> None:
```
aerorahul, line 260: Now that we have a repetition of this method here and in `atm_analysis.py`, we need to not repeat.

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -0,0 +1,332 @@
...
+        Analysis: parent class for GDAS task
+
+        Returns
+        ----------
+        None
+        """
+        super().initialize()
+
+        # stage CRTM fix files
+        crtm_fix_list_path = os.path.join(self.task_config.HOMEgfs, 'parm', 'parm_gdas', 'atm_crtm_coeff.yaml')
```
aerorahul, line 83: I notice this repeats from `atm_analysis.py`.

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -0,0 +1,332 @@
...
+        logger.debug(f"Staging CRTM fix files from {crtm_fix_list_path}")
+        crtm_fix_list = parse_yamltmpl(crtm_fix_list_path, self.task_config)
+        FileHandler(crtm_fix_list).sync()
+
+        # stage fix files
+        jedi_fix_list_path = os.path.join(self.task_config.HOMEgfs, 'parm', 'parm_gdas', 'atm_jedi_fix.yaml')
```
aerorahul, line 89: and this too.  
we should ask if we will need different yaml's for crtm, jedi_fix, etc. between `var` and `ens` modes?

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -0,0 +1,332 @@
...
+        save_as_yaml(ensda_yaml, self.task_config.fv3jedi_yaml)
+        logger.info(f"Wrote ensemble da YAML to: {self.task_config.fv3jedi_yaml}")
+
+        # link executable to DATA/ directory
+        exe_src = self.task_config.JEDIENSEXE
+        logger.debug(f"Link executable {exe_src} to DATA/")  # TODO: linking is not permitted per EE2.  Needs work in JEDI to be able to copy the exec.
+        exe_dest = os.path.join(self.task_config.DATA, os.path.basename(exe_src))
+        if os.path.exists(exe_dest):
+            rm_p(exe_dest)
+        os.symlink(exe_src, exe_dest)
```
aerorahul: I think we can do this better by defining `JEDIEXE` in `config.atmanalrun` and `config.atmensanalrun` etc, and lift this to `analysis.py` as a method to `link_jediexe` or something.

aerorahul (changes requested):
most of my comments are in `atmens_analysis.py`. 
I can work with you on those and bring them in via patch if necessary.
The changes in other files are good.

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: Thank you @aerorahul for the good suggestion.  Change committed at [f902370](https://github.com/RussTreadon-NOAA/global-workflow/commit/f902370ad57b553bf4d4fa5861250ae23a68650c).

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: Remove `bkgdir` and `FileHandler` since not needed.   Done at [d587193](https://github.com/RussTreadon-NOAA/global-workflow/commit/d5871934cb75c09d8d07884c98398767b838868c).

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -0,0 +1,325 @@
...
+                template = f'{to_fv3time(self.task_config.current_cycle)}.fv_{ftype}.res.tile{{tilenum}}.nc'
+                for itile in range(1, task_config.ntiles + 1):
+                    basename = template.format(tilenum=itile)
+                    bkglist.append([os.path.join(rst_dir, basename), os.path.join(run_dir, basename)])
+
+            for ftype in ['phy_data', 'sfc_data']:
+                template = f'{to_fv3time(self.task_config.current_cycle)}.{ftype}.tile{{tilenum}}.nc'
+                for itile in range(1, task_config.ntiles + 1):
+                    basename = template.format(tilenum=itile)
+                    bkglist.append([os.path.join(rst_dir, basename), os.path.join(run_dir, basename)])
```
```suggestion
            template = f'{to_fv3time(self.task_config.current_cycle)}.{ftype}.tile{{tilenum}}.nc'
            for ftype in ['fv_core.res', 'fv_srf_wnd.res', 'fv_tracer.res', 'phy_data', 'sfc_data']:                
                for itile in range(1, task_config.ntiles + 1):
                    basename = template.format(tilenum=itile)
                    bkglist.append([os.path.join(rst_dir, basename), os.path.join(run_dir, basename)])
```

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: I should not reuse the method name `jedi2fv3inc`.    The `jedi2fv3inc` in `atmens_analysis.py` differs from the same named method in `atm_analysis.py`.   Differences are
- atmens does not need to select between different resolution `atmges_fv3` files.  atm selects `atmges_fv3` based on analysis resolution.
- atmens paths include and vary with ensemble member.   atm paths are fixed with no ensembles.

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: Not sure if this is what you have in mind @aerorahul , but here is what I did in a working copy of `ush/python/pygfs/task/analysis.py` and `ush/python/pygfs/task/atmens_analysis.py`

The following was added to the `initialize` method in `analysis.py`
```
         obs_dict = self.get_obs_dict()
         FileHandler(obs_dict).sync()

+        # all analyses need to stage fix files
+        jedi_fix_list_path = os.path.join(self.task_config.HOMEgfs, 'parm', 'parm_gdas', 'atm_jedi_fix.yaml')
+        logger.debug(f"Staging JEDI fix files from {jedi_fix_list_path}")
+        jedi_fix_list = parse_yamltmpl(jedi_fix_list_path, self.task_config)
+        FileHandler(jedi_fix_list).sync()
+
+        # some analyses need to stage crtm files
+        crtm_fix_list_path = os.path.join(self.task_config.HOMEgfs, 'parm', 'parm_gdas', 'atm_crtm_coeff.yaml')
+        logger.debug(f"Staging CRTM fix files from {crtm_fix_list_path}")
+        crtm_fix_list = parse_yamltmpl(crtm_fix_list_path, self.task_config)
+        FileHandler(crtm_fix_list).sync()
+
         # some analyses need to stage bias corrections
         bias_dict = self.get_bias_dict()
         FileHandler(bias_dict).sync()
```
The following was deleted from `atmens_analysis.py`
```
         super().initialize()

-        # stage CRTM fix files
-        crtm_fix_list_path = os.path.join(self.task_config.HOMEgfs, 'parm', 'parm_gdas', 'atm_crtm_coeff.yaml')
-        logger.debug(f"Staging CRTM fix files from {crtm_fix_list_path}")
-        crtm_fix_list = parse_yamltmpl(crtm_fix_list_path, self.task_config)
-        FileHandler(crtm_fix_list).sync()
-
-        # stage fix files
-        jedi_fix_list_path = os.path.join(self.task_config.HOMEgfs, 'parm', 'parm_gdas', 'atm_jedi_fix.yaml')
-        logger.debug(f"Staging JEDI fix files from {jedi_fix_list_path}")
-        jedi_fix_list = parse_yamltmpl(jedi_fix_list_path, self.task_config)
-        FileHandler(jedi_fix_list).sync()
-
         # stage backgrounds
         FileHandler(self.get_bkg_dict(AttrDict(self.task_config))).sync()
```
Tests in the prufsatm show that the above changes work.

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: See comment above for staging of crtm files.  If the changes above are acceptable, I'll remove the jedi and crtm fix file staging from `atm_analysis.py`.

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: I looked in `parm/parm_gdas`.  I see that we have different crtm_coeff yamls for UFS-DA Aerosol and ATM.  My attempt to move the staging of crtm files from `atmens_analysis.py` to `analysis.py` is not sufficiently general.   It works for UFS-DA ATM and ENS but not Aerosol.

RussTreadon-NOAA: The same comment applies for jedi_fix.    `aero_jedi_fix.yaml` and `atm_jedi_fix.yaml` reference different fieldmetadata files.    The jedif_fix change to my working copy of `analysis.py` only works for UFS-DA ATM and ENS.  It won't work for UFS-DA Aerosol.

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: Test implementation in working copy of `RussTreadon-NOAA:feature/refactor_jediens` does not cover UFS-DA Aerosol case.   Need to take a step back and think more holistically.   Generalization will likely require changes to UFS-DA Aerosl, ATM, and ENS applications.   Would be good to capture this in a new issue and subsequent PR.

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: Suggested change made in working copies of `analysis.py`, `atmens_analysis.py`, and `config.atmensanl`.   Made the following modifications
- Use `JEDIEXE` to define executable in `config.atmensanl`.   Same can be done in `config.atmanl` and `config.aeroanl`.    
- Remove executable link from `atmens_analysis.py` and `atm_analysis.py`.   
- Add the following method to `analysis.py`
```
+    @logit(logger)
+    def link_jediexe(self: Task) -> None:
+        """Compile a dictionary of background error files to copy
+
+        This method links a JEDI executable to the run directory
+
+        Parameters
+        ----------
+        Task: GDAS task
+
+        Returns
+        ----------
+        None
+        """
+        exe_src = self.task_config.JEDIEXE
+
+        # TODO: linking is not permitted per EE2.  Needs work in JEDI to be able to copy the exec.
+        logger.debug(f"Link executable {exe_src} to DATA/")
+        exe_dest = os.path.join(self.task_config.DATA, os.path.basename(exe_src))
+        if os.path.exists(exe_dest):
+            rm_p(exe_dest)
+        os.symlink(exe_src, exe_dest)
+
+        return
```
This method is executed from `initialize` via
```
+        # link jedi executable to run directory
+        self.link_jediexe()
```

Tested both UFS-DA ATM and ENS in prufsatm.  The modified scripts and configs work.  I did not test UFS-DA Aerosol.   

If the above is acceptable, I'll commit the changes to `RussTreadon-NOAA:feature/refactor_jediens`

RussTreadon-NOAA: Given that this change can work for UFS-DA Aerosol, ATM, and ENS apps, I committed the required changes to `RussTreadon-NOAA:feature/refactor_jediens` at [4b61129](https://github.com/RussTreadon-NOAA/global-workflow/commit/4b61129abbe17220c51ef028f66d945b1ad40758).

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: Recommend that unification of fix file staging across UFS-DA applications be the subject of a new g-w issue.   What do you think @aerorahul and @CoryMartin-NOAA ?

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: Recommend that unification of fix file staging across UFS-DA applications be the subject of a new g-w issue.   What do you think @aerorahul and @CoryMartin-NOAA ?

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: Shall I rename the ensemble `jedi2fv3inc` in `atmens_analysis.py` to distinguish it from its deterministic counterpart in `atm_analysis.py`?    The `atmens_analysis.py` method could be named `jedi2fv3inc_ens`.

Potential unification of the ensemble and deterministic methods could be the subject of a new g-w issue.   It's possible that at a future date we will not need either method if the UFS weather model can directly read tiled analysis increments.

Thoughts?  Comments?  @aerorahul and @CoryMartin-NOAA

ush/python/pygfs/task/atmens_analysis.py
CoryMartin-NOAA: That is what I would suggest, since we may want to discuss whether we should link *all* CRTM fix files per job, or just the ones of obs that are used, for example. Plus in FV3-JEDI, there will be different fieldsets between the ATM and AERO analysis jobs.

parm/config/config.atmensanl
```diff
@@ -0,0 +1,23 @@
...
+
+########## config.atmensanl ##########
+# configuration common to all atm ens analysis tasks
+
+echo "BEGIN: config.atmensanl"
+
+export OBS_YAML_DIR=${HOMEgfs}/sorc/gdas.cd/parm/atm/obs/config/
+export OBS_LIST=${HOMEgfs}/sorc/gdas.cd/parm/atm/obs/lists/lgetkf_prototype.yaml
+export ATMENSYAML=${HOMEgfs}/sorc/gdas.cd/parm/atm/lgetkf/lgetkf.yaml
+export FV3JEDI_FIX=${HOMEgfs}/fix/gdas
```
aerorahul: Is this variable used anywhere?

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -0,0 +1,311 @@
...
+        ]
+        FileHandler({'mkdir': newdirs}).sync()
+
+        # Make directories for member analysis files
+        for imem in range(1, self.task_config.NMEM_ENKF + 1):
+            memchar = f"mem{imem:03d}"
+            anldir = [
+                os.path.join(self.task_config.DATA, 'anl', memchar)
+            ]
+            FileHandler({'mkdir': anldir}).sync()
```
aerorahul: Make 1 call to `FileHandler` instead of `NMEM_ENKF`.
```suggestion
        anldir = []
        for imem in range(1, self.task_config.NMEM_ENKF + 1):
            anldir.append(os.path.join(self.task_config.DATA, 'anl', f'mem{imem:03d}'))
        FileHandler({'mkdir': anldir}).sync()
```

ush/python/pygfs/task/atmens_analysis.py
aerorahul: I think you can use the same name.  I only looked at the name of the method and assumed them to do the same thing.  Thank you for explaining the difference.

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -0,0 +1,311 @@
...
+            atminc_fv3 = os.path.join(self.task_config.COMOUT, memchar, 'atmos',
+                                      f"{self.task_config.CDUMP}.t{self.runtime_config.cyc:02d}z.atminc.nc")
+
+            # Execute incpy to create the UFS model atm increment file
+            cmd = Executable(incpy)
+            cmd.add_default_arg(atmges_fv3)
+            cmd.add_default_arg(atminc_jedi)
+            cmd.add_default_arg(atminc_fv3)
+            logger.debug(f"Executing {cmd}")
+            cmd(output='stdout', error='stderr')
```
aerorahul, line 312: I am not sure here and may need to think a bit more.
What is happening here is we are creating `NMEM_ENKF` objects of the `cmd` with `incpy` and running them serially one after the other.

This job is already reserving a multi-core node.  We should use that knowledge to efficiently use the resource.

ush/python/pygfs/task/atmens_analysis.py
aerorahul: I think I know what to do.  We can continue with the current approach in this PR.
We can revisit this in a future PR.  I may request assistance from @RussTreadon-NOAA @CoryMartin-NOAA to test since I haven't quite gotten a working setup for UFS-DA.  It will happen.

parm/config/config.atmensanl
RussTreadon-NOAA: Thank you, @aerorahul , for noting this.  You are correct.  We no longer use `FV3JEDI_FIX` in UFS-DA ATM var and ens.   

`FV3JEDI_FIX` removed from `config.atmensanl` and `config.atmanl` at [30db276](https://github.com/RussTreadon-NOAA/global-workflow/commit/30db276ca20e8887ee686edb7722496508a63a42).   

`FV3JEDI_FIX` remains in `config.aeroanl` since I do not have a way to test UFS-DA Aerosol.

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: Thank you @aerorahul .   Your refactoring is compact and more efficient.   Suggested change committed at [8056bf9](https://github.com/RussTreadon-NOAA/global-workflow/commit/8056bf9836ac3d8dee9c2aca453e2f671670d6db).

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: I see what you are saying @aerorahul .   We could build a list of `incpy` realizations to run.  Each realization has a unique set of arguments.  The list is executed in MPMD mode via a single APRUN command.    Extra scripting needs to be added to implement this with flexibility in the job configuration.   As `NMEM_ENKF` or the resolution of the `atm` files increases we'll need to adjust the number of nodes and/or the processes per node count.

aerorahul: For now, lets just get this in.  We can optimize in a later PR.

Do you mind putting in a comment referencing this discussion as a `TODO`.  Something like:
```
# TODO: use MPMD or parallelize with mpi4py.  See https://github.com/NOAA-EMC/global-workflow/pull/1373#discussion_r1173060656
```

RussTreadon-NOAA: `TODO` comment with link committed at [812cb46](https://github.com/RussTreadon-NOAA/global-workflow/commit/812cb466b1441c1418a7541aeec957a03050ee20).

aerorahul (approved):
Thank you for accepting the suggestions.

RussTreadon-NOAA:
@aerorahul , @WalterKolczynski-NOAA , @CoryMartin-NOAA 
The review process has identified several TODOs for subsequent issues and PRs.   This noted, are any additional changes required in this PR before it can be approved and scheduled for merger into `develop`?

RussTreadon-NOAA:
Thank you @aerorahul .   You're ahead of me ... as usual.  Thank you for patiently instructing me in how to improve my meager python skills.   I appreciate your support.

WalterKolczynski-NOAA (approved):

WalterKolczynski-NOAA:
Want to get this in before I merge the COM refactor and temporarily break UFSDA, so we have a version where it works.

RussTreadon-NOAA:
Thank you @WalterKolczynski-NOAA , for your approval and thoughtfulness.   Yes, we would like this PR to be merged into `develop` before PR #1421 for exactly the reason you state.

`RussTreadon-NOAA:feature/refactor_jediens` has been updated to the current head ([740daba](https://github.com/NOAA-EMC/global-workflow/commit/740daba8d6d34a327199701c1df7d6e10da73ec5)) of `develop`.  This was done at [77bc6dc](https://github.com/RussTreadon-NOAA/global-workflow/commit/77bc6dcaf5a40990a97df5812449aafcd1ec7e22).  UFS-DA ATM Var and Ens jobs were rerun in the prufsatm.   All jobs successfully ran to completion.   

One sticky point is that after this PR enters g-w `develop` we need to approve and merge GDASApp PR [#388](https://github.com/NOAA-EMC/GDASApp/pull/388) to ensure full g-w UFS-DA ATM Var and Ens functionality.  Once PR [#388](https://github.com/NOAA-EMC/GDASApp/pull/388) enters GDASApp `develop`, we need to come back to g-w and update the GDASApp hash in `Externals.cfg` and `checkout.sh`.   

GDASApp PR [#388](https://github.com/NOAA-EMC/GDASApp/pull/388) won't be approved and merged today.   We can't delay other g-w PRs from moving forward.   Thus, after PR #1373 enters g-w `develop`, UFS-DA ATM Var and Ens DA won't work out of the box when we clone g-w `develop`.   Developers will need to manually update the GDASApp hash in `checkout.sh`.   This isn't ideal but it's not overly cumbersome.

Tagging @CoryMartin-NOAA and @guillaumevernieres for awareness.

RussTreadon-NOAA:
@WalterKolczynski-NOAA , do we have an estimate as to when this PR will be merged into `develop`?

WalterKolczynski-NOAA:
> @WalterKolczynski-NOAA , do we have an estimate as to when this PR will be merged into `develop`?

I was giving @CoryMartin-NOAA a chance to approve, since you pinged after COB yesterday. But it is afternoon now and he's had a chance now, so I'll merge in about a half-hour.

CoryMartin-NOAA (approved):
Yep, just finished going through, looks good to me, anything we find later we can fix when we come across it. Thanks!

RussTreadon-NOAA:
> Yep, just finished going through, looks good to me, anything we find later we can fix when we come across it. Thanks!

Thanks @CoryMartin-NOAA !

WalterKolczynski-NOAA:
Thanks everybody!

RussTreadon-NOAA:
Yes indeed, **_thank you everyone!_**   This is a nice way to close out a work week.

Issue #1353, opened 2023-03-01
CDUMP preceded by 'enkf' hpssarch_gen.sh


DavidHuber-NOAA:
**Expected behavior**

The enkfgdas_grp*.txt files would be populated by ENKF files from `$ROTDIR`.  Also, if an error is reported from htar or tar, hpss_gen.sh should return a non-zero status.

**Current behavior**

`$CDUMP` is preceded by `enkf`, which for enkf members causes hpssarch_gen.sh to search for filenames starting with `enkfenkfgdas`, which do not exist, and therefore the .txt files are never populated.

When this occurs, hpssarch_gen.sh returns a 0 status for all enkfgdasearc jobs except enkfgdasearc01.

**Machines affected**

Discovered on Jet, but likely all.

**To Reproduce**

Run a gdasearc job with `HPSSARCH="YES"` or `LOCALARCH="YES"` and check the logs.

**Context**

Found during Jet testing.

**Detailed Description**

`$CDUMP` is set to `$RUN` in config.base, while `$RUN` is set at runtime by rocoto in the xml.  For enkfgdas jobs, `$RUN` is set to `enkfgdas`.  In hpssarch_gen.sh, the enkf section of the script precedes all instances of `$CDUMP` with `enkf`.  This is a holdover from the time before #1309 when the enkf jobs had a `$CDUMP`/`$RUN` of `gdas`.

**Additional Information**

This may be related to #1299.

**Possible Implementation**

`enkf` should probably be removed as a prefix from all instances of `$CDUMP` in hpssarch_gen.sh and perhaps `$CDUMP` should be replaced with `$RUN`.

DavidHuber-NOAA:
As for not exiting on an error for group 0, this is by design.  For ensemble group 0, `set +e` is specified in the archiving block covering this group ([exgdas_enkf_earc.sh:100-133](https://github.com/NOAA-EMC/global-workflow/blob/c4d05e57d5e7192ac3b8a93e6880efe5bbad2e65/scripts/exgdas_enkf_earc.sh#L100-L133); specifically line 111) so that the job can continue if this is within the first 4 cycles.  If I recall correctly, only some of the files to be archived are created in the first four cycles, which causes htar to return a non-zero status, so this is ignored.

PR #1378, opened 2023-03-07
Add a missing dollar sign in the earc script

DavidHuber-NOAA:
**Description**
Ther is a missing `$` in the earc script.  This is related to #1353 and should have been caught during testing.  It appears that this was not caught because this if block is only triggered during the 3rd full cycle, so a 2-cycle test wasn't enough.

**Type of change**

**How Has This Been Tested?**

- [x] Cycled test on Jet
  
**Checklist**


aerorahul (approved):
Thanks for catching this.

PR #1393, opened 2023-03-13
Marine DA prep j-job needs more memory

### LINKED ISSUE #1389: Marine GDAS analysis prep job is running out of memory
We're not requesting enough memory in the prep job and it fails when trying to concatenate large ioda files.

Bump that [number](https://github.com/NOAA-EMC/global-workflow/blob/8a2d5061da3c3067291e51680408339ec2efec5c/parm/config/config.resources#L275) to something between 16 and 24GB.

guillaumevernieres:
The concatenation step runs out of memory in the marine-gdas prep step. This is not optimized and probably overkill, but 24GB should cover all cases.

- fixes #1389

### DIFF
--- a/parm/config/config.resources
+++ b/parm/config/config.resources
@@ -272,7 +272,7 @@ elif [[ "${step}" = "ocnanalprep" ]]; then
     export nth_ocnanalprep=1
     npe_node_ocnanalprep=$(echo "${npe_node_max} / ${nth_ocnanalprep}" | bc)
     export npe_node_ocnanalprep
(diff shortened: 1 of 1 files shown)


aerorahul (approved):
I think this is ok. 
Are we better off asking for an exclusive node?
Probably not on WCOSS2 which has 128 cores on a node.

WalterKolczynski-NOAA (approved):
Good enough for now. Like many other resource settings, this really should be dependent on resolution. We'll get to it eventually.

PR #1406, opened 2023-03-20
Feature/args schema

HenryRWinterbottom:
This PR provides an interface to the [schema](https://github.com/keleshev/schema) package. The intention of this PR is as follows:

- A means of evaluating application (e.g., GFS, GSI, JEDI, etc.,) options such that they conform to the expectations of the respective application(s);
- A means to improve the robustness of `scripts` level application scripts which may require command line arguments; the schema interface can be used to check that the arguments are correctly formatted (i.e., valid) prior to entering the task level (e.g., `ush`); an example can be seen as follows:

```
def main() -> None:

    # Define the schema attributes.
    cls_schema = {
        "yaml_file": str,
        "cycle": Or(str, int),
        "expt_name": str,
        "work_path": str,
    }

    # Collect the command line arguments.
    cls_opts = **some_options_passed_to_script**
    check_schema(cls_schema=cls_schema, cls_opts=cls_opts)
    
    # If the schema is valid, run the application; if the schema is not
    # valid, raise a SchemaException.
    task = SomeTask()
    task.run()
```
In the above example, if any of the dictionary keys are not provided to the script via either the command-line (at the `scripts` level) or by some other means, an exception will raised. Further, if the `cycle` attribute provided to the script is neither a `str` or `int`, an exception will also be raised.

**Type of change**

**How Has This Been Tested?**

A series of unit-tests have been provided in `pygw/tests/test_schema.py`; all tests pass.
  
**Checklist**

**Note**

This update requires that the [schema](https://github.com/keleshev/schema)  package is added to the respective Python stacks. Further, this PR will be used as a basis to establish the aforementioned script-level options in a future PR.

### DIFF
--- a/ush/python/pygw/src/pygw/schema.py
+++ b/ush/python/pygw/src/pygw/schema.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python
+
+# ----
+
+"""
+Module
+------
+
+    schema.py
+
+Description
+-----------
+
+    This module provides various schema evaluation interfaces and/or
+    check applications.
+
+Functions
+---------
+
+    validate(cls_schema, cls_opts)
+
+        This function validates the calling class schema; if the
+        respective schema is not validated an exception will be
+        raised; otherwise this function is passive.
+
+Requirements
+------------
+
+schema; https://github.com/keleshev/schema
+
+Author(s)
+---------
+
+    Henry R. Winterbottom; 20 March 2023
+
+History
+-------
+
+    2023-03-20: Henry Winterbottom -- Initial implementation.
+
+"""
+
+
+from typing import Dict
+
+from pygw.exceptions import SchemaException
+from pygw.logger import Logger, logit
+from schema import Schema
+
+# ----
+
+logger = Logger(level="info", colored_log=True)
+
+__all__ = ["validate_opts"]
+
+# ----
+
+
+@logit(logger)
+def validate_opts(cls_schema: Dict, cls_opts: Dict) -> None:
+    """
+    Description
+    -----------
+
+    This function validates the calling class schema; if the
+    respective schema is not validated an exception will be raised;
+    otherwise this function is passive.
+
+    Parameters
+    ----------
+
+    cls_schema: dict
+
+        A Python dictionary containing the calling class schema.
+
+    cls_opts: dict
+
+        A Python dictionary containing the options (i.e., parameter
+        arguments, keyword arguments, etc.,) passed to the respective
+        calling class.
+
+    Raises
+    ------
+
+    SchemaException:
+
+        * raised if an exception is encountered while validating the
+          schema.
+
+    """
+
+    # Define the schema object.
+    schema = Schema([cls_schema])
+
+    # Check that the specified options (`cls_opts`) are valid; proceed
+    # accordingly.
+    try:
+
+        # Validate the schema.
+        schema.validate([cls_opts])
+
+    except Exception as errmsg:
+
+        msg = f"Schema validation failed with error {errmsg}. Aborting!!!"
+        raise SchemaException(msg=msg) from errmsg
--- a/ush/python/pygw/src/tests/test_schema.py
+++ b/ush/python/pygw/src/tests/test_schema.py
@@ -1,0 +1,182 @@
+#!/usr/bin/env python
+
+# ----
+
+"""
+Module
+------
+
+    test_schema.py
+
+Description
+-----------
+
+    This module provides a series of unit-tests for the pygw.schema
+    module.
+
+Usage
+-----
+
+    pytest tests/test_schema.py
+
+Functions
+---------
+
+    test_schema_float()
+
+        This function provides a unit test for the schema module; a
+        float-value will be passed to the schema validator that
+        accepts only string or integer instances; in order to pass,
+        this unit test must raise a a SchemaException.
+
+    test_schema_int()
+
+        This function provides a unit test for the schema module; a
+        integer-value is passed to the schema.
+
+    test_schema_opt()
+
+        This function provides a unit test for the schema module; a
+        integer-value and an optional string value `opt_var` are
+        passed to the schema.
+
+    test_schema_str()
+
+        This function provides a unit test for the schema module; a
+        string is passed to the schema.
+
+Requirements
+------------
+
+schema; https://github.com/keleshev/schema
+
+Author(s)
+---------
+
+    Henry R. Winterbottom; 20 March 2023
+
+History
+-------
+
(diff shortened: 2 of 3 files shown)


WalterKolczynski-NOAA (approved):

ush/python/pygw/src/tests/test_schema.py
```diff
@@ -0,0 +1,182 @@
...
+@logit(base_logger)
+def test_schema_float() -> None:
+    """
+    Description
+    -----------
+
+    This function provides a unit test for the schema module; a
+    float-value will be passed to the schema validator that accepts
+    only string or integer instances; in order to pass, this unit test
+    must raise a a SchemaException.
```
WalterKolczynski-NOAA: Typo

ush/python/pygw/src/tests/test_schema.py
```diff
@@ -0,0 +1,182 @@
...
+    # the specified schema.
+    cls_opts = {"check_var": 0.0000}
+
+    # Evaluate the schema.
+    try:
+        validate_opts(cls_schema=CLS_SCHEMA, cls_opts=cls_opts)
+
+    except SchemaException:
+        pass
+
```
WalterKolczynski-NOAA, line 112: This will also incorrectly pass if no exception is raised.

WalterKolczynski-NOAA (changes requested):

ush/python/pygw/src/tests/test_schema.py
HenryRWinterbottom: @WalterKolczynski-NOAA Where is the typo?

ush/python/pygw/src/tests/test_schema.py
HenryRWinterbottom: The `schema` library will also raise it's own exception (which it does for this unit-test). I am just capturing the exception in a `pygw` specific exception. I want the exception to be raised (and thus pass) since I am proving the schema incorrectly formatted attributes.

ush/python/pygw/src/tests/test_schema.py
WalterKolczynski-NOAA: "a" occurs twice in a row.

ush/python/pygw/src/tests/test_schema.py
WalterKolczynski-NOAA: I get that, that's not the problem. The issue is if `validate_opts` does NOT raise an exception, the `except` clause is passed over and the function will still return successfully. Since this unit test is confirming invalid output DOES raise an exception, the test should be failing if no exception is raised, not just if a different exception is raised. Recommend changing `pass` to `return` and then adding an `assert false` after the try/except block.

HenryRWinterbottom: Got it. Thank you for the explanation. That completely eluded me.

ush/python/pygw/src/tests/test_schema.py
HenryRWinterbottom: Thank you. Good eye!

WalterKolczynski-NOAA (approved):

HenryRWinterbottom:
@WalterKolczynski-NOAA Can you please add `schema==0.7.5` to your Python stack? That will support the external dependencies of this PR.

WalterKolczynski-NOAA:
> @WalterKolczynski-NOAA Can you please add `schema==0.7.5` to your Python stack? That will support the external dependencies of this PR.

It conflicts with the current f90nml version (1.4.3). If you can figure out a version combination that works, I'll make the update.

```
Solving environment: ...working... failed

UnsatisfiableError: The following specifications were found to be in conflict:
  - f90nml
  - schema
Use "conda search <package> --info" to see the dependencies for each package.
```

HenryRWinterbottom:
Ok. I use both packages so I am assuming I found a solution. I'll take a look in the morning MST.

ush/python/pygw/src/pygw/schema.py
```diff
@@ -0,0 +1,105 @@
...
+    validate(cls_schema, cls_opts)
+
+        This function validates the calling class schema; if the
+        respective schema is not validated an exception will be
+        raised; otherwise this function is passive.
+
+Requirements
+------------
+
+schema; https://github.com/keleshev/schema
```
CoryMartin-NOAA, line 29: is this on the approved software list? It's a little worrisome since it is just on some guy's github

CoryMartin-NOAA (approved):
Looks good assuming my WCOSS2 concern is not needed.

CoryMartin-NOAA:
@HenryWinterbottom-NOAA after approving, I realized the automated tests fail, if we are allowed to use `schema` on WCOSS, then it also needs added to this CI environment as part of this PR.

ush/python/pygw/src/pygw/schema.py
aerorahul: We might just include this into pygw.  It is a single python file, quite mature and does not get much updates.

HenryRWinterbottom:
> @HenryWinterbottom-NOAA after approving, I realized the automated tests fail, if we are allowed to use `schema` on WCOSS, then it also needs added to this CI environment as part of this PR.

@CoryMartin-NOAA Yes, I have spoken with Walter about resolving this. The GW team spoke about this issue on this morning's call and we should be able to resolve it soon.

ush/python/pygw/src/pygw/schema.py
HenryRWinterbottom: I will default to what @aerorahul thinks about this. I believe that he has a contingency for this.

aerorahul:
I am fine with the introduction of this dependency on principle.  However, we don't have a demonstrable and urgent need for it as yet.
When we are ready validate a namelist or configuration in a task or experiment, we could introduce it at that time.

`schema` is a single file package and rather than adding a `pypi` dependency, I would rather just copy the file and attribute the source and maintain it.

I recommend closing this PR until we are ready for making use of it and let that define what features we need to build.

HenryRWinterbottom:
@aerorahul I agree with closing this PR, but for a different reason.

Yes, we have no use-case at the moment. And as @WalterKolczynski-NOAA found there are conflicts when trying to install the `schema` package against existing packages. Further, I think that we can emulate the schema approach by using simple dictionaries. This will provide us additional flexibility and customization instead of hacks around a package that does do exactly what we need.

That all being said, I am happy to write the `schema` package for `pygw` when the time comes, but I agree that we do not (yet) need it.

Thoughts?

WalterKolczynski-NOAA:
We may even be able to salvage schema out of [CROW](https://github.com/NOAA-EMC/CROW).

Sounds like we are closing this for now and will approach it from a new angle later when necessary.

PR #1419, opened 2023-03-23
Add all ocean da files under ocean comrot to archiving

AndrewEichmann-NOAA:
Adds all the ocean da files (along with all the files except for RESTART, which has their own archive file) to the ocean archive file.

Depending on how big these get in full resolution, the diag files might later go into their own archive file.

File changed: ush/hpssarch_gen.sh

Tested by rerunning gdasarch in existing experiment ocn-da-lores on Hera and then extracting the respective archive file from HPSS and diffing the results with the COMROT contents.


AndrewEichmann-NOAA:
whoops wrong repository

PR #1427, opened 2023-03-27
Feature/argparser.py

HenryRWinterbottom:
**Description**

This PR provides an interface for command line arguments to be passed to driver (sometimes refererenced as caller) scripts to be used by applications within the global workflow.

**Use Case**

The current global-workflow infrastructure assumes that the configuration object is defined by the contents of the run-time environment established by the global-workflow configuration beneath `parm/config`. This PR implements a methodology for a driver or caller script (`exglobal_some_task.py`, below) to receive arguments passed via the command line in order to define a similar object -- most often specific to the application. Although not imperative at the moment, providing an interface as described below will allow for simpler implementations as refactoring continues.

Consider the following example.

``python exglobal_some_task.py``

In this instance the global-workflow application will behave as usual and define the configuration object by casting `os.environ` as a Python dictionary. However, as the global-workflow refactor evolves, passing YAML-formatted configurations specific to the respective application becomes more attractive and should be considered. The contents within this PR allow such functionality as demonstrated in the following example.

``python exglobal_some_task.py  --yaml_file /path/to/yaml_file``

The driver script (`exglobal_some_task.py`) can read the YAML-formatted file and make use of it as necessary. Further, in the context of the current global-workflow, the configuration object defined by the casting of `os.environ` dictionary could be replaced by this new functionality while defaulting to the `os.environ` if the `yaml_file` (or any command-line) attribute is not specified. 

The additional advantage of this innovation is that the caller script (i.e., `exglobal_some_task.py`) does not require modification for new command line arguments. Consider the following code snippet.

```
from somewhere import SomeTask
from pygw.argsparse import ArgsParser

main() -> None:
   
        config = ArgsParser.run()
   
        task = SomeTask()
        task.run(config)
```

In the example above, all command-line arguments, no matter the number and/or type, are captured by the `ArgsParser.run` instance. However, for the current state of global-workflow development, the following configuration should be implemented.

```
import os

from somewhere import SomeTask
from pygw.argsparse import ArgsParser
from pygw.configuration import cast_strdict_as_dtypedict

main() -> None:
   
        config = ArgsParser.run()
        if config is None:
                 config = cast_strdict_as_dtypedict(os.environ)
   
        task = SomeTask(config)
        task.run()
```

In this example, if no command line values are passed, the configuration will default to the configuration defined by the casting of `os.environ` as Python dictionary.

In addition to the command line parsing application, a helper function is included in `pygw.argsparse.py` to ensure that the arguments are formatted correctly. This is necessary since the Python interpreter assumes that all arguments passed via the `argsparse.Arguments` class are formatted as strings upon entry. Further/future extensions of this capability could involve the use of schemas to evaluate command line value validity.

**Type of change**

**How Has This Been Tested?**

All unit-tests for `pygw` pass.
  
**Checklist**

### DIFF
--- a/ush/python/pygw/src/pygw/argsparse.py
+++ b/ush/python/pygw/src/pygw/argsparse.py
@@ -1,0 +1,210 @@
+"""
+Module
+------
+
+    argsparse.py
+
+Description
+-----------
+
+    This module contains classes and methods to parse and
+    appropriately format argumenst passed to a caller script.
+
+Classes
+-------
+
+    ArgsParser()
+
+        This is the base-class object for all command line argument(s)
+        parsing.
+
+Author(s)
+---------
+
+    Henry R. Winterbottom; 27 March 2023
+
+History
+-------
+
+    2023-03-27: Henry Winterbottom -- Initial implementation.
+
+"""
+
+# ----
+
+# pylint: disable=too-many-branches
+# pylint: disable=unnecessary-list-index-lookup
+
+# ----
+
+from argparse import ArgumentParser
+from dataclasses import dataclass
+from typing import Union
+
+from pygw.attrdict import AttrDict
+from pygw.logger import Logger
+
+# ----
+
+logger = Logger(level="info", colored_log="True")
+
+# ----
+
+__author__ = "Henry R. Winterbottom"
+__maintainer__ = "Henry R. Winterbottom"
+__email__ = "henry.winterbottom@noaa.gov"
+
+# ----
+
+
+@dataclass
+class ArgsParse:
+    """
+    Description
+    -----------
+
+    This is the base-class object for all command line argument(s)
+    parsing.
+
+    Returns
+    -------
+
+    args_obj: object
+
+        A Python object containing the command-line arguments; if no
+        command-line arguments have been passed to the caller script,
+        the returned value is NoneType.
+
+    """
+
+    def __check__(self, args_obj: AttrDict) -> AttrDict:
+        """
+        Description
+        -----------
+
+        This method checks that the argument values are formatted
+        correctly; this method is necessary due to the Python
+        interpretter assuming that all arguments passed via the Python
+        `sys` module are formatted as strings.
+
+        Parameters
+        ----------
+
+        in_dict: object
+
+            A Python object containing the argument values to be
+            evaluated and formatted as neccessary.
+
+        Returns
+        -------
+
+        out_dict: object
+
+            A Python object containing the formatted argument values.
+
+        """
+
+        # Define local function to sort and format the input Python
+        # dictionary upon entry.
+        def sorted_by_keys(in_dict):
+
+            out_dict = AttrDict()
+            for (key, value) in sorted(in_dict.items(), key=lambda key: key):
+
+                if isinstance(value, dict):
+                    setattr(out_dict, key, sorted_by_keys(value))
+
+                else:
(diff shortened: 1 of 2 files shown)


ush/python/pygw/src/pygw/argsparse.py
```diff
@@ -0,0 +1,210 @@
...
+                    setattr(out_dict, key, value)
+
+            return out_dict
+
+        # Define the formatted output dictionary.
+        out_dict = sorted_by_keys(in_dict=args_obj)
+
+        return out_dict
+
+    def run(self) -> Union[None, AttrDict]:
```
CoryMartin-NOAA, line 162: Just curious, what does this provide that `argparse` does not? I albeit am not an `argparse` expert but it seems like it is doing basically the same thing where:
`python thing.py --arg1=foo`
```
print(args.arg1)
foo
```
right?

HenryRWinterbottom: @CoryMartin-NOAA Yes, you are almost right. This functionality provides a wrapper over argparse. For example, if you had a script and wanted to add arguments to it using the traditional methodology, you would need to have a `parser.get_argument` for each variable that you wanted to pass.

This functionality negates that. As long as the `ArgsParse` object is invoked, it will collect all command line arguments. The user would then no longer need to explicitly define a `parser.get_argument` for each argument they wish to pass to a given script. They could pass anything that they want and the `ArgsParse` will format and and pass it to the respective task.

The advantage of this methodology is that it keeps the driver/calling scripts clean and does not the require a user/developer to need to add cumbersome directives to the script. 

I apologize that this wasn't more clear in my original description.

aerorahul: A simple `thing.py` example.
```python
from pygw.argsparse import ArgsParse

ap = ArgsParse().run()
print(ap)
```

```bash
$> python3 thing.py  --yaml 'myyaml.yaml'
2023-03-27 17:23:47,168 - INFO     - root        : Command line argument yaml is myyaml.yaml
{'yaml': 'myyaml.yaml'}
```

However,

```bash
$> python3 thing.py  'mode' --yaml 'myyaml.yaml'
2023-03-27 17:23:37,121 - INFO     - root        : Command line argument mode is --yaml
  File "/Users/rmahajan/scratch/gwWork/reviews/global-workflow/thing.py", line 7, in <module>
    ap = ArgsParse().run()
         ^^^^^^^^^^^^^^^^^
  File "/Users/rmahajan/scratch/gwWork/reviews/global-workflow/ush/python/pygw/src/pygw/argsparse.py", line 201, in run
    msg = f"Command line argument {arg_keys[idx]} is {arg_values[idx]}"
                                                      ~~~~~~~~~~^^^^^
IndexError: list index out of range
```

`ArgsParse` can only parse named command line arguments.


I am with @CoryMartin-NOAA.  This class provides some "convenience" over the standard invocation of `argparse.add_argument` and parser.parse_args()`,  but it does so without providing a better command-line interface.  For e.g.

```bash
$> python3 thing.py  --yaml 'myyaml.yaml' -h
usage: thing.py [-h]

options:
  -h, --help  show this help message and exit
```

Whereas, if the script `thing.py` needed or had options for `--yaml`, the developer would have had to define them explicitly.

I would argue that in the attempt to make it easy to pass named command line arguments, the API has been dangerously degraded.

Issue #1432, opened 2023-03-28
Python-based GFS atmosphere-only configuration


HenryRWinterbottom:
This issue is being open to provide an overview of PRs that will contribute to the corresponding epic for this task.

**Description**

The GFS atmosphere-only configuration epic will consist of the following steps/tasks/PRs.

- Build the task-level infrastructure; this includes task scripts beneath `jobs` and `scripts` to drive the related tasks;
- Build a high-level interface for the `initialize`, `run`, and `finalize` methods for the respective (GFS) UFS forecast model;
- Build the working directory tree for the UFS (GFS) forecast experiment;
- Add capabilities to link fixed-files;
- Add capabilities to build all necessary configuration files, namely `nems.configure` and `model_configure`;
- Add capabilities to build all necessary tables;
- Add capabilities to generate the UFS forecast model namelist `input.nml`;
- Add capabilities to sync/copy initial conditions files for the UFS;
- Add capabilities to run the forecast model task;
- Add capabilities to deliver UFS forecast model task products and clean up as necessary.

**Acceptance Criteria (Definition of Done)**

The above application is able to replicate a C48 global-workflow atmosphere-only (ATM) experiment.

Additional **done** criteria are a set of unit-tests to evaluate the respective steps above.

PR #1435, opened 2023-03-28
Feature/gw atm forecast init

HenryRWinterbottom:
**Description**

This PR addresses the following bullet points for issue #1432.

- Build the task-level infrastructure; this includes task scripts beneath jobs and scripts to drive the related tasks;
- Build a high-level interface for the initialize, run, and finalize methods for the respective (GFS) UFS forecast model;
- Build the working directory tree for the UFS (GFS) forecast experiment;

**Type of change**

**How Has This Been Tested?**

No exception was raised during application and directory tree was constructed as expected.
  
**Checklist**


ush/python/pyufs/ufswm/ufswm.py
```diff
@@ -0,0 +1,114 @@
...
+    Raises
+    ------
+
+    UFSWMError:
+
+        - raised if a specified forecast model is not supported.
+
+    """
+
+    def __init__(self: Task, config: object, model: str, app: str, *args, **kwargs):
```
aerorahul, line 91:
```suggestion
    def __init__(self: Task, config: Dict, model: str, app: str, *args, **kwargs):
```

`config` is a `Dict`

ush/python/pyufs/ufswm/ufswm.py
```diff
@@ -0,0 +1,114 @@
...
+# ----
+
+from pygw.logger import Logger
+from pygw.task import Task
+from pyufs.exceptions import UFSWMError
+
+# ----
+
+
+class UFSWM(Task):
```
aerorahul, line 43: This should not be a subclass of `Task`.  
This should be considered as the base class of the ufs-weather-model.

ush/python/pyufs/ufswm/gfs.py
```diff
@@ -0,0 +1,70 @@
...
+        Description
+        -----------
+
+        Creates a new GFS object.
+
+        """
+
+        # Define the base-class attributes.
+        model = "GFS"
+        super().__init__(config=config, model=model, app=config.APP)
```
aerorahul, line 70: what is the difference between `model` and `app`?

ush/python/pyufs/ufswm/gfs.py
```diff
@@ -0,0 +1,70 @@
+"""
```
aerorahul, line 1: I don't agree w/ the way the class hierarchy is being setup and defined.

aerorahul (changes requested):
1. Why are the `analysis.py` and `aero_analysis.py` removed?
2. why is `pygfs` deleted?  Please work within the structure setup.  If you wish to propose a new structure, please setup a discussion with the core team explaining how the current structure is limiting development.
3. We do not put History log, Authors in the doc block.  Please let `git` handle those.

Please see specific inline comments.

ush/python/pyufs/ufswm/gfs.py
HenryRWinterbottom: I thought this had been resolved during the Google Hangout with Walter Kolczynski. You had stated that you liked how GFS was a sub-class of UFSWM hence the defined hierarchy.

ush/python/pyufs/ufswm/gfs.py
HenryRWinterbottom: `model` is GFS here. `app` is `ATM` for atmosphere only configuration. I admit this is redundant given `APP` can be collected from the run-time environment. But, will this always be the case?

ush/python/pyufs/ufswm/ufswm.py
HenryRWinterbottom: I am following this: https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/python/pygfs/task/analysis.py#L17

What am I misunderstanding?

ush/python/pyufs/ufswm/ufswm.py
HenryRWinterbottom: This was a mistake. It should be beneath `pyufs/tasks`. I must have missed it during the copy/rename.

ush/python/pyufs/ufswm/gfs.py
aerorahul: May be I wasn't clear.
We need a `forecast.py` task that is a generic forecast task.  It contains a `class Forecast(Task)`. IMO, we don't really need it, but if you want to introduce that separation, this would be it.

The `gfs_forecast.py` contains a `class GFSForecast(Forecast, GFS)`

`GFS` is a class from `gfs.py` that contains the methods and definitions specific to the global configuration of the  ufs-weather-model.  It is derived from `ufs.py`, such that `class GFS(UFS)` and `class UFS` is the base class for the ufs-weather-model.

HenryRWinterbottom: Is that not what I have done here: https://github.com/HenryWinterbottom-NOAA/global-workflow/blob/feature/gw_atm_forecast_init/ush/python/pyufs/ufswm/ufswm.py#L43

and here: https://github.com/HenryWinterbottom-NOAA/global-workflow/blob/feature/gw_atm_forecast_init/ush/python/pyufs/ufswm/gfs.py#L41

What am I not understanding? Is this a naming issue or is there something more fundamental?

PR #1455, opened 2023-04-12
Rework arch job dependencies

### LINKED ISSUE #1451: setup_xml.py fails if arch jobs have no dependencies
**Expected behavior**

A user should be able to disable vrfy jobs.

**Current behavior**

If the vrfy tasks are disabled (`DO_VRFY="NO"`) and no other pre-arch jobs are enabled in its place, then setup_xml.py fails as there are no dependencies for the arch job.

**Machines affected**

All

**To Reproduce**

1) Run setup_expt.py for a cycled, uncoupled experiment
2) Modify config.base, setting `DO_VRFY="NO"`
3) Run setup_xml.py

**Context**

**Detailed Description**

The failure occurs in rocoto.py, line [315](https://github.com/NOAA-EMC/global-workflow/blob/d5ae3328fa4041b177357b1133f6b92e81c859d7/workflow/rocoto/rocoto.py#L315).  `dep[0] is not None` is invalid when `dep` is length 0.  Another issue here is that without the `vrfy` tasks, the `arch` tasks would have no dependency and would run immediately.

**Possible Implementation**

There are a few issues to fix.  First, create_dependency should be rewritten to check for `len(dep) > 0` instead of `dep[0] is not None`.  The `append` calls to `strings` and `dep_condition` if blocks should be moved to be within the `if len(dep) > 0` blocks to so that empty dependencies are not created, which will result in failures when running rocoto.  A warning should also be added to either `create_dependency` or `create_task` to notify if there are no dependencies for a task.  Lastly, within the `arch` method of the `Tasks` class of workflow_tasks.py, if all of the variables `DO_VRFY`, `DO_METP` (for gfs cycles), `DO_WAVE`, and `DO_OCEAN` are set to `"NO"`, then a dependency for the `{gdas|gfs}post` metatask should be added.

DavidHuber-NOAA:
**Description**
This reworks the dependencies for arch tasks so that if there are no verification tasks selected and it is an uncoupled experiment, a dependency for the cycle's post jobs is added.  It also fixes the dependency checks in rocoto.py to check for an empty list or string instead of checking the first element of a list, which may not exist if no dependencies are given.  Lastly, it issues a warning to the user if there are no dependencies for any job.

Fixes #1451

**Type of change**

**How Has This Been Tested?**
I created three test cases, each comparing develop against the current develop branch:

1. cycled test with `DO_VRFY=YES`: .xml files were identical excepting the creation time
2. cycled test with `DO_VRFY=NO` and `DO_METP=NO`: develop fails to create .xml while this branch succeeds and lists `{gdas|gfs}post` as dependencies
3. forecast only, uncoupled test with `DO_VRFY=NO`: develop fails and this branch succeeds with `{gdas|gfs}post` listed as dependencies
4. commented out the `post` dependencies in rocoto.py, [lines 1039-1041](https://github.com/DavidHuber-NOAA/global-workflow/blob/b9487d3d15a7832a5a1c95f613d5f949956e77b4/workflow/rocoto/workflow_tasks.py#L1039-L1041) and ran setup_xml.py with this branch and `DO_VRFY=NO`: the .xml file was generated successfully without any dependencies for the `*arch` jobs and a warning was produced, as expected.

A note from testing: when `DO_VRFY=NO`, the `*vrfy` jobs are still created in the resulting rocoto XML.  This doesn't seem like expected behavior and I can tackle it as part of this PR if desired.

**Checklist**

### DIFF
--- a/workflow/rocoto/rocoto.py
+++ b/workflow/rocoto/rocoto.py
@@ -109,12 +109,14 @@ def create_task(task_dict: Dict[str, Any]) -> List[str]:
             strings.append(f'\t{e}\n')
         strings.append('\n')
 
-    if dependency is not None:
+    if len(dependency) > 0:
         strings.append('\t<dependency>\n')
         for d in dependency:
             strings.append(f'\t\t{d}\n')
         strings.append('\t</dependency>\n')
         strings.append('\n')
+    else:
+        print("WARNING: No dependencies for task " + taskname)
 
     strings.append('</task>\n')
 
@@ -293,7 +295,7 @@ def _traverse(o, tree_types=(list, tuple)):
         yield o
 
 
-def create_dependency(dep_condition=None, dep=None) -> List[str]:
+def create_dependency(dep_condition=None, dep=[]) -> List[str]:
     """
     create a compound dependency given a list of dependencies, and compounding condition
     the list of dependencies are created using add_dependency
@@ -309,10 +311,10 @@ def create_dependency(dep_condition=None, dep=None) -> List[str]:
 
     strings = []
 
-    if dep_condition is not None:
-        strings.append(f'<{dep_condition}>')
+    if len(dep) > 0:
+        if dep_condition is not None:
+            strings.append(f'<{dep_condition}>')
 
-    if dep[0] is not None:
         for d in dep:
             if dep_condition is None:
                 strings.append(f'{d}')
@@ -320,8 +322,8 @@ def create_dependency(dep_condition=None, dep=None) -> List[str]:
                 for e in _traverse(d):
                     strings.append(f'\t{e}')
 
-    if dep_condition is not None:
-        strings.append(f'</{dep_condition}>')
+        if dep_condition is not None:
+            strings.append(f'</{dep_condition}>')
 
     return strings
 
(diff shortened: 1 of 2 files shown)


aerorahul (approved):
looks good.
Thanks for adding the contingency for an empty list of dependencies.

workflow/rocoto/rocoto.py
```diff
@@ -109,12 +109,14 @@ def create_task(task_dict: Dict[str, Any]) -> List[str]:
             strings.append(f'\t{e}\n')
         strings.append('\n')
 
-    if dependency is not None:
+    if len(dependency) > 0:
```
WalterKolczynski-NOAA: You are checking the length of `dependency`, but if it is `None` it has no length and will throw an exception. I tested in an interactive session to make sure this wasn't an instance of it being considered "truthy" enough to work:

```
>>> d=None
>>> if len(d) > 0:
...     print(True)
... else:
...     print(False)
...
  File "<stdin>", line 1, in <module>
TypeError: object of type 'NoneType' has no len()
```

The default value for `dependency` should be an empty list now instead of `None`. Also need to check all the calls to this function to make sure none of the tasks are passing `None` in.

WalterKolczynski-NOAA (changes requested):

WalterKolczynski-NOAA (approved):

PR #1456, opened 2023-04-12
Move Fit2Obs to stand-alone job

### LINKED ISSUE #1058: FITS submitted from vrfy.sh
**Expected behavior**
`config.vrfy` allows the user to specify the `QUEUE` to use for the FITS job.   
```
if [ $VRFYFITS = "YES" ]; then

    export fit_ver="newm.1.3"
    export fitdir="$BASE_GIT/verif/global/Fit2Obs/${fit_ver}/batrun"
    export PRVT=$HOMEgfs/fix/gsi/prepobs_errtable.global
    export HYBLEVS=$HOMEgfs/fix/am/global_hyblev.l${LEVS}.txt
    export CUE2RUN=$QUEUE
```
Therefore, expect the FITS job to run in the user specified queue.

**Current behavior**
While `subfit_orion_netcdf` sets `CUE2RUN` based on user input
```
ACCOUNT=${ACCOUNT:-CFS-T2O}
CUE2RUN=${CUE2RUN:-dev}
KEEPDATA=${KEEPDATA:-NO}
```
it does not use the user set `CUE2RUN`.  Instead it hardwires the FITS job to use the batch queue.
```
#!/bin/bash --login
#SBATCH --job-name=FITS.$EXPNAM.$CDATE --time=20:00
#SBATCH --nodes=1 --ntasks=4
#SBATCH --exclusive
#SBATCH --output=$COMDAY/FITS.$EXPNAM.$CDATE.$$
#SBATCH --account=$ACCOUNT
#SBATCH --partition=orion
#SBATCH --qos batch
```

**Machines affected**
Orion

**To Reproduce**
Set up the `develop` g-w on Orion.   As a test, set `QUEUE=debug` in `config.base`.   Submit `gdasvrfy`.   The FITS job will be submitted to the batch queue even though `CUE2RUN=QUEUE=debug` is exported to `subfits_orion_netcdf` .  

**Additional Information**
Script `subits_orion_netcdf` resides [NOAA-EMC/Fit2Obs](https://github.com/NOAA-EMC/Fit2Obs).  Resolution of this issue will require changes in the Fit2Obs repo.    The issue is opened here since g-w submits the FITS job.

### LINKED ISSUE #1232: GDAS verify does not work on WCOSS2
**Expected behavior**
The gdasvrfy job should complete successfully on WCOSS2.

**Current behavior**
The gdasvrfy job fails on WCOSS2.

**Machines affected**
Currently a blocker for WCOSS2, but updating to newm 1.5 will impact all machines.

**To Reproduce**
1.  Setup a cycled experiment on WCOSS2
2. Watch gdasvrfy fail during the first full cycle

**Context**
There is no entry in the `config.vrfy` for the `$PREPQFITSH` (Fit2Obs) script. Develop is still using newm 1.3. While this version is on WCOSS2, it does not have a `subfits_` script for wcoss2, so it wouldn't run there anyway.

**Possible Implementation**
Update to newm 1.5, which is present on all three primary machines and add an entry for WCOSS2. The structure of the git directory has changed for 1.5, removing the `verif/global` from the path (instead of `${BASE_GIT}/verif/global/Fit2Obs/` it is now simply `${BASE_GIT}/Fit2Obs`. Also, the name of the script has changed, so instead of a big elif, the path can simply be `${fitdir}/subfits_${machine,,}`.

An initial test suggests additional downscript updates are required to update to 1.5, so it is not simply just pointing to the new path.

### LINKED ISSUE #1405: Create dedicated Fit2Obs job
**Description**

The fit2obs package currently runs as part of the vrfy job and is submitted from within that job. Move fit2obs out of the vrfy job and into its own job. This work should help resolve issue #1232 and broken functionality on R&Ds machines.

- [x] Create `parm/config/config.fit2obs`
- [x] Create `jobs/rocoto` script (`fit2obs.sh`)
- [x] Create JJOB script (`JGDAS_FIT2OBS`)
- [x] Add `DO_FIT2OBS` switch in `config.base`
- [x] Add fit2obs task to gdas tasks in setup scripts
- [x] Add fit2obs task to `config.resources`
- [x] Add fit2obs task to env files.
- [x] Test new job on WCOSS2
- [x] Test new job on Hera
- [x] Test new job on Orion
- [x] Check if resolution still needed for #311

**Requirements**

Updates to fit2obs repo needed for functionality on WCOSS2. See https://github.com/NOAA-EMC/Fit2Obs/issues/4.

Fit2obs package is run and produces correct output from new separate job.

**Dependencies**

- [x] Completion of https://github.com/NOAA-EMC/Fit2Obs/issues/4.
- [x] Install new Fit2Obs `wflow.1.0` tag on supported platforms (WCOSS2s, Hera, Orion, Jet).

KateFriedman-NOAA:
**Description**

This PR moves the Fit2Obs invocation out of the vrfy job and into its own dedicated `fit2obs` job in the gdas suite. This new dedicated job uses the latest Fit2Obs tag `wflow.1.0`.

The Fit2Obs requires a type of spin-up. The job looks back `VBACKUP_FITS` hrs and needs available inputs for that lookback cycle in the `ROTDIR`. The `jobs/JGDAS_FIT2OBS` script will first check that `xdate` (`CDATE` - `VBACKUP_FITS`) > `SDATE` and then if met, will check that the needed inputs exist.
- If `xdate>SDATE` is not yet satisfied, the job will exit 0 with "Too early for FIT2OBS to run. Exiting.". The conditional is greater-than and not greater-than-or-equal since the first half cycle generally does not have some of the needed inputs (e.g. prepbufr). Thus the first half cycle is not included in the valid lookback cycles. This avoids erroneous job failures for the first cycle to run the Fit2Obs package. Additional logic could be introduced to include the half cycle if all available inputs are available.
- If any of the needed inputs are missing the job will abort with "FATAL ERROR: FILE MISSING: ${file}". 

This spin-up means that the first cycles will run the job but exit 0 immediately. The 6th cycle (if `VBACKUP_FITS=24`) will be the first cycle to run the Fit2Obs package and produce output in the online archive.

Changes:

1. Remove fit2obs variables and settings from `config.vrfy` and into newly created `config.fit2obs` for `fit2obs` job.
2. Remove fit2obs submission/invocation from `jobs/rocoto/vrfy.sh`.
3. Create new `fit2obs` job scripts: `jobs/rocoto/fit2obs.sh` and `jobs/JGDAS_FIT2OBS`
4. Add new `fit2obs` job to setup scripts: `workflow/applications.py` and `workflow/rocoto/workflow_tasks.py`
5. Add new `fit2obs` job to all env files.
6. Add new `fit2obs` job into `config.resources` (use 1 node on WCOSS2 and 3 nodes elsewhere).
7. Add `export DO_FIT2OBS="YES"` to `config.base.emc.dyn`.

Resolves #1405
Resolves #1232
Resolves #1058

**Type of change**

**How Has This Been Tested?**

- [x] Clone and build tests on all supported platforms
- [x] Cycled tests on Hera, Orion, WCOSS2, and Jet (Jet tests done by @DavidHuber-NOAA )
  
**Checklist**


env/HERA.env
```diff
@@ -291,4 +292,14 @@ elif [[ "${step}" = "gempak" ]]; then
...
     export APRUN="${launcher} -n ${npe_gempak} ${mpmd_opt}"
+
+
+elif [[ "${step}" = "fit2obs" ]]; then
+
+    nth_max=$((npe_node_max / npe_node_fit2obs))
+
+    export NTHREADS_FIT2OBS=${nth_fit2obs:-1}
+    [[ ${NTHREADS_FIT2OBS} -gt ${nth_max} ]] && export NTHREADS_FIT2OBS=${nth_max}
+    export MPIRUN="${launcher} -n ${npe_fit2obs}"
```
aerorahul, line 303: too bad they use `MPIRUN` instead of `APRUN`

jobs/JGDAS_FIT2OBS
```diff
@@ -0,0 +1,83 @@
...
+else
+
+  echo "Too early for FIT2OBS to run. Exiting."
+
+fi
+
+##########################################
+# Remove the Temporary working directory
+##########################################
+cd "${DATAROOT}" || (echo "${DATAROOT} does not exist. ABORT!"; exit 1)
```
aerorahul:
```suggestion
cd "${DATAROOT}" || (echo "FATAL ERROR: ${DATAROOT} does not exist. ABORT!"; exit 1)
```

KateFriedman-NOAA:
Working to resolve shellcheck warnings...

parm/config/config.fit2obs
```diff
@@ -0,0 +1,30 @@
...
+
+# Get task specific resources
+. $EXPDIR/config.resources fit2obs
+
+export fit_ver="wflow.1.0"
+export fitdir="${BASE_GIT}/Fit2Obs/${fit_ver}"
+
+export HOMEcfs=${fitdir}
+export EXECcfs=${HOMEcfs}/exec
+export USHcfs=${HOMEcfs}/ush
```
aerorahul, line 16: Are these even used?

jobs/JGDAS_FIT2OBS
WalterKolczynski-NOAA, line 1: A lot of this should be pushed down out of the j-job to an exscript.

jobs/JGDAS_FIT2OBS
```diff
@@ -0,0 +1,83 @@
...
+
+
+##############################################
+# Set variables used in the script
+##############################################
+
+export CDATE=${CDATE:-${PDY}${cyc}}
+export CDUMP=${CDUMP:-${RUN:-"gdas"}}
+
+export xdate=$(${NDATE} -${VBACKUP_FITS} ${CDATE})
```
aerorahul:
```suggestion
export xdate=$(${NDATE} -${VBACKUP_FITS} ${PDY}${cyc})
```

jobs/JGDAS_FIT2OBS
```diff
@@ -0,0 +1,83 @@
...
+export CDATE=${xdate} # Reset CDATE to xdate for lookback
+export vday=$(echo ${xdate} | cut -c1-8)
+export vcyc=$(echo ${xdate} | cut -c9-10)
```
WalterKolczynski-NOAA:
```suggestion
export vday=${xdate:0:8}
export vcyc=${xdate:8:2}
```

jobs/JGDAS_FIT2OBS
```diff
@@ -0,0 +1,83 @@
...
+
+source "${HOMEgfs}/ush/preamble.sh"
+source "${HOMEgfs}/ush/jjob_header.sh" -e "fit2obs" -c "base fit2obs"
+
+
+##############################################
+# Set variables used in the script
+##############################################
+
+export CDATE=${CDATE:-${PDY}${cyc}}
```
aerorahul:
```suggestion
```

jobs/JGDAS_FIT2OBS
```diff
@@ -0,0 +1,83 @@
...
+export CDUMP=${CDUMP:-${RUN:-"gdas"}}
+
+export xdate=$(${NDATE} -${VBACKUP_FITS} ${CDATE})
+export CDATE=${xdate} # Reset CDATE to xdate for lookback
```
aerorahul:
```suggestion
export CDATE=${xdate} # set CDATE to xdate for use in fit2obs package
```

WalterKolczynski-NOAA: Is the external package using `$CDATE` or `$xdate`? Or neither? Don't export things that aren't needed downstream.

jobs/JGDAS_FIT2OBS
```diff
@@ -0,0 +1,83 @@
...
+export vday=$(echo ${xdate} | cut -c1-8)
+export vcyc=$(echo ${xdate} | cut -c9-10)
+
+export COM_INA=${ROTDIR}/gdas.${vday}/${vcyc}/atmos
+export COM_INF='$ROTDIR/vrfyarch/gfs.$fdy/$fzz'
+export COM_PRP=${ROTDIR}/gdas.${vday}/${vcyc}/obs
+
+export PRPI=${COM_PRP}/${CDUMP}.t${vcyc}z.prepbufr
+export sig1=${COM_INA}/${CDUMP}.t${vcyc}z.atmanl.nc
+export sfc1=${COM_INA}/${CDUMP}.t${vcyc}z.atmanl.nc
```
aerorahul: Is this a mistake?
```suggestion
export sfc1=${COM_INA}/${CDUMP}.t${vcyc}z.sfcanl.nc
```

jobs/JGDAS_FIT2OBS
WalterKolczynski-NOAA:
```suggestion
```

jobs/JGDAS_FIT2OBS
```diff
@@ -0,0 +1,83 @@
...
+source "${HOMEgfs}/ush/preamble.sh"
+source "${HOMEgfs}/ush/jjob_header.sh" -e "fit2obs" -c "base fit2obs"
+
+
+##############################################
+# Set variables used in the script
+##############################################
+
+export CDATE=${CDATE:-${PDY}${cyc}}
+export CDUMP=${CDUMP:-${RUN:-"gdas"}}
```
WalterKolczynski-NOAA:
```suggestion
export CDUMP=${RUN/enkf}
```

jobs/JGDAS_FIT2OBS
```diff
@@ -0,0 +1,83 @@
...
+
+export xdate=$(${NDATE} -${VBACKUP_FITS} ${CDATE})
+export CDATE=${xdate} # Reset CDATE to xdate for lookback
+export vday=$(echo ${xdate} | cut -c1-8)
+export vcyc=$(echo ${xdate} | cut -c9-10)
+
+export COM_INA=${ROTDIR}/gdas.${vday}/${vcyc}/atmos
+export COM_INF='$ROTDIR/vrfyarch/gfs.$fdy/$fzz'
```
aerorahul, line 23: I notice this is in `'` quotes, so `$fdy` and `$fzz` won't get resolved here.
Where are these used?

jobs/JGDAS_FIT2OBS
```diff
@@ -0,0 +1,83 @@
...
+
+export FIT_DIR=${ARCDIR}/fits
+[[ ! -d "${FIT_DIR}" ]] && mkdir -p "${FIT_DIR}"
+export HORZ_DIR=${ARCDIR}/horiz
+[[ ! -d "${HORZ_DIR}" ]] && mkdir -p "${HORZ_DIR}"
+export COMLOX=${DATA}/fitx
+[[ ! -d "${COMLOX}" ]] && mkdir -p "${COMLOX}"
+
+echo "echo err_chk">${DATA}/err_chk; chmod 755 ${DATA}/err_chk
+echo "echo postmsg">${DATA}/postmsg; chmod 755 ${DATA}/postmsg
```
WalterKolczynski-NOAA: Why?

jobs/JGDAS_FIT2OBS
```diff
@@ -0,0 +1,83 @@
...
+  # Final processing
+  ##############################################
+  if [[ -e "${pgmout}" ]] ; then
+    cat "${pgmout}"
+  fi
+
+else
+
+  echo "Too early for FIT2OBS to run. Exiting."
+
```
aerorahul, line 79: I would flip this other way around and right up front after `xdate` is calculated like so:

```bash
if [[ "${xdate}" -le "${SDATE:-}" ]]; then
    echo "Too early for FIT2OBS to run. Exiting."
fi
``` 

And then here, you can remove the big `if-block`

aerorahul (changes requested):
Some comments.  Some suggestions, and a couple questions.
Looks good otherwise.

parm/config/config.resources
```diff
@@ -623,6 +623,14 @@ elif [ ${step} = "vrfy" ]; then
     fi
     export is_exclusive=True
 
+elif [ ${step} = "fit2obs" ]; then
+
+    export wtime_fit2obs="00:20:00"
+    export npe_fit2obs=3
+    export nth_fit2obs=1
+    export npe_node_fit2obs=1
+    if [[ ${machine} == "WCOSS2" ]]; then export npe_node_fit2obs=3 ; fi
```
WalterKolczynski-NOAA, line 632: Why does WCOSS require three full nodes vs only 3 CPU on others?

workflow/rocoto/workflow_tasks.py
```diff
@@ -1020,6 +1034,9 @@ def arch(self):
         if self.app_config.do_vrfy:
             dep_dict = {'type': 'task', 'name': f'{self.cdump}vrfy'}
             deps.append(rocoto.add_dependency(dep_dict))
+        if self.app_config.do_fit2obs and self.cdump in ['gdas']:
+            dep_dict = {'type': 'task', 'name': f'{self.cdump}fit2obs'}
+            deps.append(rocoto.add_dependency(dep_dict))
```
WalterKolczynski-NOAA, line 1039: Does metp actually depend on fit2obs?

jobs/JGDAS_FIT2OBS
WalterKolczynski-NOAA:
```suggestion
export xdate=$(${NDATE} -"${VBACKUP_FITS}" "${PDY}${cyc}")
```

WalterKolczynski-NOAA (changes requested):

jobs/JGDAS_FIT2OBS
```diff
@@ -0,0 +1,82 @@
...
+export xdate=$(${NDATE} -"${VBACKUP_FITS}" "${CDATE}")
+export CDATE=${xdate} # Reset CDATE to xdate for lookback
+export vday=$(echo "${xdate}" | cut -c1-8)
+export vcyc=$(echo "${xdate}" | cut -c9-10)
+
+export COM_INA=${ROTDIR}/gdas.${vday}/${vcyc}/atmos
+export COM_INF="$ROTDIR/vrfyarch/gfs.$fdy/$fzz"
+export COM_PRP=${ROTDIR}/gdas.${vday}/${vcyc}/obs
+
+export PRPI=${COM_PRP}/${CDUMP}.t${vcyc}z.prepbufr
```
WalterKolczynski-NOAA: All these file names should use `${RUN}`, not `${CDUMP}`

jobs/JGDAS_FIT2OBS
WalterKolczynski-NOAA: Also: is `$CDUMP` even needed?

jobs/JGDAS_FIT2OBS
KateFriedman-NOAA: This was lifted right from the Fit2Obs scripts: https://github.com/NOAA-EMC/Fit2Obs/blob/develop/batrun/excfs_gdas_vrfyfits.sh.ecf#L64
It may be a mistake in Fit2Obs package. @jack-woollen should the `sfc1` file be the `sfcanl.nc` file instead of the `atmanl.nc` file?

jobs/JGDAS_FIT2OBS
KateFriedman-NOAA: The `jobs/JGDAS_ENKF_ARCHIVE` and `jobs/JGLOBAL_ARCHIVE` scripts should also have a similar update. I will commit the suggestion into `JGDAS_FIT2OBS` though.

parm/config/config.fit2obs
KateFriedman-NOAA: They are used within the Fit2Obs package.

jobs/JGDAS_FIT2OBS
jack-woollen: @KateFriedman-NOAA Well its not exactly correct, but a do no harm workaround. I need to write code to process the netcdf sfcanl, not a high priority at the moment.

parm/config/config.fit2obs
aerorahul: ok. These may need a new definition as fit2obs conforms to EE2.  But for a later PR.

jack-woollen: I'll work on changing the fit2obs package to conform in the not too distant future.

jobs/JGDAS_FIT2OBS
KateFriedman-NOAA: Done at bb7471f2. Thanks!

jobs/JGDAS_FIT2OBS
KateFriedman-NOAA: The `COM_INF` needs to be passed into the Fit2Obs package with single quotes and then be evaluated there when `fdy` and `fzz` are set. I tried setting this with double quotes (based on shellcheck warning) but then it tries to evaluate it at the workflow level which fails since `fdy` and `fzz` aren't known yet.

jobs/JGDAS_FIT2OBS
```diff
@@ -0,0 +1,82 @@
...
+
+export CDUMP=${RUN/enkf}
+
+export xdate=$(${NDATE} -"${VBACKUP_FITS}" "${CDATE}")
+export CDATE=${xdate} # set CDATE to xdate for use in fit2obs package
+export vday=$(echo "${xdate}" | cut -c1-8)
+export vcyc=$(echo "${xdate}" | cut -c9-10)
+
+export COM_INA=${ROTDIR}/gdas.${vday}/${vcyc}/atmos
+export COM_INF='$ROTDIR/vrfyarch/gfs.$fdy/$fzz'
```
WalterKolczynski-NOAA, line 23:
```suggestion
# We want to defer variable expansion, so ignore warning about single quotes
# shellcheck disable=SC2016
export COM_INF='$ROTDIR/vrfyarch/gfs.$fdy/$fzz'
```

jobs/JGDAS_FIT2OBS
WalterKolczynski-NOAA: Made a suggestion to suppress the shellcheck warning

KateFriedman-NOAA: Works for me, thanks @WalterKolczynski-NOAA !

jobs/JGDAS_FIT2OBS
KateFriedman-NOAA: This would need an `exit 0` too, right? If we exit right here then the `DATAROOT` that is created by `jjob_header.sh` at the start of the script won't be removed at the end of `JGDAS_FIT2OBS`. That's why I set up the if-block like I did.

jobs/JGDAS_FIT2OBS
KateFriedman-NOAA: I'm treating the fit2obs job somewhat similarly to how we run the obsproc/prepobs package and invoking the external package driver script (`"${fitdir}/batrun/excfs_gdas_vrfyfits.sh.ecf"`) from the JJOB script. For the prep job we invoke the JJOB from that package from our `prep.sh` but there isn't a JJOB for fit2obs package so I created one and included the settings we need before invoking `excfs_gdas_vrfyfits.sh.ecf`. I didn't see a need to add another exscript and making the JJOB essentially do nothing but invoke it. Let me know if I'm missing a greater point though.

env/HERA.env
KateFriedman-NOAA: Perhaps something we can change in a later Fit2Obs PR?

jobs/JGDAS_FIT2OBS
KateFriedman-NOAA: Yes, the Fit2Obs package uses `CDATE` throughout so we have to set it from above.

parm/config/config.resources
KateFriedman-NOAA: WCOSS2 only uses 1 node (npe_fit2obs / npe_node_fit2obs = 3/3 = 1), whereas the R&Ds use 3 nodes (3/1 =3). This is to get enough memory.

workflow/rocoto/workflow_tasks.py
KateFriedman-NOAA: This is me adding fit2obs to the archive job dependency, not metp.

jobs/JGDAS_FIT2OBS
```diff
@@ -0,0 +1,84 @@
...
+source "${HOMEgfs}/ush/jjob_header.sh" -e "fit2obs" -c "base fit2obs"
+
+
+##############################################
+# Set variables used in the script
+##############################################
+
+export CDUMP=${RUN/enkf}
+
+export xdate=$(${NDATE} -"${VBACKUP_FITS}" "${CDATE}")
```
WalterKolczynski-NOAA:
```suggestion
# Ignore spelling warning; nothing is misspelled
# shellcheck disable=SC2155
xdate=$(${NDATE} -"${VBACKUP_FITS}" "${PDY}${cyc}")
export xdate
```

jobs/JGDAS_FIT2OBS
KateFriedman-NOAA: The Fit2Obs package uses `CDATE` throughout whereas `xdate` is just used at the workflow level to reset `CDATE` to the lookback cycle and to set `vday` and `vcyc` before invoking the Fit2Obs scripts. A lot of this is lifted from how we were doing it in the vrfy job, it can be rethought if needed.

WalterKolczynski-NOAA: I would just replace xdate with CDATE then to avoid the duplicate variable, although I don't like fit2obs using CDATE rather than VDATE.

parm/config/config.resources
WalterKolczynski-NOAA: I always forget it is defined backwards. When we rewrite the resources, we need to improve the variable name.

workflow/rocoto/workflow_tasks.py
WalterKolczynski-NOAA: d'oh, missed the code break where GH skipped lines

jobs/JGDAS_FIT2OBS
WalterKolczynski-NOAA: Do you mean `CDUMP`? If they are using it for paths where they should be using `${RUN}`, that could cause problems if they are looking at ensemble members.

jobs/JGDAS_FIT2OBS
KateFriedman-NOAA: Committed change, thanks @WalterKolczynski-NOAA !

jobs/JGDAS_FIT2OBS
KateFriedman-NOAA: Good call, done at 3b48ae11. Thanks!

jobs/JGDAS_FIT2OBS
KateFriedman-NOAA: I took this from the Fit2Obs subfits scripts, which we now go around. I ran without these two lines and it fails. something in the fit2obs scripts needs this. Can explore in a follow-up PR how to do this better in fit2obs package and remove these two lines.

WalterKolczynski-NOAA (approved):

parm/config/config.fit2obs
aerorahul: for now, I can define these in the fit2obs modulefile so they don't need to appear here.

aerorahul (approved):
looks good.

PR #1465, opened 2023-04-14
Update buoys file and fix boundary point jobs

### LINKED ISSUE #1464: Change wave buoy file to use symlink and point to full buoy list
**Description**
The current wave buoy file used ([wave_gfs.buoys](https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/wave/wave_gfs.buoys)) is just a copy of [wave_gfs.buoys.dat](https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/wave/wave_gfs.buoys.dat). Keeping a duplicate file is both wasteful and allows the two copies to become out-of-sync.

Also, that file is an abbreviated version of the full buoy list that excludes all boundary points, which prevents testing of boundary point jobs in the workflow.

**Requirements**

Replace the copy of wave_gfs.buoys.dat with a symlink to wave_gfs.buoys.full. This will eliminate the duplicate file, keep the provenance of wave_gfs.buoys, and allow testing of wave boundary jobs. Users wishing to run with the shortened buoy list can just change the symlink.

WalterKolczynski-NOAA:
**Description**
The buoy file used by the wave jobs ([wave_gfs.buoys](https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/wave/wave_gfs.buoys)) was just a copy of [wave_gfs.buoys.dat](https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/wave/wave_gfs.buoys.dat). In addition to being a duplicate, the file was a truncated version without any boundary points, causing boundary point jobs to fail.

The duplicate file has been removed and replaced by a symlink to the full buoy list [wave_gfs.buoys.full](https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/wave/wave_gfs.buoys.full). This maintains the provenance of the file and prevents the former duplicate from becoming out-of-sync. Users who still want to use the truncated buoy can change the target of the symlink to wave_gfs.buoys.dat.

There are also a few minor bugs fixes that were necessary to get boundary point jobs to run:
- `FHMAX_WAV_IBP` had been set in the bndpnt config file but not used in the j-job. This was invisible unless a user changed the value, since the config and j-job used the same default.
- Checks against `FHMAX_WAV` would set the unused `FHMAX_WAV_IBP` to the max value instead of the `FHMAX_WAV_PNT` used for the loop. This is a problem when running for less than 180 h (the default value).
- The boundary point bulletin job was not in the env job list for Orion (other machines have it).
    
Now the boundary point jobs set `FHMAX_WAV_PNT` to `$FHMAX_WAV_IBP`. `FHMAX_WAV_IBP` was moved from the bndpnt config to config.wave so it is visible to bndpntbll as well.

Fixes #1464

**Type of change**

**How Has This Been Tested?**
- [x] Coupled test on Orion
  
**Checklist**


JessicaMeixner-NOAA (approved):
Thanks for pulling these changes into this separate PR!

KateFriedman-NOAA (approved):

Issue #1434, opened 2023-03-28
Addition of GEFS WAV grid


NeilBarton-NOAA:
**Description**
Addition of WAV grid used for GEFS (0.25 lat/lon)

**Requirements**
- inclusion of the wave grid inp file and scripting updates to run a long forecast with the GEFS wave grid
- additional wave .inp file in the fix directories

**Acceptance Criteria (Definition of Done)**
- S2SW fcst-only run with the updated grid

PR #1488, opened 2023-04-19
adding GEFS wave grid in S2SW

NeilBarton-NOAA:
**Description**

Addition of a WAV grid for GEFSv13. Addresses issue https://github.com/NOAA-EMC/global-workflow/issues/1434

- Note, new wave grid files are needed in the fix directory before these updates will be successful 

 

**Type of change**

**How Has This Been Tested?**

This code was tested on hera using the S2SW app in forecast-only mode. Restarts can be found at /scratch2/NCEPDEV/stmp3/Neil.Barton/ICs/2019120300. No wave restart was used.

- [X] Forecast-only test on Hera

  
**Checklist**


parm/config/config.defaults.s2sw
```diff
@@ -16,13 +16,13 @@ use_cice_alb=".true."
 
 # config.wave
 
-waveGRD='mx025'
```
JessicaMeixner-NOAA: Were ICs added so that this will run the C384 tests out of the box?  We need to add ICs or add this as an option instead of the default.  We really need to add a more configurable wave grid option as GFS needs this to be mx025 versus the GEFS option.

scripts/exgfs_wave_post_gridded_sbs.sh
```diff
@@ -402,7 +402,7 @@
     cd $DATA
 
     FHINCG=$(( DTFLD_WAV / 3600 ))
-    if [ $fhr = $fhrg ]
+    if [ $fhr = $fhrg ] && ([ "$DOGRB_WAV" = 'YES' ] || [ "$DOGRI_WAV" = 'YES' ])
```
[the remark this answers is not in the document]
JessicaMeixner-NOAA: @NeilBarton-NOAA can you help me understand the updates to this file? I was expecting lines adding  gefsv13_025 but not these.

parm/config/config.defaults.s2sw
WalterKolczynski-NOAA: If this is just for GEFS, the setting changes in this file should be made to GEFS and not here. Other changes to handle the grid should be here though.

CC: @XianwuXue-NOAA

parm/config/config.ufs
```diff
@@ -354,6 +354,9 @@ if [[ "${skip_ww3}" == "false" ]]; then
     "mx025")
       ntasks_ww3=80
       ;;
+    "gefsv13_025")
+      ntasks_ww3=262
```
WalterKolczynski-NOAA: Probably this one too.

WalterKolczynski-NOAA (changes requested):
Plus the shellcheck warnings.

How is gefsv13_025 different from mx025? The name points towards them both being nominally ¼°, but gefs_v13 is requiring triple the resources, which seems like a pretty steep increase. Is this just balancing the load or does it actually consume that much more compute? GEFS already seems to be having trouble staying within the allotted footprint.

parm/config/config.ufs
JessicaMeixner-NOAA: I think it'd be nice to have this setting here even though it's "for GEFS".    The resources are probably to get it to run at production speed versus "just finish in some reasonable sort of time frame"//be small enough resources for quick testing.

JessicaMeixner-NOAA:
> How is gefsv13_025 different from mx025? 

gefsv13_025 is the regular lat/lon grid planned for GEFS where as mx025 is a tripolar grid we were using for HR1.  mx025 would require too many resources to be used for operations, therefore the new grid was made.

scripts/exgfs_wave_post_gridded_sbs.sh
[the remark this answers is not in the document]
NeilBarton-NOAA: @JessicaMeixner-NOAA DOGRB_WAV and DOGRI_WAV are both no, but the script would still double check for the grib2 file and error out. Specifically errors out on GRDNAME not being defined. This line says only double check for the grib2 files if DOGRB_WAV and DOGRI_WAV is YES.

parm/config/config.defaults.s2sw
NeilBarton-NOAA: Should I change the default back or what for further discussion?

scripts/exgfs_wave_post_gridded_sbs.sh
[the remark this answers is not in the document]
JessicaMeixner-NOAA: I see thanks for the explanation.  Quick question if both DOGRB_WAV and DOGRI_WAV are no, I'm curious as to why we would run the gridded post processing job? For your case, I would think DOGRB_WAV=YES while DOGRI_WAV=NO.   Do I need to add some text to the scripts to better describe these variables?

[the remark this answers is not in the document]
NeilBarton-NOAA: I think the variables are easy enough to follow, but agree on a better solution maybe to remove it from the workflow.  Any thoughts @WalterKolczynski-NOAA ?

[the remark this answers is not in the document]
JessicaMeixner-NOAA: @NeilBarton-NOAA can you help me understand why we don't see an added line for the new grid near line 327?  I saw it in your earlier scripts but don't see it here.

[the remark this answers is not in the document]
NeilBarton-NOAA: @JessicaMeixner-NOAA I didn't committed it because that part of the script wasn't being used. I'm happy to add. Are the below the correct values?

GRDNAME='global' ; GRDRES=0p50 ; GRIDNR=255  ; MODNR=10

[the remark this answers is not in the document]
JessicaMeixner-NOAA: Can you send me what atm resolution/IC date you are using so I can run a test on my end? I really thought we needed this line and would like to understand better.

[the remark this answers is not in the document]
WalterKolczynski-NOAA: > I think the variables are easy enough to follow, but agree on a better solution maybe to remove it from the workflow. Any thoughts @WalterKolczynski-NOAA ?

If at all practical, I'd prefer not to have the job run in the first place if it is not going to do anything.

parm/config/config.defaults.s2sw
WalterKolczynski-NOAA: Change the settings in _this_ file back. You can work with @XianwuXue-NOAA to get them as the default in GEFS. To run g-w directly with the GEFS grid, people will have to manually edit the config file.

WalterKolczynski-NOAA:
> > How is gefsv13_025 different from mx025?
> 
> gefsv13_025 is the regular lat/lon grid planned for GEFS where as mx025 is a tripolar grid we were using for HR1. mx025 would require too many resources to be used for operations, therefore the new grid was made.

Thanks @JessicaMeixner-NOAA, that clears it up. Can we get a more descriptive name like reg_025 or latlon_025 then, if that is what the grid is?

JessicaMeixner-NOAA:
> > > How is gefsv13_025 different from mx025?
> > 
> > 
> > gefsv13_025 is the regular lat/lon grid planned for GEFS where as mx025 is a tripolar grid we were using for HR1. mx025 would require too many resources to be used for operations, therefore the new grid was made.
> 
> Thanks @JessicaMeixner-NOAA, that clears it up. Can we get a more descriptive name like reg_025 or latlon_025 then, if that is what the grid is?

I think gefsv13_025 is okay personally.  reg_025 probably shouldn't be used as those are other regular lat/lon grids without the cutoff at 82 N for GEFS.  latlon_025 would work as well.

parm/config/config.ufs
WalterKolczynski-NOAA: Okay, keeping the setting doesn't really hurt anything.

WalterKolczynski-NOAA:
> > > > How is gefsv13_025 different from mx025?
> > > 
> > > 
> > > gefsv13_025 is the regular lat/lon grid planned for GEFS where as mx025 is a tripolar grid we were using for HR1. mx025 would require too many resources to be used for operations, therefore the new grid was made.
> > 
> > 
> > Thanks @JessicaMeixner-NOAA, that clears it up. Can we get a more descriptive name like reg_025 or latlon_025 then, if that is what the grid is?
> 
> I think gefsv13_025 is okay personally. reg_025 probably shouldn't be used as those are other regular lat/lon grids without the cutoff at 82 N for GEFS. latlon_025 would work as well.

Polling the CM team on what they think wrt names.

JessicaMeixner-NOAA:
> Polling the CM team on what they think wrt names.

Do wave model developers get a say or just g-w CM?

parm/config/config.defaults.s2sw
```diff
@@ -16,13 +16,13 @@ use_cice_alb=".true."
...
-waveinterpGRD='reg025'
-waveuoutpGRD='mx025'
-MESH_WAV='mesh.mx025.nc'
+waveGRD='gefsv13_025'
```
aerorahul: Could you explain what is the difference between the grids?
Both are "quarter-degree" I assume.
Adding application specific names has got us into trouble before, so I would like to know more what the actual differences are then propose a solution that will work across the GFS and GEFS infrastructures.

parm/config/config.defaults.s2sw
```diff
@@ -16,13 +16,13 @@ use_cice_alb=".true."
...
+waveinterpGRD=' '
+waveuoutpGRD='gefsv13_025'
```
aerorahul: We will seek consistency when declaring names that belong to a category.
Some places we see `mx025`, some places just `025` for e.g.

parm/config/config.ufs
aerorahul: This needs an explanation. 
The case statement is for grid resolution or type.
If someone wants to hard-wire something that just works, this is not the place.

parm/config/config.wavepostsbs
```diff
@@ -17,7 +17,11 @@ export DOIBP_WAV='NO' # Input boundary points
 export DOFLD_WAV='YES' # Field data
 export DOPNT_WAV='YES' # Station data
 export DOGRB_WAV='YES' # Create grib2 files
-export DOGRI_WAV='YES' # Create interpolated grids
+if [[ -z ${waveinterpGRD} ]]; then
+    export DOGRI_WAV='YES' # Create interpolated grids
+else
+    export DOGRI_WAV='NO' # Create interpolated grids
```
aerorahul: Shouldn't the comment be "Do not create interpolated grids"?

aerorahul (changes requested):
There is some explanation needed for these suggested changes.

parm/config/config.defaults.s2sw
WalterKolczynski-NOAA: Jessica explained the difference between mx and gefsv13 above.

WalterKolczynski-NOAA:
> > Polling the CM team on what they think wrt names.
> 
> Do wave model developers get a say or just g-w CM?

I think it is useful for us to come to a consensus first and make sure I'm not an outlier in preferring latlon_025 before further discussing with wave developers whether to change it.

parm/config/config.defaults.s2sw
aerorahul: Apologies, I did not refresh by browser to see comments from the previous reviews.
Instead of `gefsv13_`, I would propose something more generic.  If `mx` stands for "tripolar" what would an appropriate prefix for "regular" be?

parm/config/config.defaults.s2sw
JessicaMeixner-NOAA: Then all the wave model grids need to be renamed.    If you tell me your preferred naming convention, then I'll update all the wave grids to match.  This will be a major fix file update FYI.  For grids already in operations, perhaps we can make exceptions though?

parm/config/config.ufs
JessicaMeixner-NOAA: This is for a specific grid.  I'm a little confused by the comment of hard-coding or how what @NeilBarton-NOAA added here is any different than any of the above cases.

parm/config/config.defaults.s2sw
aerorahul: `develop` is not going into **current** operations, so I don't think `gfsv16.x` needs to worry about this.

We don't know yet if **all** wave model grids will need updating either.
If we know what are the current wave model grids are, we can come to a consensus on the naming convention, retaining whatever consistency we have.
Please create an issue and this can be discussed there.

A list of the current `fix/wave` are:
```
.../fix/wave/
├── mesh.gefsv13_025.nc
├── mesh.gwes_30m.nc
├── rmp_src_to_dst_conserv_002_001.nc
├── rmp_src_to_dst_conserv_003_001.nc
├── weights_rtofs_to_r4320x2160.nc
├── WHTGRIDINT.bin.ak_10m
├── WHTGRIDINT.bin.glo_15mext
├── WHTGRIDINT.bin.glo_15mxt
├── WHTGRIDINT.bin.glo_30m
├── WHTGRIDINT.bin.reg025
├── ww3_grid.inp.ak_10m
├── ww3_grid.inp.ant_9km
├── ww3_grid.inp.ao_20m
├── ww3_grid.inp.aoc_9km
├── ww3_grid.inp.at_10m
├── ww3_grid.inp.ep_10m
├── ww3_grid.inp.gefsv13_025
├── ww3_grid.inp.glix_10m
├── ww3_grid.inp.glo_10m
├── ww3_grid.inp.glo_15m
├── ww3_grid.inp.glo_15mxt
├── ww3_grid.inp.glo_30m
├── ww3_grid.inp.glox_10m
├── ww3_grid.inp.gnh_10m
├── ww3_grid.inp.gsh_15m
├── ww3_grid.inp.gwes_30m
├── ww3_grid.inp.icean_5m
├── ww3_grid.inp.mx025
├── ww3_grid.inp.mx025gefs
├── ww3_grid.inp.mx050
├── ww3_grid.inp.mx100
├── ww3_grid.inp.points
├── ww3_grid.inp.reg025
├── ww3_grid.inp.reg050
├── ww3_grid.inp.reg100
├── ww3_grid.inp.rtofs_5m
├── ww3_grid.inp.so_20m
└── ww3_grid.inp.wc_10m
```

Most of them seem to have a consistent naming convention, with the exception of a couple.  We can catalog what these are, whether they are used or not, and provide backward compatibility for a year or two.

JessicaMeixner-NOAA:
I have created an issue for the naming of wave grids: https://github.com/NOAA-EMC/global-workflow/issues/1493  To be fair to @NeilBarton-NOAA  we had previously discussed the gefsv12_025 name of the grid and I thought it sounded pretty reasonable to me.

parm/config/config.ufs
aerorahul: The comment is about naming of the specific grid.  The other names in the case block are based on basin, resolution, and not an application.

NeilBarton-NOAA:
I have no major opinions about the grid name. Whatever makes sense to the wave and g-w teams. Just let me know and I'll edit the scripts and rename the files.

parm/config/config.defaults.s2sw
JessicaMeixner-NOAA: FYI -   I am going to move several of these variables to config.ufs for each wave grid, that way when someone switches between grids, we have this set of grid dependent variables set and only waveGRD then needs to be set.   Fix file update which is required before this and the other changes can go in will be requested shortly, with a PR to @NeilBarton-NOAA's branch for this PR to update things for the gefs grid.

WalterKolczynski-NOAA: @JessicaMeixner-NOAA As long as they are variables ONLY used by the forecast job, since `config.ufs` isn't used by anything else, while `config.wave` (which these are overriding) is used in several jobs. And, if you are moving anything to `config.ufs` from this section, it should be removed from `config.wave` as well (or wherever the setting noramlly is).

With our move towards running S2S more typically, we should also probably look to retire this config and fold the settings back in to other places (in a future PR).

JessicaMeixner-NOAA: Glad I mentioned this as these variables are used elsewhere and this will save some time as those variables are used for things other than the forecast.   I am going to have some variable groups defined based on waveGRD, which will help in switching grids.   TBD if it's this file or config.wave or where.  Thank you @WalterKolczynski-NOAA !!!

JessicaMeixner-NOAA:
@NeilBarton-NOAA I'm ready to push my branch with updates for this here, however there are many files that change, I think mostly because this is out of date with the top of develop at this point.  I think there's likely 2 ways to proceed here: 

1. You merge develop into your branch here and then I make a PR to your branch for this PR 
2. I just make a new PR to develop 

Please let me know what you prefer.  Either option requires PR #1605 to be merged first.

NeilBarton-NOAA:
@JessicaMeixner-NOAA I'll update my branch with develop after your branch is merged.

JessicaMeixner-NOAA:
@NeilBarton-NOAA I'd like to share the updates required for running the GEFS grid and post as soon as possible.  As I do not know the timeline for merging of the other PR, would you mind going ahead and updating with develop now so I can proceed with updating this branch (I have the other PR updates in there as well).  Another solution could be I just create a new PR. 
 Please let me know what your preference is and thanks in advance for your help.

Issue #1418, opened 2023-03-23
Archiving fails in S2S trying to save sfluxgrb forecast files that are not created


AndrewEichmann-NOAA:
**Expected behavior**

Archiving task completes successfully

**Current behavior**

When running S2S, archiving fails on trying to archive sfluxgrb forecast files on "off-hours" (not divisible by 3), as these are currently not created. 

**Machines affected**

Hera

**To Reproduce**

Can be seen in `/scratch1/NCEPDEV/stmp4/Andrew.Eichmann/COMROT/ocn-da-lores/logs/2021032412/gdasarch.log.1 `starting on line 1001:

```
ERROR: /apps/hpss/bin/htar: error -1 stat-ing "./gdas.20210324/12/atmos/gdas.t12z.sfluxgrbf001.grib2"
*** Warning: No matching names located for './gdas.20210324/12/atmos/gdas.t12z.sfluxgrbf001.grib2'
ERROR: /apps/hpss/bin/htar: error -1 stat-ing "./gdas.20210324/12/atmos/gdas.t12z.sfluxgrbf001.grib2.idx"
*** Warning: No matching names located for './gdas.20210324/12/atmos/gdas.t12z.sfluxgrbf001.grib2.idx'
ERROR: /apps/hpss/bin/htar: error -1 stat-ing "./gdas.20210324/12/atmos/gdas.t12z.sfluxgrbf002.grib2"
*** Warning: No matching names located for './gdas.20210324/12/atmos/gdas.t12z.sfluxgrbf002.grib2'
ERROR: /apps/hpss/bin/htar: error -1 stat-ing "./gdas.20210324/12/atmos/gdas.t12z.sfluxgrbf002.grib2.idx"
*** Warning: No matching names located for './gdas.20210324/12/atmos/gdas.t12z.sfluxgrbf002.grib2.idx'
ERROR: /apps/hpss/bin/htar: error -1 stat-ing "./gdas.20210324/12/atmos/gdas.t12z.sfluxgrbf004.grib2"
*** Warning: No matching names located for './gdas.20210324/12/atmos/gdas.t12z.sfluxgrbf004.grib2'
ERROR: /apps/hpss/bin/htar: error -1 stat-ing "./gdas.20210324/12/atmos/gdas.t12z.sfluxgrbf004.grib2.idx"
*** Warning: No matching names located for './gdas.20210324/12/atmos/gdas.t12z.sfluxgrbf004.grib2.idx'
```

[...]

PR #1498, opened 2023-04-21
ignore archiving sfluxgrbf00[124578] files when running WCDA

AndrewEichmann-NOAA:
**Description**

When running WCDA (S2S app), task gdasarch fails because sfluxgrb files for non-divisible-by-3 forecast hours are not in comrot. This change wraps this section of ush/hpssarch_gen.sh in an if/then block for DO_JEDIOCNVAR=NO to avoid this problem.

Addresses issue https://github.com/NOAA-EMC/global-workflow/issues/1418

**Type of change**

**How Has This Been Tested?**

Tested with cycling on Hera
  
**Checklist**

### DIFF
--- a/ush/hpssarch_gen.sh
+++ b/ush/hpssarch_gen.sh
@@ -341,13 +341,15 @@ if [[ ${type} = "gdas" ]]; then
     echo  "${dirname}${head}sfcf${fhr}.nc              " >>gdas.txt
     fh=$((fh+3))
   done
-  if [[ ${DO_JEDIOCNVAR} = "NO" ]]; then
-    flist="001 002 004 005 007 008"
-    for fhr in ${flist}; do
-      echo  "${dirname}${head}sfluxgrbf${fhr}.grib2      " >>gdas.txt
-      echo  "${dirname}${head}sfluxgrbf${fhr}.grib2.idx  " >>gdas.txt
-    done
-  fi
+  flist="001 002 004 005 007 008"
+  for fhr in ${flist}; do
+    file="${dirname}${head}sfluxgrbf${fhr}.grib2"
+    # Only add to list if file is present.
+    if [[ -s "${file}" ]]; then
+      echo  "${file}"      >>gdas.txt
+      echo  "${file}$.idx" >>gdas.txt
+    fi
+  done
   
 
 
--- a/ush/hpssarch_gen.sh
+++ b/ush/hpssarch_gen.sh
@@ -347,7 +347,7 @@ if [[ ${type} = "gdas" ]]; then
     # Only add to list if file is present.
     if [[ -s "${file}" ]]; then
       echo  "${file}"      >>gdas.txt
-      echo  "${file}$.idx" >>gdas.txt
+      echo  "${file}.idx"  >>gdas.txt
     fi
   done
   
--- a/ush/hpssarch_gen.sh
+++ b/ush/hpssarch_gen.sh
@@ -341,11 +341,13 @@ if [[ ${type} = "gdas" ]]; then
     echo  "${dirname}${head}sfcf${fhr}.nc              " >>gdas.txt
     fh=$((fh+3))
   done
-  flist="001 002 004 005 007 008"
-  for fhr in ${flist}; do
-    echo  "${dirname}${head}sfluxgrbf${fhr}.grib2      " >>gdas.txt
-    echo  "${dirname}${head}sfluxgrbf${fhr}.grib2.idx  " >>gdas.txt
-  done
+  if [[ ${DO_JEDIOCNVAR} = "NO" ]]; then
+    flist="001 002 004 005 007 008"
+    for fhr in ${flist}; do
+      echo  "${dirname}${head}sfluxgrbf${fhr}.grib2      " >>gdas.txt
+      echo  "${dirname}${head}sfluxgrbf${fhr}.grib2.idx  " >>gdas.txt
+    done
+  fi
   
 
 


AndrewEichmann-NOAA:
FYI @guillaumevernieres @CoryMartin-NOAA

CoryMartin-NOAA:
Question, why aren't they there? Is that by design, or a bug?

guillaumevernieres:
> Question, why aren't they there? Is that by design, or a bug?

@CoryMartin-NOAA , our DA history collection/diag_table is quite minimalist, I suspect that is the issue but we probably should have checked where these files are coming from originally.

CoryMartin-NOAA:
I guess my point of my question was: is this treating the symptom or the actual "disease"? I am hesitant to just not archive something because we don't have it when running with DA, does it also fail without DA in develop with the coupled model?

guillaumevernieres:
> I guess my point of my question was: is this treating the symptom or the actual "disease"? I am hesitant to just not archive something because we don't have it when running with DA, does it also fail without DA in develop with the coupled model?

Yes, that's a fair point @CoryMartin-NOAA , and the same offline head scratching comments that @AndrewEichmann-NOAA and myself had. This probably should be a draft PR, since @AndrewEichmann-NOAA 's solution may not be the correct one. 

@AndrewEichmann-NOAA , I think we need to do more "homework":
- Identify what model component originate these files
- Figure out if this should be saved during the cycle. @JessicaMeixner-NOAA would know maybe?

... A problem for nextr week as far as I'm concerned!

JessicaMeixner-NOAA:
These files are atm files.  

I'm confused how this is failing for the gdas app and not in other cycled runs?

ush/hpssarch_gen.sh
```diff
@@ -341,11 +341,13 @@ if [[ ${type} = "gdas" ]]; then
...
-    echo  "${dirname}${head}sfluxgrbf${fhr}.grib2      " >>gdas.txt
-    echo  "${dirname}${head}sfluxgrbf${fhr}.grib2.idx  " >>gdas.txt
-  done
+  if [[ ${DO_JEDIOCNVAR} = "NO" ]]; then
+    flist="001 002 004 005 007 008"
+    for fhr in ${flist}; do
+      echo  "${dirname}${head}sfluxgrbf${fhr}.grib2      " >>gdas.txt
+      echo  "${dirname}${head}sfluxgrbf${fhr}.grib2.idx  " >>gdas.txt
+    done
+  fi
```
aerorahul: Would this work instead?
```suggestion
  flist="001 002 004 005 007 008"
  for fhr in ${flist}; do
    file="${dirname}${head}sfluxgrbf${fhr}.grib2"
    # Only add to list if file is present.
    if [[ -s "${file}" ]]; then
      echo  "${file}"      >>gdas.txt
      echo  "${file}$.idx" >>gdas.txt
    fi
  done
```

aerorahul (commented):
Rather than adding a condition on application, would testing on the existence of the file work for you?

aerorahul:
These files are written by the inline post.
This is likely failing because, in SOCA, they are not doing the backgrounds in 4D at every hour.
They are likely writing output at hours 3, 6, 9 and not hourly.
Either way, the check should be if the file is present, add it to the archival, and if not, don't. 
Whether the whole set belongs in the archive or not is another q.

guillaumevernieres:
> These files are written by the inline post. This is likely failing because, in SOCA, they are not doing the backgrounds in 4D at every hour. They are likely writing output at hours 3, 6, 9 and not hourly. Either way, the check should be if the file is present, add it to the archival, and if not, don't. Whether the whole set belongs in the archive or not is another q.

We are writing every hour @aerorahul , but currently only using the 3 hourly history file. I forgot why we did that, the early implementation was hourly, but when you and I merged my early work, we decided to only use the 3 hourly history. Maybe we should revisit this issue in this PR ?

aerorahul:
> > These files are written by the inline post. This is likely failing because, in SOCA, they are not doing the backgrounds in 4D at every hour. They are likely writing output at hours 3, 6, 9 and not hourly. Either way, the check should be if the file is present, add it to the archival, and if not, don't. Whether the whole set belongs in the archive or not is another q.
> 
> We are writing every hour @aerorahul , but currently only using the 3 hourly history file. I forgot why we did that, the early implementation was hourly, but when you and I merged my early work, we decided to only use the 3 hourly history. Maybe we should revisit this issue in this PR ?

@guillaumevernieres 
The output from "diag_table" is hourly for the ocean files.  I agree.  It is hard-wired there.
```
"ocn_da%4yr%2mo%2dy%2hr", 1, "hours", 1, "hours", "time", 1,  "hours"
```

But that is not what the atmosphere inline post is seeing.
In the `model_configure` file, can you check if `output_fh` is hourly or 3-hourly?
This determines the atmosphere output frequency.

guillaumevernieres:
@aerorahul , from the forecast log, the output is 3 hourly.
```gdasfcst.log:+ forecast_predet.sh[88]: OUTPUT_FH=' 0 3 6 9'```

File: ush/hpssarch_gen.sh (hunk 1 above)
Comment (guillaumevernieres) [reply ush/hpssarch_gen.sh]: Seems like a better solution. Merge and move on @AndrewEichmann-NOAA ?

**Comment by aerorahul** [comment]
Can I commit the suggestion and merge this PR @guillaumevernieres @AndrewEichmann-NOAA?
I would like to get this in before the com reorg PR gets in.

**Review by aerorahul — APPROVED** [review]
I can merge with my suggested change.

**Comment by CoryMartin-NOAA** [comment]
I say merge with the suggested change

**Review by guillaumevernieres — APPROVED** [review]

File: ush/hpssarch_gen.sh [hunk 2]
```diff
@@ -343,8 +343,12 @@ if [[ ${type} = "gdas" ]]; then
   done
   flist="001 002 004 005 007 008"
   for fhr in ${flist}; do
-    echo  "${dirname}${head}sfluxgrbf${fhr}.grib2      " >>gdas.txt
-    echo  "${dirname}${head}sfluxgrbf${fhr}.grib2.idx  " >>gdas.txt
+    file="${dirname}${head}sfluxgrbf${fhr}.grib2"
+    # Only add to list if file is present.
+    if [[ -s "${file}" ]]; then
+      echo  "${file}"      >>gdas.txt
+      echo  "${file}$.idx" >>gdas.txt
```
aerorahul:
```suggestion
      echo  "${file}.idx"  >>gdas.txt
```
fix the typo.

PR #1538, opened 2023-04-26
Update UFSDA ATM ens for new COM directory structure 

### LINKED ISSUE #1518: Update UFSDA ATM EnKF for COM refactor
**Description**
UFSDA jobs were not included in the initial COM refactor work (#1421). These jobs will all need to be updated to use the new structure and templates before they can run again.

**Requirements**
UFSDA ATM ensemble analysis jobs that utilize the new COM templates to generate COM directory paths.

UFSDA ATM ensemble jobs that work as well as before the COM refactor

There are two approaches we can take. The first is to define the full paths in the j-job using the templates, as is done for most of the existing global-workflow jobs, and update the python to expect full paths (with dates and cycles). The second is to use the templates directly in the python (reading the `*_TMPL` variables from the environment) and using Rahul's Template class to fill in the variables.

Recommend the first for now where feasible, though the latter may be needed if iterating over members in python (similar to how the EnKF exscripts have to do it). The Template class will need a minor update to proceed with option 2: The templates reply on the bash substitution replacing undefined variables with an empty string (mainly to replacing `$MEMDIR` for the deterministic files), but the class currently leaves any variables it doesn't have substitutions for as-is. It will need to be updated to provide the option to replace undefined variables with an empty string.

RussTreadon-NOAA:
**Description**
g-w PR #1421 changed the GFS COM directory structure. This PR updates UFSDA ATM ensemble analysis jobs and python script to work with the updated GFS COM directory structure.

Fixes #1518 

**Type of change**

**How Has This Been Tested?**

- [x] Cycled test on Hera
  
**Checklist**


RussTreadon-NOAA:
@WalterKolczynski-NOAA , the linter failure is

```
✋ Defects, NEEDS INSPECTION
Error: SHELLCHECK_WARNING:
./jobs/JGLOBAL_ATMENS_ANALYSIS_FINALIZE:11:28: note[SC2153]: Possible misspelling: PDY may not be assigned. Did you mean gPDY?

Error: SHELLCHECK_WARNING:
./jobs/JGLOBAL_ATMENS_ANALYSIS_INITIALIZE:10:28: note[SC2153]: Possible misspelling: PDY may not be assigned. Did you mean gPDY?
```

For both jobs the offending line is

` 
 GDATE=$(date +%Y%m%d%H -d "${PDY} ${cyc} - ${assim_freq} hours")
 `

Both the initialize and finalize jobs use files from the previous cycle. The rocoto workflow defines `PDY`.

What is the preferred solution to resolve the linter failure?

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -234,8 +232,8 @@ def get_bkg_dict(self, task_config: Dict[str, Any]) -> Dict[str, List[str]]:
             # accumulate directory list for member restart files
             dirlist.append(os.path.join(task_config.DATA, 'bkg', memchar))
 
-            # get FV3 RESTART files, this will be a lot simpler when using history files
-            rst_dir = os.path.join(task_config.comin_ges_atmens, memchar, 'atmos/RESTART')
+            # get FV3 restart files, this will be a lot simpler when using history files
+            rst_dir = os.path.join(task_config.COM_TOP_ENS_GES, memchar, 'model_data/atmos/restart')
```
WalterKolczynski-NOAA: Instead of doing this, which basically reconstructs the atmos restart template, let's use `pygw.Template` and build paths here. Use this as a starting point, and then modify the config dictionaries that contain 'ROTDIR', etc. if I've gotten any wrong.

At the top of the script, you'll need to add the following import:
```
from pygw.template import Template, TemplateConstants
```

Then, change this part:
```suggestion
        template = task_config.COM_ATMOS_RESTART_TMPL
        tmpl_dict = {
            'ROTDIR': task_config.ROTDIR,
            'RUN': self.runtime_config.RUN,
            'YMD': self.runtime_config.PDY,
            'HH': self.runtime_config.cyc,
            }
        for imem in range(1, task_config.NMEM_ENKF + 1):
            memchar = f"mem{imem:03d}"
            # accumulate directory list for member restart files
            dirlist.append(os.path.join(task_config.DATA, 'bkg', memchar))

            # get FV3 restart files, this will be a lot simpler when using history files
            tmpl_dict['MEMDIR'] = memchar
            rst_dir = Template.substitute_structure(template, TemplateConstants.DOLLAR_CURLY_BRACE, tmpl_dict.get)
            os.makedirs(rst_dir, mode=0o775, exist_ok=True)
```
This will take the template and use the `tmpl_dict` dictionary to replace variables (defined as being within `${ ... }`) with entries from the dictionary.

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -290,15 +288,15 @@ def jedi2fv3inc(self: Analysis) -> None:
 
             # make output directory for member increment
             incdir = [
-                os.path.join(self.task_config.COMOUT, memchar, 'atmos')
+                os.path.join(self.task_config.COM_TOP_ENS, memchar, 'analysis/atmos')
```
WalterKolczynski-NOAA: Modify these similarly to above, using the appropriate template. Basically, usage of `COM_TOP` should just be avoided. It's only being used in two places where we need to cut above the member level.

jobs/JGLOBAL_ATMENS_ANALYSIS_FINALIZE
```diff
@@ -8,19 +8,27 @@ source "${HOMEgfs}/ush/jjob_header.sh" -e "atmensanlfinal" -c "base atmensanl at
...
+    COM_ATMOS_ANALYSIS_ENS:COM_ATMOS_ANALYSIS_TMPL
 
-# COMIN_GES and COMIN_GES_ENS are used in script
-export COMIN_GES="${ROTDIR}/${GDUMP}.${GDATE:0:8}/${GDATE:8:2}/atmos"
-export COMIN_GES_ENS="${ROTDIR}/enkf${GDUMP}.${GDATE:0:8}/${GDATE:8:2}"
+RUN=${GDUMP_ENS} YMD=${PDY} HH=${cyc} generate_com -rx \
+    COM_TOP_ENS:COM_TOP_TMPL
+
+RUN=${GDUMP_ENS} YMD=${gPDY} HH=${gcyc} generate_com -rx \
+    COM_TOP_ENS_GES:COM_TOP_TMPL
```
WalterKolczynski-NOAA: See below about removing the need for using COM_TOP, which should be avoided.
```suggestion
```

jobs/JGLOBAL_ATMENS_ANALYSIS_INITIALIZE
```diff
@@ -7,19 +7,33 @@ source "${HOMEgfs}/ush/jjob_header.sh" -e "atmensanlinit" -c "base atmensanl atm
...
+    COM_ATMOS_ANALYSIS_PREV:COM_ATMOS_ANALYSIS_TMPL 
+
+RUN=${GDUMP_ENS} YMD=${PDY} HH=${cyc} generate_com -rx \
+    COM_TOP_ENS:COM_TOP_TMPL
+
+RUN=${GDUMP_ENS} YMD=${gPDY} HH=${gcyc} generate_com -rx \
+    COM_TOP_ENS_GES:COM_TOP_TMPL
+
+
+mkdir -m 775 -p "${COM_TOP_ENS}"
```
WalterKolczynski-NOAA:
```suggestion
```
Can add the mkdir in the python where you generate the path.

WalterKolczynski-NOAA (changes requested):

ush/python/pygfs/task/atmens_analysis.py
WalterKolczynski-NOAA: Oops, this was supposed to replace back to the start of the loop.

RussTreadon-NOAA:
Thank you @WalterKolczynski-NOAA for your review.  I followed the option 1 approach you outlined in #1518.   This is the  minimalist approach to restore functionality ASAP.   I thought option 2 could be pursued via a subsequent g-w issue and PR.   If this isn't the case, I'll work to fold your requested changes into this PR.

aerorahul:
> Thank you @WalterKolczynski-NOAA for your review. I followed the option 1 approach you outlined in #1518. This is the minimalist approach to restore functionality ASAP. I thought option 2 could be pursued via a subsequent g-w issue and PR. If this isn't the case, I'll work to fold your requested changes into this PR.

@WalterKolczynski-NOAA is correct.  We should not be using `COM_TOP` unless absolutely there is no other way.
Since the JEDI ATM DA is not a priority for v17, I recommend we take the proper solution and make this an example for others (looking at SOCA) to follow.
I don't think there is a significant loss of time and resources to implement option 2 compared to option 1.
Besides, once in, we know it rarely gets the priority for improvement.

RussTreadon-NOAA:
Understood.  Thank you @aerorahul and @WalterKolczynski-NOAA .  An attempt to make the required changes is underway.  I suspect both of you can make the changes faster.   If you want to do so, please let me know.

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: Line 250 aborts with the traceback

```
  File "/scratch1/NCEPDEV/da/Russ.Treadon/git/global-workflow/newcom/scripts/exglobal_atmens_analysis_initialize.py", line 25, in <module>
    AtmEnsAnl.initialize()
  File "/scratch1/NCEPDEV/da/Russ.Treadon/git/global-workflow/newcom/ush/python/pygw/src/pygw/logger.py", line 264, in wrapper
    retval = func(*args, **kwargs)
  File "/scratch1/NCEPDEV/da/Russ.Treadon/git/global-workflow/newcom/ush/python/pygfs/task/atmens_analysis.py", line 94, in initialize
    FileHandler(self.get_bkg_dict(AttrDict(self.task_config))).sync()
  File "/scratch1/NCEPDEV/da/Russ.Treadon/git/global-workflow/newcom/ush/python/pygw/src/pygw/logger.py", line 264, in wrapper
    retval = func(*args, **kwargs)
  File "/scratch1/NCEPDEV/da/Russ.Treadon/git/global-workflow/newcom/ush/python/pygfs/task/atmens_analysis.py", line 257, in get_bkg_dict
    rst_dir = Template.substitute_structure(template, TemplateConstants.DOLLAR_CURLY_BRACE, tmpl_dict.get)
  File "/scratch1/NCEPDEV/da/Russ.Treadon/git/global-workflow/newcom/ush/python/pygw/src/pygw/template.py", line 115, in substitute_structure
    get_value)
  File "/scratch1/NCEPDEV/da/Russ.Treadon/git/global-workflow/newcom/ush/python/pygw/src/pygw/template.py", line 87, in substitute_string
    variable, v)
TypeError: an integer is required (got type str)
```
This occurs while processing `mem001`.  The contents of `tmpl_dict` are 

`{'ROTDIR': '/scratch2/NCEPDEV/stmp1/Russ.Treadon/comrot/prufscom', 'RUN': 'enkfgdas', 'YMD': datetime.datetime(2021, 12, 21, 0, 0), 'HH': 0, 'MEMDIR': 'mem001'}
`

RussTreadon-NOAA: Variant of the suggested change committed at [`3253bfd`](https://github.com/RussTreadon-NOAA/global-workflow/commit/3253bfd97c8063c6b881e906da682b2f822802b6).    Tested in Hera prufscom `2021122100 enkfgdasatmensanlinit`.   Job successfully ran to completion.

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: Requested change committed at [`dc0f196`](https://github.com/RussTreadon-NOAA/global-workflow/commit/dc0f1963c9088ed946661c29e196c6fba42ecf5c)

jobs/JGLOBAL_ATMENS_ANALYSIS_FINALIZE
RussTreadon-NOAA: Requested change committed at [`dc0f196`](https://github.com/RussTreadon-NOAA/global-workflow/commit/dc0f1963c9088ed946661c29e196c6fba42ecf5c)

jobs/JGLOBAL_ATMENS_ANALYSIS_INITIALIZE
RussTreadon-NOAA: Requested change committed at [`dc0f196`](https://github.com/RussTreadon-NOAA/global-workflow/commit/dc0f1963c9088ed946661c29e196c6fba42ecf5c)

RussTreadon-NOAA:
An attempt has been made to address requested changes.   `RussTreadon-NOAA:feature/ufsda_atmens_com` at [`dc0f196`](https://github.com/RussTreadon-NOAA/global-workflow/commit/dc0f1963c9088ed946661c29e196c6fba42ecf5c) was used to run the prufscom `2021122100` `enkfgdasatmens` `init`, `run`, and `final` jobs.   All three jobs successfully ran to completion.

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -234,8 +233,20 @@ def get_bkg_dict(self, task_config: Dict[str, Any]) -> Dict[str, List[str]]:
...
+                'ROTDIR': task_config.ROTDIR,
+                'RUN': self.runtime_config.RUN,
+                'YMD': to_YMD(task_config.previous_cycle),
+                'HH': task_config.previous_cycle.strftime('%H')
+            }
+
+            # get FV3 restart files, this will be a lot simpler when using history files
+            tmpl_dict['MEMDIR'] = memchar
+            rst_dir = Template.substitute_structure(template, TemplateConstants.DOLLAR_CURLY_BRACE, tmpl_dict.get)
+            os.makedirs(rst_dir, mode=0o775, exist_ok=True)
```
aerorahul: While this works, the function here is to return a list of model background files to copy.
Now this function is also tasked w/ creating member directories.

We should move this to `analysis.py` and into its own method.

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -234,8 +233,20 @@ def get_bkg_dict(self, task_config: Dict[str, Any]) -> Dict[str, List[str]]:
...
-            # get FV3 RESTART files, this will be a lot simpler when using history files
-            rst_dir = os.path.join(task_config.comin_ges_atmens, memchar, 'atmos/RESTART')
+            # get FV3 restart files, this will be a lot simpler when using history files
+            template = task_config.COM_ATMOS_RESTART_TMPL
+            tmpl_dict = {
+                'ROTDIR': task_config.ROTDIR,
+                'RUN': self.runtime_config.RUN,
+                'YMD': to_YMD(task_config.previous_cycle),
+                'HH': task_config.previous_cycle.strftime('%H')
+            }
```
WalterKolczynski-NOAA: This doesn't change from member to member, so can be moved before the loop.

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: I will begin work on this request later tonight or tomorrow.

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -289,17 +300,30 @@ def jedi2fv3inc(self: Analysis) -> None:
             memchar = f"mem{imem:03d}"
 
             # make output directory for member increment
-            incdir = [
-                os.path.join(self.task_config.COMOUT, memchar, 'atmos')
-            ]
-            FileHandler({'mkdir': incdir}).sync()
+            template = self.task_config.COM_ATMOS_ANALYSIS_TMPL
+            tmpl_dict = {
```
WalterKolczynski-NOAA: Since other than `MEMDIR`, these do not change from member to member, can we define two dicts before the loop (one for current cycle and one for previous cycle) and then just change the `MEMDIR` of each inside each iteration?

WalterKolczynski-NOAA (changes requested):
Looking good. Just a recommendation to eliminate the redundant dictionary definitions.

ush/python/pygfs/task/atmens_analysis.py
WalterKolczynski-NOAA: Sorry Russ, this is on me for saying you should just add it here instead of suggesting a new method to create directories.

RussTreadon-NOAA: Requested change committed at [`da20985`](https://github.com/RussTreadon-NOAA/global-workflow/commit/da20985506a351ff6b13201f261bde5df12e99d3).

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: This code is now in `ush/python/pygfs/task/analysis.py`.   Requested change made at [`37fbac4`](https://github.com/RussTreadon-NOAA/global-workflow/commit/37fbac4bb01b74a9a642d2fbb3a871f2aafc6432)

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: Requested change made at [`37fbac4`](https://github.com/RussTreadon-NOAA/global-workflow/commit/37fbac4bb01b74a9a642d2fbb3a871f2aafc6432)

RussTreadon-NOAA:
Incorporate patch from @aerorahul (see PR [#4](https://github.com/RussTreadon-NOAA/global-workflow/pull/4)) into `RussTreadon-NOAA:feature/ufsda_atmens_com`.  Done at [`8126791`](https://github.com/RussTreadon-NOAA/global-workflow/commit/812679157606fc6d9994f56f419b6f98eaa843b2).    Rerun prufscom 2021122112 `enkfgdas` `init`, `run`, and `final` jobs.   All jobs successfully ran to completion.

Thank you, @aerorahul , for the patch and sorry for misunderstanding your original request.

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -285,21 +235,38 @@ def jedi2fv3inc(self: Analysis) -> None:
...
             memchar = f"mem{imem:03d}"
 
             # make output directory for member increment
-            incdir = [
-                os.path.join(self.task_config.COMOUT, memchar, 'atmos')
-            ]
-            FileHandler({'mkdir': incdir}).sync()
+            tmpl_inc_dict['MEMDIR'] = memchar
+            incdir = Template.substitute_structure(template_inc, TemplateConstants.DOLLAR_CURLY_BRACE, tmpl_inc_dict.get)
+            FileHandler({'mkdir': [incdir]}).sync()
```
aerorahul: We should try to collect all the actions (in this case mkdir for the different members) into a list and then call `FileHandler` once, instead of calling it in a loop over the ensemble members.
`FileHandler` will have the ability to handle data asynchronously.

Should we pull this out at the top of `initialize` where `DATA/bkg/memXXX` and `DATA/anl/memXXX` are being created?

RussTreadon-NOAA: Good point @aerorahul .    Suggestion committed at [`292acce`](https://github.com/RussTreadon-NOAA/global-workflow/commit/292accebbb14dace68286c00d35dfe95e3fbdd57).

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -285,21 +249,37 @@ def jedi2fv3inc(self: Analysis) -> None:
         # Reference the python script which does the actual work
         incpy = os.path.join(self.task_config.HOMEgfs, 'ush/jediinc2fv3.py')
 
+        # create template dictionaries
+        template_inc = self.task_config.COM_ATMOS_ANALYSIS_TMPL
+        tmpl_inc_dict = {
+            'ROTDIR': self.task_config.ROTDIR,
+            'RUN': self.runtime_config.RUN,
```
aerorahul: `task_config` should have this attribute.
```suggestion
            'RUN': self.task_config.RUN,
```

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -285,21 +249,37 @@ def jedi2fv3inc(self: Analysis) -> None:
...
+            'YMD': to_YMD(self.task_config.current_cycle),
+            'HH': self.task_config.current_cycle.strftime('%H')
+        }
+
+        template_ges = self.task_config.COM_ATMOS_HISTORY_TMPL
+        tmpl_ges_dict = {
+            'ROTDIR': self.task_config.ROTDIR,
+            'RUN': self.runtime_config.RUN,
```
aerorahul: Same suggestion.
```suggestion
            'RUN': self.task_config.RUN,
```

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -285,21 +249,37 @@ def jedi2fv3inc(self: Analysis) -> None:
...
+            # create output path for member analysis increment
+            tmpl_inc_dict['MEMDIR'] = memchar
+            incdir = Template.substitute_structure(template_inc, TemplateConstants.DOLLAR_CURLY_BRACE, tmpl_inc_dict.get)
 
             # rewrite UFS-DA atmens increments
-            atmges_fv3 = os.path.join(self.task_config.COMIN_GES_ENS, memchar, 'atmos',
-                                      f"{self.task_config.CDUMP}.t{self.runtime_config.previous_cycle.hour:02d}z.atmf006.nc")
+            tmpl_ges_dict['MEMDIR'] = memchar
+            gesdir = Template.substitute_structure(template_ges, TemplateConstants.DOLLAR_CURLY_BRACE, tmpl_ges_dict.get)
+            atmges_fv3 = os.path.join(gesdir, f"{self.task_config.CDUMP}.t{self.runtime_config.previous_cycle.hour:02d}z.atmf006.nc")
```
```suggestion
            atmges_fv3 = os.path.join(gesdir, f"{self.task_config.CDUMP}.t{self.task_config.previous_cycle.hour:02d}z.atmf006.nc")
```

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -285,21 +249,37 @@ def jedi2fv3inc(self: Analysis) -> None:
...
             atminc_jedi = os.path.join(self.task_config.DATA, 'anl', memchar, f'atminc.{cdate_inc}z.nc4')
-            atminc_fv3 = os.path.join(self.task_config.COMOUT, memchar, 'atmos',
-                                      f"{self.task_config.CDUMP}.t{self.runtime_config.cyc:02d}z.atminc.nc")
+            atminc_fv3 = os.path.join(incdir, f"{self.task_config.CDUMP}.t{self.runtime_config.cyc:02d}z.atminc.nc")
```
```suggestion
            atminc_fv3 = os.path.join(incdir, f"{self.task_config.CDUMP}.t{self.task_config.cyc:02d}z.atminc.nc")
```

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -79,6 +78,27 @@ def initialize(self: Analysis) -> None:
         """
         super().initialize()
 
+        # Make member directories in DATA for background and in DATA and ROTDIR for analysis files
+        # create template dictionary for output member analysis directories
+        template_inc = self.task_config.COM_ATMOS_ANALYSIS_TMPL
+        tmpl_inc_dict = {
+            'ROTDIR': self.task_config.ROTDIR,
+            'RUN': self.runtime_config.RUN,
```
```suggestion
            'RUN': self.task_config.RUN,
```

aerorahul (changes requested):
looks good.
just one set of consistent changes requested.
Use `task_config` in the methods since we have created this attribute to contain everything this task should need on line 58.

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: `self.runtime_config` replaced with `self.task_config`.   Done at [`4402e25`](https://github.com/RussTreadon-NOAA/global-workflow/commit/4402e255ee53b5059c5cfdad8364a80101fd6b08).   Thank you @aerorahul .  I'm still learning what all these different `self.*` mean and when/where we should one or the other.

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: `self.runtime_config` replaced with `self.task_config`.   Done at [`4402e25`](https://github.com/RussTreadon-NOAA/global-workflow/commit/4402e255ee53b5059c5cfdad8364a80101fd6b08).

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: `self.runtime_config` replaced with `self.task_config`.   Done at [`4402e25`](https://github.com/RussTreadon-NOAA/global-workflow/commit/4402e255ee53b5059c5cfdad8364a80101fd6b08).

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: `self.runtime_config` replaced with `self.task_config`.   Done at [`4402e25`](https://github.com/RussTreadon-NOAA/global-workflow/commit/4402e255ee53b5059c5cfdad8364a80101fd6b08).

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: `self.runtime_config` replaced with `self.task_config`.   Done at [`4402e25`](https://github.com/RussTreadon-NOAA/global-workflow/commit/4402e255ee53b5059c5cfdad8364a80101fd6b08).

aerorahul (approved):
Looks good.

ush/python/pygfs/task/analysis.py
```diff
@@ -10,6 +10,7 @@
 from pygw.template import Template, TemplateConstants
 from pygw.logger import logit
 from pygw.task import Task
+from pygw.timetools import to_fv3time, to_YMD
```
aerorahul: I think this is a left-over from when `get_bkg_ens_dict` was in `analysis.py`.
```suggestion
```

RussTreadon-NOAA:
Thank you @aerorahul for [36d733d](https://github.com/NOAA-EMC/global-workflow/pull/1538/commits/36d733d3a61e867ab11b900b65dc4dcd4884ffc1).   You are right.   This was a leftover which I missed.   Good to remove unused stuff.

WalterKolczynski-NOAA (approved):

Issue #241, opened 2021-01-15
HOTFIX: new EMC_verif-global tag to replace use of /tmp space


KateFriedman-NOAA:
WCOSS admins reported a "Compute Node /tmp usage Violation" to several users. It was determined that the MET configs were using the "/tmp" space on WCOSS (and other platforms). @malloryprow has made a hotfix update to the MET configs (https://github.com/NOAA-EMC/EMC_verif-global/issues/27) in EMC_verif-global and cut a new tag: [verif_global_v1.13.1](https://github.com/NOAA-EMC/EMC_verif-global/releases/tag/verif_global_v1.13.1). 

Will update EMC_verif-global tag and let users know, particularly on WCOSS, to update to this new tag in their clones.

KateFriedman-NOAA:
Updated EMC_verif-global tag in develop and release/gfsv16.0.0 branches. Sent announcement to glopara announce listserv.

Issue #233, opened 2021-01-05
Update GLDAS tag to gldas_gfsv16_release.v1.13.0 in develop


KateFriedman-NOAA:
A new GLDAS tag ([gldas_gfsv16_release.v1.13.0](https://github.com/NOAA-EMC/GLDAS/releases/tag/gldas_gfsv16_release.v1.13.0)) has been cut for the following:

1. Resolve build error on Orion - resolution of GLDAS [issue #18](https://github.com/NOAA-EMC/GLDAS/issues/18) reported by @CatherineThomas-NOAA 
2. Update PRCP file setting in ush/gldas_forcing.sh to use "_early" version appropriately

Test build on Orion and run GLDAS job to check correct cpc_precip setting.

KateFriedman-NOAA:
Created branch in my fork for this work:

https://github.com/KateFriedman-NOAA/global-workflow/tree/issue233

KateFriedman-NOAA:
Updated checkout.sh to use new tag. Ran checkout.sh, build_all.sh and link_fv3gfs.sh on Orion. No issues encountered. GLDAS build encountered no issues either.

Will now test GLDAS to confirm continued functionality and correct usage of PRCP files.

KateFriedman-NOAA:
Started a C192C96L127 cold started run on Orion. Will run it until GLDAS is initiated after spin-up period. Experiment cron is now enabled, will let test continue on it's own while focusing on other tasks.

clone: /work/noaa/global/kfriedma/git/issue233
EXPDIR: /work/noaa/global/kfriedma/expdir/testcyc192c
COMROT: /work/noaa/stmp/kfriedma/comrot/testcyc192c

KateFriedman-NOAA:
Test ran successfully on Orion and used correct PRCP files. Submitting PR to global-workflow develop to update GLDAS tag now.

KateFriedman-NOAA:
Sent announcement about commit to users. Closing issue now.

PR #199, opened 2020-11-30
feature/coupled-crow: Updates for consistency 

JessicaMeixner-NOAA:
This partially addresses issue #172.  This updates to use the ufs-weather-model instead of ufs-s2s-model and updates the hash the model is pointing to. 

This consolidates the MOM_inputs so there is only one per resolution and the configurable variables are consistent with the RT tests for the ufs-weather-model

PR #1541, opened 2023-04-27
Add in initial LETKFOI snow DA cycling capability and update UFSDA LAND for new COM refactor

jiaruidong2017:
**Add Jobs and Scripts for JEDI-Based Land DA**
1) To support global land initialization for future GFS implementations, we need to add to global-workflow jobs and scripts similar to what have already been created for JEDI-based aerosol/atmospheric DA, but for a separate, land analysis.

- jobs/JGDAS_GLOBAL_LAND_ANALYSIS_FINALIZE
- jobs/JGDAS_GLOBAL_LAND_ANALYSIS_INITIALIZE
- jobs/JGDAS_GLOBAL_LAND_ANALYSIS_RUN
- scripts/exgdas_global_land_analysis_finalize.py
- scripts/exgdas_global_land_analysis_initialize.py
- scripts/exgdas_global_land_analysis_run.sh

2) To support global land initialization for future GFS implementations, we need to add to global-workflow experiment setup and rocoto XML generation scripts support for including land DA related tasks. This PR adds in stubs for land DA tasks (jobs/rocoto shell scripts), the ability for setup_expt and setup_xml to include three land DA related tasks.

    Add the land analysis script and update two workflow scripts:
- ush/python/pygfs/task/land_analysis.py
- workflow/applications.py
- workflow/rocoto/workflow_tasks.py

3) G-W PR #1421 changed the GFS COM directory structure. This PR updates UFSDA LAND analysis jobs and python scripts to work with the updated GFS COM directory structure. (Issue #1520)

The prep job should:
- convert netCDF GTS snow depth obs to IODA format using python/compiled code
- stage backgrounds, fix files, and all other things necessary to run an land analysis.

Run job:
- create YAML file from template(s)
- Run fv3jedi_letkf.x

Post job:
- archive diagnostic files
- move/copy increments/analysis files and make changes as needed to stage for model initialization

Acceptance Criteria (Definition of Done)
- A 'dummy' cycling system is able to run with Land DA tasks included.

**Checklist**

- [x]  My code follows the style guidelines of this project
- [x]  I have performed a self-review of my own code
- [x]  I have commented my code, particularly in hard-to-understand areas
- [x]  My changes need updates to the documentation. I have made corresponding changes to the documentation
- [x]  My changes generate no new warnings
- [x]  New and existing tests pass with my changes
- [x]  Any dependent changes have been merged and published


jiaruidong2017:
@aerorahul @WalterKolczynski-NOAA Would you please assign reviewers to this PR? Thanks. 

The above shellnorms errors don't require a fix as below: 
image

WalterKolczynski-NOAA:
> @aerorahul @WalterKolczynski-NOAA Would you please assign reviewers to this PR? Thanks.
> 
> The above shellnorms errors don't require a fix as below: image

Add the following on the line before the one shellcheck complains about:
```
# Ignore possible spelling error (nothing is misspelled)
# shellcheck disable=SC2153
```
That directive will tell shellcheck to ignore that error code on the next line (and the comment is so we remember why it is being ignored).

jobs/JGLOBAL_LAND_ANALYSIS_FINALIZE
```diff
@@ -9,18 +9,22 @@ source "${HOMEgfs}/ush/jjob_header.sh" -e "landanlfinal" -c "base landanl landan
...
-mkdir -p "${COMOUT}"
+RUN=${GDUMP} YMD=${gPDY} HH=${gcyc} generate_com -rx \
+    COM_ATMOS_ANALYSIS_PREV:COM_ATMOS_ANALYSIS_TMPL \
+    COM_ATMOS_HISTORY_PREV:COM_ATMOS_HISTORY_TMPL \
+    COM_ATMOS_RESTART_PREV:COM_ATMOS_RESTART_TMPL
 
-# COMIN_GES and COMIN_GES_ENS are used in script
-export COMIN_GES="${ROTDIR}/${GDUMP}.${GDATE:0:8}/${GDATE:8:2}/atmos"
-export COMIN_GES_ENS="${ROTDIR}/enkf${GDUMP}.${GDATE:0:8}/${GDATE:8:2}/atmos"
+mkdir -m 775 -p "${COM_ATMOS_ANALYSIS}"
```
WalterKolczynski-NOAA: Probably only need to create the directory once in the initialize job.

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+# It assumes the runtime directory has already been staged with the appropriate
+# input files and YAML configuration (by the initialize script) before execution.
+#
+################################################################################
+#  Utilities
+export NCP=${NCP:-"/bin/cp"}
+export NMV=${NMV:-"/bin/mv"}
+export NLN=${NLN:-"/bin/ln -sf"}
+CASE=${CASE:-C96}
+RES=$(echo "${CASE}" | cut -c 2-)
```
WalterKolczynski-NOAA: All but `RES` are defined in config.base and don't really need to be redefined here. The `RES` definition can be written without using cut:
```suggestion
RES=${CASE:1}
```

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+ANLDIR=${DATA}/anl
+
+#  Filenames
+export CREATEENS=${CREATEENS:-"${HOMEgfs}/sorc/gdas.cd/ush/land/letkf_create_ens.py"}
+export IMS_IODA=${IMS_IODA:-"${HOMEgfs}/sorc/gdas.cd/build/bin/imsfv3_scf2ioda.py"}
+export CALCFIMS=${CALCFIMS:-"${HOMEgfs}/sorc/gdas.cd/build/bin/calcfIMS.exe"}
+export ADDJEDIINC=${ADDJEDIINC:-"${HOMEgfs}/sorc/gdas.cd/build/bin/apply_incr.exe"}
+
+#export TPATH="/scratch2/NCEPDEV/land/data/fix/C${RES}/"
+#export TPATH=${FIXgfs:-"${HOMEgfs}/fix/orog/C${RES}"}
```
WalterKolczynski-NOAA: The deactivated code can be removed.

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+export TPATH="${HOMEgfs}/fix/orog/C${RES}"
+export TSTUB="C${RES}_oro_data"
+export GFSv17=${GFSv17:-"NO"}
```
WalterKolczynski-NOAA: Not a fan of this variable name at all. Can you use something more meaningful as to what the setting actually is?

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+FILEDATE=${PDY}.${cyc}0000
+DOY=$(date +%j -d "${PDY} + 1 day")
+YYYY=$(echo "${PDY}" | cut -c1-4)
```
WalterKolczynski-NOAA:
```suggestion
YYYY=${PDY:0:4}
```

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+################################################################################
+# Create ensemble member
+WORKDIR=${BKGDIR}
+B=30  # background error std for LETKFOI
```
WalterKolczynski-NOAA:
```suggestion
BERR_STD=30  # background error std for LETKFOI
```

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+if [[ ${GFSv17} == "YES" ]]; then
+    SNOWDEPTHVAR="snodl"
+else
+    SNOWDEPTHVAR="snwdph"
+fi
+
+echo "DONG:GFSv17: ${GFSv17}"
+echo "DONG:SNOWDEPTHVAR: ${SNOWDEPTHVAR}"
+# FOR LETKFOI, CREATE THE PSEUDO-ENSEMBLE
+cd "${WORKDIR}" || exit
```
WalterKolczynski-NOAA: The exit should be non-zero, since this is an error.

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+B=30  # background error std for LETKFOI
+
+if [[ ${GFSv17} == "YES" ]]; then
+    SNOWDEPTHVAR="snodl"
+else
+    SNOWDEPTHVAR="snwdph"
+fi
+
+echo "DONG:GFSv17: ${GFSv17}"
+echo "DONG:SNOWDEPTHVAR: ${SNOWDEPTHVAR}"
```
WalterKolczynski-NOAA: Are these debug statements that you forgot to remove? If you want to keep them, change to `echo "variable=${variable}"`.

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+    done
+    ${NCP} "${WORKDIR}/${FILEDATE}.coupler.res" "${WORKDIR}/mem${ens}/${FILEDATE}.coupler.res"
+done
+
+echo 'do_landDA: calling create ensemble'
+
+python "${CREATEENS}" "${FILEDATE}" "${SNOWDEPTHVAR}" "${B}" "${WORKDIR}"
+
+for ens in 001 002
+do
```
WalterKolczynski-NOAA:
```suggestion
nmem_land=2
ntiles=6
for ens in $(seq -f "03g" 1 ${nmem_land}); do
```

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+    do
+      ${NMV} "${WORKDIR}/mem${ens}/${FILEDATE}.sfc_data.tile${tile}.nc" "${WORKDIR}/mem${ens}/RESTART/"
+    done
+    ${NMV} "${WORKDIR}/mem${ens}/${FILEDATE}.coupler.res" "${WORKDIR}/mem${ens}/RESTART/"
+done
+################################################################################
+# IMS proc
+if [[ ${cyc} == "18" ]]; then
+  WORKDIR=${DATA}/obs
+  cd "${WORKDIR}" || exit
```
WalterKolczynski-NOAA: Should be non-zero exit

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+
+echo 'do_landDA: calling create ensemble'
+
+python "${CREATEENS}" "${FILEDATE}" "${SNOWDEPTHVAR}" "${B}" "${WORKDIR}"
+
+for ens in 001 002
+do
+    mkdir -p "${WORKDIR}/mem${ens}/RESTART"
+    for tile in 1 2 3 4 5 6
+    do
```
WalterKolczynski-NOAA:
```suggestion
    for tile in $(seq 1 ${ntiles}); do
```

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+  imsformat=2,
+  imsversion=1.3,
+  IMS_OBS_PATH="${WORKDIR}/",
+  IMS_IND_PATH="${WORKDIR}/"
+  /
+EOF
+
+  # stage restarts
+  for tile in 1 2 3 4 5 6
+  do
```
WalterKolczynski-NOAA:
```suggestion
for tile in $(seq 1 ${ntiles}); do
```

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+fi
+################################################################################
+# run executable
+export pgm=${JEDIEXE}
+. prep_step
+${APRUN_LANDANL} "${DATA}/fv3jedi_letkf.x" "${DATA}/${CDUMP}.t${cyc}z.letkfoi.yaml" 1>&1 2>&2
+################################################################################
+# add jedi increment
+WORKDIR=${ANLDIR}
+cd "${WORKDIR}" || exit
```
WalterKolczynski-NOAA: Non-zero exit

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+ res=${RES}
+ frac_grid=${GFSv17}
+ orog_path="${TPATH}"
+ otype="${TSTUB}"
+/
+EOF
+
+# stage restarts
+for tile in 1 2 3 4 5 6
+do
```
WalterKolczynski-NOAA:
```suggestion
for tile in $(seq 1 ${ntiles}); do
```

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+  if [[ ! -e ${FILEDATE}.sfc_data.tile${tile}.nc ]]; then
+    ${NCP} "${BKGDIR}/${FILEDATE}.sfc_data.tile${tile}.nc"  "${WORKDIR}/${FILEDATE}.sfc_data.tile${tile}.nc"
+  fi
+done
+
+# change names of the increments
+for tile in 1 2 3 4 5 6
+do
```
WalterKolczynski-NOAA:
```suggestion
  # change names of the increments
```

workflow/rocoto/workflow_tasks.py
```diff
@@ -543,6 +549,52 @@ def aeroanlfinal(self):
 
         return task
 
+    def landanlinit(self):
+
+        dump_suffix = self._base["DUMP_SUFFIX"]
+        dmpdir = self._base["DMPDIR"]
+        atm_hist_path = self._template_to_rocoto_cycstring(self._base["COM_ATMOS_HISTORY_TMPL"], {'RUN': 'gdas'})
+        dump_path = self._template_to_rocoto_cycstring(self._base["COM_OBSDMP_TMPL"],
+                                                       {'DMPDIR': dmpdir, 'DUMP_SUFFIX': dump_suffix})
```
WalterKolczynski-NOAA: You are adding a dependency on COM_OBSDMP, but the script is using COM_OBS. I think you should probably be using COM_OBS_TMPL here instead.
```suggestion
        dump_path = self._template_to_rocoto_cycstring(self._base["COM_OBS_TMPL"] , {'RUN': self.cdump})
```

ush/python/pygfs/task/land_analysis.py
```diff
@@ -0,0 +1,265 @@
...
+                'GPREFIX': f"gdas.t{self.runtime_config.previous_cycle.hour:02d}z.",
+                'fv3jedi_yaml': _fv3jedi_yaml,
+            }
+        )
+
+        # task_config is everything that this task should need
+        self.task_config = AttrDict(**self.config, **self.runtime_config, **local_dict)
+
+    @logit(logger)
+    def initialize(self: Analysis) -> None:
```
WalterKolczynski-NOAA, line 160: Let @aerorahul confirm, but I think we may want to make each part of this a separate functions that all get called from here, e.g. `stage()`, a `generate_yaml()`, etc.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -0,0 +1,265 @@
...
+            exec_cmd()
+        except OSError:
+            raise OSError(f"Failed to execute {exec_cmd}")
+        except Exception:
+            raise WorkflowException(f"An error occured during execution of {exec_cmd}")
+
+        pass
+
+    @logit(logger)
+    def finalize(self: Analysis) -> None:
```
WalterKolczynski-NOAA, line 320: Same as initialize, may want to divide this up into pieces. CC: @aerorahul

ush/python/pygfs/task/land_analysis.py
```diff
@@ -0,0 +1,265 @@
...
+        This method will initialize a global land analysis using JEDI.
+        This includes:
+        #- staging CRTM fix files
+        - staging FV3-JEDI fix files
+        #- staging B error files
+        - staging model backgrounds
+        - generating a YAML file for the JEDI executable
+        - linking the JEDI executable (TODO make it copyable, requires JEDI fix)
+        - creating output directories
+        """
```
WalterKolczynski-NOAA, line 178: All of the function docstrings should be updated to the Numpy standard: https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

You can see what Russ did in atmens as an example (there are also others in workflow: https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/python/pygfs/task/atmens_analysis.py)

WalterKolczynski-NOAA (changes requested):

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
+#! /usr/bin/env bash
```
aerorahul, line 1: I don't understand what is going in this script, but seems like we are deviating from the `execute` method from `LandAnalysis`.
@CoryMartin-NOAA 
Do you have an understanding of the requirements for the `execute`?

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+
+  # stage observations
+  ${NCP} "${COMIN_OBS}/gdas.t${cyc}z.ims${YYYY}${DOY}_4km_v1.3.nc" "${WORKDIR}/ims${YYYY}${DOY}_4km_v1.3.nc"
+  ${NCP} "${COMIN_OBS}/gdas.t${cyc}z.IMS4km_to_FV3_mapping.C${RES}_oro_data.nc" "${WORKDIR}/IMS4km_to_FV3_mapping.C${RES}_oro_data.nc"
+
+  ulimit -Ss unlimited
+  ${CALCFIMS}
+
+##  export PYTHONPATH=${PYTHONPATH}:${gdasapp_dir}/build/lib/python${Python3_VERSION_MAJOR}.${Python3_VERSION_MINOR}/pyioda/
+  export PYTHONPATH=${PYTHONPATH}:${gdasapp_dir}/iodaconv/src/:${gdasapp_dir}/build/lib/python3.7/pyioda/
```
aerorahul: what is all this and why is this in an exscript?

scripts/exglobal_land_analysis_run.sh
```diff
@@ -0,0 +1,175 @@
...
+    if [[ ! -e ${FILEDATE}.sfc_data.tile${tile}.nc ]]; then
+      ${NCP} "${BKGDIR}/${FILEDATE}.sfc_data.tile${tile}.nc" "${WORKDIR}"
+    fi
+  done
+
+  # stage observations
+  ${NCP} "${COMIN_OBS}/gdas.t${cyc}z.ims${YYYY}${DOY}_4km_v1.3.nc" "${WORKDIR}/ims${YYYY}${DOY}_4km_v1.3.nc"
+  ${NCP} "${COMIN_OBS}/gdas.t${cyc}z.IMS4km_to_FV3_mapping.C${RES}_oro_data.nc" "${WORKDIR}/IMS4km_to_FV3_mapping.C${RES}_oro_data.nc"
+
+  ulimit -Ss unlimited
```
aerorahul: why is this in an exscript?

scripts/exglobal_land_analysis_run.sh
aerorahul: This script needs to be reworked significantly.
Is it possible to outline what is being done in this script in a flow diagram so we can suggest a solution?

ush/python/pygfs/task/land_analysis.py
```diff
@@ -0,0 +1,265 @@
...
+
+        try:
+            logger.debug(f"Executing {exec_cmd}")
+            exec_cmd()
+        except OSError:
+            raise OSError(f"Failed to execute {exec_cmd}")
+        except Exception:
+            raise WorkflowException(f"An error occured during execution of {exec_cmd}")
+
+        pass
```
aerorahul: why `pass`?
`pass` is used when the method isn't ready for use and is just a stub.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -0,0 +1,265 @@
...
+        # need output dir for diags and anl
+        logger.debug("Create empty output [anl, diags] directories to receive output from executable")
+        newdirs = [
+            os.path.join(self.task_config['DATA'], 'anl'),
+            os.path.join(self.task_config['DATA'], 'diags'),
+        ]
+        FileHandler({'mkdir': newdirs}).sync()
+
+    @logit(logger)
+    def execute(self: Analysis) -> None:
```
aerorahul, line 219: what is the use of this method if a shell script is being used?

ush/python/pygfs/task/land_analysis.py
```diff
@@ -0,0 +1,265 @@
...
+            src = os.path.join(self.task_config.DATA, 'anl', sfcdata)
+            dest = os.path.join(self.task_config.COM_ATMOS_ANALYSIS, sfcdata)
+            inclist.append([src, dest])
+        FileHandler({'copy': inclist}).sync()
+
+    def clean(self):
+        super().clean()
+
+    @logit(logger)
+    def _add_fms_cube_sphere_increments(self: Analysis) -> None:
```
aerorahul, line 411: where is this method being used?

ush/python/pygfs/task/land_analysis.py
```diff
@@ -0,0 +1,265 @@
...
+        for itile in range(1, self.task_config.ntiles + 1):
+            sfcdata = template.format(tilenum=itile)
+            src = os.path.join(self.task_config.COM_ATMOS_RESTART_PREV, sfcdata)
+            dest = os.path.join(self.task_config.COM_ATMOS_ANALYSIS, f'landges.{sfcdata}')
+            bkglist.append([src, dest])
+        FileHandler({'copy': bkglist}).sync()
+
+        # ---- add increments to RESTART files
+        # logger.info('Adding increments to RESTART files')
+        # self._add_fms_cube_sphere_increments()
```
aerorahul: This is commented out, then why is there a need for the method?

aerorahul (changes requested):
The python and the shell script needs more work.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: @WalterKolczynski-NOAA The above `for ens in $(seq -f "03g" 1 ${nmem_land}); do` doesn't seem work, and I got the following errors:

`seq: format '03g' has no % directive`

Any suggestions? Thanks.

jiaruidong2017: @WalterKolczynski-NOAA I fixed it by using `for ens in $(seq -f %03g 1 ${nmem_land}); do`

jobs/JGLOBAL_LAND_ANALYSIS_FINALIZE
jiaruidong2017: Deleted this line.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: Made changes as suggested.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: Removed.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: Made changes to all three variable names.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: Made changes as suggested.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: Done the changes.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: Done the changes as suggested.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: These are debug statements. Removed.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: Done the changes.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: Done the changes as suggested.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: Done the changes as suggested.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: Added 99 to exit.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: Done the changes as suggested.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: Done the changes as suggested.

workflow/rocoto/workflow_tasks.py
jiaruidong2017: For our DA cycling tests, we use `DMPDIR` to define for OBS data in the `config.base`. Therefore, we use `COM_OBSDMP_TMPL` here, which is the same as `ATM DA` and `AERO DA`. @CoryMartin-NOAA Am I correct? Thanks.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: Made the changes to add the function docstrings. Thanks.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: Yes, you are right. This script includes some from `initialize`, all the `execute`, and some from `finalize`. There are four separate sections in this run script `scripts/exglobal_land_analysis_run.sh` for the snow DA. They are 1) create the background ensembles; 2) process the observed IMS data; 3) conduct JEDI runs; and 4) apply the JEDI increments. We have discussed and agreed to use the shell scripts here for this snow DA implementation initially due to the limited time. 

The above four processes were implemented in the `GDASApp` separately as below: 

1. Create the background ensembles ([Issue #199](https://github.com/NOAA-EMC/GDASApp/issues/199) and [PR #207](https://github.com/NOAA-EMC/GDASApp/pull/207)): A python code is used to convert the model restart files into two separate ensembles. 

2. IMS preprocessing ([Issue #198](https://github.com/NOAA-EMC/GDASApp/issues/198) and [PR #216](https://github.com/NOAA-EMC/GDASApp/pull/216)): Code for processing IMS input ascii/netcdf files on the UFS model grid. 

- Inputs: IMS ascii/netcdf file, IMS index file (is model resolution specific) and model restart files. 
- Output: file with i) IMS-derived snow cover fraction over land on UFS model grid, and ii) snow depth derived form the IMS snow cover fraction, using an inversion of the noah model snow depletion curve. 

3. Conduct JEDI run ([Issue #241](https://github.com/NOAA-EMC/GDASApp/issues/241) and [PR #248](https://github.com/NOAA-EMC/GDASApp/pull/248)): Scripts to perform the snow DA, to assimilate GTS snow depth and IMS snow cover obs, using the JEDI LETKF and for the noahMP land surface model.

4. Apply the DA increments generated by JEDI to UFS sfc_data restart ([Issue #200](https://github.com/NOAA-EMC/GDASApp/issues/200) and [PR #233](https://github.com/NOAA-EMC/GDASApp/pull/233)). Currently, only option is to add snow depth increment to the Noah-MP land surface model. Above based on disaggregation code written by Mike Barlage.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: I need help from @CoryMartin-NOAA for this.

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: Deleted this line.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: This is the same as the atm DA and aero DA.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: The `execute` function here is a place holder, which is not used in the land DA. I can delete this function.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: This `_add_fms_cube_sphere_increments` is not used in the current land DA, which can be removed.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: The apply JEDI increments is implemented in the `scripts/exglobal_land_analysis_run.sh`. Therefore, `_add_fms_cube_sphere_increments` is not used here in the land DA and can be removed.

jiaruidong2017: Removed.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: I added this back for moving the `apply_jedi_increment` from `gdaslandanlrun` into `gdaslandanlfinal`.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: The `execute` function is removed.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: This function is removed.

jobs/JGLOBAL_LAND_ANALYSIS_FINALIZE
```diff
@@ -8,19 +8,23 @@ source "${HOMEgfs}/ush/jjob_header.sh" -e "landanlfinal" -c "base landanl landan
...
 GDATE=$(date +%Y%m%d%H -d "${PDY} ${cyc} - ${assim_freq} hours")
+gPDY=${GDATE:0:8}
+gcyc=${GDATE:8:2}
 GDUMP="gdas"
 
 ##############################################
 # Begin JOB SPECIFIC work
 ##############################################
+# Generate COM variables from templates
+YMD=${PDY} HH=${cyc} generate_com -rx COM_ATMOS_ANALYSIS
```
CoryMartin-NOAA: I think we talked about earlier in the week keeping a land analysis directory (for things like diagnostics, increments, etc.)?

parm/config/yaml/defaults.yaml
```diff
@@ -17,3 +17,8 @@ ocnanal:
   R2D2_OBS_DUMP: 's2s_v1'
   NICAS_RESOL: 1
   NICAS_GRID_SIZE: 15000
+
```
CoryMartin-NOAA: isn't this already in this file (see up like 10 lines or so)

scripts/exglobal_aero_analysis_finalize.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-# exgdas_global_aero_analysis_finalize.py
+# exglobal_aero_analysis_finalize.py
```
CoryMartin-NOAA: should all of these 9 files that have the comments changed be in this PR or separate? @aerorahul @WalterKolczynski-NOAA

scripts/exglobal_atmos_sfcanl.sh
```diff
@@ -190,8 +190,8 @@ fi
 
 # Update surface restarts at middle of window
 for n in $(seq 1 ${ntiles}); do
-    ${NCP} "${COM_ATMOS_RESTART_PREV}/${PDY}.${cyc}0000.sfc_data.tile${n}.nc" \
-            "${COM_ATMOS_RESTART}/${PDY}.${cyc}0000.sfcanl_data.tile${n}.nc"
+    ${NCP} "${COM_ATMOS_ANALYSIS}/${PDY}.${cyc}0000.sfc_data.tile${n}.nc" \
```
CoryMartin-NOAA: is this dangerous, what if there is no land DA, will there always be land DA?
General question though for @aerorahul, which is the preferred approach (for this and aero) as long as we are updating the RESTART files, can we save the GES somewhere and update the RESTART in place, or do we need to not touch previously written RESTART, write ANL, and make sure the fcst job knows where to look? Basically, sfcanl does it one way, aero does it another, and we should be consistent (this is on me for not asking this until now)

scripts/exglobal_land_analysis_run.sh
CoryMartin-NOAA: @jiaruidong2017, @aerorahul and I spoke today and we think this is probably the eventual path forward:
1 and 3 will go into the landanlrun job
2 will get a new landprep job (for eventual transition to obsproc)
4 will go to landanlfinal job
Does that make sense/sound reasonable?

scripts/exglobal_land_analysis_run.sh
CoryMartin-NOAA: @aerorahul with these python ioda-converters, we probably need a good solution for this... something to think about when designing the socaprep and landprep jobs I think

workflow/rocoto/workflow_tasks.py
CoryMartin-NOAA: @jiaruidong2017 I think the preferred way going forward @aerorahul correct me if I'm wrong, is that the prep job copies/links obs to COM_OBS and then our DA jobs grab them from there. Thus this should have the same dependency as the atmanlinit or aeroanlinit jobs.

CoryMartin-NOAA (commented):
Some comments

scripts/exglobal_aero_analysis_finalize.py
aerorahul: that's fine.

scripts/exglobal_atmos_sfcanl.sh
aerorahul: good point @CoryMartin-NOAA
I think we can handle this with a conditional on whether Land DA is exercised or not.
Or is that too simple and I missing something?

scripts/exglobal_land_analysis_run.sh
aerorahul: lets handle this in a PR that prepares the observations. 
There is a lot going on in this PR at the moment.

workflow/rocoto/workflow_tasks.py
aerorahul: @CoryMartin-NOAA is correct.

scripts/exglobal_atmos_sfcanl.sh
CoryMartin-NOAA: Conditional is fine, so that would mean you would generally prefer to leave the original RESTART untouched and instead copy/link to the analysis? perhaps this is something you and I should discuss more next week in detail?

jiaruidong2017: @CoryMartin-NOAA This is a great finding, and I will try to work on this.

workflow/rocoto/workflow_tasks.py
jiaruidong2017: Removed the dependency on `COM_OBSDMP_TMPL`, and use `COM_OBS` for obs. Thanks @CoryMartin-NOAA

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: @CoryMartin-NOAA Can we talk how to proceed with this? Thanks.

jobs/JGLOBAL_LAND_ANALYSIS_FINALIZE
jiaruidong2017: As suggested, I added the `COM_LAND_ANALYSIS` to keep a land analysis directory for the land related increments, diagnostics, etc.

parm/config/yaml/defaults.yaml
jiaruidong2017: I just added the following in this file: 

```
landanl:
  IO_LAYOUT_X: 1
  IO_LAYOUT_Y: 1
```

scripts/exglobal_atmos_sfcanl.sh
jiaruidong2017: As suggested, I leave the original RESTART untouched and copy the land analysis to SFCANL when available. I added the following in the script:

```
    if [[ -e "${COM_LAND_ANALYSIS}/${PDY}.${cyc}0000.sfc_data.tile${n}.nc" ]]; then
        ${NCP} "${COM_LAND_ANALYSIS}/${PDY}.${cyc}0000.sfc_data.tile${n}.nc" \
               "${COM_ATMOS_RESTART}/${PDY}.${cyc}0000.sfcanl_data.tile${n}.nc"
    fi
```

scripts/exglobal_land_analysis_run.sh
jiaruidong2017: I have added a new `landanlprep` job for the IMS preprocessing. The following codes were added: 
```
        jobs/JGLOBAL_LAND_ANALYSIS_PREP
        jobs/rocoto/landanlprep.sh
        parm/config/config.landanlprep
        scripts/exglobal_land_analysis_prep.sh
```

@CoryMartin-NOAA For `4 will go to landanlfinal job` you mentioned above, I will need help on how to move this part (a shell script) into the `landanlfinal` job (a python code).  Thanks.

parm/config/yaml/defaults.yaml
CoryMartin-NOAA: Yeah I guess my question is, isn't this already in this file? Why is it here twice?

aerorahul: The way the `defaults.yaml` is set, the key (in this case `landanl`) will change the templates in `config.landanl`. It is present in this file under a different key, so it will not impact `landanl`.

aerorahul: It also allows `landanl` to have "different" layout, if they so desire.

scripts/exglobal_atmos_sfcanl.sh
```diff
@@ -192,6 +192,10 @@ fi
 for n in $(seq 1 ${ntiles}); do
     ${NCP} "${COM_ATMOS_RESTART_PREV}/${PDY}.${cyc}0000.sfc_data.tile${n}.nc" \
             "${COM_ATMOS_RESTART}/${PDY}.${cyc}0000.sfcanl_data.tile${n}.nc"
+    if [[ -e "${COM_LAND_ANALYSIS}/${PDY}.${cyc}0000.sfc_data.tile${n}.nc" ]]; then
```
CoryMartin-NOAA: I don't think this is the best solution, this is copying things 2x then. Instead it should be something like:
```
if DO_LANDDA:
  cp from COM_LAND_ANALSYS
else:
  cp from COM_ATMOS_RESTART_PREV
```

parm/config/yaml/defaults.yaml
CoryMartin-NOAA: https://github.com/NOAA-EMC/global-workflow/blob/7b3e53ebf5f779aed652f11877935d507f3736ea/parm/config/yaml/defaults.yaml 

What is the difference between lines 5-7 and lines 21-23?

aerorahul: doh. Yeah.  I did not notice that.  Thanks for clarifying.
You are right in this case.
These lines should be deleted.

scripts/exglobal_atmos_sfcanl.sh
aerorahul: I think I would like to improve this further. 
`gaussian_sfcanl` job creates the `sfcanl_data.tile${n}.nc` and it is placed in `COM_ATMOS_RESTART` for historic reasons.
It should be in `COM_LAND_ANALYSIS` regardless of whether LandDA is ON/OFF in light with the recent development.
I am sure there are cons to that approach too.

If `DO_LANDDA` is true, do we need to run `exglobal_atmos_sfcanl.sh`?

CoryMartin-NOAA: for now, yes, global_cycle does many things, of which snow DA only is doing a small bit. So we need both for at least the short term (I think that will be the case for v17 but may need to confirm with Clara)

scripts/exglobal_land_analysis_prep.sh
```diff
@@ -0,0 +1,81 @@
...
+# This script processes the IMS input ascii/netcdf files onto the UFS model grid.
+# Inputs: IMS ascii/netcdf file, IMS index file (is model resolution specific) 
+#         and model restart files.
+# Outputs: file with i) IMS-derived snow cover fraction over land on UFS model 
+#         grid, and ii) snow depth derived form the IMS snow cover fraction, 
+#         using an inversion of the noah model snow depletion curve.
+#
+################################################################################
+#  Utilities
+RES=${CASE:1}
```
CoryMartin-NOAA: @aerorahul after talking with @ClaraDraper-NOAA it turns out this IMS snow cover processing is not just dependent on resolution but the snow background itself, does that change your thinking in terms of a separate 'prep' step? Or should we keep it separate a la prepbufr?

scripts/exglobal_atmos_sfcanl.sh
jiaruidong2017: @CoryMartin-NOAA Thanks for your review and suggestions. Based on your suggestion, I made the following changes:

```
    if [[ ${DO_JEDILANDDA} = "YES" ]]; then
        ${NCP} "${COM_LAND_ANALYSIS}/${PDY}.${cyc}0000.sfc_data.tile${n}.nc" \
               "${COM_ATMOS_RESTART}/${PDY}.${cyc}0000.sfcanl_data.tile${n}.nc"
    else
        ${NCP} "${COM_ATMOS_RESTART_PREV}/${PDY}.${cyc}0000.sfc_data.tile${n}.nc" \
               "${COM_ATMOS_RESTART}/${PDY}.${cyc}0000.sfcanl_data.tile${n}.nc"
    fi
```
I conducted the test run, and it worked well.

scripts/exglobal_land_analysis_prep.sh
jiaruidong2017: The current implementation is as following: 

image

I just move the IMS processing into the `gdaslandanlprep` job, and the other three processes (create ensembles, jedi runs and apply increments) are still in the `gdaslandanlrun` job. @CoryMartin-NOAA @aerorahul Please let me know if you want the current implementation or want me to move the IMS processing back to the `gdaslandanlrun` job. Thanks.

aerorahul: Thank you @jiaruidong2017 
Moving the IMS processing to `gdaslandanlprep` is sufficient.

Presently, applying the increments is being done in the `final` jobs for aerosols and atmosphere, since that executable does not need the resources of the same magnitude as the `run` job.

At present, I am inclined to keep the `create ensembles` and `run jedi executable` in the `run` job, if that is acceptable to you and @CoryMartin-NOAA

scripts/exglobal_land_analysis_prep.sh
```diff
@@ -0,0 +1,81 @@
...
+
+  # stage restarts
+  for tile in $(seq 1 "${ntiles}"); do
+    if [[ ! -e ${FILEDATE}.sfc_data.tile${tile}.nc ]]; then
+      ${NCP} "${BKGDIR}/${FILEDATE}.sfc_data.tile${tile}.nc" "${WORKDIR}"
+    fi
+  done
+
+  # stage observations
+  ${NCP} "${COM_OBS}/gdas.t${cyc}z.ims${YYYY}${DOY}_4km_v1.3.nc" "${WORKDIR}/ims${YYYY}${DOY}_4km_v1.3.nc"
```
aerorahul: Please use `DATA` instead of `WORKDIR`

scripts/exglobal_land_analysis_prep.sh
```diff
@@ -0,0 +1,81 @@
...
+if [[ ${cyc} == "18" ]]; then
+  WORKDIR=${DATA}/obs
+  cd "${WORKDIR}" || exit 99
+  if [[ -e fims.nml ]]; then
+    rm fims.nml
+  fi
+
+cat >> fims.nml << EOF
+ &fIMS_nml
+  idim=${RES}, jdim=${RES},
```
aerorahul:
```suggestion
  idim=${CASE/C/}, jdim=${CASE/C/},
```

scripts/exglobal_land_analysis_prep.sh
```diff
@@ -0,0 +1,81 @@
...
+  done
+
+  # stage observations
+  ${NCP} "${COM_OBS}/gdas.t${cyc}z.ims${YYYY}${DOY}_4km_v1.3.nc" "${WORKDIR}/ims${YYYY}${DOY}_4km_v1.3.nc"
+  ${NCP} "${COM_OBS}/gdas.t${cyc}z.IMS4km_to_FV3_mapping.C${RES}_oro_data.nc" "${WORKDIR}/IMS4km_to_FV3_mapping.C${RES}_oro_data.nc"
+
+  ${CALCFIMS}
+
+##  export PYTHONPATH=${PYTHONPATH}:${gdasapp_dir}/build/lib/python${Python3_VERSION_MAJOR}.${Python3_VERSION_MINOR}/pyioda/
+  export PYTHONPATH=${PYTHONPATH}:${gdasapp_dir}/iodaconv/src/:${gdasapp_dir}/build/lib/python3.7/pyioda/
```
aerorahul: @CoryMartin-NOAA 
We need to find a solution for this and soon.

scripts/exglobal_land_analysis_prep.sh
```diff
@@ -0,0 +1,81 @@
...
+# Outputs: file with i) IMS-derived snow cover fraction over land on UFS model 
+#         grid, and ii) snow depth derived form the IMS snow cover fraction, 
+#         using an inversion of the noah model snow depletion curve.
+#
+################################################################################
+#  Utilities
+RES=${CASE:1}
+
+#  Directories.
+export gdasapp_dir=${gdasapp_dir:-"${HOMEgfs}/sorc/gdas.cd/"}
```
aerorahul: Why is `gdasapp_dir` needed?

jobs/JGLOBAL_LAND_ANALYSIS_PREP
```diff
@@ -0,0 +1,39 @@
...
+# Begin JOB SPECIFIC work
+##############################################
+# Generate COM variables from templates
+YMD=${PDY} HH=${cyc} generate_com -rx COM_OBS
+
+
+###############################################################
+# Run relevant script
+
+EXSCRIPT=${GDASLANDPREPSH:-${HOMEgfs}/scripts/exglobal_land_analysis_prep.sh}
```
aerorahul: I would prefer to write all exscripts in python, similar to the aerosol and atmos scripts.

scripts/exglobal_atmos_sfcanl.sh
```diff
@@ -190,8 +190,13 @@ fi
 
 # Update surface restarts at middle of window
 for n in $(seq 1 ${ntiles}); do
-    ${NCP} "${COM_ATMOS_RESTART_PREV}/${PDY}.${cyc}0000.sfc_data.tile${n}.nc" \
-            "${COM_ATMOS_RESTART}/${PDY}.${cyc}0000.sfcanl_data.tile${n}.nc"
+    if [[ ${DO_JEDILANDDA} = "YES" ]]; then
```
aerorahul:
```suggestion
    if [[ ${DO_JEDILANDDA:-"NO"} = "YES" ]]; then
```

scripts/exglobal_land_analysis_prep.sh
```diff
@@ -0,0 +1,81 @@
...
+BKGDIR=${DATA}/bkg
+
+#  Filenames
+export IMS_IODA=${IMS_IODA:-"${HOMEgfs}/sorc/gdas.cd/build/bin/imsfv3_scf2ioda.py"}
+export CALCFIMS=${CALCFIMS:-"${HOMEgfs}/sorc/gdas.cd/build/bin/calcfIMS.exe"}
+
+export OROGTYPE="C${RES}_oro_data"
```
aerorahul: Why is this being exported?
```suggestion
OROGTYPE="${CASE}_oro_data"
```

scripts/exglobal_land_analysis_prep.sh
```diff
@@ -0,0 +1,81 @@
...
+FILEDATE=${PDY}.${cyc}0000
+DOY=$(date +%j -d "${PDY} + 1 day")
+YYYY=${PDY:0:4}
+ntiles=6
+################################################################################
+# IMS proc
+if [[ ${cyc} == "18" ]]; then
+  WORKDIR=${DATA}/obs
+  cd "${WORKDIR}" || exit 99
```
aerorahul: Please don't just exit.  Please put out an error message with FATAL ERROR and then exit.  Look at other examples.

scripts/exglobal_land_analysis_prep.sh
```diff
@@ -0,0 +1,81 @@
...
+  ${NCP} "${COM_OBS}/gdas.t${cyc}z.IMS4km_to_FV3_mapping.C${RES}_oro_data.nc" "${WORKDIR}/IMS4km_to_FV3_mapping.C${RES}_oro_data.nc"
+
+  ${CALCFIMS}
+
+##  export PYTHONPATH=${PYTHONPATH}:${gdasapp_dir}/build/lib/python${Python3_VERSION_MAJOR}.${Python3_VERSION_MINOR}/pyioda/
+  export PYTHONPATH=${PYTHONPATH}:${gdasapp_dir}/iodaconv/src/:${gdasapp_dir}/build/lib/python3.7/pyioda/
+
+  echo "PYTHONPATH: ${PYTHONPATH}"
+  echo 'do_landDA: calling ioda converter'
+  python "${IMS_IODA}" -i "IMSscf.${PDY}.${OROGTYPE}.nc" -o "ioda.IMSscf.${PDY}.${OROGTYPE}.nc"
```
aerorahul: We cannot do `python ...`. The `IMS_IODA` script needs to have executable permissions.

scripts/exglobal_land_analysis_prep.sh
```diff
@@ -0,0 +1,81 @@
...
+
+  if [[ -e "ioda.IMSscf.${PDY}.C${RES}_oro_data.nc"  ]]; then
+    ${NCP} "ioda.IMSscf.${PDY}.C${RES}_oro_data.nc" "gdas.t${cyc}z.ims_snow_${PDY}${cyc}.nc4"
```
aerorahul:
```suggestion
  if [[ -e "ioda.IMSscf.${PDY}.${CASE}_oro_data.nc"  ]]; then
    ${NCP} "ioda.IMSscf.${PDY}.${CASE}_oro_data.nc" "${RUN}.t${cyc}z.ims_snow_${PDY}${cyc}.nc4"
```

scripts/exglobal_land_analysis_prep.sh
```diff
@@ -0,0 +1,81 @@
...
+  # stage restarts
+  for tile in $(seq 1 "${ntiles}"); do
+    if [[ ! -e ${FILEDATE}.sfc_data.tile${tile}.nc ]]; then
+      ${NCP} "${BKGDIR}/${FILEDATE}.sfc_data.tile${tile}.nc" "${WORKDIR}"
+    fi
+  done
+
+  # stage observations
+  ${NCP} "${COM_OBS}/gdas.t${cyc}z.ims${YYYY}${DOY}_4km_v1.3.nc" "${WORKDIR}/ims${YYYY}${DOY}_4km_v1.3.nc"
+  ${NCP} "${COM_OBS}/gdas.t${cyc}z.IMS4km_to_FV3_mapping.C${RES}_oro_data.nc" "${WORKDIR}/IMS4km_to_FV3_mapping.C${RES}_oro_data.nc"
```
aerorahul: Please use `${RUN}` instead of hard-wiring `gdas`.
```suggestion
  ${NCP} "${COM_OBS}/${RUN}.t${cyc}z.IMS4km_to_FV3_mapping.${CASE}_oro_data.nc" "${WORKDIR}/IMS4km_to_FV3_mapping.${CASE}_oro_data.nc"
```

scripts/exglobal_land_analysis_prep.sh
aerorahul: Please do not hard-wire `RUN` variable.
```suggestion
  ${NCP} "${COM_OBS}/${RUN}.t${cyc}z.ims${YYYY}${DOY}_4km_v1.3.nc" "${WORKDIR}/ims${YYYY}${DOY}_4km_v1.3.nc"
```

scripts/exglobal_land_analysis_prep.sh
```diff
@@ -0,0 +1,81 @@
+#! /usr/bin/env bash
```
aerorahul: We prefer all new exscripts added to be in python unless there is a very strong argument to not do so.

scripts/exglobal_land_analysis_prep.sh
```diff
@@ -0,0 +1,81 @@
...
+#  Utilities
+RES=${CASE:1}
+
+#  Directories.
+export gdasapp_dir=${gdasapp_dir:-"${HOMEgfs}/sorc/gdas.cd/"}
+BKGDIR=${DATA}/bkg
+
+#  Filenames
+export IMS_IODA=${IMS_IODA:-"${HOMEgfs}/sorc/gdas.cd/build/bin/imsfv3_scf2ioda.py"}
+export CALCFIMS=${CALCFIMS:-"${HOMEgfs}/sorc/gdas.cd/build/bin/calcfIMS.exe"}
```
aerorahul: Why are these not in `ush/` and `exec/` respectively?

scripts/exglobal_land_analysis_prep.sh
aerorahul: Why do we need `RES`?  It can be easily inferred as `${CASE/C/}`

scripts/exglobal_land_analysis_prep.sh
aerorahul: Also, they don't need to be exported

scripts/exglobal_land_analysis_prep.sh
jiaruidong2017: Thanks @aerorahul for your review and efforts, and I will make the changes as you suggested.

CoryMartin-NOAA:
@jiaruidong2017 except for 2-3 scripts I think this PR is ready. Would it be possible for you to pull out the 'remaining' files into a new branch/PR and reduce this one to the things @aerorahul and I have no concerns with? Do you know how to do that or would you like @aerorahul to take the lead on pruning this PR for that purpose?

scripts/exglobal_land_analysis_prep.sh
jiaruidong2017: Replaced all the `WORKDIR` by `DATA`.

scripts/exglobal_land_analysis_prep.sh
jiaruidong2017: Replaced `RES` with `CASE` as suggested.

scripts/exglobal_land_analysis_prep.sh
jiaruidong2017: If the above `PYTHONPATH` issue is solved, this `gdasapp_dir` will be removed.

jobs/JGLOBAL_LAND_ANALYSIS_PREP
jiaruidong2017: This is on the schedule after I move the `apply_jedi_increments` into the `gdaslandanlfinal` job.

scripts/exglobal_atmos_sfcanl.sh
jiaruidong2017: Made the changes as suggested.

scripts/exglobal_land_analysis_prep.sh
jiaruidong2017: Removed the `export `and made the changes.

scripts/exglobal_land_analysis_prep.sh
jiaruidong2017: I checked other scripts (e.g., `exglobal_atmos_sfcanl.sh`), and they just use `cd ${DATA} || exit 99`. Do you have any specific suggestion so that I can make the correct changes. Thanks.

scripts/exglobal_land_analysis_prep.sh
jiaruidong2017: @aerorahul I don't quite follow this. Would you please guide me how to make the corresponding changes? Thank you.

scripts/exglobal_land_analysis_prep.sh
jiaruidong2017: Done the changes.

scripts/exglobal_land_analysis_prep.sh
jiaruidong2017: Done the changes.

scripts/exglobal_land_analysis_prep.sh
jiaruidong2017: Done the changes.

scripts/exglobal_land_analysis_prep.sh
jiaruidong2017: As mentioned above, this work can be done after moving the `apply_jedi_increments` into the `gdaslandanlfinal` job.

scripts/exglobal_land_analysis_prep.sh
jiaruidong2017: The `imsfv3_sfc2ioda.py` and the `calcfIMS.exe` are from GDASApp. We might to link/copy them into `ush/` and `exec/`. Great suggestions, and I will check the `link_workflow.sh` for this. Thanks.

scripts/exglobal_land_analysis_prep.sh
jiaruidong2017: Removed this.

jiaruidong2017:
> @jiaruidong2017 except for 2-3 scripts I think this PR is ready. Would it be possible for you to pull out the 'remaining' files into a new branch/PR and reduce this one to the things @aerorahul and I have no concerns with? Do you know how to do that or would you like @aerorahul to take the lead on pruning this PR for that purpose?

Thanks @CoryMartin-NOAA. It would be great if @aerorahul can take the lead on pruning this PR and guide me how to proceed. I appreciated the great efforts from you all.

scripts/exglobal_land_analysis_prep.sh
jiaruidong2017: Done this.

aerorahul:
> > @jiaruidong2017 except for 2-3 scripts I think this PR is ready. Would it be possible for you to pull out the 'remaining' files into a new branch/PR and reduce this one to the things @aerorahul and I have no concerns with? Do you know how to do that or would you like @aerorahul to take the lead on pruning this PR for that purpose?
> 
> Thanks @CoryMartin-NOAA. It would be great if @aerorahul can take the lead on pruning this PR and guide me how to proceed. I appreciated the great efforts from you all.

@jiaruidong2017 
Sounds good.  I will start with the hash [ba3aaaf](https://github.com/NOAA-EMC/global-workflow/pull/1541/commits/ba3aaaf922447261d8031a6b57cc8e43c51cf847).
Please don't push anything more to this branch till I open a PR.

jiaruidong2017:
> > > @jiaruidong2017 except for 2-3 scripts I think this PR is ready. Would it be possible for you to pull out the 'remaining' files into a new branch/PR and reduce this one to the things @aerorahul and I have no concerns with? Do you know how to do that or would you like @aerorahul to take the lead on pruning this PR for that purpose?
> > 
> > 
> > Thanks @CoryMartin-NOAA. It would be great if @aerorahul can take the lead on pruning this PR and guide me how to proceed. I appreciated the great efforts from you all.
> 
> @jiaruidong2017 Sounds good. I will start with the hash [ba3aaaf](https://github.com/NOAA-EMC/global-workflow/pull/1541/commits/ba3aaaf922447261d8031a6b57cc8e43c51cf847). Please don't push anything more to this branch till I open a PR.

I just merged to develop before I saw your message.  Would you please use the new hash [4a30ee5](https://github.com/NOAA-EMC/global-workflow/pull/1541/commits/4a30ee5cab024b710fe7074b32d519e93932687d)? Thanks.

aerorahul:
@jiaruidong2017 
I can help you next week to finish off this work.

jiaruidong2017:
> @jiaruidong2017 I can help you next week to finish off this work.

Great! Thanks @aerorahul

aerorahul:
@jiaruidong2017 
Can this draft PR be closed as a result of #1635?

jiaruidong2017:
Yes, sure. Thanks @aerorahul

PR #1551, opened 2023-04-28
Update aerosol DA to use new COM structure

### LINKED ISSUE #1516: Update UFSDA Aerosols for COM refactor
**Description**
UFSDA jobs were not included in the initial COM refactor work (#1421). These jobs will all need to be updated to use the new structure and templates before they can run again.

**Requirements**
UFSDA Aerosol jobs that utilize the new COM templates to generate COM directory paths.

UFSDA Aerosol jobs that work as well as before the COM refactor

There are two approaches we can take. The first is to define the full paths in the j-job using the templates, as is done for most of the existing global-workflow jobs, and update the python to expect full paths (with dates and cycles). The second is to use the templates directly in the python (reading the `*_TMPL` variables from the environment) and using Rahul's Template class to fill in the variables.

Recommend the first for now where feasible, though the latter may be needed if iterating over members in python (similar to how the EnKF exscripts have to do it). The Template class will need a minor update to proceed with option 2: The templates reply on the bash substitution replacing undefined variables with an empty string (mainly to replacing `$MEMDIR` for the deterministic files), but the class currently leaves any variables it doesn't have substitutions for as-is. It will need to be updated to provide the option to replace undefined variables with an empty string.

CoryMartin-NOAA:
**Description**

Fixes #1516 

This PR updates the j-jobs and python classes for aerosol DA to use the new COM directory structure.
This PR also includes removing of the chem history staging for the ICSDIR in setup_expt.py. The aerosol fields are treated as FV3 tracers, so they are either available (warm start) with the fv_tracer files or start at 0 (cold start) and must be spun up from emissions.

**Type of change**

**How Has This Been Tested?**

- [x] Cycled test on Hera - C48 ATMA; 1.5 cycles complete from cold start, 2nd full cycle in progress

  
**Checklist**

### DIFF
--- a/jobs/JGLOBAL_AERO_ANALYSIS_FINALIZE
+++ b/jobs/JGLOBAL_AERO_ANALYSIS_FINALIZE
@@ -8,7 +8,10 @@ source "${HOMEgfs}/ush/jjob_header.sh" -e "aeroanlfinal" -c "base aeroanl aeroan
 ##############################################
 # Set variables used in the script
 ##############################################
+# shellcheck disable=SC2153
 GDATE=$(date +%Y%m%d%H -d "${PDY} ${cyc} - ${assim_freq} hours")
+gPDY=${GDATE:0:8}
+gcyc=${GDATE:8:2}
 GDUMP="gdas"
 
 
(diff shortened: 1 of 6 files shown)


aerorahul (approved):
Looks great.

WalterKolczynski-NOAA (approved):

PR #1555, opened 2023-05-01
Feature/pygw schema

HenryRWinterbottom:
This draft PR provides an implementation example for a schema validator to be considered for `pygw`. It is intended to initiate a discussion for a possible path forward for such an implementation.

The example application, for an abbreviate UFS `model_configure` file configuration, can be executed as follows from within the `schema_pkg` path.

```
export PYTHONPATH=/path/to/pygw/src
python test_schema.py
```

The above application attempts to validate the attributes defined within `TEST_SCHEMA` below against those defined within the YAML-formatted file path `test_files/model_configure.schema.yaml`. The latter defines the valid configuration for the `model_configure` example schema.

```
TEST_SCHEMA = {"DT_ATMOS": 1200,
               "FHMAX": 120,
               "IMO": 384,
               "JMO": 192,
               "SYEAR": 2000,
               "SMONTH": 1,
               "SDAY": 1,
               "SHOUR": 0,
               "WRITE_GROUP": 1,
               "WRTTASK_PER_GROUP": 4,

               # TODO: This needs to be transformed via `f90_bool`.
               "WRITE_DOPOST": True,
               }
```

Next, this application seeks any optional variables for the established schema. If they not are defined within `TEST_SCHEMA` the validator class `check_schema/CheckSchema` updates the dictionary using the default values (defined in `test_files/model_configure.schema.yaml) prior validation. The resulting dictionary (i.e., configuration) is used to build an abbreviated `model_configure.out` using the `test_files/model_configure.template` file. The template file and the resulting rendered `model_configure.out` file respecctively appear as follows.

```
start_year:              @[SYEAR]
start_month:             @[SMONTH]
start_day:               @[SDAY]
start_hour:              @[SHOUR]
start_minute:            0
start_second:            0
nhours_fcst:             @[FHMAX]

dt_atmos:                @[DT_ATMOS]
output_1st_tstep_rst:    .false.

write_groups:            @[WRITE_GROUP]
write_tasks_per_group:   @[WRTTASK_PER_GROUP]
output_history:          @[OUTPUT_HISTORY]
write_dopost:            @[WRITE_DOPOST]
write_nsflip:            @[WRITE_NSFLIP]
num_files:               @[NUM_FILES]
filename_base:           @[FILENAME_BASE]
imo:                     @[IMO]
jmo:                     @[JMO]
```

```
start_year:              2000
start_month:             1
start_day:               1
start_hour:              0
start_minute:            0
start_second:            0
nhours_fcst:             120

dt_atmos:                1200
output_1st_tstep_rst:    .false.

write_groups:            1
write_tasks_per_group:   1
output_history:          True
write_dopost:            True
write_nsflip:            True
num_files:               2
filename_base:           'atm' 'sfc'
imo:                     384
jmo:                     192
```

Note that the optional variables `OUTPUT_HISTORY`, `WRITE_DOPOST`, `WRITE_NSFLIP`, `NUM_FILES`, and `FILENAME_BASE`, that are not defined within the `TEST_SCHEMA` dictionary. These values are populated using the default values for the respective variable (defined within `test_files/model_configure.schema.yaml`).

Additional `Schema` validators such as `Or` maybe considered and as demonstrated [here](https://github.com/HenryWinterbottom-NOAA/ufs_pyutils/blob/develop/utils/schema_interface.py#L270). However, no use case has yet been demonstrated within the context of the global-workflow applications for such a feature.

### DIFF
--- a/ush/python/pygw/src/schema_pkg/build_schema.py
+++ b/ush/python/pygw/src/schema_pkg/build_schema.py
@@ -1,0 +1,142 @@
+import logging
+from pydoc import locate
+from typing import Dict
+
+import pygw.yaml_file
+from pygw.attrdict import AttrDict
+from pygw.logger import logit
+from pygw.schema import Optional, Schema
+
+# ----
+
+logger = logging.getLogger(__name__.split(".")[-1])
+
+# ----
+
+__all__ = ["BuildSchema"]
+
+# ----
+
+
+class BuildSchema:
+    """
+    Description
+    -----------
+
+    This is the base-class object for setting up and defining the
+    schema for a give application; it is based on
+    https://tinyurl.com/pyutils-schema-build.
+
+    Parameters
+    ----------
+
+    yaml_file: str
+
+        A Python string containg the YAML-formatted file path
+        containing the schema; the YAML-formatted file containing the
+        schema attributes should be formatted similar to the example
+        below.
+
+        required:
+            variable1: bool
+            variable2: float
+            variable3: int
+
+        optional:
+            variable4:
+                type: bool
+                default: False
+            variable5:
+                type: bool
+                default: True
+            variable6:
+                type: float
+                default: 1.0
+
+    """
+
+    @logit(logger, name="BuildSchema")
+    def __init__(self: object, yaml_file: str):
+        """
+        Description
+        -----------
+
+        Creates a new BuildSchema object.
+
+        """
+
+        # Define the base-class attributes.
+        self.schema = {}
+        self.data = pygw.yaml_file.parse_yaml(path=yaml_file)
+
+    @logit(logger)
+    def reqschema(self: object) -> None:
+        """
+        Description
+        -----------
+
+        Collect any required schema attributes from the schema YAML
+        file.
+
+        """
+
+        try:
+            reqdict = self.data["required"]
+            if len(reqdict) > 0:
+                for (key, value) in reqdict.items():
+                    self.schema[key] = locate(value)
+        except KeyError:
+            # TODO: A logger warning informing the user that no
+            # required attributes were determined from the YAML
+            # file; this would be useful for debugging.
+            pass
+
+    @logit(logger)
+    def optschema(self: object) -> None:
+        """
+        Description
+        -----------
+
+        Collect any optional schema attributes and the default values
+        from the schema YAML file.
+
+        """
+
+        try:
+            optdict = self.data["optional"]
+            if len(optdict) > 0:
+                for (key, value) in optdict.items():
+                    vardef = value["default"]
+                    vartyp = value["type"]
+                    self.schema[Optional(key, default=vardef)] = locate(vartyp)
+        except KeyError:
+            # TODO: A logger warning informing the user that no
+            # optional attributes were determined from the YAML file;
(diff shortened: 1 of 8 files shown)


ush/python/pygw/src/schema_pkg/build_schema.py
```diff
@@ -0,0 +1,142 @@
...
+        optional:
+            variable4:
+                type: bool
+                default: False
+            variable5:
+                type: bool
+                default: True
+            variable6:
+                type: float
+                default: 1.0
```
aerorahul, line 54: Instead of keys `required` and `optional`, I would recommend that a `required` schema attribute be defined.
If the attribute is not present, it is understood that the variable is `optional`.  If `required` is present, its value can be `True|False` making it `Required|Optional`.  If `required` is `False`, then a `default` value must be provided.

E.g.
```
variable1:
    type: bool
     required: True
variable2:
    type: int
    default: 4
    required: False
variable3:
    type: float
    default: 0.25
```

ush/python/pygw/src/schema_pkg/build_schema.py
```diff
@@ -0,0 +1,142 @@
...
+        Creates a new BuildSchema object.
+
+        """
+
+        # Define the base-class attributes.
+        self.schema = {}
+        self.data = pygw.yaml_file.parse_yaml(path=yaml_file)
+
+    @logit(logger)
+    def reqschema(self: object) -> None:
```
aerorahul, line 73: In the light of the above comment, I think these methods need an update.

ush/python/pygw/src/schema_pkg/check_schema.py
```diff
@@ -0,0 +1,98 @@
+import logging
```
aerorahul, line 1: I am not quite clear on the need of `build_schema.py` and `check_schema.py`. 
They are convenient, but I am not sold on their need yet.
I'll think about it some more.

ush/python/pygw/src/schema_pkg/test_files/model_configure.schema.yaml
```diff
@@ -0,0 +1,70 @@
+required:
```
aerorahul, line 1: These are good tests.
My only suggestion would be to turn these into pytest fixtures with a more variety of type checking.
There is no need to have a test with 10+ `int` checks and not a single `float` or `complex` or `array` etc.

aerorahul (commented):
All in all this is very much a desirable feature.

ush/python/pygw/src/schema_pkg/build_schema.py
HenryRWinterbottom: Good idea. That makes it much cleaner. I did note that we could have schema options i.e., `Or` or 'And`. I haven't seen a use case for it yet but maybe something to keep in mind.

ush/python/pygw/src/schema_pkg/check_schema.py
HenryRWinterbottom: Agreed. I don't have a good idea how to handle this yet (i.e., what is the simpliest/cleanest way).

HenryRWinterbottom:
Closing as this is redundant.

PR #1509, opened 2023-04-25
Update UFS to develop as of 2023 Apr 17

WalterKolczynski-NOAA:
**Description**
Updates the UFS model hash to the version as of 2023 Apr 17.

Some associated changes accompany this update
- Restart filenames for MOM6 (ufs-community/ufs-weather-model#1599)
- Remove store coriolis setting from MOM6 namelist (ufs-community/ufs-weather-model#1599)
- Change in atm 'log' file names (ufs-community/ufs-weather-model#1704)
- Additions to diag_table for frozen species (ufs-community/ufs-weather-model#1529)
- Restart quilting (ufs-community/ufs-weather-model#1633)
- Update to post itag (ufs-community/ufs-weather-model#1690)

The switch to restart quilting adds an additional constraint on the size of write groups to be divisible by the number of tiles, so all were increased to the next multiple of 6.

In the process of updating the diag tables, unused tables were removed.

Closes #1279 as moot
Closes #1445
Closes #1499
Partially addresses #1277

**Type of change**
- [x] New feature (non-breaking change which adds functionality)

**How Has This Been Tested?**
- [x] Cycled test on Orion
- [x] P8 test on Orion
  
**Checklist**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing tests pass with my changes
- [x] Any dependent changes have been merged and published

PR #1556, opened 2023-05-01
Remove remnant WAVE_RUN from archive

### LINKED ISSUE #1548: WAVE_RUN is not defined in archive job
**Expected behavior**
Archive job should run with success. 

**Current behavior**
Failure at line  line [182 of exglobal_archive.sh](https://github.com/NOAA-EMC/global-workflow/blob/develop/scripts/exglobal_archive.sh#L182). 

**Machines affected**
Any with HPSS access. 
**To Reproduce**
Found testing an HR1 like configuration in PR #1509 

**Possible Implementation**
Line  [182 of exglobal_archive.sh](https://github.com/NOAA-EMC/global-workflow/blob/develop/scripts/exglobal_archive.sh#L182).  should be updated from: 
if [ "${DO_WAVE}" = "YES" ] && [ "${WAVE_RUN}" != "gdas" ]; then
to 
if [ "${DO_WAVE}" = "YES" ] ; then

as we've already checked $RUN is gfs before, so we shouldn't need that check here and WAVE_RUN is not defined and causes an error in the archive.

WalterKolczynski-NOAA:
**Description**
`$WAVE_RUN` is no longer used in workflow, but one reference remained in the archive job, which would cause failures. The conditional did not need to be replaced by `$RUN` since `$RUN` has already been checked at that point.

Fixes #1548

**Type of change**

**How Has This Been Tested?**
Tested during testing of #1509
  
**Checklist**

### DIFF
--- a/scripts/exglobal_archive.sh
+++ b/scripts/exglobal_archive.sh
@@ -179,7 +179,7 @@ if [[ ${HPSSARCH} = "YES" || ${LOCALARCH} = "YES" ]]; then
             fi
         fi
 
-        if [ "${DO_WAVE}" = "YES" ] && [ "${WAVE_RUN}" != "gdas" ]; then
+        if [ "${DO_WAVE}" = "YES" ]; then
             targrp_list="${targrp_list} gfswave"
         fi
 


aerorahul (commented):
I don't really know the impact of this change.  Someone else with wave superpowers is better situated to comment/approve.

JessicaMeixner-NOAA (approved):
Thanks for fixing this!

aerorahul (approved):

PR #1559, opened 2023-05-02
Avoid parsing group name when checking RadMon diagnostic files

### LINKED ISSUE #1515: RadMon diagnostic check fails on S4
**Expected behavior**

The RadMon diagnostic checking script should, among other things, identify if any diagnostic files have a file size of 0.  This is done by parsing the output of `tar -tvf ${radstat_file} | grep '_ges'` with `gawk` ([see lines 116-149 of radmon_diag_ck.sh](https://github.com/NOAA-EMC/global-workflow/blob/efa5180462f71ec476aeb6c5de4ba074a9d38a29/ush/radmon_diag_ck.sh#L116-L149)).

**Current behavior**

On S4, the primary group for users is `domain users`.  Since there is a space in the group name, this causes `gawk '{print $3}'` to pull out `users` rather than the file size, which causes the size checking algorithm to fail.  This does not cause a crash, but does fail to correctly identify any 0-sized files.

**Machines affected**

S4

**Possible Implementation**

Before parsing the output of `tar -tvf` with `gawk`, cleanse the output of `tar`.  The output is in the form
`<permissions> <user>/<group> <file size> <creation date> <creation time> <file name>`.  A `sed` command could be used to clear out permissions and user/group.  The permissions can be an assortment of letters (I think just `rwxtsd`, but it's probably safer to assume any `[a-zA-Z]`) or `-`.  Usernames and groups can have alphanumeric plus several special characters, including ` `s.  The complete list of characters found in Hera, Orion, and S4 usernames and groups shows the following `[ -_.a-zA-Z0-9]`.  This leads me to believe the following `sed` commands could be used
```
vc=`echo ${vc} | sed 's/  */ /g' | sed 's/^[-a-zA-Z]\+ [-_.a-zA-Z0-9]\+\/[-_.a-zA-Z0-9]\+ *[-_.a-zA-Z]*//`
```
This will trim the permissions, username, and group from the output, which should make the output of `echo $vc | gawk '{print $1}` be the file size.  There are two assumptions with this: 1) a group has no more than one space in its name and 2) the second word of the group name contains no numbers.  These are both true for S4, Hera, and Orion, but it is entirely possible to create a group that violates these assumptions.

A foolproof, albeit extraneous, solution would be to extract the contents of `radstat` file and run `du` or `ls --size` on each output file to get the sizes.

DavidHuber-NOAA:
**Description**

This changes the way the RadMon diagnostic file checker determines if a diagnostic file is 0-sized.  Instead of attempting to parse the output of `tar -tv`, the script now extracts the `radstat` file and runs `du -b` on each gzipped diagnostic file.  On S4, the primary group is `domain users`, which, containing a space, caused issues with the previous method.

Fixes #1515

**Type of change**

**How Has This Been Tested?**

- [x] Cycled test on S4
- [x] Cycled test on Orion
- [x] Cycled test on Hera
  
**Checklist**


EdwardSafford-NOAA (approved):
Looks good to me.

ush/radmon_diag_ck.sh
```diff
@@ -105,51 +105,38 @@ echo "--> radmon_diag_ck.sh"
...
-   #  
+   #
    #  So for this check we'll assume anything in the radstat
    #  file with a size of > 1000 bytes is suspect.  (That's 
    #  overkill, 100 is probably sufficient, but I'm the 
    #  nervous type.) So we'll extract, uncompress, and check 
    #  the actual file size of those.  Anything with an 
    #  uncompressed size of 0 goes on the zero_len_diag list.
    #
-   verbose_contents=`tar -tvf ${radstat_file} | grep '_ges'`
```
WalterKolczynski-NOAA, line 116:
```suggestion
# File sizes contain only digits and immediately precede the date
sizes=($(tar -vtf gdas.t00z.radstat --wildcards '*_ges*' | grep -P -o '(\d)+(?= \d{4}-\d{2}-\d{2})'))
# Filenames are the last group of non-whitespace characters
filenames=($(tar -vtf gdas.t00z.radstat --wildcards '*_ges*' | grep -P -o '\S+$'))
```

ush/radmon_diag_ck.sh
```diff
@@ -105,51 +105,38 @@ echo "--> radmon_diag_ck.sh"
...
+         if [[ ${uz_file_size} -eq 0 ]]; then
+            sat=$(echo "${ges}" | gawk -F"diag_" -F"_ges" '{print $1}')
 
             zero_len_diag="${zero_len_diag} ${sat}"
          fi
 
-         rm -f ${unzipped_file}
+         rm -f "${ges}"
       fi
+      rm -f "${gz_ges}"
```
WalterKolczynski-NOAA:
```suggestion
   for file_num in ${!file_names[@]}; do
      file_name="${filenames[${file_num}]}"
      file_size="${sizes[${file_num}]}"
      
      if (( file_size <= 1000 )); then
        tar -xf ${radstat_file} ${file_name}
        gunzip ${file_name}
        uz_file_name="${file_name%.*}"
        uz_file_size=$(stat -c "%s" "${uz_file_name}")
        
        if (( uz_file_size <= 0 )); then
           # Remove leading diag_
           sat=${uz_file_name#diag_}
           # Remove trailing _ges*
           sat=${sat%_ges*}
           
           zero_len_diag="${zero_len_diag} ${sat}"
        fi
        
        rm -f ${uz_file_name}
```

WalterKolczynski-NOAA (changes requested):
Try this alternative way that does not require adding a full untar just for S4 and also eliminates the need for gawk.

ush/radmon_diag_ck.sh
DavidHuber-NOAA: @WalterKolczynski-NOAA these changes work perfectly with just some minor modifications.  However, these two lines result in `shellcheck SC2207` warnings:

`sizes=($(tar -vtf gdas.t00z.radstat --wildcards '*_ges*' | grep -P -o '(\d)+(?= \d{4}-\d{2}-\d{2})'))`
`filenames=($(tar -vtf gdas.t00z.radstat --wildcards '*_ges*' | grep -P -o '\S+$'))`

The suggested fix is in the form

```
array=()
mycommand | while IFS="" read -r line; do array+=("$line"); done
```

but I cannot seem to make this work.  In every case, I am getting empty `sizes` and `filenames` arrays.  Do you have a suggestion on how to make this work or should I add `# shellcheck disable=SC2207 ... # shellcheck disable=` before/after these lines or do you have a suggestion on how to get shellcheck's suggestion to work?

WalterKolczynski-NOAA: Try the other suggested form:
```
IFS=" " read -r -a array <<< "$(mycommand)"
```

If that doesn't work, add a shellcheck disable directive along with a TODO to look into it later (as opposed to others we have that are permanent).

DavidHuber-NOAA: @WalterKolczynski-NOAA Unfortunately, that didn't work either.  I will go ahead with the original solution and add the TODO and shellcheck disable directive.  For reference, these are the options I tried:

```
   sizes=()
   tar -vtf ${radstat_file} --wildcards '*_ges*' | \
      grep -P -o '(\d)+(?= \d{4}-\d{2}-\d{2})' | \
      while IFS="" read -r size; do sizes+=("$size"); done
```
and
```
   sizes=()
   IFS=" " read -r -a sizes <<< "$(tar -vtf ${radstat_file} --wildcards '*_ges*' | \
      grep -P -o '(\d)+(?= \d{4}-\d{2}-\d{2})')"
```

The first returned an empty array and the second contained only one element even though the tarball contains ~35 files.

WalterKolczynski-NOAA (approved):

WalterKolczynski-NOAA:
@DavidHuber-NOAA Thanks for making the change.

PR #1601, opened 2023-05-11
Add GEFS support to rocoto XML generation

### LINKED ISSUE #1577: Add initial GEFS support to XML workflow creation
**Description**
Add the initial capability to produce rocoto XML for GEFS.

**Requirements**
Must create a valid rocoto XML that runs a single metatask for the GEFS forecast job.

Ability to create a functional rocoto XML that can run the ensemble forecast job.

WalterKolczynski-NOAA:
**Description**

Adds GEFS support to rocoto XML generation.

To facilitate this, a few classes were refactored into abstract classes with child classes for each of the three `$NET`/`$MODE` combinations (gfs/cycled, gfs/forecast-only, and gefs). These child classes then implement the details particular to their configuration. A generic factory class from @aerorahul was added to pygw and is used to generate objects for the appropriate class based on `$NET` and `$MODE`.

GEFS is implemented as only forecast jobs right now: a fcst job intended for the control and a metatask of efcs jobs intended for the perturbed members. A couple of additional config files were needed in order to parse the configs; these were copied from the gfs versions. Beyond that, changes have not yet been made to update any configs for GEFS. For instance, in contrast to GFS, even the control member of GEFS will want `MEMDIR` defined (to 'mem000').

Also resolves all shellcheck issues in `config.fcst` and `config.nsst`.

Fixes #1577

**Type of change**

**How Has This Been Tested?**

- [x] GFS cycled experiment produces identical workflow XMLs
- [x] GFS forecast-only experiment produces identical workflow XMLs
- [x] GEFS experiment produces a workflow XML that looks correct on visual inspection
- [x] GFS cycled on Orion
- [x] GFS P8 on Orion

**Checklist**


WalterKolczynski-NOAA:
With all the config modifications now to fix all the shellcheck warnings, I'm actually running the forecast.

parm/config/gefs/config.fcst
```diff
@@ -0,0 +1,366 @@
...
+export d2_bg_k1=0.20
+export d2_bg_k2=0.04
+export dz_min=6
+export n_sponge=42
+if (( LEVS == 128 )) && [[ "${CDUMP}" =~ "gdas" ]]; then
+   export tau=5.0
+   export rf_cutoff=1.0e3
+   export d2_bg_k1=0.20
+   export d2_bg_k2=0.0
+fi
```
XianwuXue-NOAA, line 149: Does GEFS need this?

parm/config/gefs/config.fcst
```diff
@@ -0,0 +1,366 @@
...
+        ;;
+    *) echo "Unknown microphysics option, ABORT!" ;;
+esac
+
+# Stochastic physics
+export DO_SPPT=${DO_SPPT:-"NO"}
+export DO_SKEB=${DO_SKEB:-"NO"}
+export DO_SHUM=${DO_SHUM:-"NO"}
+export DO_LAND_PERT=${DO_LAND_PERT:-"NO"}
+export DO_CA=${DO_CA:-"YES"}
```
XianwuXue-NOAA, line 249: For mem000, DO_CA should be NO or YES?

parm/config/gefs/config.fcst
XianwuXue-NOAA: I mean lines 144-149

parm/config/gefs/config.fcst
```diff
@@ -0,0 +1,366 @@
...
+export ishuffle=0
+# compression for RESTART files written by FMS
+export shuffle=1
+export deflate_level=1
+
+#---------------------------------------------------------------------
+# Disable the use of coupler.res; get model start time from model_configure
+export USE_COUPLER_RES="NO"
+
+if [[ "${CDUMP}" =~ "gdas" ]] ; then # GDAS cycle specific parameters
```
XianwuXue-NOAA, line 280: Can CDUMP=gefs? For the conditions Lines 280-349, it should keep only for GEFS part. Or I will do this.

parm/config/gfs/config.fcst
```diff
@@ -276,10 +277,10 @@ export deflate_level=1
 # Disable the use of coupler.res; get model start time from model_configure
 export USE_COUPLER_RES="NO"
 
-if [[ "$CDUMP" =~ "gdas" ]] ; then # GDAS cycle specific parameters
+if [[ "${CDUMP}" =~ "gdas" ]] ; then # GDAS cycle specific parameters
```
XianwuXue-NOAA, line 280: Same things for these conditions. It should only keep gefs parts

XianwuXue-NOAA (commented):
I am still working on the tests, just post some comments

parm/config/gefs/config.fcst
WalterKolczynski-NOAA: These are just copied from GFS for now. They will be pruned later.

parm/config/gefs/config.fcst
WalterKolczynski-NOAA: See above, these are just copied from GFS for now. Future PR can update them (so we can see the changes).

When we get there, set cellular automata to whatever v12 is using unless/until we get guidance from MDAB.

parm/config/gfs/config.fcst
WalterKolczynski-NOAA: This is the GFS one.

parm/config/gefs/config.fcst
WalterKolczynski-NOAA: Same as above. We'll strip it down later.

WalterKolczynski-NOAA:
> I am still working on the tests, just post some comments

Note this probably does not work to actually run the GEFS forecast. But you should get a workflow with just a fcst task (for the control) and an efcs metatask (for the perturbed members)

XianwuXue-NOAA:
> > I am still working on the tests, just post some comments
> 
> Note this probably does not work to actually run the GEFS forecast. But you should get a workflow with just a fcst task (for the control) and an efcs metatask (for the perturbed members)

Why uses fcst task instead of efcs for all members since all members should use the same settings, including mem000.

WalterKolczynski-NOAA:
> > > I am still working on the tests, just post some comments
> > 
> > 
> > Note this probably does not work to actually run the GEFS forecast. But you should get a workflow with just a fcst task (for the control) and an efcs metatask (for the perturbed members)
> 
> Why uses fcst task instead of efcs for all members since all members should use the same settings, including mem000.

Because it doesn't. The control doesn't use stochastic physics, for one. Plus, it will be easier from a code management standpoint because it means the same thing as in the GFS.

workflow/rocoto/workflow_tasks.py
```diff
@@ -189,6 +184,41 @@ def get_task(self, task_name, *args, **kwargs):
...
+
+        cdump: str
+               CDUMP to generate tasks for (currently a RUN stand-in, will be changed
+               soon).
+
+        Returns
+        -------
+        Tasks: New object of the appropriate type for the net.
+        '''
+        if app_config.net in ['gfs']:
```
aerorahul: I think this should work:
```python
me = sys.modules[__name__]
class_name = f"{app_config.net.upper()Tasks}"
try:
    return getattr(me, class_name)(app_config, cdump)
except AttributeError:
    raise AttributeError(f"{class_name} is not a valid class in {me}"
```

It would help you everywhere you have `if-else` conditions.

aerorahul (commented):
looks good.
A suggestion if you want to consider making the factory truly agnostic.

workflow/rocoto/workflow_tasks.py
WalterKolczynski-NOAA: I think the better approach would be to just change it to the uwtools Factory. Not sure if we should do that now, or get this in so things can move forward and then refactor the factories later.

workflow/rocoto/workflow_tasks.py
```diff
@@ -27,6 +28,8 @@ class Tasks:
                    'waveawipsbulls', 'waveawipsgridded', 'wavegempak', 'waveinit',
                    'wavepostbndpnt', 'wavepostbndpntbll', 'wavepostpnt', 'wavepostsbs', 'waveprep']
 
+    tasks_factory = Factory(__qualname__)
```
aerorahul: I think you want this out of the `class Tasks` rather than a class attribute.

workflow/rocoto/workflow_tasks.py
```diff
@@ -1417,6 +1426,40 @@ def earc(self):
...
+
+        resources = self.get_resource('efcs')
+        task = create_wf_task('efcs', resources, cdump=self.cdump, envar=efcsenvars, dependency=dependencies,
+                              metatask='efmn', varname='grp', varval=groups, cycledef='gefs')
+
+        return task
+
+
+Tasks.tasks_factory.register('gfs', GFSTasks)
+Tasks.tasks_factory.register('gefs', GEFSTasks)
```
aerorahul: This is not the same as:
`tasks_factory.register('gfs', GFSTasks)` as it should be.
When you call `Tasks.tasks_factory...`, you are spinning up another instance of `Tasks`. The memory address of this instance will be different than the one for `gefs`.
I don't know how this is working.

workflow/rocoto/workflow_tasks.py
```diff
@@ -1456,7 +1499,7 @@ def get_wf_tasks(app_config: AppConfig) -> List:
     tasks = []
     # Loop over all keys of cycles (CDUMP)
     for cdump, cdump_tasks in app_config.task_names.items():
-        task_obj = Tasks(app_config, cdump)  # create Task object based on cdump
+        task_obj = Tasks.tasks_factory.create(app_config.net, app_config, cdump)  # create Task object based on cdump
```
aerorahul: This should also be `tasks_factory.create` and not `Tasks.tasks_factory`.

aerorahul (changes requested):
Sorry I started reviewing in the morning but got distracted.

workflow/rocoto/workflow_tasks.py
WalterKolczynski-NOAA: `tasks_factory` is a static field of Tasks. It doesn't create a new Tasks object at all. A single factory is created for Tasks at class definition time, which is what is accessed here.

aerorahul: I see that, and see my previous comment.  `tasks_factory` cannot be a member of the class `Tasks` as `tasks_factory` is and should be an instance of `class TasksFactory`, which is inherited from `Factory`.  Such that:
```python
class TasksFactory(Factory):
```
and 
```python
tasks_factory = TasksFactory()
tasks_factory.register('gfs', GFSTasks)
```

ush/python/pygw/src/pygw/factory.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+        Parameters
+        ----------
+        name : Name of factory
+        """
+        self._name = f'{name}Factory'
+        self._builders = {}
+
+        # Register {name}Factory as importable from uwtools.factory
```
aerorahul:
```suggestion
        # Register {name}Factory as importable from pygw.factory
```

workflow/rocoto/workflow_tasks.py
aerorahul: Are we going to use `Factory` the way it was intended to be used or keep using it this way?

workflow/rocoto/workflow_tasks.py
aerorahul: I would like to see it used in the way it was supposed/designed, rather than using it any other way because it "works"

aerorahul (changes requested):
still waiting on requested updates + one correction.

WalterKolczynski-NOAA:
I'm working on the factory stuff. Need to make a new one for all three uses. Also splitting out the new child classes into separate files as we discussed.

WalterKolczynski-NOAA:
@TerrenceMcGuinness-NOAA I have no indication what failed about experiment setup. I can't even find a log in the CI directory where that information would be held.

TerrenceMcGuinness-NOAA:
Here is the EXPDIR:` /scratch1/NCEPDEV/global/Terry.McGuinness/GFS_CI_ROOT/PR/1601/RUNTESTS/C96_atm3DVar/EXPDIR/C96_atm3DVar` Yes I see there is no error reporting or logging on the `setup_exp.py` fail.  Let's put in a PR request on this for sure.

TerrenceMcGuinness-NOAA:
Let me try to reproduce the error.

TerrenceMcGuinness-NOAA:
Here's the error.  It was with setup_xml.py: 

```shell
(gfs_workflow) Terry.McGuinness (hfe01) workflow (feature/gefs_workflow) $ ./setup_xml.py /scratch1/NCEPDEV/global/Terry.McGuinness/GFS_CI_ROOT/PR/1601/RUNTESTS/C96_atm3DVar/EXPDIR/C96_atm3DVar
  File "/scratch1/NCEPDEV/global/Terry.McGuinness/GFS_CI_ROOT/PR/1601/global-workflow/workflow/./setup_xml.py", line 9, in <module>
    from applications.application_factory import app_config_factory
ModuleNotFoundError: No module named 'applications.application_factory'; 'applications' is not a package
(gfs_workflow) Terry.McGuinness (hfe01) workflow (feature/gefs_workflow) $ 
```

WalterKolczynski-NOAA:
Thanks Terry. The new directory did not appear in `git status`, so I overlooked adding it to the commit. Fixed now.

workflow/rocoto/gefs_xml.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+# Copy of GFSForecastOnlyRocotoXML for now, other than changing cycledef names from 'gfs' to 'gefs'
+#   If it remains this way, we can consolidate into a single forecast-only class
+class GEFSRocotoXML(RocotoXML):
+
+    def __init__(self, app_config: AppConfig, rocoto_config: Dict) -> None:
+        super().__init__(app_config, rocoto_config)
+
+    def _get_cycledefs(self):
```
aerorahul: This shouldn't be `_get_cycledefs` any more as it is a concrete implementation *and* no longer a private

workflow/rocoto/rocoto_xml_factory.py
```diff
@@ -0,0 +1,16 @@
+from pygw.factory import Factory
+from rocoto.gfs_cycled_xml import GFSCycledRocotoXML
+from rocoto.gfs_forecast_only_xml import GFSForecastOnlyRocotoXML
+from rocoto.gefs_xml import GEFSRocotoXML
+
+
+class RocotoXMLFactory(Factory):
+
+    def __init__(self, name: str):
+        super().__init__(name)
```
aerorahul: Not sure why you need this.

workflow/rocoto/rocoto_xml_factory.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+rocoto_xml_factory = RocotoXMLFactory('RocotoXML')
```
aerorahul:
```suggestion
rocoto_xml_factory = Factory('RocotoXML')
```

workflow/rocoto/rocoto_xml_factory.py
WalterKolczynski-NOAA: Factory sub-classes have been removed.

workflow/rocoto/rocoto_xml_factory.py
WalterKolczynski-NOAA: Done.

aerorahul (approved):
looks good.
thanks for adding GEFS XML support.

PR #1609, opened 2023-05-12
Prepare snow depth observations for JEDI-based Land DA.

aerorahul:
**Description**
This PR:
- adds a job to prepare IMS snow depth observations as a task in the workflow.  This task depends on the `prep.sh` job to bring IMS data.  To test this type of data `DMPDIR` in `config.base` needed to be pointed to `"/scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GlobalWorkflow/para_gfs/glopara_dump"`
- `land_analysis.py` introduces a method `prepare_IMS` for this type of data.
- Updates are necessary in the `GDASApp` repo.  See companion PR https://github.com/NOAA-EMC/GDASApp/pull/472

This job only runs at the `18z` cycle in the workflow.  This is controlled in the XML and not in the script.
Consequently, updates are in the `workflow/` scripts to make this happen.

Other updates included:
- Adds a method (and associated test) to provide a datetime in the form of Julian day of the year.  The method is `to_julian` in `timetools.py`
- `executable.py` was updated to accept a `list` of arguments (in addition to `str`).

**Note:**
- `calcfIMS.exe` is a serial executable and yet requires to be run with `srun`.  I recommend compiling this executable without MPI. @jiaruidong2017 @CoryMartin-NOAA 
- There are several deprecation warnings from the python IODA converter that will need to be addressed.  I recommend creating an issue to be resolved later. @jiaruidong2017 @CoryMartin-NOAA 

Part of Land DA work for snow depth assimilation, for which there is no issue open.

**Type of change**

**How Has This Been Tested?**
A standalone test was created on Hera at:
`/scratch1/NCEPDEV/stmp2/Rahul.Mahajan/EXPDIR/landda`
See `preplandobs.log` for details.
  
**Checklist**


sorc/checkout.sh
```diff
@@ -165,7 +165,7 @@ if [[ ${checkout_gsi} == "YES" ]]; then
 fi
 
 if [[ ${checkout_gdas} == "YES" ]]; then
-  checkout "gdas.cd" "https://github.com/NOAA-EMC/GDASApp.git" "aaf7caa"; errs=$((errs + $?))
+  checkout "gdas.cd" "https://github.com/NOAA-EMC/GDASApp.git" "feature/prepland"; errs=$((errs + $?))
```
CoryMartin-NOAA: Note: to update this to whatever the stable nightly hash will be tomorrow AM.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -0,0 +1,182 @@
...
+        self.task_config = AttrDict(**self.config, **self.runtime_config, **local_dict)
+
+    @logit(logger)
+    def prepare_IMS(self: Analysis) -> None:
+        """Prepare the IMS data for a global land analysis
+
+        This method will prepare IMS data for a global land analysis using JEDI.
+        This includes:
+        - staging model backgrounds
+        - generating a card file for IMS processing
```
CoryMartin-NOAA: what is a card file?

workflow/rocoto/workflow_xml.py
```diff
@@ -118,6 +118,12 @@ def _get_cycledefs_cycled(self):
         sdate = sdate + to_timedelta(interval)
         strings.append(f'\t<cycledef group="gdas">{sdate.strftime("%Y%m%d%H%M")} {edate.strftime("%Y%m%d%H%M")} {interval}</cycledef>')
 
+        if self._app_config.do_jedilandda:
+            sdate_land = sdate.replace(hour=18, minute=0, second=0)
+            edate_land = edate
+            if edate >= sdate:
+                strings.append(f'\t<cycledef group="gdas_land_prep">{sdate_land.strftime("%Y%m%d%H%M")} {edate_land.strftime("%Y%m%d%H%M")} 24</cycledef>')
```
CoryMartin-NOAA: Workflow Q for you. So landanlinit, run, final will run every cycle, the prep only 18z (for the IMS). Where does the different dependency depending on cycle come into play in the XML generation?

ush/python/pygfs/task/land_analysis.py
aerorahul: No idea. I assume it is the output from `calcfIMS.exe`.  @jiaruidong2017 would know for sure.

workflow/rocoto/workflow_xml.py
aerorahul: I haven't reviewed the `landanlinit` and `landanlrun` scripts yet on how they use the observation data created from `preplandobs`.  Once I see what that is, I may be able to add a dependency of `preplandobs` on the 18z cycle of `landanlinit` only.
I don't have an answer until I can review the actual use of this observation.

ush/python/pygfs/task/land_analysis.py
CoryMartin-NOAA: I think this probably means `generating a namelist for IMS processing` ?

sorc/checkout.sh
aerorahul: will update to the hash on Monday.

ush/python/pygfs/task/land_analysis.py
aerorahul: possible.  I'll update the comment on Monday with the hash when more reviews come in.

workflow/rocoto/workflow_xml.py
CoryMartin-NOAA: got it, okay we will hold for now

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: Yes, it is namelist for IMS processing

jiaruidong2017 (changes requested):
@aerorahul First thank you very much for your great efforts. The `prepare_IMS` in the `land_analysis.py` is so nice and so pretty to me, and I have learned a lot from these codes. I cloned your branch and compiled it. When I tried to setup the experiment and workflow, it is as below: 

image

I didn't find the `preplandobs` job there. Did I miss something?

aerorahul:
@jiaruidong2017 
you should see that job only in the 18z cycle.

jiaruidong2017:
> @jiaruidong2017 you should see that job only in the 18z cycle.

Thanks @aerorahul 
But it seems the `landanlinit` job wont start because there are not `preplandobs` job available at other cycles. Do we need to modify the `landanlinit` dependencies in the workflow_tasks.py to start the `landanlinit` job after `prep` in the other cycles?

jiaruidong2017:
@aerorahul I merged other land DA jobs (`landanlinit`, `landanlrun` and `landanlfinal`) into your `feature/prepland` branch, and I conducted the test runs as below: 

image
image
image
image
image
image
image

See the job monitoring processes above, the `preplandobs` job runs successful at the `18z` cycle. However, the workflow seemed weirds after the `18z` cycle, and they were `202103231801`, `202103231802`, and `202103231803` for `preplandobs` jobs. Do you know what these `preplandobs` processes are? Thanks.

WalterKolczynski-NOAA:
@jiaruidong2017 Can you point me to the rocoto XML?

jiaruidong2017:
@WalterKolczynski-NOAA it is /scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GlobalWorkflow/para_gfs/landid8888/landid8888.xml on hera. Thanks.

workflow/rocoto/workflow_xml.py
WalterKolczynski-NOAA:
```suggestion
                strings.append(f'\t<cycledef group="gdas_land_prep">{sdate_land.strftime("%Y%m%d%H%M")} {edate_land.strftime("%Y%m%d%H%M")} 24:00:00</cycledef>')
```

WalterKolczynski-NOAA: I think this is the problem: the interval in the cycledef doesn't have the zeros for minutes and seconds. Compare to the others in the XML. So there is a prep land obs job for almost every minute (I'm not sure how '24' translates into 'most minutes'.)

jiaruidong2017 (approved):
Thanks again for your efforts.

workflow/rocoto/workflow_tasks.py
```diff
@@ -510,8 +510,23 @@ def aeroanlfinal(self):
...
+
+        resources = self.get_resource('preplandobs')
+        task = create_wf_task('preplandobs', resources, cdump=self.cdump, envar=self.envars, dependency=dependencies,
+                              cycledef=f'{self.cdump}_land_prep')
+
+        return task
+
     def landanlinit(self):
 
+        # TODO: Q: what from gdaspost -6H or atmf009.nc does landanlinit depend on?
```
aerorahul: @jiaruidong2017 
Can you please answer this question?
Please take a look at the next 2 dependencies that were added to this job.

jiaruidong2017: @aerorahul I just followed the settings from `aeroanlinit`. I think the `landanlinit` may just require model background from `prep`. @CoryMartin-NOAA Do you have any suggestions? Thanks.

CoryMartin-NOAA: @jiaruidong2017 I believe it depends on the 6 hour RESTART files, no? which I guess would be the atmf009 for simplicity sake

aerorahul: Ok. Thanks.
Completion of `gdasprep` should be enough to capture that, instead of the data dependency on `atmf009.nc` from previous cycle.
I will update the dependencies.

I assume this will also be true for the `aeroanlinit` @CoryMartin-NOAA?

jiaruidong2017: @CoryMartin-NOAA Yes, it depends on the 6-hr RESTART files.

CoryMartin-NOAA: @aerorahul yes gdasprep should be all aeroanlinit needs

aerorahul: I have resolved and simplified the dependencies for both `landanlinit` and `aeroanlinit`.

workflow/rocoto/workflow_xml.py
aerorahul: I think we resolved this.

workflow/rocoto/workflow_tasks.py
```diff
@@ -466,21 +466,9 @@ def atmanlfinal(self):
...
-        dump_path = self._template_to_rocoto_cycstring(self._base["COM_OBSDMP_TMPL"],
-                                                       {'DMPDIR': dmpdir, 'DUMP_SUFFIX': dump_suffix})
-
         deps = []
-        dep_dict = {'type': 'metatask', 'name': 'gdaspost', 'offset': '-06:00:00'}
-        deps.append(rocoto.add_dependency(dep_dict))
-        data = f'{atm_hist_path}/gdas.t@Hz.atmf009.nc'
-        dep_dict = {'type': 'data', 'data': data, 'offset': '-06:00:00'}
-        deps.append(rocoto.add_dependency(dep_dict))
         dep_dict = {'type': 'task', 'name': f'{self.cdump}prep'}
```
WalterKolczynski-NOAA, line 470: Why the changes to the aerosol analysis dependencies?

WalterKolczynski-NOAA (changes requested):
Looks really good. Question about aerosol dependencies, and still need a hash for GDASapp.

workflow/rocoto/workflow_tasks.py
aerorahul: because they are not needed. And I am breaking my own rule of not adding unrelated code updates.
Do you want me to pull these out in its own PR?

WalterKolczynski-NOAA: I'm fine leaving the changes as long as we are sure the dependencies are not needed. I guess `prep` already has the same ones?

CoryMartin-NOAA: yeah the dependencies are 1) the previous forecast completing and 2) observations being available. `prep` should take care of those I think

sorc/checkout.sh
aerorahul: GDASApp hash has been updated.

aerorahul:
> Looks really good. Question about aerosol dependencies, and still need a hash for GDASapp.

@WalterKolczynski-NOAA GDASapp hash has been updated.

CoryMartin-NOAA (approved):
Thanks @aerorahul for working on this!

WalterKolczynski-NOAA (approved):

PR #1635, opened 2023-05-19
Perform JEDI LETKFOI analysis for snow depth

aerorahul:
**Description**

This PR will allow users to run JEDI-based LETKFOI update for snow depth analysis in the global-workflow.
These jobs are run in a single `landanl` task instead of 3 separate tasks.  The workflow is updated.
Some improvements are made to pull out running JEDI executable as well as tarring diag files in `analysis.py` 
The `initialize`, `execute` and `finalize` methods for `LandAnalysis` are filled up.
static methods to create ensemble and adding increments are created, so they can be used elsewhere if desired.

Updates to `pygw` included in this PR:
-  documentation is added to `fsutils.chdir` to make it clear that it is not a replacement for `os.chdir`.
- a filter `to_f90bool` is added to `jinja` to allow type conversion from python bool to fortran bool.
- documentation in `jinja.py` is slightly better in this PR.

This PR depends on:
https://github.com/NOAA-PSL/land-apply_jedi_incr/pull/3
https://github.com/NOAA-EMC/GDASApp/pull/490

**Type of change**

**How Has This Been Tested?**
Ran a standalone `landanl.sh` job on Hera.
  
**Checklist**


parm/config/gfs/config.resources
```diff
@@ -236,22 +236,13 @@ elif [[ "${step}" = "landanlinit" || "${step}" = "landanlrun"  || "${step}" = "l
...
-       export npe_node_landanlrun
-       export is_exclusive=True
-   fi
+   export wtime_landanl="00:15:00"
+   npe_landanl=$(echo "${layout_x} * ${layout_y} * 6" | bc)
+   export npe_landanl
+   export nth_landanl=1
+   npe_node_landanl=$(echo "${npe_node_max} / ${nth_landanl}" | bc)
+   export npe_node_landanl
+   export is_exclusive=True
```
aerorahul: @jiaruidong2017 @CoryMartin-NOAA Do we really need exclusive node?

parm/parm_gdas/land_jedi_fix.yaml
```diff
@@ -0,0 +1,7 @@
+mkdir:
+- $(DATA)/fv3jedi
+copy:
+- [$(HOMEgfs)/fix/gdas/fv3jedi/fv3files/akbk$(npz).nc4, $(DATA)/fv3jedi/akbk.nc4]
+- [$(HOMEgfs)/fix/gdas/fv3jedi/fv3files/fmsmpp.nml, $(DATA)/fv3jedi/fmsmpp.nml]
+- [$(HOMEgfs)/fix/gdas/fv3jedi/fv3files/field_table_gfdl, $(DATA)/fv3jedi/field_table]
+- [$(HOMEgfs)/fix/gdas/fv3jedi/fieldmetadata/gfs-land.yaml, $(DATA)/fv3jedi/gfs-land.yaml]
```
aerorahul, line 7: I could add the `orog` data in this yaml file in the future.
This is a TODO.

parm/config/gfs/config.landanl
```diff
@@ -1,23 +1,35 @@
...
+# Name of the JEDI executable and its yaml template
+export JEDIEXE="${HOMEgfs}/exec/fv3jedi_letkf.x"
+export JEDIYAML="${HOMEgfs}/sorc/gdas.cd/parm/land/letkfoi/letkfoi.yaml"
+
+# Ensemble member properties
+export SNOWDEPTHVAR="snwdph"  # if FRACGRID=NO, else "snodl"
+export BESTDDEV="30."  # Background Error Std. Dev. for LETKFOI
+
+# Name of the executable that applies increment to bkg and its namelist template
+export FRACGRID="yes"
```
WalterKolczynski-NOAA: Is there any circumstance where this would be different than the `$FRAC_GRID` set in `config.base`, other than one uses Fortran boolean (`.true.`/`.false.`) and one is yes/no? If not, we should just use the one in `config.base`.

parm/config/gfs/config.landanl
```diff
@@ -1,23 +1,35 @@
...
 export OBS_LIST=${HOMEgfs}/sorc/gdas.cd/parm/land/obs/lists/${obs_list_name}
-export LANDVARYAML=${HOMEgfs}/sorc/gdas.cd/parm/land/letkfoi/letkfoi.yaml
-export FV3JEDI_FIX=${HOMEgfs}/fix/gdas
+
+# Name of the JEDI executable and its yaml template
+export JEDIEXE="${HOMEgfs}/exec/fv3jedi_letkf.x"
+export JEDIYAML="${HOMEgfs}/sorc/gdas.cd/parm/land/letkfoi/letkfoi.yaml"
+
+# Ensemble member properties
+export SNOWDEPTHVAR="snwdph"  # if FRACGRID=NO, else "snodl"
```
WalterKolczynski-NOAA: If this is determined by `$FRACGRID` (per comment), why is it being set independently?

ush/python/pygfs/task/analysis.py
```diff
@@ -192,10 +196,75 @@ def link_jediexe(self: Task) -> None:
...
+            Launcher command e.g. mpirun -np <ntasks> or srun, etc.
+        jedi_exec : str
+            Name of the JEDI executable e.g. fv3jedi_var.x
+        jedi_yaml : str | os.PathLike
+            Name of the yaml file to feed the JEDI executable e.g. fv3jedi_var.yaml
+
+        Raises
+        ------
+        OSError
+           Failure due to OS issues
```
WalterKolczynski-NOAA: Indentation here does not match the rest of the docblock (needs another space).

ush/python/pygfs/task/analysis.py
```diff
@@ -192,10 +196,75 @@ def link_jediexe(self: Task) -> None:
...
         exe_dest = os.path.join(self.task_config.DATA, os.path.basename(exe_src))
         if os.path.exists(exe_dest):
             rm_p(exe_dest)
         os.symlink(exe_src, exe_dest)
 
         return
+
+    @staticmethod
+    @logit(logger)
+    def execute_jediexe(DATA: Union[str, os.PathLike], aprun_cmd: str, jedi_exec: str, jedi_yaml: str) -> None:
```
WalterKolczynski-NOAA: I know `DATA` here is in all caps to match the env var, but we don't do this with other python variables to match their shell counterpart, and it violates standard python conventions. Would like to see this changed to `data`, but won't insist on it.

ush/python/pygfs/task/analysis.py
```diff
@@ -192,10 +196,75 @@ def link_jediexe(self: Task) -> None:
...
+        except OSError:
+            logger.exception(f"FATAL ERROR: Failed to execute {exec_cmd}")
+            raise OSError(f"{exec_cmd}")
+        except Exception:
+            logger.exception(f"FATAL ERROR: Error occured during execution of {exec_cmd}")
+            raise WorkflowException(f"{exec_cmd}")
+
+    @staticmethod
+    @logit(logger)
+    def tgz_diags(statfile, DATA):
```
WalterKolczynski-NOAA: Same as above RE `DATA`. Also, there are no type hints in the method signature, which does not match the other methods.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -129,26 +146,149 @@ def prepare_IMS(self: Analysis) -> None:
...
             logger.exception(f"{self.task_config.IMS2IODACONV} failed to produce {output_file}")
-            raise FileNotFoundError(f"{os.path.join(self.task_config.DATA, output_file)}")
+            raise FileNotFoundError(f"{os.path.join(localconf.DATA, output_file)}")
         else:
             logger.info(f"Copy {output_file} to {self.task_config.COM_OBS}")
             FileHandler(prep_ims_config.ims2ioda).sync()
 
     @logit(logger)
-    def get_bkg_dict(self, config: Dict) -> Dict[str, List[str]]:
+    def initialize(self: Analysis) -> None:
```
WalterKolczynski-NOAA: I don't think type hints are necessary for `self`, but if you are going to use them, I think they should match the class, so this would be `LandAnalysis`, not `Analysis`.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -129,26 +146,149 @@ def prepare_IMS(self: Analysis) -> None:
...
+        # need output dir for diags and anl
+        logger.info("Create empty output [anl, diags] directories to receive output from executable")
+        newdirs = [
+            os.path.join(localconf.DATA, "anl"),
+            os.path.join(localconf.DATA, "diags"),
+        ]
+        FileHandler({'mkdir': newdirs}).sync()
+
+    @logit(logger)
+    def execute(self: Analysis) -> None:
```
WalterKolczynski-NOAA: Same as above.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -180,3 +320,168 @@ def get_bkg_dict(self, config: Dict) -> Dict[str, List[str]]:
...
+
+        # 2 ens members
+        offset = bestddev / np.sqrt(LandAnalysis.NMEM_LANDENS)
+
+        logger.info(f"Creating ensemble for LETKFOI by offsetting with {offset}")
+
+        workdir = os.path.join(config.DATA, 'bkg')
+
+        sign = [1, -1]
+        ens_dirs = ['mem001', 'mem002']
```
WalterKolczynski-NOAA, line 404: There should be some sort of comment here that these are currently hard-coded for a 2-mem ensemble.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -180,3 +320,168 @@ def get_bkg_dict(self, config: Dict) -> Dict[str, List[str]]:
...
+
+    @staticmethod
+    @logit(logger)
+    def create_ensemble(vname: str, bestddev: float, config: Dict) -> None:
+        """Create an ensemble for Snow Depth analysis by perturbing snow depth with a prescribed variance.
+        Additionally, remove glacier locations
+
+        Parameters
+        ----------
+        vname : str
```
WalterKolczynski-NOAA, line 385: For now this works, because we are only using the method for one variable. But if additional variables were ever added, loading the same NetCDF over and over is inefficient. On the other hand, if the the doc-block is correct and this should only be used for snow depth, perhaps the method name should reflect that: `create_snow_depth_ens()`, and then I'd be less worried about this being extending to multiple variables.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -180,3 +320,168 @@ def get_bkg_dict(self, config: Dict) -> Dict[str, List[str]]:
...
+        template = f'{to_fv3time(config.current_cycle)}.sfc_data.tile{{tilenum}}.nc'
+        anllist = []
+        for itile in range(1, config.ntiles + 1):
+            filename = template.format(tilenum=itile)
+            src = os.path.join(config.COM_ATMOS_RESTART_PREV, filename)
+            dest = os.path.join(config.DATA, "anl", filename)
+            anllist.append([src, dest])
+        FileHandler({'copy': anllist}).sync()
+
+        config.FRACGRID = ".true." if config.FRACGRID else ".false."
```
WalterKolczynski-NOAA: So here we are converting from yes/no into a fortran bool. Only reinforces my above question about using the `$FRAC_GRID` in `config.base`, which would also eliminate the need for this conversion.

WalterKolczynski-NOAA (dismissed):

ush/python/pygfs/task/land_analysis.py
aerorahul: I am glad you noted this.  I agree the instance of the object is not necessary to be type defined; it is implied.
Will remove.

ush/python/pygfs/task/analysis.py
aerorahul: Good point on the use of `DATA`.
I will replace with `workdir` and henceforth the python rendition of `DATA` will be `workdir`.  `data` is too generic a word and not easy to distinguish from say `datadir` or something.  `rundir` would also work.

ush/python/pygfs/task/analysis.py
aerorahul: Will fix.

ush/python/pygfs/task/land_analysis.py
aerorahul: `config.FRACGRID` has been cast as `True|False` from `YES|NO` when it entered `python`.
Here it is necessary to be cast as a `F90Bool` for parsing into the Fortran namelist.

ush/python/pygfs/task/land_analysis.py
aerorahul: Since the method is part of `land_analysis.py`, it is not necessary to say `create_snow_depth_ens`.
I would wait till there is more information on what other state variables will they perturb and how to create separate methods or turn this into an interface.

ush/python/pygfs/task/land_analysis.py
aerorahul: The doc-block says, a Fake 2 member ensemble.  Isn't that sufficient?

parm/config/gfs/config.landanl
aerorahul: I don't know.  @jiaruidong2017 and @CoryMartin-NOAA may have a plan for this.

ush/python/pygfs/task/land_analysis.py
WalterKolczynski-NOAA: It's not in the doc-block for this method, and either way it is nice to identify which parts need to be replaced later and which parts don't. But I'll leave it up to you.

ush/python/pygfs/task/analysis.py
WalterKolczynski-NOAA: Yeah, `workdir` or `rundir` is even better. I hate the `DATA` name at least as much as I hate using `CASE` for resolution.

parm/config/gfs/config.landanl
CoryMartin-NOAA: I'm guessing it will be the same as the model, so `config.base` but we need to double check with @barlage and/or @ClaraDraper-NOAA

jiaruidong2017: I agree with @CoryMartin-NOAA because there are many `frac_grid` found from GFS physics model. In the namelist of `apply_incr_nml`, the `frac_grid` is assigned with `$FRACGRID` (defined as yes/no). With the changes to the variable `frac_grid` from yes/no to Fortran boolean made by @aerorahul for the `apply_jedi_incr` code at PR https://github.com/NOAA-PSL/land-apply_jedi_incr/pull/3, we can replace the `$FRACGRID` with `$FRAC_GRID` (defined at `config.base`). We will double check with @barlage and @ClaraDraper-NOAA

parm/config/gfs/config.resources
jiaruidong2017: I have no idea on this. @CoryMartin-NOAA would you please answer this?

CoryMartin-NOAA: probably not but I don't know how much memory it uses

parm/config/gfs/config.landanl
jiaruidong2017: The `SNOWDEPTHVAR` (snwdph/snodl) can be defined by `$FRACGRID` (no/yes).

aerorahul: updated.

ush/python/pygfs/task/land_analysis.py
aerorahul: resolved.  But we still need to do Python to F90 boolean transform.

parm/config/gfs/config.landanl
aerorahul: defining `FRAC_GRID` is done in `config.base`.  This instance is now removed from `config.landanl`.

parm/config/gfs/config.resources
aerorahul: I will remove the `is_exclusive=True`.  If you need more memory, please determine how much rather than reserving the full node for a 6 processor job.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -180,3 +320,168 @@ def get_bkg_dict(self, config: Dict) -> Dict[str, List[str]]:
...
+        template = f'{to_fv3time(config.current_cycle)}.sfc_data.tile{{tilenum}}.nc'
+        anllist = []
+        for itile in range(1, config.ntiles + 1):
+            filename = template.format(tilenum=itile)
+            src = os.path.join(config.COM_ATMOS_RESTART_PREV, filename)
+            dest = os.path.join(config.DATA, "anl", filename)
+            anllist.append([src, dest])
+        FileHandler({'copy': anllist}).sync()
+
+        config.FRAC_GRID = ".true." if config.FRAC_GRID else ".false."
```
jiaruidong2017: If the first `config.FRAC_GRID` is the same as the second `config.FRAC_GRID`, this line is not necessary, right?

jiaruidong2017 (commented):
The `landanl` directory needs to be deleted after the job `landanl` is successfully completed.

ush/python/pygfs/task/land_analysis.py
aerorahul: It is not.
The first `config.FRAC_GRID` is `True|False`.  The one being passed to the jinja template is a boolean and needs to be `.true.|.false.`
Alternately, I can write a jinja2 filter `to_F90bool` and update the template to be `{{ FRAC_GRID | to_F90bool }}`

ush/python/pygfs/task/land_analysis.py
```diff
@@ -129,26 +146,149 @@ def prepare_IMS(self: Analysis) -> None:
...
+        - creates the JEDI LETKF yaml from the template
+        - stages backgrounds, observations and ensemble members
+
+        Parameters
+        ----------
+        self : Analysis
+            Instance of the LandAnalysis object
+        """
+
+        super().initialize()
```
jiaruidong2017, line 170: This copies the observation data in the IODA format based the predefined observation yaml file. 
However, the IMS observation data in the final IODA format (`gdas.t{HH}z.ims_snow_{PDY}{HH}.nc4`) is not available yet, and will be computed in `prepare_IMS`. The name convention of the original IMS data is `ims{YYYY}{DOY}_4km_v1.3.nc`. Therefore, the following errors were caused: 

```
  File "/scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GlobalWorkflow/global-workflow.landid8888/ush/python/pygw/src/pygw/fsutils.py", line 83, in cp
    shutil.copyfile(source, target)
  File "/scratch1/NCEPDEV/da/python/opt/core/miniconda3/4.6.14/envs/gdasapp/lib/python3.7/shutil.py", line 120, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: '/scratch1/NCEPDEV/global/Jiarui.Dong/ptmp/landid8888/gdas.20210323/18/obs/gdas.t18z.ims_snow_2021032318.nc4'

During handling of the above exception, another exception occurred:

  File "/scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GlobalWorkflow/global-workflow.landid8888/scripts/exglobal_land_analysis.py", line 23, in <module>
    anl.initialize()
  File "/scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GlobalWorkflow/global-workflow.landid8888/ush/python/pygw/src/pygw/logger.py", line 264, in wrapper
    retval = func(*args, **kwargs)
  File "/scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GlobalWorkflow/global-workflow.landid8888/ush/python/pygfs/task/land_analysis.py", line 170, in initialize
    super().initialize()
  File "/scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GlobalWorkflow/global-workflow.landid8888/ush/python/pygfs/task/analysis.py", line 37, in initialize
    FileHandler(obs_dict).sync()
  File "/scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GlobalWorkflow/global-workflow.landid8888/ush/python/pygw/src/pygw/file_utils.py", line 39, in sync
    sync_factory[action](files)
  File "/scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GlobalWorkflow/global-workflow.landid8888/ush/python/pygw/src/pygw/file_utils.py", line 59, in _copy_files
    cp(src, dest)
  File "/scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GlobalWorkflow/global-workflow.landid8888/ush/python/pygw/src/pygw/fsutils.py", line 85, in cp
    raise OSError(f"unable to copy {source} to {target}")
OSError: unable to copy /scratch1/NCEPDEV/global/Jiarui.Dong/ptmp/landid8888/gdas.20210323/18/obs/gdas.t18z.ims_snow_2021032318.nc4 to /scratch1/NCEPDEV/stmp2/Jiarui.Dong/RUNDIRS/landid8888/landanl.135218/obs/gdas.t18z.ims_snow_2021032318.nc4
```
One solution is to rename the original IMS data name from `ims{YYYY}{DOY}_4km_v1.3.nc` to `gdas.t{HH}z.ims_snow_{PDY}{HH}.nc4`. Any other solutions? Thanks.

aerorahul: I can't reproduce this error.  Can we discuss offline what the issue is?  
I thought the observations are provided in a yaml.

jiaruidong2017: The observation is provided at the yaml file sorc/gdas.cd/parm/land/obs/config/ims_snow.yaml:

```
  obsdatain:
    engine:
      type: H5File
      obsfile: $(DATA)/obs/$(OPREFIX)ims_snow_{{ current_cycle | to_YMDH }}.nc4
```
In the t00z, t06z, and t12z cycles, `ims_snow.yaml` is not called. Only in the t18z cycle, the ims snow data is required.

jiaruidong2017: In ush/python/pygfs/task/analysis.py:

```
def get_obs_dict(self) -> Dict[str, Any]:

        # get observers from master dictionary
        observers = obs_list_config['observers']
        copylist = []
        for ob in observers:
            obfile = ob['obs space']['obsdatain']['engine']['obsfile']
            basename = os.path.basename(obfile)
            copylist.append([os.path.join(self.task_config['COM_OBS'], basename), obfile])
        obs_dict = {
            'mkdir': [os.path.join(self.runtime_config['DATA'], 'obs')],
            'copy': copylist
        }
        return obs_dict
```

aerorahul: > The observation is provided at the yaml file sorc/gdas.cd/parm/land/obs/config/ims_snow.yaml:
> 
> ```
>   obsdatain:
>     engine:
>       type: H5File
>       obsfile: $(DATA)/obs/$(OPREFIX)ims_snow_{{ current_cycle | to_YMDH }}.nc4
> ```
> 
> In the t00z, t06z, and t12z cycles, `ims_snow.yaml` is not called. Only in the t18z cycle, the ims snow data is required.

It seems that the observation file needs to be cycle dependent.

aerorahul: Or you can link the `ims{YYYY}{DOY}_4km_v1.3.nc` as `gdas.t{HH}z.ims_snow_{PDY}{HH}.nc4` for all cycles except `18z`.
This will allow you to retain the same observation yaml.

jiaruidong2017: We already made the observation yaml file as cycle dependent in config.landanl as: 

```
obs_list_name=gdas_land_adpsfc_only.yaml
if [[ "${cyc}" = "18" ]]; then
    obs_list_name=gdas_land_prototype.yaml
fi
```

While:
**gdas_land_adpsfc_only.yaml:**
```
observers:
- !INC ${OBS_YAML_DIR}/adpsfc_snow.yaml
```

**gdas_land_prototype.yaml:**
```
observers:
- !INC ${OBS_YAML_DIR}/adpsfc_snow.yaml
- !INC ${OBS_YAML_DIR}/ims_snow.yaml
```

Therefore, there are no any issues for the t00z, t06z and t12z cycles. The problem is only at the t18z cycle when `ims_snow.yaml` is included. As you suggested, I can link the `ims{YYYY}{DOY}_4km_v1.3.nc` as `gdas.t{HH}z.ims_snow_{PDY}{HH}.nc4` to avoid this problem, and later the `gdas.t{HH}z.ims_snow_{PDY}{HH}.nc4` will be replaced by the computed file from prepare_IMS. @CoryMartin-NOAA Do you think it is okay for this approach? Thanks.

CoryMartin-NOAA: What we really need is IODA to handle missing obs... I would rather not link a wrong file here because that is asking for trouble. We can create a fake file if you want, but I think that is more workflow effort than just doing what @jiaruidong2017 mentioned above and use a different list YAML depending on cycle.

aerorahul: I suspect simply linking `ims{YYYY}{DOY}_4km_v1.3.nc` as `gdas.t{HH}z.ims_snow_{PDY}{HH}.nc4` won't be enough.  IODA is not just looking for the file to be present, but it also expects it to be a IODA file.  `ims{YYYY}{DOY}_4km_v1.3.nc` is not a IODA file and so the program will abort. 

We can proceed with the approach of including the `ims_snow.yaml` only for the 18z cycle.

@CoryMartin-NOAA Is there an issue open in IODA that communicates this requirement for IODA to handle missing IODA files?  If not, I would recommend opening an issue.

CoryMartin-NOAA: Yes https://github.com/JCSDA-internal/ioda/issues/541

jiaruidong2017: This issue has been fixed with the change at https://github.com/NOAA-EMC/GDASApp/pull/490#issuecomment-1563363699

jiaruidong2017 (approved):
The snow DA test run succeeded for all four cycles with the fix at https://github.com/NOAA-EMC/GDASApp/pull/490#issuecomment-1563363699

The `landanl` and `preplandobs` directories need to be deleted after the jobs are successfully completed.

ush/python/pygfs/task/land_analysis.py
WalterKolczynski-NOAA: Would prefer a jinja2 filter. Worried about unintended side effects of altering the config value here. For instance, if this line got called twice and started as False, it would go False → '.false.' → '.true.' because the non-null string would be "truthy" on the second pass:

```
>>> a=False
>>> a = ".true." if a else ".false."
>>> print(a)
.false.
>>> a = ".true." if a else ".false."
>>> print(a)
.true.
```

WalterKolczynski-NOAA (commented):
I think I'm happy once the FRAC_GRID conversion to fortran namelist bool gets worked out and the GDAS version gets updated after the dependency PRs are merged.

aerorahul:
@jiaruidong2017 I have implemented a change that @WalterKolczynski-NOAA requested regarding the handling of the type conversion from python to Fortran.
I don't expect this to break as I have tested this function in another branch.  
Please feel free to check.
An update to the GDASApp is included in the PR https://github.com/NOAA-EMC/GDASApp/pull/499

jiaruidong2017:
@aerorahul Your recent change for handling the type conversion works well for my tests. 

I noticed one change need to be reverted in `workflow/applications/gfs_cycled.py` as below: 

```
Plase make a change from: 
           configs += ['preplandobs', 'landanlinit', 'landanlrun', 'landanlfinal']
to: 
           configs += ['preplandobs', 'landanl']

And make a change from:
            gdas_gfs_common_tasks_before_fcst += ['preplandobs', 'landanlinit', 'landanlrun', 'landanlfinal']
to: 
            gdas_gfs_common_tasks_before_fcst += ['preplandobs', 'landanl']
```
I think the changes may come from the "merge to develop". Please revert the above changes.

aerorahul:
> @aerorahul Your recent change for handling the type conversion works well for my tests.
> 
> I noticed one change need to be reverted in `workflow/applications/gfs_cycled.py` as below:
> 
> ```
> Plase make a change from: 
>            configs += ['preplandobs', 'landanlinit', 'landanlrun', 'landanlfinal']
> to: 
>            configs += ['preplandobs', 'landanl']
> 
> And make a change from:
>             gdas_gfs_common_tasks_before_fcst += ['preplandobs', 'landanlinit', 'landanlrun', 'landanlfinal']
> to: 
>             gdas_gfs_common_tasks_before_fcst += ['preplandobs', 'landanl']
> ```
> 
> I think the changes may come from the "merge to develop". Please revert the above changes.

Thanks @jiaruidong2017 for catching this.  I have updated this.

aerorahul:
@CoryMartin-NOAA 
gdasApp did not update the hash in feature/stable-nightly following the merge of PR NOAA-EMC/GDASApp#499

jiaruidong2017 (approved):
This PR is good to me, and hope it can be merged soon. Thanks.

aerorahul:
> This PR is good to me, and hope it can be merged soon. Thanks.

Yes. As soon as the GDASApp hash in `feature/stable-nightly` is updated to include commits from NOAA-EMC/GDASApp#499, I will make that commit and merge.

aerorahul:
And ofcourse a review from @CoryMartin-NOAA

CoryMartin-NOAA:
@aerorahul commit is [1da2e63](https://github.com/NOAA-EMC/GDASApp/commit/1da2e6350609149cec8f21be4585e88494daf9ce)

I will review later this AM

Externals.cfg
```diff
@@ -50,7 +50,7 @@ protocol = git
 required = False
 
 [GDASApp]
-hash = 81675c9
+hash = c9cd07f
```
CoryMartin-NOAA: update to [1da2e63](https://github.com/NOAA-EMC/GDASApp/commit/1da2e6350609149cec8f21be4585e88494daf9ce)

sorc/checkout.sh
```diff
@@ -156,7 +156,7 @@ if [[ ${checkout_gsi} == "YES" ]]; then
 fi
 
 if [[ ${checkout_gdas} == "YES" ]]; then
-  checkout "gdas.cd" "https://github.com/NOAA-EMC/GDASApp.git" "81675c9"; errs=$((errs + $?))
+  checkout "gdas.cd" "https://github.com/NOAA-EMC/GDASApp.git" "c9cd07f"; errs=$((errs + $?))
```
CoryMartin-NOAA: [1da2e63](https://github.com/NOAA-EMC/GDASApp/commit/1da2e6350609149cec8f21be4585e88494daf9ce)

ush/python/pygfs/task/analysis.py
```diff
@@ -192,10 +196,75 @@ def link_jediexe(self: Task) -> None:
...
+        except OSError:
+            logger.exception(f"FATAL ERROR: Failed to execute {exec_cmd}")
+            raise OSError(f"{exec_cmd}")
+        except Exception:
+            logger.exception(f"FATAL ERROR: Error occured during execution of {exec_cmd}")
+            raise WorkflowException(f"{exec_cmd}")
+
+    @staticmethod
+    @logit(logger)
+    def tgz_diags(statfile: str, diagdir: str) -> None:
```
CoryMartin-NOAA, line 250: note to myself and @RussTreadon-NOAA we can use this in aero/atm once this PR is merged

ush/python/pygfs/task/land_analysis.py
```diff
@@ -23,22 +25,37 @@ class LandAnalysis(Analysis):
     """
     Class for global land analysis tasks
     """
+
+    NMEM_LANDENS = 2  # The size of the land ensemble is fixed at 2.  Does this need to be a variable?
```
CoryMartin-NOAA, line 29: good question, probably not? but let's leave it for now

ush/python/pygw/src/pygw/jinja.py
```diff
@@ -94,27 +94,49 @@ def render(self, data: Dict = None) -> str:
...
         """
+
         env = jinja2.Environment(loader=loader, undefined=self.undefined)
         env.filters["strftime"] = lambda dt, fmt: strftime(dt, fmt)
         env.filters["to_isotime"] = lambda dt: to_isotime(dt) if not isinstance(dt, SilentUndefined) else dt
         env.filters["to_fv3time"] = lambda dt: to_fv3time(dt) if not isinstance(dt, SilentUndefined) else dt
         env.filters["to_YMDH"] = lambda dt: to_YMDH(dt) if not isinstance(dt, SilentUndefined) else dt
         env.filters["to_YMD"] = lambda dt: to_YMD(dt) if not isinstance(dt, SilentUndefined) else dt
         env.filters["to_julian"] = lambda dt: to_julian(dt) if not isinstance(dt, SilentUndefined) else dt
+        env.filters["to_f90bool"] = lambda bool: ".true." if bool else ".false."
```
CoryMartin-NOAA, line 133: this is nice!

CoryMartin-NOAA (commented):
Everything looks good, thanks @aerorahul , just will approve once GDASApp hash is updated

Externals.cfg
aerorahul: done.

sorc/checkout.sh
aerorahul: done.

CoryMartin-NOAA (approved):

PR #1667, opened 2023-06-06
Add AWS functionality to workflow setup and forecast job

HenryRWinterbottom:
**Description**

This PR provides support for running UFS weather model forecasts, using the global-workflow, on the AWS Parallel Works (PW) platform. This PR accomplishes the following:

- Building the UFS weather model within the global-workflow build system;
- Running a UFS weather model atmosphere-only deterministic forecast using the global-workflow infrastructure.

Follow-up PRs will include the following:

- Adding documentation for building and using the AWS PW UFSWM spack-stack environment; this is heavily borrowed from Sadegh Sadeghi Tabas's presentation 03 May 2023;
- Addressing remaining `TODO` issues within this PR; several notes are made throughout as to how to better streamline the connections between the UFSWM and the global-workflow for the respective cloud platforms;
- Adding data-transfer capabilities within the global-workflow for cloud deployments;
- Testing additional UFSWM capabilities (e.g., coupling, etc.,) and resolutions (this PR has been tested at C48 and has not been configured for other cubed-sphere resolutions).

No dependencies other than the loading of the appropriate AWS PW modules are required. These are included in this PR. 

**Type of change**

**How Has This Been Tested?**

This PR has been tested as follows:
- The global-workflow fork/branch has been deployed to the AWS PW platform;
- The UFSWM is built using the global-workflow build system;
- A C48 atmosphere-only forecast has been run to completion. 
  
**Checklist**


ush/detect_machine.sh
```diff
@@ -38,6 +38,16 @@ case $(hostname -f) in
   *) MACHINE_ID=UNKNOWN ;;  # Unknown platform
 esac
 
+if [[ ${MACHINE_ID} == "UNKNOWN" ]]; then 
+   case $(dnsdomainname -f) in
```
aerorahul:
```suggestion
   case ${PW_CSP:-} in
```

ush/detect_machine.sh
```diff
@@ -38,6 +38,16 @@ case $(hostname -f) in
...
+
+      # AWS Parallel Works
+      *pw-noaa*pw.local) MACHINE_ID=aws ;;
```
aerorahul:
```suggestion
      "aws" | "google" | "azure" ) MACHINE_ID=noaacloud ;; 
```

ush/module-setup.sh
```diff
@@ -102,6 +102,14 @@ elif [[ ${MACHINE_ID} = discover* ]]; then
     export PATH=${PATH}:${SPACK_ROOT}/bin
     . "${SPACK_ROOT}"/share/spack/setup-env.sh
 
+# TODO: This can likely be made more general once other cloud
+# platforms come online.
+elif [[ ${MACHINE_ID} = aws_pw ]]; then
```
aerorahul:
```suggestion
elif [[ ${MACHINE_ID} = "noaacloud" ]]; then
```

ush/module-setup.sh
```diff
@@ -102,6 +102,14 @@ elif [[ ${MACHINE_ID} = discover* ]]; then
...
+
+    export SPACK_ROOT=/contrib/global-workflow/spack-stack/spack
+    export PATH=${PATH}:${SPACK_ROOT}/bin
+    . "${SPACK_ROOT}"/share/spack/setup-env.sh
+
```
aerorahul, line 112: See https://github.com/ufs-community/ufs-weather-model/blob/2f71156c9b56a8e5f26cb900dfa4f57b50df2fe9/tests/module-setup.sh#LL47C29-L52C29

workflow/hosts.py
```diff
@@ -15,14 +15,15 @@ class Host:
     """
 
     SUPPORTED_HOSTS = ['HERA', 'ORION', 'JET',
-                       'WCOSS2', 'S4', 'CONTAINER']
+                       'WCOSS2', 'S4', 'CONTAINER', 'AWSPW']
```
aerorahul, line 18:
```suggestion
                       'WCOSS2', 'S4', 'CONTAINER', 'NOAACLOUD']
```

aerorahul: May be there is a need to distinguish between Azure, Google and AWS later.
But for now, lets keep with the ufs-weather-model

sorc/build_ufs.sh
```diff
@@ -25,17 +25,47 @@ while getopts ":da:v" option; do
...
+  mv "./tests/modules.fv3_${COMPILE_NR}.lua" ./tests/modules.ufs_model.lua
+  cp "./modulefiles/ufs_common.lua" ./tests/ufs_common.lua
+fi
+
+if [[ "${MACHINE_ID}" == "noaacloud" ]]; then
+  case $(dnsdomainname -f) in   
+     # TODO: Add Google and Azure platforms.
+      *pw-noaa*pw.local) CLOUD_MACHINE_ID="aws.${RT_COMPILER}" ;;
+      *) ;; 
+  esac
```
aerorahul: we shouldn't be doing detection again.  MACHINE ID has been detected once. UFSWM is using a `PW_CSP` variable. Lets use that and if necessary use `PW_CSP` to distinguish.

sorc/build_ufs.sh
```diff
@@ -25,17 +25,47 @@ while getopts ":da:v" option; do
...
+  if [[ "${CLOUD_MACHINE_ID}" == "aws.intel" ]]; then
+    module use /contrib/spack-stack/envs/ufswm/install/modulefiles/Core
+    module load stack-intel
+    module load stack-intel-oneapi-mpi
+    module load ufs-weather-model-env/1.0.0
+    # TODO: It is still uncertain why this is the only module that is 
+    # missing; check the spack build as this needed to be added manually.
+    module load w3emc/2.9.2 # TODO: This has similar issues for the EPIC stack.
+    module list
+  fi
```
aerorahul: I don't think this is necessary as `ufs_noaacloud.intel.lua` contains the information.  Can we not use that?

aerorahul (changes requested):
Sorry for getting late on this PR.  
I am noticing redefinitions between the global-workflow and the ufs-weather-model e.g. PW_CSP and some of the module loading.
Please take a look at the comments.

HenryRWinterbottom:
@aerorahul I will test your comments/suggestions next week. For some reason I think that I needed to be explicit with the platforms but I could be mistaken. I will keep you posted. 

Thank you for the review.

sorc/build_ufs.sh
HenryRWinterbottom: @aerorahul Can you point me to a use case within the UFSWM of the `PW_CSP` instance?

For clarification, the above detection was needed to determine which CSP the user is on. The DNS strings should be unique. This allows the GW (and build system) to determine which host they are on and thus make the downstream decisions accordingly.

For example, AWS PW use `mpiexec` as it's parallel executable launcher. However, Azure or Google could use another. This may ultimately wind up being redundant but since I haven't see the `PW_CSP` use case I felt it unsafe to make such an assumption.

aerorahul: See https://github.com/ufs-community/ufs-weather-model/blob/develop/tests/detect_machine.sh
I would recommend a `PW_CSP` as a separate configurable if you wish to identify a CSP available on NOAACloud.

sorc/build_ufs.sh
HenryRWinterbottom: The path that the module file points to, an EPIC stack, does not exist. As a result, the build environment cannot be defined using the module that you noted above.

```
[Henry.Winterbottom@hwufscpldcld-2 modulefiles]$ ls /contrib/EPIC/spack-stack/spack-stack-1.3.0/
ls: cannot access /contrib/EPIC/spack-stack/spack-stack-1.3.0/: No such file or directory
```

Note that this comes from the `develop` branch.

I suggest that we include this bit for now. Once EPIC provides a working stack we can create a new PR and make the update.

aerorahul: Understood.  I did not know that the EPIC stack is not available for use by us.  This certainly makes porting an issue.
I will discuss with @arunchawla-NOAA on how to enable that.

ush/detect_machine.sh
HenryRWinterbottom: Addressed in latest commit/update.

ush/detect_machine.sh
HenryRWinterbottom: Addressed in latest commit/update.

ush/module-setup.sh
HenryRWinterbottom: Addressed in latest commit/update.

aerorahul:
@SadeghTabas-NOAA Would you please review the changes in the build and run script requirements of the ufs-weather-model?

sadeghitabas:
@aerorahul Sure, I'll do it today. Thanks

jobs/rocoto/fcst.sh
```diff
@@ -30,6 +32,25 @@ if [[ "${MACHINE_ID}" = "hera" ]]; then
 #elif [[ "${MACHINE_ID}" = "wcoss2" ]]; then
 #  module load "python/3.7.5"
 fi
+if [[ "${MACHINE_ID}" == "noaacloud" ]]; then
+   if [[ "${PW_CSP}" == "aws" ]]; then
```
aerorahul:
```suggestion
   if [[ "${PW_CSP:-}" = "aws" ]]; then
```

jobs/rocoto/fcst.sh
```diff
@@ -12,9 +12,11 @@ source "${HOMEgfs}/ush/preamble.sh"
 source "${HOMEgfs}/ush/detect_machine.sh"
 set +x
 source "${HOMEgfs}/ush/module-setup.sh"
-module use "${HOMEgfs}/sorc/ufs_model.fd/tests"
-module load modules.ufs_model.lua
-module load prod_util
+if [[ "${MACHINE_ID}" != "awspw" ]]; then
+  module use "${HOMEgfs}/sorc/ufs_model.fd/tests"
+  module avail
```
aerorahul:
```suggestion
```
Don't need this

sorc/build_ufs.sh
```diff
@@ -21,7 +21,7 @@ while getopts ":da:v" option; do
   esac
 done
 
-cd "${cwd}/ufs_model.fd"
+cd "${cwd}/ufs_model.fd" # TODO: This is redundant; see below.
```
aerorahul:
```suggestion
cd "${cwd}/ufs_model.fd"
```
No it is not!  If you delete this line, line 26 will fail.

sorc/build_ufs.sh
```diff
@@ -32,10 +32,38 @@ COMPILE_NR=0
...
-cp "./modulefiles/ufs_common.lua" ./tests/ufs_common.lua
-cp "./modulefiles/ufs_common_spack.lua" ./tests/ufs_common_spack.lua
+if [[ "${MACHINE_ID}" != "noaacloud" ]]; then
+  ./tests/compile.sh "${MACHINE_ID}" "${MAKE_OPT}" "${COMPILE_NR}" "${CLEAN_BEFORE}" "${CLEAN_AFTER}"
+  mv "./tests/fv3_${COMPILE_NR}.exe" ./tests/ufs_model.x
+  mv "./tests/modules.fv3_${COMPILE_NR}.lua" ./tests/modules.ufs_model.lua
+  cp "./modulefiles/ufs_common.lua" ./tests/ufs_common.lua
+fi
+
+if [[ "${MACHINE_ID}" == "noaacloud" ]]; then
```
aerorahul:
```suggestion
else
```

sorc/build_ufs.sh
```diff
@@ -32,10 +32,38 @@ COMPILE_NR=0
 CLEAN_BEFORE=YES
 CLEAN_AFTER=NO
 
-./tests/compile.sh "${MACHINE_ID}" "${MAKE_OPT}" "${COMPILE_NR}" "intel" "${CLEAN_BEFORE}" "${CLEAN_AFTER}"
-mv "./tests/fv3_${COMPILE_NR}.exe" ./tests/ufs_model.x
-mv "./tests/modules.fv3_${COMPILE_NR}.lua" ./tests/modules.ufs_model.lua
-cp "./modulefiles/ufs_common.lua" ./tests/ufs_common.lua
-cp "./modulefiles/ufs_common_spack.lua" ./tests/ufs_common_spack.lua
+if [[ "${MACHINE_ID}" != "noaacloud" ]]; then
+  ./tests/compile.sh "${MACHINE_ID}" "${MAKE_OPT}" "${COMPILE_NR}" "${CLEAN_BEFORE}" "${CLEAN_AFTER}"
```
aerorahul:
```suggestion
  ./tests/compile.sh "${MACHINE_ID}" "${MAKE_OPT}" "${COMPILE_NR}" "intel" "${CLEAN_BEFORE}" "${CLEAN_AFTER}"
```
Note the signature of `compile.sh` has changed!

sorc/build_ufs.sh
```diff
@@ -32,10 +32,38 @@ COMPILE_NR=0
...
+    module load stack-intel
+    module load stack-intel-oneapi-mpi
+    module load ufs-weather-model-env/1.0.0
+    # TODO: It is still uncertain why this is the only module that is 
+    # missing; check the spack build as this needed to be added manually.
+    module load w3emc/2.9.2 # TODO: This has similar issues for the EPIC stack.
+    module list
+  fi
+
+  cd "${cwd}/ufs_model.fd" # TODO: This is redundant; see line #28.
```
aerorahul:
```suggestion
```
You are already in this directory.

sorc/build_ufs.sh
```diff
@@ -32,10 +32,38 @@ COMPILE_NR=0
...
+  export CMAKE_FLAGS="${MAKE_OPT}"
+  ./build.sh 
+  mv "${cwd}/ufs_model.fd/build/ufs_model" "${cwd}/ufs_model.fd/tests/ufs_model.x"
+ 
+  # TODO: This is hack? Where is this step performed in the build system?
+  mkdir -p "${HOMEgfs}/exec"
+  cp "${cwd}/ufs_model.fd/tests/ufs_model.x" "${HOMEgfs}/exec/ufs_model.x"
```
aerorahul:
```suggestion
```
Indeed a hack and not necessary.
This is performed in ......

sorc/build_ufs.sh
```diff
@@ -32,10 +32,38 @@ COMPILE_NR=0
...
+  ./build.sh 
+  mv "${cwd}/ufs_model.fd/build/ufs_model" "${cwd}/ufs_model.fd/tests/ufs_model.x"
+ 
+  # TODO: This is hack? Where is this step performed in the build system?
+  mkdir -p "${HOMEgfs}/exec"
+  cp "${cwd}/ufs_model.fd/tests/ufs_model.x" "${HOMEgfs}/exec/ufs_model.x" 
+fi
+
+echo  "${MACHINE_ID}"
+
```
aerorahul:
```suggestion
```

ush/detect_machine.sh
```diff
@@ -38,6 +38,13 @@ case $(hostname -f) in
   *) MACHINE_ID=UNKNOWN ;;  # Unknown platform
 esac
 
+if [[ ${MACHINE_ID} == "UNKNOWN" ]]; then 
+   case ${PW_CSP:-} in
+      "aws" | "google" | "azure") MACHINE_ID=noaacloud ;;
+      *) MACHINE_ID=UNKNOWN ;; # Unknown platform
```
aerorahul:
```suggestion
      *) PW_CSP="UNKNOWN"
```
`MACHINE_ID` already is `UNKNOWN`, so perhaps a better option would be to put `PW_CSP` as "UNKNOWN" since that is the check in this case block?

aerorahul (changes requested):
I only went through the shell scripts here.
Take a look if you agree with these suggestions.

WalterKolczynski-NOAA (commented):
No additional requests beyond what Rahul has pointed out. I would have rather kept `$MACHINE` as the specific cloud platform, but understand the decision to follow what UFS has done.

sorc/build_ufs.sh
HenryRWinterbottom: I still don't see (cannot find) where this step is performed.

jobs/rocoto/fcst.sh
```diff
@@ -12,9 +12,10 @@ source "${HOMEgfs}/ush/preamble.sh"
 source "${HOMEgfs}/ush/detect_machine.sh"
 set +x
 source "${HOMEgfs}/ush/module-setup.sh"
-module use "${HOMEgfs}/sorc/ufs_model.fd/tests"
-module load modules.ufs_model.lua
-module load prod_util
+if [[ "${MACHINE_ID}" != "awspw" ]]; then
```
aerorahul:
```suggestion
if [[ "${MACHINE_ID}" != "noaacloud" ]]; then
```

sorc/build_ufs.sh
aerorahul: Line 216-217 of `sorc/link_workflow.sh`

workflow/hosts.py
```diff
@@ -47,6 +48,12 @@ def detect(cls):
         elif container is not None:
             machine = 'CONTAINER'
 
+        # TODO: Note that this requires the environment variable to be
+        # defined prior to calling this script; is there a
+        # better/preferred way to do this?
+        elif os.getenv('MACHINE') is not None:
+            machine = os.getenv('MACHINE')
+
```
aerorahul:
```suggestion
```
Instead of this, how about `machine = os.getenv('MACHINE', "NOTFOUND")` on Line 35 above?

aerorahul (approved):
looks good.
You will have to show us how to run this beast on the cloud.

WalterKolczynski-NOAA (approved):

PR #1670, opened 2023-06-08
Use test cases from PR for CI

### LINKED ISSUE #1662: Running New Cases from PR in CI
**Description**

As a developer, I  would like the option of running new cases in CI from the PR branch that is being Labeled Ready for CI testing so I can include novel tests that are being introduced by the updates in the PR itself.

The new cases in the PR are created and run from the CI cron jobs when the option is  selected
How the option is selected will be explored as part of this effort.

**Suggest A Solution**

Use an auxiliary Label, **CI-PR-Cases**, to cause the above to happen.

TerrenceMcGuinness-NOAA:
**Description**

This feature changes the CI system to use the test cases from the PR in question. This allows the PR to add or modify tests and have them applied to itself to ensure they work.

Fixes #1662

- [ ] Bug fix (non-breaking change which fixes an issue)

**How Has This Been Tested?**

- [ ] Clone and Build tests on WCOSS Dell P3
- [ ] Cycled test on Orion
- [x] Forecast-only test **C48_S2S** on Hera from a case file within the PR


aerorahul (approved):
looks simple enough

ci/scripts/driver.sh
```diff
@@ -86,11 +88,19 @@ for pr in ${pr_list}; do
     done
     rm -Rf "${pr_dir}"
   fi
+  # Check to see if this PR is Labeled CI-PR-Cases to get cases from itself
```
WalterKolczynski-NOAA: Is there any reason to not just always pull cases from the PR?

ci/scripts/pr_list_database.py
```diff
@@ -1,12 +1,20 @@
...
 import sys
+import os
 from pathlib import Path
-from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter, REMAINDER
+from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter, REMAINDER, ZERO_OR_MORE
 import sqlite3
 
 
-def sql_connection(filename: Path) -> sqlite3.Connection:
+def dir_path(string):
```
WalterKolczynski-NOAA: As dumb as it seems for so simple a function, there should have a doc-block for this.

WalterKolczynski-NOAA: Since this will return the absolute path for a file as well, should it be named `abs_path` of `full_path` instead?

WalterKolczynski-NOAA (changes requested):

ci/scripts/driver.sh
TerrenceMcGuinness-NOAA: Not sure I understand the question, the PR is pulled when it is built.

ci/scripts/pr_list_database.py
TerrenceMcGuinness-NOAA: This was to make sure it was one or the other and nothing else, I guess you are asking me to over document and split this into two types.

ci/scripts/driver.sh
WalterKolczynski-NOAA: Right, but you are relying on whether there is a label set to use the cases from the PR rather than what you already have. Why not always use the ones in the PR? That eliminates the need to set yet another label, and also avoids problems caused by tests being added or changed and the label not being set.

TerrenceMcGuinness-NOAA: Oh, I understand the question now, yes I guess we could just always get cases from the PR, I'm going to refer this question to Raul.

TerrenceMcGuinness-NOAA: The problem I was told is that users can then override the intent of the tests by leaving them out, but like you said, that would be obvious to you.

ci/scripts/pr_list_database.py
TerrenceMcGuinness-NOAA: Oh wait, I see now, you are asking to simply change the name of the function, yes that makes more sense, will do.

TerrenceMcGuinness-NOAA (commented):
updated name of type function for full path (may not be a file)

ci/scripts/pr_list_database.py
WalterKolczynski-NOAA: Yeah, those were two different things about the same function:
- It needs a doc-block
- If it handles more than just directory paths, the name shouldn't imply it is only for directories

TerrenceMcGuinness-NOAA: - Why does this need a doc block, doesn't that defeat the purpose of useful documentation?
- I did rename it.

WalterKolczynski-NOAA: Because every function/method should have a doc-block. When we finally get around to parsing these to create API documentation, it will be needed. It should be very straightforward and short.

TerrenceMcGuinness-NOAA: ok

TerrenceMcGuinness-NOAA (commented):
added doc block and check the file is present except when being created

TerrenceMcGuinness-NOAA (commented):
added doc block and checks

ci/scripts/pr_list_database.py
```diff
@@ -7,7 +7,22 @@
...
+
+    Parameters
+    ----------
+    string: string
+        argument sting from --dbfile
+
+    Returns
+    -------
+    os.path.abspath
+        abspath type of input string
```
WalterKolczynski-NOAA:
```suggestion
    Gets the absolute path of the given file and confirms the directory exists

    Parameters
    ----------
    string : str
        Path to a file

    Returns
    --------
    str
        Absolute path of input path
        
    Raises
    -------
    NotADirectoryError
        If the target directory for the file does not exist.
```

WalterKolczynski-NOAA (approved):

PR #1689, opened 2023-06-13
Fix wave gridded post logic for glo_025 

JessicaMeixner-NOAA:
**Description**

A recent fix of another issue meant that wave config settings were set incorrectly for glo_025 resolution.  This removes pass through options which caused the issue since a few variables were intentionally set to nothing but then were set incorrectly to another default.  

**Type of change**

**How Has This Been Tested?**

Testing now w/C384/glo_025
  
**Checklist**

### DIFF
--- a/parm/config/gfs/config.wave
+++ b/parm/config/gfs/config.wave
@@ -81,10 +81,8 @@ case "${waveGRD}" in
     export waveuoutpGRD=${waveGRD}
     ;;
   *)
-    echo "No grid specific wave config values"  
-    export waveinterpGRD=${waveinterpGRD:-'glo_15mxt at_10m ep_10m wc_10m glo_30m'} # Grids that need to be interpolated from native 
-    export wavepostGRD=${wavepostGRD:-${waveGRD}} # Native computational grids that will be post-processed (grib2)
-    export waveuoutpGRD=${waveGRD}
+    echo "No grid specific wave config values exist.  Aborting."
+    exit 
     ;;  
 esac
 
--- a/parm/config/gfs/config.wave
+++ b/parm/config/gfs/config.wave
@@ -54,42 +54,43 @@ case "${waveGRD}" in
     #Grid used for P8
     export waveinterpGRD=''
     export wavepostGRD='gwes_30m'
+    export waveuoutpGRD=${waveGRD}
     ;;
   "mx025")
     #Grid used for HR1 (tripolar 1/4 deg)
     export waveinterpGRD='reg025'
-    export wavepostGRD=' '
+    export wavepostGRD=''
+    export waveuoutpGRD=${waveGRD}
     ;;
   "glo_025") 
     #GEFSv13 regular lat/lon 1/4 deg grid 
     export waveinterpGRD=''
     export wavepostGRD='glo_025'
+    export waveuoutpGRD=${waveGRD}
     ;;
   "glo_200")
     #Global regular lat/lon 2deg deg grid 
     export waveinterpGRD=''
     export wavepostGRD='glo_200'
+    export waveuoutpGRD=${waveGRD}
     ;;
   "glo_500")
     #Global regular lat/lon 5deg deg grid 
     export waveinterpGRD=''
     export wavepostGRD='glo_500'
+    export waveuoutpGRD=${waveGRD}
     ;;
   *)
     echo "No grid specific wave config values"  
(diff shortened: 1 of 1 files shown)


JessicaMeixner-NOAA:
The test is now running the correct wave post.  EXP/COM can be found here: /work2/noaa/marine/jmeixner/Lapenta2023/test01/lt01 

Please let me know if any additional testing or explanation is needed.

WalterKolczynski-NOAA (approved):
Approved pending CI test success.

parm/config/gfs/config.wave
```diff
@@ -54,43 +54,44 @@ case "${waveGRD}" in
...
     ;;
   "glo_500")
     #Global regular lat/lon 5deg deg grid 
     export waveinterpGRD=''
     export wavepostGRD='glo_500'
+    export waveuoutpGRD=${waveGRD}
     ;;
   *)
     echo "No grid specific wave config values"  
+    export waveinterpGRD=${waveinterpGRD:-'glo_15mxt at_10m ep_10m wc_10m glo_30m'} # Grids that need to be interpolated from native
```
aerorahul: Can you elaborate what value of `waveGRD` will result into this case?
I don't think this block will ever be entered.

aerorahul (commented):
Code changes look ok, but I do have a question.

parm/config/gfs/config.wave
JessicaMeixner-NOAA: If you choose a waveGRD that is not one of the specified values above. This would be more of a case where you could try to use a grid that's not one of the set-grids here but you have fewer places to change things.  I guess this could be removed completely instead of just moved here.

aerorahul: I guess the more appropriate question is, is there a use case for this `*`?  Should we instead abort saying this `waveGRD` is not supported and ask the user to choose one from the above available options?

JessicaMeixner-NOAA: My opinion is no, but if that's the desired path forward I can make those changes.

aerorahul: I am of the same opinion.  Just abort then with a message saying this is an unsupported option for `waveGRD`.

JessicaMeixner-NOAA: Apologies, I mispoke here.  My opinion is no we should not exit.   But since that's what you want I'll  go ahead and make the change so that this bug fix can be made available as soon as possible.

parm/config/gfs/config.wave
```diff
@@ -81,10 +81,8 @@ case "${waveGRD}" in
     export waveuoutpGRD=${waveGRD}
     ;;
   *)
-    echo "No grid specific wave config values"  
-    export waveinterpGRD=${waveinterpGRD:-'glo_15mxt at_10m ep_10m wc_10m glo_30m'} # Grids that need to be interpolated from native 
-    export wavepostGRD=${wavepostGRD:-${waveGRD}} # Native computational grids that will be post-processed (grib2)
-    export waveuoutpGRD=${waveGRD}
+    echo "No grid specific wave config values exist.  Aborting."
+    exit
```
WalterKolczynski-NOAA:
```suggestion
    echo "FATAL ERROR: Unsupported wave grid ${waveGRD}. Aborting."
    exit 5
```

JessicaMeixner-NOAA: Fatal error and the specific wave grid has been added.

parm/config/gfs/config.wave
JessicaMeixner-NOAA: This update has been done as requested.

parm/config/gfs/config.wave
```diff
@@ -54,43 +54,42 @@ case "${waveGRD}" in
...
   "glo_500")
     #Global regular lat/lon 5deg deg grid 
     export waveinterpGRD=''
     export wavepostGRD='glo_500'
+    export waveuoutpGRD=${waveGRD}
     ;;
   *)
-    echo "No grid specific wave config values"  
+    echo "FATAL ERROR: No grid specific wave config values exist for ${waveGRD}. Aborting."
+    exit
```
aerorahul:
```suggestion
    exit 1
```
A non-zero exit is required.

JessicaMeixner-NOAA: Fixed.

WalterKolczynski-NOAA (approved):

aerorahul (approved):
Thank you for elaborating and addressing comments.

Issue #923, opened 2022-07-21
Port NCL scripts to python


WalterKolczynski-NOAA:
**Description**
NCL was deprecated in Sep 2019 and discontinued the follow summer. As such, it is no longer available on the operational machine and we should be moving to non-NCL alternatives for all scripts.

For global-workflow directly, this is mostly the new ice and ocean post. Otherwise. there are some gsi_utility plotting scripts and others are located in UFS components, none of which I think we use.

```
./sorc/ufs_model.fd/AQM/src/model/CMAQ/PREP/wbdust/scripts/monthly_to_daily.ncl
./sorc/ufs_model.fd/MOM6-interface/MOM6/pkg/CVMix-src/reg_tests/shear/plot_PP_diff_coeffs.ncl
./sorc/ufs_model.fd/MOM6-interface/MOM6/pkg/CVMix-src/reg_tests/shear/plot_LMD_diff_coeffs.ncl
./sorc/ufs_model.fd/MOM6-interface/MOM6/pkg/CVMix-src/reg_tests/double_diff/plot_diff_coeffs.ncl
./sorc/ufs_model.fd/MOM6-interface/MOM6/pkg/CVMix-src/reg_tests/kpp/plot_flux_profiles.ncl
./sorc/ufs_model.fd/MOM6-interface/MOM6/pkg/CVMix-src/reg_tests/kpp/compare_interpolation_methods.ncl
./sorc/ufs_model.fd/MOM6-interface/MOM6/pkg/CVMix-src/reg_tests/kpp/plot_bulk_Rich.ncl
./sorc/ufs_model.fd/MOM6-interface/MOM6/pkg/CVMix-src/reg_tests/tidal-Simmons/plot_diff_coeffs.ncl
./sorc/ufs_model.fd/MOM6-interface/MOM6/pkg/CVMix-src/reg_tests/tidal-Simmons/plot_tidal_energy_map.ncl
./sorc/ufs_model.fd/MOM6-interface/MOM6/pkg/CVMix-src/reg_tests/common/plot_settings.ncl
./sorc/ufs_model.fd/MOM6-interface/MOM6/pkg/CVMix-src/reg_tests/common/version_consistency.ncl
./sorc/ufs_model.fd/MOM6-interface/MOM6/pkg/CVMix-src/reg_tests/Bryan-Lewis/plot_diff_coeffs.ncl
./sorc/gsi_utils.fd/src/radar_process/reflectivity/plot_compositeRef.ncl
[...]

PR #1736, opened 2023-07-13
Replace ocean/ice post NCL programs with Python program 

EricSinsky-NOAA:
**Description**

The ocean and ice post NCL programs can no longer be used in the global-workflow because NCL is obsolete. Therefore, the ocean and ice post NCL programs have been replaced by one Python program. The ocean/ice post Python program has the same capabilities as the NCL programs.

Refs #923 

**Type of change**

**How Has This Been Tested?**

- [x] EP4 on WCOSS2
  
**Checklist**


ush/ocnpost.py
```diff
@@ -0,0 +1,646 @@
...
+
+        end subroutine
+    '''
+
+    try:
+        import esmfwghtinterp_f90subroutine
+    except ImportError or ModuleNotFoundError:
+        print('Weighted interpolation fortran subroutine has not been compiled. Compiling...')
+        np.f2py.compile(fsource1, modulename='esmfwghtinterp_f90subroutine', verbose=False)
+        import esmfwghtinterp_f90subroutine
```
aerorahul, line 142: One cannot be compiling everytime this script is run.

ush/ocnpost.py
```diff
@@ -0,0 +1,646 @@
...
+
+def esmfmanualregrid(src,S,row,col,frac_b,n_s,varmeth,slatd,slond,dlatd,dlond):
+    import esmfwghtinterp_f90subroutine
```
aerorahul, line 145: all imports must be at the top

WalterKolczynski-NOAA:
@NeilBarton-NOAA @jiandewang Hold off on reviewing until all the checks passed and Rahul or I have given the go-ahead

ush/ocnpost.py
```diff
@@ -0,0 +1,646 @@
...
+
+        if 1 == 1:        
+            dst_oned=esmfwghtinterp_f90subroutine.bilin(src_oned,S,row,col,frac_b,dlatd*dlond)
+
+        dst[0,i,:,:] = np.transpose(np.reshape(dst_oned,(dlond,dlatd),'F'))
+    return dst
+
+def main():
+
+    print('Main program has started')
```
aerorahul, line 170: all new python programs must use `wxflow` to setup logging, execution etc.

ush/ocnpost.py
```diff
@@ -0,0 +1,646 @@
...
+
+    if len(sys.argv) == 6:
+        infiles0=sys.argv[1]
+        outfiles0=sys.argv[2]
+        nemsrc=sys.argv[3]
+        parmfile=sys.argv[4]
+        dstgrds=[sys.argv[5]]
+    else:
+        print('Incorrect number of arguments.')
+        exit()
```
aerorahul, line 184: This needs to be handled with `argparse`
Use `raise` for exception handling and exit with an error message that is clear.

ush/ocnpost.py
```diff
@@ -0,0 +1,646 @@
...
+    if infiles0.endswith('.nc') and outfiles0.endswith('.nc'):
+        print('Input and output files are netcdf.')
+        infiles1=[sys.argv[1]]
+        outfiles1=[sys.argv[2]]
+        dimfiles=1
+    else:
+        with open(infiles0) as f1:
+            infiles1 = f1.read().splitlines()
+        with open(outfiles0) as f2:
+            outfiles1 = f2.read().splitlines()
```
aerorahul, line 195: Please provide comments or describe what the code/section is doing.
Break code into methods for readability and testing.
Add tests.

ush/ocnpost.py
```diff
@@ -0,0 +1,646 @@
...
+    regridtype='manual' #manual or auto
+
+    #variables to be regridded with the native tripole stagger location
+    #and dimensionality 
+    #first BM contained only field "mld", which was actually ePBL
+    #the remaining BMs contain ePBL, MLD_003 and MLD_0125 
+    #the following NCO command will be issued at the end 
+    #to rename the variable mld to ePBL if the variable mld is found
+
+    ncocmd=['ncrename -O -v MLD_003,mld']
```
aerorahul: Where is this used?  Remove unused codes

ush/ocnpost.py
```diff
@@ -0,0 +1,646 @@
...
+        model='MOM'
+    elif 'Tsfc_d' in ncvarlist or 'Tsfc_h' in ncvarlist:
+        model='CICE'
+        if 'Tsfc_d' in ncvarlist:
+            samplevar='Tsfc_d'
+        if 'Tsfc_h' in ncvarlist:
+            samplevar='Tsfc_h' 
+    else:
+        print('Product not supported. Exiting program...')
+        exit()
```
aerorahul, line 268: This should be turned into methods for testing and extensibility

ush/ocnpost.py
```diff
@@ -0,0 +1,646 @@
...
+    print('Post-processing '+str(model))
+
+    #Extract Source Grid coordinates
+    if model == 'MOM':
+        dimxh=ocnf.dimensions['xh'].size
+        dimyh=ocnf.dimensions['yh'].size
+    if model == 'CICE':
+        dimxh=ocnf.dimensions['ni'].size
+        dimyh=ocnf.dimensions['nj'].size
```
aerorahul, line 277: Can `model` be both `MOM` and `CICE` at the same time?

ush/ocnpost.py
```diff
@@ -0,0 +1,646 @@
...
+            cosrot=ocnf['cos_rot']
+        else:
+            cosrot=ocnf['cosrot']
+
+        z_l=ocnf['z_l']
+        z_i=ocnf['z_i']
+        nlevs=len(z_l)
+
+    if model == 'CICE':
+        angleT=ocnf['ANGLET']
```
aerorahul, line 298: Handle `MOM` in `MOM` related methods and `CICE` likewise.

ush/ocnpost.py
```diff
@@ -0,0 +1,646 @@
+#------------------------------------------------------------------
```
aerorahul, line 1: There are too many questions from this script.
1. Is this really efficient?  Is python the right choice for doing this or is a compiled language (e.g. Fortran) better suited for the operations of this size.
2. There are no tests for this.  Every python program that is added must include tests and pass standards.
3. Why are we using those `parm` files?  Can that information not be described in a better form e.g. `yaml` or `namelist` (if using Fortran).

aerorahul (changes requested):
See first cut at the review comments.

ush/ocnpost.py
EricSinsky-NOAA: I agree. As of now, a compilation is only performed if the .so file does not exist in any Python path. However, the compile step can be removed completely from the code if that is preferred.

ush/ocnpost.py
EricSinsky-NOAA: Sure. I wasn't entirely sure if argparse was supported on WCOSS2. I will use argparse instead.

ush/ocnpost.py
aerorahul: I don't understand why this can't just be this python function:
```python
def bilin(src, S, row, col, frac_b, dst_len):
    dst = [0.0] * dst_len  # Initialize destination field to 0.0

    # Apply weights
    for i in range(len(row)):
        dst[row[i]] += S[i] * src[col[i]]

    # Apply fraction correction
    for i in range(dst_len):
        if frac_b[i] != 0.0:
            dst[i] /= frac_b[i]

    return dst
```

ush/ocnpost.py
WalterKolczynski-NOAA: argparse is core python

ush/ocnpost.py
```diff
@@ -0,0 +1,646 @@
...
+            latd = np.arange(-90.0, 90.1, 1.0)
+
+        wgtsfile1 = nemsrc+'/'+'tripole.mx025.Ct.to.'+dsttype[0]+dstgrds[jj]+'.bilinear.nc'        
+        wgtsfile2 = nemsrc+'/'+'tripole.mx025.Ct.to.'+dsttype[0]+dstgrds[jj]+'.conserve.nc'
+        if model == 'MOM':
+            wgtsfile3 = nemsrc+'/'+'tripole.mx025.Cu.to.Ct.bilinear.nc'    
+            wgtsfile4 = nemsrc+'/'+'tripole.mx025.Cv.to.Ct.bilinear.nc'
+        if model == 'CICE':
+            wgtsfile3 = nemsrc+'/'+'tripole.mx025.Bu.to.Ct.bilinear.nc'    
+            wgtsfile4 = nemsrc+'/'+'tripole.mx025.Bu.to.Ct.bilinear.nc'
```
aerorahul, line 389: these seem to be hard-wired here. Why?

ush/ocnpost.py
```diff
@@ -0,0 +1,646 @@
...
+        col3=rgrdf3['col'][:]
+        frac_b3=rgrdf3['frac_b'][:]
+        n_s3=len(S3)
+
+        rgrdf4=nc.Dataset(wgtsfile4)
+        S4=rgrdf4['S'][:].copy()
+        row4=rgrdf4['row'][:]
+        col4=rgrdf4['col'][:]
+        frac_b4=rgrdf4['frac_b'][:]
+        n_s4=len(S4)
```
aerorahul, line 423: `netCDF4` affords lazy-loading of the dataset.  Why are we extracting that data upfront?  This doesn't seem like a good use of the library.

ush/ocnpost.py
```diff
@@ -0,0 +1,646 @@
...
+        latitude = f.createVariable('lat', 'f4', 'lat')  
+        latitude.long_name = 'latitude'
+        latitude.units = 'degrees_north'
+
+        rgmask2dn = f.createVariable('rgmask2d1', 'f4', ('time', 'lat', 'lon'))
+        if model == 'MOM':
+            rgmask3dn = f.createVariable('rgmask3d', 'f4', ('time', 'z_l','lat', 'lon')) 
+        time = f.createVariable('Time', 'i4', 'time')
+
+        longitude[:] = lond
```
aerorahul, line 472: these are things you would do in a Fortran code.
This is not proper use of the python language.

ush/ocnpost.py
EricSinsky-NOAA: This program can only process MOM and CICE one at a time, therefore `model` can only be either `MOM` or `CICE`.

ush/ocnpost.py
```diff
@@ -0,0 +1,646 @@
...
+            print('beginning the regridding for mask3d')                                                                                                                                                                                                 
+            if regridtype == 'manual':
+               rgmask3d=esmfmanualregrid(mask3d,S1,row1,col1,frac_b1,n_s1,'bilinear',dimyh,dimxh,len(latd),len(lond))
+            print('Data has been regridded for mask3d') 
+            rgmask3d = np.where(rgmask3d>0.0,  1.0, 0.0)       
+
+        #Create Mask NETCDF File#########      
+        testfile='masks_'+dstgrds[jj]+'.nc'
+        os.system('rm -vf '+testfile)
+        f = nc.Dataset(testfile,'w', format='NETCDF4')
```
aerorahul, line 446: use `contextmanagers`

ush/ocnpost.py
```diff
@@ -0,0 +1,646 @@
...
+            ocnf=nc.Dataset(infile0)
+            print('Input file is '+infile0)
+            print('Output file is '+outfile0)
+        #Extract time
+            intime=ocnf['time']
+                
+        #Create output regridded netcdf file 
+            FILENAME_REGRID = outfile0
+            os.system('rm -f '+FILENAME_REGRID)
+            outcdf = nc.Dataset(FILENAME_REGRID,'w', format='NETCDF3_CLASSIC')
```
aerorahul, line 498: why are we writing out `NETCDF3` version filed?

ush/ocnpost.py
aerorahul: In that case, this program needs to be written in a way that leverages OOP.

ush/ocnpost.py
EricSinsky-NOAA: 1.  I definitely agree that Fortran is more efficient with computation, which is why most of the intense computation is performed using a Fortran to Python interface generator (f2py) provided by Numpy. 
2. Sure, formal tests will be performed. 
3. There's no particular reason. I can change the format of the parm files to yaml.

ush/ocnpost.py
EricSinsky-NOAA: Yes, I will fix this so that the filenames not hard-wired. Thanks for catching this.

ush/ocnpost.py
EricSinsky-NOAA: Thanks for catching this. I will remove this since it will not be needed.

ush/ocnpost.py
WalterKolczynski-NOAA: All these things that are different for MOM and CICE should ideally be moved to a yaml file. The python shouldn't need to know what it is working on, just a yaml with all the settings.

ush/ocnpost.py
EricSinsky-NOAA: I tried something very similar to this. The program ran less efficiently when this was a python function.

ush/ocnpost.py
EricSinsky-NOAA: Agreed. yaml files will be used instead of the preliminary parm files that are being used now.

ush/ocnpost.py
aerorahul: There are efficient bilinear interpolation methods in `numpy`.  Did you try any of those? 
If this is the meat and potatoes of this program, IMO, it makes sense for this to be a Fortran program.  The rest is just IO.

ush/ocnpost.py
EricSinsky-NOAA: I believe NETCDF3 was the format that was used in the ocean and ice post NCL programs. However, there is no particular reason we need to keep it this way. I'll change this to NETCDF4, or this can be a user-configurable option in a yaml file.

jiandewang: > I believe NETCDF3 was the format that was used in the ocean and ice post NCL programs. However, there is no particular reason we need to keep it this way. I'll change this to NETCDF4, or this can be a user-configurable option in a yaml file.

it is netcdf4

ush/ocnpost.py
EricSinsky-NOAA: I tried [esmpy.Regrid](https://earthsystemmodeling.org/esmpy_doc/release/ESMF_8_0_1/html/regrid.html), but this also turned out to be less efficient.

ush/ocnpost.py
EricSinsky-NOAA: Got it, thanks Jiande.

ush/ocnpost.py
aerorahul: I meant something from `numpy` or `scipy` e.g. https://numpy.org/devdocs/reference/generated/numpy.interp.html

EricSinsky-NOAA: To my understanding, `np.interp` does not have an option for a weighted bilinear interpolation. This Fortran subroutine, which uses the ESMF regridding weights in a bilinear or conservative interpolation is referenced [here](https://earthsystemmodeling.org/docs/release/ESMF_8_0_1/ESMF_refdoc/node3.html#SECTION03029300000000000000).

ush/ocnpost.py
HenryRWinterbottom: 1. From my experience, FORTRAN is going to be the most efficient way to do this. Further, there is no need to wrap the FORTRAN with `f2py`. If you have the weights already, you should only need to loop over the coefficients to do the remapping. Using Python only complicates things and makes the application less portable.

2. A simple unit test would be using a static ESMF coefficients file, remapping from 5p0 to a comparable Gaussian grid nominal resolution and taking the difference of the fields. I would expect the different to be on the order of 0 aside from rounding error.

3. This is the perfect scenario for which to use a YAML configuration. The configuration file you provide is difficult to interpret. A YAML variation would be much more readable and easier to digest.

ush/ocnpost.py
HenryRWinterbottom: I would personally avoid `numpy` and/or `scipy`. I have also tried these methods and they are slow and do not take into account issues such as landmask and fractional grids.

To reiterate, once you have the ESMF weights, all you need to do is add a short subroutine to read in the source grid variable and remap it to the destination grid. It should be short (on the order of 10 lines).

ush/ocnpost.py
HenryRWinterbottom: The program should be agnostic to the model. All it requires if the ESMF coefficients and a FORTRAN bit to remap the variables. ESMF is also agnostic to the model. All it cares about is mapping points from a source grid to a destination grid. The only scenario that I can think of in which you would need to specify the model is if you are generating a MOM6 or CICE tri-polar projection from scratch.

ush/ocnpost.py
HenryRWinterbottom: This is not necessary. Both MOM6 and CICE are on the same tri-polar grid for the UFS. Further, there is no rotation necessary for CICE since the variables are all Arakawa-C mass points.

ush/ocnpost.py
HenryRWinterbottom: netCDF3 is an artifact of NGGPS.

ush/ocnpost.py
HenryRWinterbottom, line 1: Anytime that you need to run over large loops in Python, consideration should be given to using a lower-level language (e.g., FORTRAN). Even with efficient use of Python list-comprehensions FORTRAN will still be a faster solution.

HenryRWinterbottom (commented):
I think this application can be leaned out considerably.

1. Use YAML files as noted throughout the comments.

2. Take the remapping step out. If you already have the ESMF remapping coefficient files, all you need is a FORTRAN program to read the source variable files, do the interpolation, and write the remapped variable out.

3. The MOM6 velocity vector Earth -> grid and grid-> Earth relative rotations can also be done more efficiently in FORTRAN.

4. As is, this application limits portability.

ush/ocnpost.py
EricSinsky-NOAA: @HenryWinterbottom-NOAA Thank you for your feedback. Just to confirm, is this the short subroutine that you are referring to? 

           ! Initialize destination field to 0.0
           do i=1,dst_len
             dst(i)=0.0
           enddo

           ! Apply weights
           do i=1, n_s
             dst(row(i))=dst(row(i))+S(i)*src(col(i))
           enddo
           
           do i=1, dst_len
             if (frac_b(i) .ne. 0.0) then
               dst(i)=dst(i)/frac_b(i)
             endif
           enddo

HenryRWinterbottom: @EricSinsky-NOAA Yes.

ush/ocnpost.py
JessicaMeixner-NOAA: Can we confirm with @DeniseWorthen who wrote the original NCL scripts.  While I don't remember specific details, I do know that Denise is an expert on the CICE grid and carefully confirmed every variable was being interpolated the correct way and I'm assuming Eric is following what was done in the original NCL scripts here. Although maybe somethings changed since then. 

PS Thank you @EricSinsky-NOAA for this PR.  I know this has been a long standing need (https://github.com/NOAA-EMC/global-workflow/issues/923 was created a year ago) and it's greatly appreciated.

EricSinsky-NOAA: Thanks @JessicaMeixner-NOAA. The vector rotation is being performed the same way as in Denise's NCL code. It is possible, however, that something has changed since then (as you pointed out).

PR #1739, opened 2023-07-14
Change JEDI jobs to grab fieldmetadata from GDASApp rather than glopara fix

CoryMartin-NOAA:
**Description**

I think out of convenience before, the glopara/fix/gdas directories contain several text files that do not need to be maintained by the glopara group, and can instead store them in GDASApp or global-workflow.

This PR is the first step towards that, by moving the FV3-JEDI IO fieldmetadata files into GDASApp. Subsequent work will need to determine if things like FV3 diag tables can be shared between FV3-JEDI and the UFS.

This is a complimentary PR to https://github.com/NOAA-EMC/GDASApp/pull/537

Before this can be merged, the above PR must be merged and this PR will need an updated GDASApp hash added in a commit.

**Type of change**

**How Has This Been Tested?**
- [x] GDASApp workflow-based ctests
- [ ] Cycled test on Hera
  
**Checklist**

### DIFF
--- a/parm/parm_gdas/aero_jedi_fix.yaml
+++ b/parm/parm_gdas/aero_jedi_fix.yaml
@@ -7,5 +7,5 @@ copy:
   - !ENV ${DATA}/fv3jedi/fmsmpp.nml
 - - !ENV ${FV3JEDI_FIX}/fv3jedi/fv3files/field_table_gfdl
   - !ENV ${DATA}/fv3jedi/field_table
-- - !ENV ${FV3JEDI_FIX}/fv3jedi/fieldmetadata/gfs-aerosol.yaml
-  - !ENV ${DATA}/fv3jedi/gfs-restart.yaml
+- - !ENV $(HOMEgfs)/sorc/gdas.cd/parm/io/fv3jedi_fieldmetadata_restart.yaml
+  - !ENV ${DATA}/fv3jedi/fv3jedi_fieldmetadata_restart.yaml
--- a/parm/parm_gdas/atm_jedi_fix.yaml
+++ b/parm/parm_gdas/atm_jedi_fix.yaml
@@ -4,4 +4,4 @@ copy:
 - [$(HOMEgfs)/fix/gdas/fv3jedi/fv3files/akbk$(npz).nc4, $(DATA)/fv3jedi/akbk.nc4]
 - [$(HOMEgfs)/fix/gdas/fv3jedi/fv3files/fmsmpp.nml, $(DATA)/fv3jedi/fmsmpp.nml]
 - [$(HOMEgfs)/fix/gdas/fv3jedi/fv3files/field_table_gfdl, $(DATA)/fv3jedi/field_table]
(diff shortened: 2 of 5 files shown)


parm/parm_gdas/aero_jedi_fix.yaml
```diff
@@ -7,5 +7,5 @@ copy:
   - !ENV ${DATA}/fv3jedi/fmsmpp.nml
 - - !ENV ${FV3JEDI_FIX}/fv3jedi/fv3files/field_table_gfdl
   - !ENV ${DATA}/fv3jedi/field_table
-- - !ENV ${FV3JEDI_FIX}/fv3jedi/fieldmetadata/gfs-aerosol.yaml
-  - !ENV ${DATA}/fv3jedi/gfs-restart.yaml
+- - !ENV $(HOMEgfs)/sorc/gdas.cd/parm/io/fv3jedi_fieldmetadata_restart.yaml
```
CoryMartin-NOAA, line 10: I'm guessing @aerorahul will want this to be instead linked at the link_workflow.sh stage and put somewhere else. Question is is that an 'eventually' thing or a 'now' thing?

aerorahul: Correct.  This should be linked to `${HOMEgfs}/parm/gdas/io/fv3jedi_fieldmetadata_restart.yaml` in `sorc/link_workflow.sh` at the time the workflow is being setup for vertical structure.

From this location, it can be copied to `DATA` at job runtime.

parm/parm_gdas/atm_jedi_fix.yaml
```diff
@@ -4,4 +4,4 @@ copy:
 - [$(HOMEgfs)/fix/gdas/fv3jedi/fv3files/akbk$(npz).nc4, $(DATA)/fv3jedi/akbk.nc4]
 - [$(HOMEgfs)/fix/gdas/fv3jedi/fv3files/fmsmpp.nml, $(DATA)/fv3jedi/fmsmpp.nml]
 - [$(HOMEgfs)/fix/gdas/fv3jedi/fv3files/field_table_gfdl, $(DATA)/fv3jedi/field_table]
-- [$(HOMEgfs)/fix/gdas/fv3jedi/fieldmetadata/gfs-restart.yaml, $(DATA)/fv3jedi/gfs-restart.yaml]
+- [$(HOMEgfs)/sorc/gdas.cd/parm/io/fv3jedi_fieldmetadata_restart.yaml, $(DATA)/fv3jedi/fv3jedi_fieldmetadata_restart.yaml]
```
aerorahul, line 7: same with this and others

parm/parm_gdas/aero_jedi_fix.yaml
CoryMartin-NOAA: Per offline convo, Rahul and or I will do this, just not in this PR.

aerorahul (approved):
looks good.
No need to test as there are no tests for GDASApp.

PR #1755, opened 2023-07-18
Support coupled GEFS forecast, use mem000 for GEFS control

### LINKED ISSUE #1754: Add coupled forecast support for GEFS
**Description**
GEFS will need to support coupled forecasts for GEFS v13.

**Requirements**
Coupled forecasts can run in GEFS mode 

Meets requirements without disrupting non-GEFS activity.

WalterKolczynski-NOAA:
**Description**

Makes the necessary updates to allow the workflow to run (non-wave) coupled forecasts. Some of the coupled functions in postdet had to be updated because they only checked for gfs and gdas runs, so necessary code was never executed.

Also updates so that the GEFS control uses `mem000` as the memdir so it appears at the same level as the perturbed members. Relatedly, the default member number in coupled files is changed to 000 instead of 01.

Finally, an unrelated error in the earc job is corrected where a character from the middle of the line was moved to the beginning.

Resolves #1754

**Type of change**

- [x] GEFS coupled forecast-only on Orion
- [ ] P8 coupled on Orion
- [x] Cycled on Orion
  
**Checklist**


WalterKolczynski-NOAA:
Low-res test pass. C384 fails for a reason that seems unrelated (I tried develop and it failed too).

parm/config/gefs/config.efcs
```diff
@@ -6,10 +6,10 @@
...
 
 # TODO: the _ENKF counterparts need to be defined in config.base
-export DO_AERO=${DO_AERO_ENKF:-"NO"}
-export DO_OCN=${DO_OCN_ENKF:-"NO"}
-export DO_ICE=${DO_ICE_ENKF:-"NO"}
-export DO_WAVE=${DO_WAVE_ENKF:-"NO"}
+# export DO_AERO=${DO_AERO_ENKF:-"NO"}
+# export DO_OCN=${DO_OCN_ENKF:-"NO"}
+# export DO_ICE=${DO_ICE_ENKF:-"NO"}
+# export DO_WAVE=${DO_WAVE_ENKF:-"NO"}
```
aerorahul: Just delete these options to inherit from `config.fcst`

ush/forecast_postdet.sh
```diff
@@ -736,11 +736,11 @@ MOM6_postdet() {
   [[ ! -d ${COM_OCEAN_HISTORY} ]] && mkdir -p "${COM_OCEAN_HISTORY}"
 
   # Link output files
-  if [[ "${RUN}" =~ "gfs" ]]; then
+  if [[ "${RUN}" =~ "gfs" || "${RUN}" =~ "gefs" ]]; then
     # Link output files for RUN = gfs
 
     # TODO: get requirements on what files need to be written out and what these dates here are and what they mean
-    export ENSMEM=${ENSMEM:-01}
+    export MEMNUM=${MEMNUM:-00}
```
aerorahul: elevate this to `function common_predet()` in `forecast_predet.sh` and remove from `function CICE_postdet()`
Also note that it is 2 digits here and 3 digits in `CICE_postdet` line 923.

aerorahul (changes requested):
first pass.

parm/config/gefs/config.efcs
WalterKolczynski-NOAA: Done.

ush/forecast_postdet.sh
WalterKolczynski-NOAA: Done (and three digits).

parm/config/gefs/config.base.emc.dyn
```diff
@@ -276,6 +276,11 @@ export DO_MERGENSST="NO"
 # Hybrid related
 export NMEM_ENS=@NMEM_ENS@
 
+# set default member number memdir for control
+# this will be overwritten for the perturbed members
+export ENSMEM="000"
+export MEMDIR="mem${ENSMEM}"
```
aerorahul, line 282: why is this needed in `config.base`?

parm/config/gefs/config.wave
```diff
@@ -0,0 +1,202 @@
+#! /usr/bin/env bash
```
aerorahul: Can we just link to `config.wave` from `parm/config/gfs`?  Is there any distinction for wave between gfs and gefs for waves?

scripts/exgdas_enkf_earc.sh
```diff
@@ -234,7 +234,7 @@ if [[ "${ENSGRP}" -eq 0 ]]; then
                         # shellcheck disable=SC2312
                         templates=$(compgen -A variable | grep 'COM_OCEAN_.*_TMPL')
                         for template in ${templates}; do
-                            YMEMDIR="${mem}" MD="${gPDY}" HH="${gcyc}" generate_com "directory:${template}"
+                            MEMDIR="${mem}" YMD="${gPDY}" HH="${gcyc}" generate_com "directory:${template}"
```
aerorahul, line 237: oops!

aerorahul (changes requested):
looks better.
A couple of questions:
1. see the review
2. can we just link the `config.XYZ` files in `parm/config/gefs` from `parm/config/gfs` that are not different?  I don't imagine things like `config.ocn`, `config.ice`, `config.ufs`, `config.wave` to be different. These are setting the various properties of the model for a choice of resolution and `wavgrd` set in `config.base`.

scripts/exgdas_enkf_earc.sh
WalterKolczynski-NOAA: Yeah, found this one completely randomly while I was looking for instances of `MEMDIR`.

parm/config/gefs/config.base.emc.dyn
WalterKolczynski-NOAA: Usually we only set `MEMDIR` when calling `generate_com()` (when needed). For the GFS deterministic, that means `MEMDIR` is just an empty string and no member level is created in COM. But for GEFS, we want the control to also have a memdir; otherwise the tree looks like this:

```
gefs.20211221/00/
- model_data
- mem001
  - model_data
- mem002
  - model_data
```

So this allows us to set a blanket default `MEMDIR` that will be used for the control member of GEFS without needing to otherwise modify the code and having to account for whether we are running GFS or GEFS at the j-job level. And anywhere it was already being set (for other members or ensstat) still apply and override this setting.

parm/config/gefs/config.wave
WalterKolczynski-NOAA: Not sure if we will need to use different settings for GEFS. Can certain do that (and for aero/ice/ocn) until such time as we find we need to have different settings. Will leave it up to you.

parm/config/gefs/config.base.emc.dyn
aerorahul: Not sure I understand.
`generate_com` is called in `exgdas_enkf_fcst.sh` and is looping over `ENSBEG ENSEND`.
When they are both 0 (control member), wouldn't those ensure that the appropriate directory paths are constructed.

We can discuss offline.

parm/config/gefs/config.wave
aerorahul: I would prefer to link in `gefs` to force them to be the same as gfs and really understand the requirements when they are presented why they need to be different.  Otherwise there will be a tendency to just cram gefs specifics into this application.
I am not against the flexibility, but I would like to understand the reason and justification when that is provided.
This does not prevent us from maintaining a separate configuration for GEFS in the future.

ush/forecast_predet.sh
```diff
@@ -52,6 +52,7 @@ common_predet(){
   CDUMP=${CDUMP:-gdas}
   CASE=${CASE:-C768}
   CDATE=${CDATE:-2017032500}
+  ENSMEM=${ENSMEM:-000}
```
aerorahul, line 55: I think this should be blank.
I need to inspect what is being doing in `forecast_postdet.sh` and raise the question why is this even relevant in the GFS application.  This seems like an remnant from P8 development where the output was hard-wired.
```suggestion
  ENSMEM=${ENSMEM:-000}
```
```suggestion
  ENSMEM=${ENSMEM:-}
```

parm/config/gefs/config.wave
WalterKolczynski-NOAA: Done.

parm/config/gefs/config.base.emc.dyn
WalterKolczynski-NOAA: The control member is run like the deterministic GFS and never calls exgdas_enkf_fcst.sh.

parm/config/gefs/config.defaults.s2sw
```diff
@@ -0,0 +1,29 @@
+#! /usr/bin/env bash
```
aerorahul: This can also be linked from `parm/config/gfs`

parm/config/gefs/config.base.emc.dyn
aerorahul: Ok.  Following our offline discussion, we can proceed with this for now.

aerorahul (approved):
Looks good.
There is one comment on linking `config.defaults.s2sw` from `gfs` into `gefs`.
I'll leave it to you if you want to address it here or later or not.

WalterKolczynski-NOAA:
> Looks good. There is one comment on linking `config.defaults.s2sw` from `gfs` into `gefs`. I'll leave it to you if you want to address it here or later or not.

Took care of this one.

PR #1790, opened 2023-08-11
Fix f-o mode not sourcing postsnd when soundings are on

### LINKED ISSUE #1431: Sounding task in job mesh but config not loaded in forecast-only
**Expected behavior**
The bufr sounding job (postsnd) should either not be in the task list in forecast-only mode even if `DO_BUFRSND` is set to `"YES"`, or the `config.postsnd` should be sourced during XML generation

**Current behavior**
The postsnd task is added to the task list as long as `DO_BUFRSND` is `"YES"` no matter which forecast mode is used, but the config file is not loaded by AppConfig for forecast-only jobs.

**Machines affected**
All

**To Reproduce**
1. Setup a forecast-only experiment with `workflow/setup_expt.py`
2. Change `DO_BUFRSND` to `"YES"` in `config.base`
3. Run `workflow/setup_xml.py`

**Context**
This isn't a pressing issue, as soundings are off by default, so most people don't make them, and making them in forecast-only mode is even rarer. However, doing so should not cause an exception in XML generation.

**Detailed Description**

**Additional Information**

**Possible Implementation**
Either:
- Add postsnd to the config list for forecast-only as it is in cycled mode
- Only add postsnd to the task list for cycled mode

WalterKolczynski-NOAA:
**Description**

Soundings are allowed in forecast-only mode, but the config file would not be sourced during experiment creation. Now the config file is sourced if bufr soundings (`DO_BUFRSND`) are on.

Resolves #1431

**Type of change**

**How Has This Been Tested?**

- [x] Forecast-only experiment creation with `DO_BUFRSND=YES` on Orion
  
**Checklist**

### DIFF
--- a/workflow/applications/gfs_forecast_only.py
+++ b/workflow/applications/gfs_forecast_only.py
@@ -29,6 +29,9 @@ class GFSForecastOnlyAppConfig(AppConfig):
         if self.do_atm and self.do_metp:
             configs += ['metp']
 
+        if self.do_bufrsnd:
+            configs += ['postsnd']
+
         if self.do_gempak:
             configs += ['gempak']
 


HenryRWinterbottom (approved):
Seems simple and all CI/CD tests passed.

Approved.

PR #1797, opened 2023-08-12
Copy land increments and land DA confs to COM

jiaruidong2017:
`The NCEP 'dump' convention is the opposite of the JEDI IODA convention. For example, NCEP is [t03z-t09z), whereas IODA is (t03z-t09z], so the observations right at 03:00 and 09:00 will be missing (https://github.com/JCSDA-internal/ufo/issues/2233). This PR solved this issue by shifting the DA window_begin 1 second backward. `
**The above DA window manipulation is being handled somewhere else.**

This PR created `COM_CONF_TMPL`, copied the generated YAML file from initialize to the COM and copied the increment files to the COM. 

**Checklist**

### DIFF
--- a/ush/python/pygfs/task/land_analysis.py
+++ b/ush/python/pygfs/task/land_analysis.py
@@ -34,7 +34,7 @@ class LandAnalysis(Analysis):
 
         _res = int(self.config['CASE'][1:])
         _window_begin = add_to_datetime(self.runtime_config.current_cycle, -to_timedelta(f"{self.config['assim_freq']}H") / 2)
-        _window_begin = add_to_datetime(_window_begin, -to_timedelta(f"2S"))
+        _window_begin = add_to_datetime(_window_begin, -to_timedelta(f"1S"))
         _letkfoi_yaml = os.path.join(self.runtime_config.DATA, f"{self.runtime_config.RUN}.t{self.runtime_config['cyc']:02d}z.letkfoi.yaml")
 
         # Create a local dictionary that is repeatedly used across this class
@@ -45,7 +45,7 @@ class LandAnalysis(Analysis):
                 'npz_ges': self.config.LEVS - 1,
                 'npz': self.config.LEVS - 1,
                 'LAND_WINDOW_BEGIN': _window_begin,
-                'LAND_WINDOW_LENGTH': f"PT{self.config['assim_freq']}H1S",
+                'LAND_WINDOW_LENGTH': f"PT{self.config['assim_freq']}H2S",
                 'OPREFIX': f"{self.runtime_config.RUN}.t{self.runtime_config.cyc:02d}z.",
                 'APREFIX': f"{self.runtime_config.RUN}.t{self.runtime_config.cyc:02d}z.",
                 'jedi_yaml': _letkfoi_yaml
--- a/ush/python/pygfs/task/land_analysis.py
+++ b/ush/python/pygfs/task/land_analysis.py
@@ -45,7 +45,7 @@ class LandAnalysis(Analysis):
                 'npz_ges': self.config.LEVS - 1,
                 'npz': self.config.LEVS - 1,
                 'LAND_WINDOW_BEGIN': _window_begin,
-                'LAND_WINDOW_LENGTH': f"PT{self.config['assim_freq']}H2S",
+                'LAND_WINDOW_LENGTH': f"PT{self.config['assim_freq']}H1S",
                 'OPREFIX': f"{self.runtime_config.RUN}.t{self.runtime_config.cyc:02d}z.",
                 'APREFIX': f"{self.runtime_config.RUN}.t{self.runtime_config.cyc:02d}z.",
                 'jedi_yaml': _letkfoi_yaml
--- a/ush/python/pygfs/task/land_analysis.py
+++ b/ush/python/pygfs/task/land_analysis.py
@@ -34,6 +34,7 @@ class LandAnalysis(Analysis):
 
         _res = int(self.config['CASE'][1:])
         _window_begin = add_to_datetime(self.runtime_config.current_cycle, -to_timedelta(f"{self.config['assim_freq']}H") / 2)
+        _window_begin = add_to_datetime(_window_begin, -to_timedelta(f"1S") )
         _letkfoi_yaml = os.path.join(self.runtime_config.DATA, f"{self.runtime_config.RUN}.t{self.runtime_config['cyc']:02d}z.letkfoi.yaml")
 
         # Create a local dictionary that is repeatedly used across this class
@@ -323,8 +324,10 @@ class LandAnalysis(Analysis):
     def finalize(self) -> None:
         """Performs closing actions of the Land analysis task
         This method:
-        - copies analysis back to COM/ from DATA/
-        - tar and gzips the JEDI diagnostic files
+        - tarring up output diag files and place in COM/
+        - copying the generated YAML file from initialize to the COM/
+        - copying the analysis files to the COM/
+        - copying the increment files to the COM/
 
         Parameters
         ----------
@@ -336,6 +339,15 @@ class LandAnalysis(Analysis):
         statfile = os.path.join(self.task_config.COM_LAND_ANALYSIS, f"{self.task_config.APREFIX}landstat.tgz")
         self.tgz_diags(statfile, self.task_config.DATA)
 
+        logger.info("Copy full YAML  to COM")
+        src = os.path.join(self.task_config['DATA'], f"{self.task_config['CDUMP']}.t{self.runtime_config['cyc']:02d}z.letkfoi.yaml")
+        dest = os.path.join(self.task_config.COM_LAND_ANALYSIS, f"{self.task_config['CDUMP']}.t{self.runtime_config['cyc']:02d}z.letkfoi.yaml")
+        yaml_copy = {
+            'mkdir': [self.task_config.COM_LAND_ANALYSIS],
+            'copy': [[src, dest]]
+        }
+        FileHandler(yaml_copy).sync()
+
         logger.info("Copy analysis to COM")
         template = f'{to_fv3time(self.task_config.current_cycle)}.sfc_data.tile{{tilenum}}.nc'
         anllist = []
@@ -345,6 +357,16 @@ class LandAnalysis(Analysis):
             dest = os.path.join(self.task_config.COM_LAND_ANALYSIS, filename)
             anllist.append([src, dest])
         FileHandler({'copy': anllist}).sync()
+
+        logger.info('Copy increments to COM')
+        template = f'landinc.{to_fv3time(self.task_config.current_cycle)}.sfc_data.tile{{tilenum}}.nc'
+        inclist = []
+        for itile in range(1, self.task_config.ntiles + 1):
+            filename = template.format(tilenum=itile)
+            src = os.path.join(self.task_config.DATA, 'anl', filename)
+            dest = os.path.join(self.task_config.COM_LAND_ANALYSIS, filename)
+            inclist.append([src, dest])
+        FileHandler({'copy': inclist}).sync()
 
     @staticmethod
     @logit(logger)
--- a/ush/python/pygfs/task/land_analysis.py
+++ b/ush/python/pygfs/task/land_analysis.py
@@ -34,7 +34,7 @@ class LandAnalysis(Analysis):
 
         _res = int(self.config['CASE'][1:])
         _window_begin = add_to_datetime(self.runtime_config.current_cycle, -to_timedelta(f"{self.config['assim_freq']}H") / 2)
-        _window_begin = add_to_datetime(_window_begin, -to_timedelta(f"1S") )
+        _window_begin = add_to_datetime(_window_begin, -to_timedelta(f"1S"))
         _letkfoi_yaml = os.path.join(self.runtime_config.DATA, f"{self.runtime_config.RUN}.t{self.runtime_config['cyc']:02d}z.letkfoi.yaml")
 
         # Create a local dictionary that is repeatedly used across this class
--- a/ush/python/pygfs/task/land_analysis.py
+++ b/ush/python/pygfs/task/land_analysis.py
@@ -340,8 +340,8 @@ class LandAnalysis(Analysis):
         self.tgz_diags(statfile, self.task_config.DATA)
 
         logger.info("Copy full YAML to COM")
-        src = os.path.join(self.task_config['DATA'], f"{self.task_config['CDUMP']}.t{self.runtime_config['cyc']:02d}z.letkfoi.yaml")
-        dest = os.path.join(self.task_config.COM_CONF, f"{self.task_config['CDUMP']}.t{self.runtime_config['cyc']:02d}z.letkfoi.yaml")
+        src = os.path.join(self.task_config['DATA'], f"{self.task_config.APREFIX}letkfoi.yaml")
+        dest = os.path.join(self.task_config.COM_CONF, f"{self.task_config.APREFIX}letkfoi.yaml")
         yaml_copy = {
             'mkdir': [self.task_config.COM_CONF],
             'copy': [[src, dest]]
--- a/ush/python/pygfs/task/land_analysis.py
+++ b/ush/python/pygfs/task/land_analysis.py
@@ -34,7 +34,7 @@ class LandAnalysis(Analysis):
 
         _res = int(self.config['CASE'][1:])
         _window_begin = add_to_datetime(self.runtime_config.current_cycle, -to_timedelta(f"{self.config['assim_freq']}H") / 2)
-        _window_begin = add_to_datetime(_window_begin, -to_timedelta(f"1S"))
+        _window_begin = add_to_datetime(_window_begin, -to_timedelta(f"2S"))
         _letkfoi_yaml = os.path.join(self.runtime_config.DATA, f"{self.runtime_config.RUN}.t{self.runtime_config['cyc']:02d}z.letkfoi.yaml")
 
         # Create a local dictionary that is repeatedly used across this class
--- a/ush/python/pygfs/task/land_analysis.py
+++ b/ush/python/pygfs/task/land_analysis.py
@@ -34,7 +34,6 @@ class LandAnalysis(Analysis):
 
         _res = int(self.config['CASE'][1:])
         _window_begin = add_to_datetime(self.runtime_config.current_cycle, -to_timedelta(f"{self.config['assim_freq']}H") / 2)
-        _window_begin = add_to_datetime(_window_begin, -to_timedelta(f"1S"))
         _letkfoi_yaml = os.path.join(self.runtime_config.DATA, f"{self.runtime_config.RUN}.t{self.runtime_config['cyc']:02d}z.letkfoi.yaml")
 
         # Create a local dictionary that is repeatedly used across this class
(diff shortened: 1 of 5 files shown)


CoryMartin-NOAA:
@jiaruidong2017 did you test this? I was under the impression that for many cases the background *has* to be at the center of the window... I forget how the MetOffice gets around this issue, but they do something.

jiaruidong2017:
@CoryMartin-NOAA Yes, I did the test runs in the global-workflow, and it worked perfectly. The updated window setup did pick up the lost observations. 

As I remembered, the MetOffice used the same approach by moving the DA window backward 1 second. What they did is to do this shift in the yaml file. However, their approach only works in the HofX, but not in the letkf.x.

CoryMartin-NOAA:
@jiaruidong2017 ok thanks, just wanted to double check.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, config):
 
         _res = int(self.config['CASE'][1:])
         _window_begin = add_to_datetime(self.runtime_config.current_cycle, -to_timedelta(f"{self.config['assim_freq']}H") / 2)
+        _window_begin = add_to_datetime(_window_begin, -to_timedelta(f"1S"))
```
CoryMartin-NOAA: to do this "correctly, now don't we need to make the window_length 6hours + 1 second?

ush/python/pygfs/task/land_analysis.py
```diff
@@ -336,6 +339,15 @@ def finalize(self) -> None:
         statfile = os.path.join(self.task_config.COM_LAND_ANALYSIS, f"{self.task_config.APREFIX}landstat.tgz")
         self.tgz_diags(statfile, self.task_config.DATA)
 
+        logger.info("Copy full YAML  to COM")
```
CoryMartin-NOAA: too many spaces

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: @CoryMartin-NOAA I conducted and compared two runs: one with window length 6 hours and the other with 6 hours + 1 second. They do show some difference from the land increment files as below: 

```
Tile 1: 
Variable Group Count        Sum     AbsSum          Min        Max       Range        Mean      StdDev
snodl    /       432 0.00534215 0.00534219 -4.60635e-09 0.00056797 0.000567975 1.23661e-05 6.58902e-05

Tile 2: 
Variable Group Count       Sum  AbsSum      Min     Max   Range         Mean   StdDev
snodl    /      1213 -0.570553 65.0728 -3.67096 3.36201 7.03297 -0.000470365 0.306991

Tile 3: 
Variable Group Count Sum AbsSum     Min     Max   Range Mean StdDev
snodl    /       982 nan    nan -94.825 98.1387 192.964  nan    nan
vtype    /       105 nan    nan     inf    -inf    -inf  nan    nan
slmsk    /       105 nan    nan     inf    -inf    -inf  nan    nan

Tile 4, 5, 6 are identical. 
```
Let's talk at tomorrow's meeting.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: Removed the space

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: @CoryMartin-NOAA @barlage When I conducted the snow DA cycling run with window length `6 hours + 2 seconds` which make the DA window centering at the cycle time, all the `nan` values were gone. Therefore, the DA window centering at the cycle time is important.

jiaruidong2017: CORRECTIONS: 

My above statements are incorrect, because there are still `nan` values when the above setting with the DA window beginning at 1 second back and window length 6 hours + 2 seconds. 

The above with no nan values came from the experiment with the DA window beginning at 2 second back and window length 6 hours + 1 seconds.

barlage: Since @CoryMartin-NOAA questioned in the meeting earlier, we looked at an increment file and not all of the grids are NaN, only about 100-200 and they look to be in locations where there are snow observations, which can seen from the above nccmp. Also, there must be some increment limits in the add_increment code since these do not affect the analysis.

barlage: @jiaruidong2017 after a quick look, I'm not seeing any code that limits the increment, so I'm not understanding why the analysis doesn't have bad values.

jiaruidong2017: Thanks @barlage for checking the add_increment  code. I am thinking both `slmsk` and `vtype` were listed at state variables in the yaml file as `state variables: [snodl,vtype,slmsk]`. I am guessing there may be some changes made to them at somewhere in JEDI from 0/1 (water/land) value to the missing value (ice?). I need to conduct more tests for this. Please let me know if you have any suggestions.

jiaruidong2017: @CoryMartin-NOAA @barlage The `nan` values from the previous runs result from the missing bugfix in JEDI fv3-bundle (https://github.com/JCSDA-internal/fv3-jedi/pull/980) when building the GDASApp with `feature/stable-nightly` branch. 

After I separately rebuilt the GDASApp with `develop` branch including the bugfix from JEDI fv3-bundle, the following experiments were conducted including: 
**(1) 1S0S:** 
window begin: '2021-10-31T20:59:59Z'
window end: '2021-11-01T02:59:59Z'
**(2) 1S1S**
window begin: '2021-10-31T20:59:59Z'
window end: '2021-11-01T03:00:00Z'
**(3) 1S2S** (centered at `2021-11-01T00:00:00Z`)
window begin: '2021-10-31T20:59:59Z'
window end: '2021-11-01T03:00:01Z'

```
Files "rundir/landanl.2021110100_1S2S/anl/landinc.20211101.000000.sfc_data.tile1.nc" and "rundir/landanl.2021110100_1S1S/anl/landinc.20211101.000000.sfc_data.tile1.nc" are identical.
Files "rundir/landanl.2021110100_1S2S/anl/landinc.20211101.000000.sfc_data.tile2.nc" and "rundir/landanl.2021110100_1S1S/anl/landinc.20211101.000000.sfc_data.tile2.nc" are identical.
Files "rundir/landanl.2021110100_1S2S/anl/landinc.20211101.000000.sfc_data.tile3.nc" and "rundir/landanl.2021110100_1S1S/anl/landinc.20211101.000000.sfc_data.tile3.nc" are identical.
Files "rundir/landanl.2021110100_1S2S/anl/landinc.20211101.000000.sfc_data.tile4.nc" and "rundir/landanl.2021110100_1S1S/anl/landinc.20211101.000000.sfc_data.tile4.nc" are identical.
Files "rundir/landanl.2021110100_1S2S/anl/landinc.20211101.000000.sfc_data.tile5.nc" and "rundir/landanl.2021110100_1S1S/anl/landinc.20211101.000000.sfc_data.tile5.nc" are identical.
Files "rundir/landanl.2021110100_1S2S/anl/landinc.20211101.000000.sfc_data.tile6.nc" and "rundir/landanl.2021110100_1S1S/anl/landinc.20211101.000000.sfc_data.tile6.nc" are identical.
Files "rundir/landanl.2021110100_1S2S/anl/landinc.20211101.000000.sfc_data.tile1.nc" and "rundir/landanl.2021110100_1S0S/anl/landinc.20211101.000000.sfc_data.tile1.nc" are identical.
Files "rundir/landanl.2021110100_1S2S/anl/landinc.20211101.000000.sfc_data.tile2.nc" and "rundir/landanl.2021110100_1S0S/anl/landinc.20211101.000000.sfc_data.tile2.nc" are identical.
Files "rundir/landanl.2021110100_1S2S/anl/landinc.20211101.000000.sfc_data.tile3.nc" and "rundir/landanl.2021110100_1S0S/anl/landinc.20211101.000000.sfc_data.tile3.nc" are identical.
Files "rundir/landanl.2021110100_1S2S/anl/landinc.20211101.000000.sfc_data.tile4.nc" and "rundir/landanl.2021110100_1S0S/anl/landinc.20211101.000000.sfc_data.tile4.nc" are identical.
Files "rundir/landanl.2021110100_1S2S/anl/landinc.20211101.000000.sfc_data.tile5.nc" and "rundir/landanl.2021110100_1S0S/anl/landinc.20211101.000000.sfc_data.tile5.nc" are identical.
Files "rundir/landanl.2021110100_1S2S/anl/landinc.20211101.000000.sfc_data.tile6.nc" and "rundir/landanl.2021110100_1S0S/anl/landinc.20211101.000000.sfc_data.tile6.nc" are identical.
```
All three experiments have the identical increment values regardless the setup of the window end.

CoryMartin-NOAA: ah! great! We have a new feature/stable-nightly commit overnight, can we use that commit hash and add it in for this PR?

jiaruidong2017: This means we can freely setup the window length to either `6 hours`, or `6 hours +1 second`, or `6 hours + 2 seconds`, and the increment values are identical.

jiaruidong2017: Thanks Cory, and I will update the commit hash and include it in this PR.

CoryMartin-NOAA: In that case then I think the 6 hr 1 sec window is the one that would be the 'best' in this scenario, since it would include all obs in a NCEP dump.

jiaruidong2017: @CoryMartin-NOAA Okay, I will change the window length to 6hr 1sec.

CoryMartin-NOAA (approved):

ush/python/pygfs/task/land_analysis.py
aerorahul: I don't understand, if the results are the same for 6hr, 6hr-1s, 6hr+1s, then why make the change?
Also, shouldn't this be handled in the observation processing part rather than in the workflow?
Consider the case where the observations are in a single file for 24h (for e.g.), shouldn't this be teased out in the yaml?

CoryMartin-NOAA: @aerorahul this is because the 'JEDI' window convention is the opposite of the 'NCEP' window convention, and we lost that battle with Yannick a while ago. While it may be identical in this case, it shouldn't be. Observations at exactly 03z would not make it into the 06z analysis without this change. I don't quite understand you comment in the second sentence, this is what defines the window variable that goes into the templated YAML.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -336,6 +339,15 @@ def finalize(self) -> None:
...
         self.tgz_diags(statfile, self.task_config.DATA)
 
+        logger.info("Copy full YAML to COM")
+        src = os.path.join(self.task_config['DATA'], f"{self.task_config['CDUMP']}.t{self.runtime_config['cyc']:02d}z.letkfoi.yaml")
+        dest = os.path.join(self.task_config.COM_LAND_ANALYSIS, f"{self.task_config['CDUMP']}.t{self.runtime_config['cyc']:02d}z.letkfoi.yaml")
+        yaml_copy = {
+            'mkdir': [self.task_config.COM_LAND_ANALYSIS],
+            'copy': [[src, dest]]
+        }
+        FileHandler(yaml_copy).sync()
```
aerorahul: Why? 
`COM` is for `data` not configuration.  The UFS-weather-model has set a precedent for copying namelists and configuration stuff into `COM`.  If there is a need to copy this data for every cycle, we need a proper solution with a well defined set of requirements.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -346,6 +358,16 @@ def finalize(self) -> None:
             anllist.append([src, dest])
         FileHandler({'copy': anllist}).sync()
 
+        logger.info('Copy increments to COM')
+        template = f'landinc.{to_fv3time(self.task_config.current_cycle)}.sfc_data.tile{{tilenum}}.nc'
```
aerorahul, line 361: Can we not alter the format of the file?
i.e. can we retain;
`YYYYMMDD.HH0000.landinc.tile{{tilenum}}.nc`
It is clear then, that this file goes with `YYYYMMDD.HH0000.sfc_data.tile{{tilenum}}.nc`

aerorahul (changes requested):
Please see comments regarding observation cutoff, copying configuration data and increments to COM.

ush/python/pygfs/task/land_analysis.py
CoryMartin-NOAA: Why not? This is a few kb file that is a record of the exact configuration needed to reproduce results. Especially in the land case, this is different at 18z than it is at other cycles. When observations and QC have to be turned on/off these YAMLs will not always be identical and should be retained.

ush/python/pygfs/task/land_analysis.py
CoryMartin-NOAA: seems reasonable

ush/python/pygfs/task/land_analysis.py
aerorahul: I understand.  Lets discuss offline.

jiaruidong2017: Thanks @CoryMartin-NOAA for the detailed explanations. 
@aerorahul Without this change, about several hundreds of observations will be lost in each cycle. For example, the observations increased from 1718 to 2161 with this changes.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: Thanks @CoryMartin-NOAA. In addition, I include copying the YAML files for archive to keep consistent with other DA component such as aero DA.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: I will work to change the format of the increment files.

jiaruidong2017: I checked other JEDI-based DA components such as `atm DA` and `aero DA`, and the formats of their increment files are as `atminc.YYYYMMDD.HH0000.tile{{tilenum}}.nc` and `aeroinc.YYYYMMDD.HH0000.tile{{tilenum}}.nc`. @aerorahul @CoryMartin-NOAA Do I still need to make changes to `YYYYMMDD.HH0000.landinc.tile{{tilenum}}.nc` or keep consistent with other components?

aerorahul: Thanks @jiaruidong2017 
I take my comment back. We can retain the current convention and move on.

ush/python/pygfs/task/land_analysis.py
aerorahul: An update to this comment:

After discussions w/ NCO, if we wish to retain cycle specific configuration artifacts such as namelists, etc. that are not already part of `HOMEgfs`, we can do so in a single location in `COM`.  @CoryMartin-NOAA and @aerorahul discussed a good place for said configuration artifacts would be `ROTDIR/RUN.PDY/cyc/conf`.
The jobs will need to ensure that they identify the artifacts uniquely; for e.g. `input.nml` from the ufs-weather-model should be saved as `ufs_input.nml` 

This will require creating a `COM_CONF_TMPL` in `config.com` and then filling the template with appropriate substitutions in the j-jobs that wish to save the artifacts.

Please let me know if you have any questions.
@WalterKolczynski-NOAA FYI

jiaruidong2017: Thanks @aerorahul for the updates. 
@CoryMartin-NOAA @aerorahul Do you want me to include this change in this PR by putting the yaml file into the `ROTDIR/RUN.PDY/cyc/conf`?

jiaruidong2017: As suggested, I create `COM_CONF_TMPL` in `config.com` and copy the YAML file into `ROTDIR/RUN.PDY/cyc/conf`. I still keep using the previous name (e.g., gdas.t18z.letkfoi.yaml). What do you suggest the change for the file name?

ush/python/pygfs/task/land_analysis.py
```diff
@@ -336,6 +339,15 @@ def finalize(self) -> None:
         statfile = os.path.join(self.task_config.COM_LAND_ANALYSIS, f"{self.task_config.APREFIX}landstat.tgz")
         self.tgz_diags(statfile, self.task_config.DATA)
 
+        logger.info("Copy full YAML to COM")
+        src = os.path.join(self.task_config['DATA'], f"{self.task_config['CDUMP']}.t{self.runtime_config['cyc']:02d}z.letkfoi.yaml")
```
aerorahul:
```suggestion
        src = os.path.join(self.task_config['DATA'], f"{self.task_config['RUN']}.t{self.runtime_config['cyc']:02d}z.letkfoi.yaml")
```
Please don't use `CDUMP` anymore.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -336,6 +339,15 @@ def finalize(self) -> None:
...
+        dest = os.path.join(self.task_config.COM_CONF, f"{self.task_config['CDUMP']}.t{self.runtime_config['cyc']:02d}z.letkfoi.yaml")
```
aerorahul:
```suggestion
        dest = os.path.join(self.task_config.COM_CONF, f"{self.task_config['RUN']}.t{self.runtime_config['cyc']:02d}z.letkfoi.yaml")
```

ush/python/pygfs/task/land_analysis.py
aerorahul: Also, isn't this already defined as `self.task_config.APREFIX`?   Why do we need to reconstruct that again here?

ush/python/pygfs/task/land_analysis.py
aerorahul: Also, isn't this already defined as `self.task_config.APREFIX`?   Why do we need to reconstruct that again here?

aerorahul (changes requested):
A quick review on the yaml copying.  Thanks for including the changes in `config.com`.

I am working with @CoryMartin-NOAA on the window specification, but unfortunately I have little availability at the moment to work on this new requirement.  I'll get back to it ASAP.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: Thanks @aerorahul for quick review. Good catch, and I made the changes as suggested.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: Made the changes as suggested.

aerorahul:
@jiaruidong2017 
Can I suggest you remove this +/- 1sec business from this PR and then we can merge this PR.
Meanwhile, I will work on a better solution for the window manipulation in the yaml rather than hard-wiring it in the python code.
Please let me know when you have this change.
Thanks

jiaruidong2017:
> @jiaruidong2017 Can I suggest you remove this +/- 1sec business from this PR and then we can merge this PR. Meanwhile, I will work on a better solution for the window manipulation in the yaml rather than hard-wiring it in the python code. Please let me know when you have this change. Thanks

@aerorahul I removed the changes for the window manipulations. Thanks for your suggestions and your efforts.

aerorahul (approved):
Looks good. 
Thank you for accepting the suggestions. 
I will get the solution for windows in ASAP. 
I will reach out to get a setup to test the changes.

aerorahul:
@WalterKolczynski-NOAA 
The changes in this PR are not covered in the tests. So I don't think there is a reason to run the CI on this PR. 
This can be merged as is.

WalterKolczynski-NOAA (approved):

WalterKolczynski-NOAA:
Do the title/description need to be updated now that the window is being handled elsewhere?

jiaruidong2017:
> Do the title/description need to be updated now that the window is being handled elsewhere?

@WalterKolczynski-NOAA I updated the title and descriptions for your review. I prefer to keep the descriptions of the DA window manipulation there for future references.

PR #575, opened 2022-01-13
Update ecf scripts for memory and exclusivity; update release notes

KateFriedman-NOAA:
This PR includes the following updates:

* update ecf scripts for jobs that will run shared to include a memory request
* add the "excl" tag for the gfs_atmos_gempak and gfs_forecast jobs
* update release notes instructions to include step for linking ecf scripts

Refs: #399

PR #1801, opened 2023-08-16
Create and populate ensemble directory for UFS-based ATM DA

### LINKED ISSUE #1799: Add hybrid option to UFS-DA ATM variational analysis
**Description**
GSI based atmospheric DA can augment static B with an ensemble to produce a hybrid-variational (hybvar) analysis.  The same capability should exist for variational atmospheric UFS-DA.   Attempts to execute UFS-DA in hybvar mode failed.   The `gdasatmanlinit` job does not populate the run directory with ensemble members.   This option needs to be added to atmospheric analysis scripts in `ush/python/pygfs/task`.  

**Requirements**
Extend atmospheric analysis python scripts to include the option to copy specified ensemble members to the run directory when `gdasatmanlinit` is run.  Additional changes may also be needed to correctly construct a yaml file to run `fv3jedi_var.x` in hybvar mode.

`gdasatmanlrun` can run atmospheric UFS-DA in 3dvar (static-B only) or hybvar (static-B plus ensemble) mode.

RussTreadon-NOAA:
**Description**
This PR adds scripting to stage ensemble files for use in hybrid variational and ensemble UFS-based atmospheric DA.   Three files are modified:
- `ush/python/pygfs/task/analysis.py` - add method `get_ens_dict` to construct dictionary of ensemble member RESTART files to copy
- `ush/python/pygfs/task/atm_analysis.py` - invoke `get_ens_dict` to stage ensemble members in UFS-based ATM variational DA runtime directory 
- `ush/python/pygfs/task/atmens_analysis.py` - invoke `get_ens_dict` to stage ensemble members in UFS-based ATM ensemble DA runtime directory

Fixes #1799

**Type of change**

**How Has This Been Tested?**

- [x] Cycled test on Hera

Clone and install g-w `feature/ufsda_hybvar` on Hera.  Configure `EXPDIR` to run UFS-based atmospheric DA using hybrid background error.   Run `gdasatmanlinit`, `gdasatmanlrun`, and `gdasatmanlfinal` jobs for 2021081418.   Confirm that `init` job populated run directory with correct ensemble member restarts.   Subsequent `run` job successfully ran to completion as did `final` job.

Repeat the above steps for `enkfgdasatmensanlinit`, `enkfgdasatmensanlrun`, and `enkfgdasatmensanlfinal`.  Confirm that modified scripts work as intended.
  
**Checklist**


ush/python/pygfs/task/atm_analysis.py
```diff
@@ -394,6 +400,67 @@ def _get_berror_dict_gsibec(config: Dict[str, Any]) -> Dict[str, List[str]]:
         }
         return berror_dict
 
+    @logit(logger)
+    def get_berror_ens_dict(self, config: Dict[str, Any]) -> Dict[str, List[str]]:
```
aerorahul: https://github.com/RussTreadon-NOAA/global-workflow/blob/8e7d56568d2308fd091d2ee6f3b4f70f7ccf7fd6/ush/python/pygfs/task/atmens_analysis.py#L291

aerorahul: Need to find a way to not repeat large sections of the code.

RussTreadon-NOAA: I do not understand your comment.   A new method was added to `atm_analysis.py`.   It uses templates to create and populate the `ens` directory if `DOHYBVAR=YES`.   It does not repeat other sections of `atm_analysis.py`.  If you have a preferred way to implement the new method, I'm willing to learn.

jobs/JGLOBAL_ATM_ANALYSIS_INITIALIZE
```diff
@@ -27,8 +27,8 @@
     COM_ATMOS_HISTORY_PREV:COM_ATMOS_HISTORY_TMPL \
     COM_ATMOS_RESTART_PREV:COM_ATMOS_RESTART_TMPL
 
-MEMDIR='ensstat' RUN=${GDUMP_ENS} YMD=${gPDY} HH=${gcyc} generate_com -rx \
-    COM_ATMOS_HISTORY_ENS_PREV:COM_ATMOS_HISTORY_TMPL
+MEMDIR='${MEMDIR}' RUN=${GDUMP_ENS} YMD=${gPDY} HH=${gcyc} generate_com -rx \
```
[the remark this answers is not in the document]
RussTreadon-NOAA: If `MEMDIR='${MEMDIR}'` is replaced with `MEMDIR="${MEMDIR}"` or `MEMDIR="'${MEMDIR}'"`, job `atmanlinit` aborts with an `unbound variable` message
```
/scratch1/NCEPDEV/da/role.jedipara/git/global-workflow/test0/jobs/JGLOBAL_ATM_ANALYSIS_INITIALIZE: line 30: MEMDIR: unbound variable
++ JGLOBAL_ATM_ANALYSIS_INITIALIZE[1]: postamble JGLOBAL_ATM_ANALYSIS_INITIALIZE 1692209724 1
++ preamble.sh[68]: set +x
End JGLOBAL_ATM_ANALYSIS_INITIALIZE at 18:15:28 with error code 1 (time elapsed: 00:00:04)
```

How do we pass the shellcheck without breaking existing functionality?

ush/python/pygfs/task/atm_analysis.py
aerorahul: The new method is a duplicate of the method in the `atmens_analysis.py`.
Is there a reason why this method from `atmens_analysis.py` cannot be used by making it available to `atm_analysis.py`?

RussTreadon-NOAA: Agreed.  The ensemble directory pieces are 99.9% identical.  `atmens_analysis.py` populates `bkg/` whereas `atm_analysis.py` populates `ens/`.  I always start from the simplest implementation, get it to work, and generalize from there.   

A possible refactor would be to 
- remove `get_berror_ens_dict` from `atm_analysis.py`
- remove `get_bkg_dict` from `atmens_analysis.py`
- add generic `get_ens_dict` to `analysis.py`.   Pass `bkg` or `ens` into method to create and populate correct directory.  Alternatively, figure out another way to distinguish between deterministic (`ens`) and ensemble (`bkg`) jobs.
- add `get_ens_dict` to `atm_analysis.py` and `atmens_analysis.py`

Is this acceptable?

aerorahul: Thank you @RussTreadon-NOAA for iterating with me.
Elevating to `analysis.py` and calling it `get_fv3_ens_dict` would be acceptable.

The trouble you are having with `generate_com` is because you are trying to create a template from a template.

RussTreadon-NOAA: What's the recommended fix, then, for the shellcheck error?   What is currently scripted in `JGLOBAL_ATM_ANALYSIS_INITIALIZE` works ... but it fails shellcheck.

jobs/JGLOBAL_ATM_ANALYSIS_INITIALIZE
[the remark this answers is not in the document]
aerorahul: I think you want to disable the error since shellcheck does not know that you are intentionally trying to create the template.
Adding a line:
```
shellcheck disable=SC2016
```
 before this line will inform shellcheck to ignore this error.

ush/python/pygfs/task/atm_analysis.py
WalterKolczynski-NOAA: You should be able to use `COM_ATMOS_RESTART_TMPL` directly instead of running it through `generate_com()`. All generate_com does is fill in the template and assign the result to a variable. Just be sure to supply the correct variable assignments when you fill it in in python.

RussTreadon-NOAA: Attempt to address reviewer comments committed to `feature/ufsda_hybvar` at [`1f9b889`](https://github.com/RussTreadon-NOAA/global-workflow/commit/1f9b88909549a957ada848b71f265038114a45bc)

ush/python/pygfs/task/analysis.py
```diff
@@ -200,6 +201,74 @@ def link_jediexe(self) -> None:
 
         return
 
+    @logit(logger)
+    def get_ens_dict(self, task_config: Dict[str, Any]) -> Dict[str, List[str]]:
```
aerorahul: My suggestion would be to name this method `get_fv3ens_dict` or `get_atmens_dict` since that is what this method is doing.
I can anticipate similar requirements for ocean, ice, aerosols in the future.

ush/python/pygfs/task/atmens_analysis.py
```diff
@@ -108,7 +108,8 @@ def initialize(self: Analysis) -> None:
         FileHandler(jedi_fix_list).sync()
 
         # stage backgrounds
-        FileHandler(self.get_bkg_dict()).sync()
+        logger.debug(f"Stage ensemble member background files")
+        FileHandler(Analysis.get_ens_dict(self, self.task_config)).sync()
```
```suggestion
        FileHandler(self.get_ens_dict(self.task_config)).sync()
```
`atmens_analysis.py` inherits from `analysis.py`.

ush/python/pygfs/task/atm_analysis.py
```diff
@@ -82,11 +82,15 @@ def initialize(self: Analysis) -> None:
...
-        # stage berror files
-        # copy static background error files, otherwise it will assume ID matrix
+        # stage static background error files, otherwise it will assume ID matrix
         logger.debug(f"Stage files for STATICB_TYPE {self.task_config.STATICB_TYPE}")
         FileHandler(self.get_berror_dict(self.task_config)).sync()
 
+        # stage ensemble files for use in hybrid background error
+        if self.task_config.DOHYBVAR:
+            logger.debug(f"Stage ensemble files for DOHYBVAR {self.task_config.DOHYBVAR}")
+            FileHandler(Analysis.get_ens_dict(self, self.task_config)).sync()
```
aerorahul:
```suggestion
            FileHandler(sel.f.get_ens_dict(self.task_config)).sync()
```
`atm_analysis.py` inherits from `analysis.py`, so this method is already in scope.

ush/python/pygfs/task/analysis.py
aerorahul: The documentation is calling this `config` and not `task_config`.
It should be `config` as `task_config` is an attribute of `self` and we do not want it to confused with this `"config"`.
```suggestion
    def get_ens_dict(self, config: Dict[str, Any]) -> Dict[str, List[str]]:
```

ush/python/pygfs/task/analysis.py
```diff
@@ -200,6 +201,74 @@ def link_jediexe(self) -> None:
...
+            'HH': prev_cycle.strftime('%H'),
+            'MEMDIR': None
+        }
+
+        # set directory type based on RUN
+        if self.task_config.RUN in ['enkfgdas', 'enkfgfs']:
+            dirtype = 'bkg'
+        else:
+            tmpl_res_dict['RUN'] = 'enkf' + self.task_config.RUN
+            dirtype = 'ens'
```
aerorahul: Now that I have had a chance to look at this, I would suggest turning this method in to a `staticmethod` and pass everything that this method needs through `config`, rather than relying on the global variable `self.task_config`.
This would also eliminate this `if-else` and raise it to the class that defines this specific need.

aerorahul (changes requested):
Thank you for this PR @RussTreadon-NOAA 
I have a few comments.
If there is anything that is confusing or unclear, please let me know.
If you wish, I can do some of the work that I describe; `staticmethod` next week.

Also, it would be great to get a test for JEDI Atm 3dvar, EnKF and now 4DEnVar when you are ready to ensure these capabilities are retained.

ush/python/pygfs/task/analysis.py
RussTreadon-NOAA: Rename as `get_fv3ens_dict`.  Done at 5e88371.

ush/python/pygfs/task/atm_analysis.py
RussTreadon-NOAA: Accept suggestion.  Done at 5e88371.

ush/python/pygfs/task/atmens_analysis.py
RussTreadon-NOAA: Accept suggestion.  Done at 5e88371.

ush/python/pygfs/task/analysis.py
RussTreadon-NOAA: Agree that scripting does not agree with documentation.   Both `atm_analysis.py` and `atmens_analysis.py` pass `self.config_task` to `get_fv3ens_dict` so it seems documentation should be updated.  That is, replace `config` with `task_config` in the documentation

```
        Parameters
        ----------
        task_config: Dict
            a dictionary containing all of the configuration needed for the task
```
This change was committed at 5e88371.   Not sure if this is acceptable.

ush/python/pygfs/task/analysis.py
RussTreadon-NOAA: I think I see what you are saying.  Let me see if I can correctly implement this suggestion.  I'll update the method documentation accordingly once I get the suggested change implemented.

RussTreadon-NOAA: `get_fv3ens_dict` has been changed to a `staticmethod`.   The `if-else` in `get_fv3ens_dict` has been removed and necessary additions made in the classes which invoke the method.   Done at [`51200a0`](https://github.com/RussTreadon-NOAA/global-workflow/commit/51200a0d9459f9c67c45bd6cb284163603ca51e7).

ush/python/pygfs/task/atm_analysis.py
```diff
@@ -82,11 +82,17 @@ def initialize(self: Analysis) -> None:
...
-        # copy static background error files, otherwise it will assume ID matrix
+        # stage static background error files, otherwise it will assume ID matrix
         logger.debug(f"Stage files for STATICB_TYPE {self.task_config.STATICB_TYPE}")
         FileHandler(self.get_berror_dict(self.task_config)).sync()
 
+        # stage ensemble files for use in hybrid background error
+        if self.task_config.DOHYBVAR:
+            logger.debug(f"Stage ensemble files for DOHYBVAR {self.task_config.DOHYBVAR}")
+            self.task_config.RUN = 'enkf' + self.task_config.RUN
+            self.task_config.dirname ='ens'
```
aerorahul: These are dangerous.
After these lines, one would have altered the `self.task_config`   This is forbidden.
What you should do here instead, is make a deepcopy of `self.task_config`, extract the attributes that you need to send to `self.get_fv3ens_dict`, instead of sending in the entire task configuration.

RussTreadon-NOAA: I guess having this in `get_fv3ens_dict` was safer.  I could move it back but I don't think you want this.  I don't know how to do what you outline.  Is there an example somewhere in g-w that I can follow.

aerorahul: see `land_analysis.py`.
You could follow something along these lines (not what this method needs), but an example of creating a local dict and pulling relevant info from `self.task_config`.
```
        localconf = AttrDict()
        keys = ['DATA', 'current_cycle', 'COM_OBS', 'COM_ATMOS_RESTART_PREV',
                'OPREFIX', 'CASE', 'ntiles']
        for key in keys:
            localconf[key] = self.task_config[key]
        self.get_fv3ens_dict(localconf)
```

Having them in `self.get_fv3ens_dict` is not really a good solution as it makes that method bring in the entire global environment. One should know exactly what a method requires and pass that explicitly.  One could say it is the same argument that people frown on using global variables in Fortran.

RussTreadon-NOAA: Thanks for the guidance.  Back to refactoring.

RussTreadon-NOAA: Create, populate, and pass `localconf` to `get_fv3ens_dict`.   Change made in both `atm_analysis.py` and `atmens_analysis.py`.  Done at [`a2ab183`](https://github.com/RussTreadon-NOAA/global-workflow/commit/a2ab18339287dc73cb36d269e86b641d54445f55)

aerorahul: perfect @RussTreadon-NOAA 
This way, when the method runs, the logger will print out exactly what is going in and coming out.
Thank you~

aerorahul (approved):
Looks good.
Two comments:
1. please update the description of the PR to reflect the actual changes
2. can you provide test cases and configurations so we can run the UFS-based atmosphere DA as part of our CI tests?

RussTreadon-NOAA:
PR description updated to note that this PR impacts the staging of ensemble files for both hybrid variational and ensemble UFS-DA atmospheric analyses.  Is this acceptable?

My preference is to provide test cases and configurations so we can run the UFS-based atmosphere DA as part of our CI tests via a new g-w issue and subsequent PR.  Is this acceptable?

Where are the files for current g-w CI tests?  I'd like to see if I can leverage or augment what we already have instead of adding new cases to the g-w CI database.

aerorahul:
> PR description updated to note that this PR impacts the staging of ensemble files for both hybrid variational and ensemble UFS-DA atmospheric analyses. Is this acceptable?

sure

> My preference is to provide test cases and configurations so we can run the UFS-based atmosphere DA as part of our CI tests via a new g-w issue and subsequent PR. Is this acceptable?

yes

> Where are the files for current g-w CI tests? I'd like to see if I can leverage or augment what we already have instead of adding new cases to the g-w CI database.

On Hera: `/scratch1/NCEPDEV/global/glopara/data/ICSDIR`
On Orion: `/work/noaa/global/glopara/data/ICSDIR`
These are the cases that are currently part of the automated testing:
https://github.com/NOAA-EMC/global-workflow/tree/develop/ci/cases

RussTreadon-NOAA:
OK.  We can work cycle `ICSDIR/C96C48`.  This directory is for cold-starts.  Nothing wrong with this but we have not yet added to g-w the ability to run UFS-DA ATM parallels from the existing GDA.  

Initial steps toward this goal are in GDASApp PR [#575](https://github.com/NOAA-EMC/GDASApp/pull/575).   Once UFS-DA ATM parallels can directly use GDA bufr files we can cold start from 20211220/18 `ICSDIR/C96C48` and cycle through 20211221/00.

PR #1871, opened 2023-09-20
CDO based post-processing application

### LINKED ISSUE #923: Port NCL scripts to python

HenryRWinterbottom:
# Description
This PR addresses issue #923. A `bash` script is provided that leverages the `CDO` and `NCO` packages to remap specified variable fields using the CDO supported remapping types. This PR is a necessary upgrade to the existing NCL based scripts due to NCL reaching EOL (i.e., it is no-longer supported). No special Python libraries or utilities are required. Leveraging the capabilities of CDO and NCO allow support across all development and production platforms.

  Resolves #923

# Type of change

# Change characteristics
- This is a non-breaking change.
- This change does not require any documentation updates.

# How has this been tested?
Tested using ice and ocean forecast files provided by @NeilBarton-NOAA and @jiandewang and executed on RDHPCS Hera, NOAA CSP AWS, and OSX.

# Checklist


WalterKolczynski-NOAA:
Please use the PR template (check your comment delineators).

ush/cdo_post.sh
```diff
@@ -0,0 +1,376 @@
...
+_calling_script=$(basename "${BASH_SOURCE[0]}")
+start_time_human=$(date -d"@${start_time}" -u)
+echo "Begin ${_calling_script} at ${start_time_human}."
+
+# Read the configuration file for the the variables to be remapped and
+# proceed accordingly.
+while IFS= read -r line; do
+
+    # Get the attributes for the respective variable(s).
+    varname=$(echo "${line}" | $(command -v awk) "{print $1}")
```
HenryRWinterbottom: The `shellnorms` linter insists on double quotes around the `awk` command; however this will cause `awk` to fail within this bash environment. Single quotes are required.

Is there an issue with the bash script linter?

HenryRWinterbottom: The error is `SC2016 (info): Expressions don't expand in single quotes, use double quotes for that.` which is incorrect and will cause the script to break.

EricSinsky-NOAA:
@HenryWinterbottom-NOAA Can you modify `scripts/run_regrid.sh` so that `ush/cdo_post.sh` is executed instead of `ush/ocnpost.ncl` and `ush/icepost.ncl`?

HenryRWinterbottom:
> @HenryWinterbottom-NOAA Can you modify `scripts/run_regrid.sh` so that `ush/cdo_post.sh` is executed instead of `ush/ocnpost.ncl` and `ush/ocnpost.ncl`?

@EricSinsky-NOAA Yes, I will do that in a separate PR. We will need an additional script beneath `scripts` for the pre-processing step as well.

EricSinsky-NOAA:
@HenryWinterbottom-NOAA Ok. When you modify `scripts/run_regrid.sh` to work with your ush scripts (e.g. `ush/cdo_post.sh`, `ush/cdo_post_ocean_format.sh`), please make sure it is still compatible with  `jobs/rocoto/ocnpost.sh`. In the current global workflow, `scripts/run_regrid.sh` is executed  in `jobs/rocoto/ocnpost.sh`.

HenryRWinterbottom:
> @HenryWinterbottom-NOAA Ok. When you modify `scripts/run_regrid.sh` to work with your ush scripts (e.g. `ush/cdo_post.sh`, `ush/cdo_post_ocean_format.sh`), please make sure it is still compatible with `jobs/rocoto/ocnpost.sh`. In the current global workflow, `scripts/run_regrid.sh` is executed in `jobs/rocoto/ocnpost.sh`.

@EricSinsky-NOAA Can you please provide me documentation as to what exactly those scripts are doing (e.g., what are the respective environment variables and what are they used for)? The scripts are not documented and I am not familiar enough with NCL to guess what is happening in the respective NCL scripts.

EricSinsky-NOAA:
@HenryWinterbottom-NOAA Sure, I can provide you more information on these scripts.

DeniseWorthen:
@HenryWinterbottom-NOAA Since the rotation angles in MOM6 are located on cell centers and the ocean velocities are on the cell edges, you must first regrid the ocean velocities to the center cell location before rotation. Is that what your CDO script does?

jiandewang:
@HenryWinterbottom-NOAA I tried offline on reamp U/V and that worked as expected.
On HERA /scratch1/NCEPDEV/climate/Jiande.Wang/working/Eric-regrid/test-vector-T
note I added "coordinator" to sin_rot and cos_rot, also note in rotate-remap.sh, I used
rotated_ssu=cos_rot*SSU+sin_rot*SSV, here I put angle in front of U/V so that rotated variable will be on center grid.
SSU-JW
SSUV-JW

HenryRWinterbottom:
> @HenryWinterbottom-NOAA Since the rotation angles in MOM6 are located on cell centers and the ocean velocities are on the cell edges, you must first regrid the ocean velocities to the center cell location before rotation. Is that what your CDO script does?

@DeniseWorthen Yes. I think the issue was the wrong rotation (i.e., earth relative versus grid relative).

DeniseWorthen:
@HenryWinterbottom-NOAA What issue? The NCL scripts and the python scripts that @EricSinsky-NOAA built from them both regridded the velocities to the center, then rotated.

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+#
+# Example usage:
+#   _comma_split_string "item1,item2 item3,item4"
+#   for element in "${global_array[@]}"; do
+#       echo "$element"
+#   done
+#
+# This example will split the input string into individual elements
+# and print each element on a separate line.
+function _comma_split_string() {
```
WalterKolczynski-NOAA: This function and the next appear in both PRs. These should be moved to a separate script that gets called to avoid duplication.

Will skip reviewing this copy since I've already reviewed those in the other PR.

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+#   nc_concat "variable_interpolated.nc"
+#
+# This example will concatenate the variable-interpolated file into
+# the output netCDF file, handling the case where the output file may
+# or may not exist.
+function nc_concat(){
+    local var_interp_path="${1}"
+
+    tmp_nc_file="${PWD}/tmp_nc.nc"
+    if test -e "${output_path}"; then
```
WalterKolczynski-NOAA:
```suggestion
    if [[ -e "${output_path}" ]]; then
```

WalterKolczynski-NOAA: The contents of this if-block is also not indented properly.

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+#   exists, the function merges `var_interp_path` into it and manages
+#   temporary files accordingly. If `output_path` does not exist, it
+#   creates a new netCDF file by renaming `var_interp_path` to
+#   `output_path`.
+#
+# Parameters:
+#   $1 - The path to the variable-interpolated netCDF file.
+#
+# Global Variables:
+#   output_path - The path to the output netCDF file.
```
WalterKolczynski-NOAA, line 100: Make output_path an argument. Don't rely on globals.

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+# or may not exist.
+function nc_concat(){
+    local var_interp_path="${1}"
+
+    tmp_nc_file="${PWD}/tmp_nc.nc"
+    if test -e "${output_path}"; then
+	echo "netCDF-formatted file path ${output_path} exists; merging ${var_interp_path}"
+	cdo merge "${output_path}" "${var_interp_path}" "${tmp_nc_file}"
+	mv "${tmp_nc_file}" "${output_path}"
+	rm "${var_interp_path}" >> /dev/null
```
WalterKolczynski-NOAA: How damaging is it if the file isn't removed, and why is STDOUT being trashed?

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+#   defined by the global variable `output_path`.
+#
+# Parameters:
+#   $1 - The name of the variable to remap.
+#   $2 - The path to the input netCDF file.
+#   $3 - The interpolation method to use (e.g., "remapbil").
+#
+# Global Variables:
+#   output_path - The path to the output netCDF file.
+#   dstgrid_config - The configuration for destination grid specifications.
```
WalterKolczynski-NOAA, line 142: Same as above, make these arguments.

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+#   cdo_remap "temperature" "input_file.nc" "remapbil"
+#
+# This example remaps the "temperature" variable using bilinear
+# interpolation from the input file and appends the result to the
+# output netCDF file.
+function cdo_remap(){
+    local varname="${1}"
+    local varfile="${2}"
+    local interp_type="${3}"
+    local var_interp_path="${PWD}/${varname}.interp.nc"
```
WalterKolczynski-NOAA, line 155: `${PWD}` is undefined.

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+#
+# Parameters:
+#   $1 - Comma-delimited variable names to rotate (e.g., "u,v").
+#   $2 - The path to the input netCDF file.
+#   $3 - The interpolation method to use (e.g., "remapbil").
+#   $4 - Comma-delimited rotation angles or trigonometric functions
+#        (e.g., "theta" or "cosang,sinang").
+#
+# Global Variables:
+#   output_path - The path to the output netCDF file.
```
WalterKolczynski-NOAA, line 183: Same as before.

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+#     that the `cos(theta)` and `sin(theta)` have been computed
+#     a'priori and assigned the input file variable names defined by
+#     `cosang` and `sinang` respectively.
+#
+function cdo_rotate(){    
+    local varnames="${1}"
+    local varfile="${2}"
+    local interp_type="${3}"
+    local angles="${4}"  
+    local var_rotate_path="${PWD}/rotate.nc"
```
WalterKolczynski-NOAA: `${PWD}` undefined again

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+    _comma_split_string "${angles}"
+    angle_array=("${global_array[@]}")
+    _strip_whitespace "${varname_array[0]}"
+    xvar="${out_string}"
+    _strip_whitespace "${varname_array[1]}"
+    yvar="${out_string}"
+    nangles="${#angle_array[@]}"
+    
+    echo "Rotating and remapping variables ${xvar} and ${yvar} from file ${varfile}."
+    if [[ "${nangles}" == 1 ]]; then
```
WalterKolczynski-NOAA:
```suggestion
    if (( nangles == 1 )); then
```

WalterKolczynski-NOAA: This if-block is also not indented properly.

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+	_strip_whitespace "${angle_array[0]}"
+	theta="${out_string}"
+	cdo -expr,"xr=${xvar}*cos(${theta})-${yvar}*sin(${theta}); yr=${xvar}*sin(${theta})+${yvar}*cos(${theta})" -selname,"${xvar}","${yvar}","${theta}" "${varfile}" "${var_rotate_path}"
+    elif [[ "${nangles}" == 2 ]]; then
```
WalterKolczynski-NOAA:
```suggestion
    elif (( nangles == 2 )); then
```

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+    else
+	echo "Vector rotations with ${nangles} attributes is not supported. Aborting!!!"
+	exit 102
+    fi
+
+    cdo_remap "xr" "${var_rotate_path}" "${interp_type}"
+    varname_update "xr" "${xvar}" "${output_path}"
+    cdo_remap "yr" "${var_rotate_path}" "${interp_type}"
+    varname_update "yr" "${yvar}" "${output_path}"
+    rm "${var_rotate_path}" >> /dev/null
```
WalterKolczynski-NOAA: Same questions as before.

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+}
+
+#######
+
+# varname_update - Rename a variable in a netCDF file and write to a new file.
+#
+# Description:
+#   This function renames a variable in a netCDF file from
+#   `old_varname` to `new_varname` and writes the result to a new
+#   netCDF file specified by `ncfile`.
```
WalterKolczynski-NOAA: Is this really a new file, or renaming within the same file? Seems like the latter since no old file is specified.

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+    echo "Renaming variable ${old_varname} to ${new_varname} and writing to file ${ncfile}."
+    ncrename -O -v "${old_varname}","${new_varname}" "${ncfile}"
+}
+
+#######
+
+start_time=$(date +%s)
+_calling_script=$(basename "${BASH_SOURCE[0]}")
+start_time_human=$(date -d"@${start_time}" -u)
+echo "Begin ${_calling_script} at ${start_time_human}."
```
WalterKolczynski-NOAA: I think I missed this in the other PR, but the preamble handles this. It should be sourced at the top of both scripts and all of this should be removed.

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+# Read the configuration file for the the variables to be remapped and
+# proceed accordingly.
+while IFS= read -r line; do
+
+    # Get the attributes for the respective variable(s).
+    varname=$(echo "${line}" | awk '{print $1}')
+    interp_type=$(echo "${line}" | awk '{print $2}')
+    srcgrid=$(echo "${line}" | awk '{print $3}')
+    rotate=$(echo "${line}" | awk '{print $4}')
+    angle=$(echo "${line}" | awk '{print $5}')
```
WalterKolczynski-NOAA: Use here-strings instead of piping an echo, e.g.:
```
    varname=$(awk '{print $1}' <<< "${line}")
```

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+    
+    if [[ "${rotate}" == 0 ]]; then
```
WalterKolczynski-NOAA:
```suggestion
    if (( rotate == 0 )); then
```

WalterKolczynski-NOAA: If-block also isn't properly indented.

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+    	# No rotation necessary; interpolate/remap the variables and
+    	# directly.
+	echo "Remapping variable ${varname} without rotation."
+	cdo_remap "${varname}" "${srcgrid}" "${interp_type}"
+	
+    elif [[ "${rotate}" = "1" ]]; then
```
WalterKolczynski-NOAA:
```suggestion
    elif (( rotate = 1 )); then
```

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+    	# Rotation necessary; rotate the respective vector quantities
+    	# relative to the source grid projection and subsequently
+    	# remap the variables to the specified destination grid.
+	echo "Remapping variables ${varname} with rotation."
+	cdo_rotate "${varname}" "${srcgrid}" "${interp_type}" "${angle}"
+	
+    else
+	echo "Rotation option ${rotate} not recognized. Aborting!!!"
```
WalterKolczynski-NOAA:
```suggestion
	echo "FATAL ERROR: Rotation option ${rotate} not recognized. Aborting!!!"
```

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+	exit 101
+    fi
+
+done < "${variable_file}"
+
+stop_time=$(date +%s)
+_calling_script=$(basename "${BASH_SOURCE[0]}")
+stop_time_human=$(date -d"@${stop_time}" -u)
+echo "End ${_calling_script} at ${stop_time_human}."
```
WalterKolczynski-NOAA: Preamble will handle this.

ush/remap.sh
```diff
@@ -0,0 +1,376 @@
...
+	cdo -expr,"xr=${xvar}*cos(${theta})-${yvar}*sin(${theta}); yr=${xvar}*sin(${theta})+${yvar}*cos(${theta})" -selname,"${xvar}","${yvar}","${theta}" "${varfile}" "${var_rotate_path}"
+    elif [[ "${nangles}" == 2 ]]; then
+	_strip_whitespace "${angle_array[0]}"
+	cosang="${out_string}"
+	_strip_whitespace "${angle_array[1]}"
+	sinang="${out_string}"
+	cdo -expr,"xr=${xvar}*${cosang}-${yvar}*${sinang}; yr=${xvar}*${sinang}+${yvar}*${cosang}" -selname,"${xvar}","${yvar}","${cosang}","${sinang}" "${varfile}" "${var_rotate_path}"
+	
+    else
+	echo "Vector rotations with ${nangles} attributes is not supported. Aborting!!!"
```
WalterKolczynski-NOAA:
```suggestion
        echo "FATAL ERROR: Vector rotations with ${nangles} attributes is not supported. Aborting!!!"
```

WalterKolczynski-NOAA (changes requested):

HenryRWinterbottom:
@jiandewang Can you please confirm that `/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/ocean_remap_rotate.nc` on RDHPCS Hera is correct relative to your results?

ush/remap.sh
HenryRWinterbottom: Can you point me to the correct way to indent these blocks?

ush/remap.sh
HenryRWinterbottom: This is to remove files that are no longer needed. Once the interpolated file, for the respective variable, is merged (i.e., appended) to the output netCDF file path it is no longer needed. If the files remain they will contribute to the disk footprint without reason.

ush/remap.sh
HenryRWinterbottom: This is defined in the environment once a shell is invoked.

Here are examples from three different hosts.

- RDHPCS Hera: 
   `[Henry.Winterbottom@hfe07 Neil.Barton]$ echo $PWD
     /scratch1/NCEPDEV/da/Henry.Winterbottom/scratch
    `
- NOAA CSP AWS:
    `[Henry.Winterbottom@globalworkflowtest-3 logs]$ echo $PWD
     /contrib/Rajendra.Panda/global-workflow/sorc/logs
    `
- OSX:
   `[leviathan:UFS/global-workflow/ush] henry.winterbottom% echo $PWD
    /Users/henry.winterbottom/trunk/UFS/global-workflow/ush
   `
- Ubuntu 20.04 Docker container:
   `root@a7c3ca831789:/home# echo $PWD
    /home
   `

ush/remap.sh
HenryRWinterbottom: See https://github.com/NOAA-EMC/global-workflow/pull/1871/files/571276e9296c4d7f3dcb92ef74f0b18f22b76658#r1337059165.

ush/remap.sh
HenryRWinterbottom: See https://github.com/NOAA-EMC/global-workflow/pull/1871/files/571276e9296c4d7f3dcb92ef74f0b18f22b76658#r1337049025

ush/remap.sh
HenryRWinterbottom: Done.

ush/remap.sh
HenryRWinterbottom: See https://github.com/NOAA-EMC/global-workflow/pull/1871/files/571276e9296c4d7f3dcb92ef74f0b18f22b76658#r1337051162.

ush/remap.sh
HenryRWinterbottom: It is the same file. Only `old_varname` is renamed to `new_varname`.

ush/remap.sh
HenryRWinterbottom: Done.

ush/remap.sh
HenryRWinterbottom: Done.

ush/remap.sh
HenryRWinterbottom: Done.

ush/remap.sh
HenryRWinterbottom: Done.

ush/remap.sh
HenryRWinterbottom: See https://github.com/NOAA-EMC/global-workflow/pull/1871/files/571276e9296c4d7f3dcb92ef74f0b18f22b76658#r1337049025.

ush/remap.sh
HenryRWinterbottom: Can you suggest where beneath `/ush` I am assuming to add this?

DeniseWorthen:
@jiandewang From our off-line conversation, you said that when you use ``rotated_ssu=cos_rotSSU+sin_rotSSV`` the SSU and SSV are internally regridded to the cos_rot, sin_rot location. 

Question: How are missing values treated? The native SSU and SSV carry missing values on land points. Those missing values need to be considered zeros when internally interpolating to the center grid points. Does CDO consider the missing values as zeros when it forms the velocities at the center grid points?

I'm also curious---does CDO use pre-generated weights? If not, what is the elapsed time to for doing a 3-dim ocean output on the tripole grid using the python scripts vs CDO?

jiandewang:
> @jiandewang Can you please confirm that `/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/ocean_remap_rotate.nc` on RDHPCS Hera is correct relative to your results?

it doesn't look correct for vector, see below figure (sorry for the low quality of the plot as I am having trouble to do scp)
SSU-0925

jiandewang:
> @jiandewang From our off-line conversation, you said that when you use `rotated_ssu=cos_rotSSU+sin_rotSSV` the SSU and SSV are internally regridded to the cos_rot, sin_rot location.
> 
> Question: How are missing values treated? The native SSU and SSV carry missing values on land points. Those missing values need to be considered zeros when internally interpolating to the center grid points. Does CDO consider the missing values as zeros when it forms the velocities at the center grid points?
> 
> I'm also curious---does CDO use pre-generated weights? If not, what is the elapsed time to for doing a 3-dim ocean output on the tripole grid using the python scripts vs CDO?

I almost have not used cdo until this work. I learn the usage by looking from Henry's script so it is being used as a blackbox for me. What I noticed is that:
(1) if you do u-rotate=u*angle......., the resuled u-rotate will be on U grid
(2) if you do u-rotate=angel*u......., the resulted u-rotate will be on T grid
I don't know how cdo internally handle this kind of remap and math for two variables that are on different grid. I noticed the above fact by looking at the meta data in the u-rotate nc files

jiandewang:
I think we need to re-consider to use Eric's approach (although I am not the right decision maker). With cdo approach we will get different land-sea mask (compare with ncl and Python's approach). While this may not be a critical issue, we (at least me) don't really have a good understanding on how cdo did the vector remap and rotation internally as they are on diferent grid, and the output file format (here I mean the variable dimension) will not be the same as from ncl/Python, and this output will not be able to feed in as input for the step right after this. On the other side, output from Python mimic what ncl did almost perfect and the output format is identical to ncl's. In ncl/Python the output file variables dimension et. al are defined by their script, but in cdo it is controlled by cdo internally and we don't have control of it.

HenryRWinterbottom:
> I think we need to re-consider to use Eric's approach (although I am not the right decision maker). With cdo approach we will get different land-sea mask (compare with ncl and Python's approach). While this may not be a critical issue, we (at least me) don't really have a good understanding on how cdo did the vector remap and rotation internally as they are on diferent grid, and the output file format (here I mean the variable dimension) will not be the same as from ncl/Python, and this output will not be able to feed in as input for the step right after this. On the other side, output from Python mimic what ncl did almost perfect and the output format is identical to ncl's. In ncl/Python the output file variables dimension et. al are defined by their script, but in cdo it is controlled by cdo internally and we don't have control of it.

Eric's method will require significant rewrite and, as is, is not acceptable for merging into the global-workflow. Further, unless someone wants to invest the time in getting the ESMPy API to work within Python, CDO is the quickest solution. NCL is EOL and therefore not supported and NCO will not support xESMF (which https://github.com/HenryWinterbottom-NOAA/ufs_tools uses). 

The CDO method works properly (as you demonstrated above). However, there are several steps that are required to make sure that the information that is interpolated correctly and require careful debugging. We (the global workflow team) are actively working on this. 

If Eric's method is your desired solution then it will need to be rewritten to comply with both the global-workflow standards as well as properly using OOP methods/packages.

If questions remain regarding CDO, I suggest reading the documentation [here](https://www.climate-cryosphere.org/wiki/index.php?title=Regridding_with_CDO).

jiandewang:
> > I think we need to re-consider to use Eric's approach (although I am not the right decision maker). With cdo approach we will get different land-sea mask (compare with ncl and Python's approach). While this may not be a critical issue, we (at least me) don't really have a good understanding on how cdo did the vector remap and rotation internally as they are on diferent grid, and the output file format (here I mean the variable dimension) will not be the same as from ncl/Python, and this output will not be able to feed in as input for the step right after this. On the other side, output from Python mimic what ncl did almost perfect and the output format is identical to ncl's. In ncl/Python the output file variables dimension et. al are defined by their script, but in cdo it is controlled by cdo internally and we don't have control of it.
> 
> Eric's method will require significant rewrite and, as is, is not acceptable for merging into the global-workflow. Further, unless someone wants to invest the time in getting the ESMPy API to work within Python, CDO is the quickest solution. NCL is EOL and therefore not supported and NCO will not support xESMF (which https://github.com/HenryWinterbottom-NOAA/ufs_tools uses).
  I recall in previous team meeting (happened several months ago) it is agreed by team leader to use non-wcoss2 platform to generate fixed files using xESMF (as it's a one time deal) and do remap on wcoss2.
> 
> The CDO method works properly (as you demonstrated above). However, there are several steps that are required to make sure that the information that is interpolated correctly and require careful debugging. We (the global workflow team) are actively working on this.
> 
> If Eric's method is your desired solution then it will need to be rewritten to comply with both the global-workflow standards as well as properly using OOP methods/packages.
  yes that's what we desired. The output file from cdo will not be able to feed in for the next step. 
> 
> If questions remain regarding CDO, I suggest reading the documentation [here](https://www.climate-cryosphere.org/wiki/index.php?title=Regridding_with_CDO).

EricSinsky-NOAA:
> @jiandewang From our off-line conversation, you said that when you use `rotated_ssu=cos_rotSSU+sin_rotSSV` the SSU and SSV are internally regridded to the cos_rot, sin_rot location.
> 
> Question: How are missing values treated? The native SSU and SSV carry missing values on land points. Those missing values need to be considered zeros when internally interpolating to the center grid points. Does CDO consider the missing values as zeros when it forms the velocities at the center grid points?
> 
> I'm also curious---does CDO use pre-generated weights? If not, what is the elapsed time to for doing a 3-dim ocean output on the tripole grid using the python scripts vs CDO?

@DeniseWorthen, CDO does have an option for the use of pre-generated weights using the SCRIP convention. More information on this capability can be found in the CDO documentation [here](https://code.mpimet.mpg.de/projects/cdo/embedded/index.html#x1-7380002.12.8).

HenryRWinterbottom:
@jiandewang Please check file `/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/test_pr.nc` on RDHPCS Hera.

Note that only the relevant momentum variables have been output.

EricSinsky-NOAA:
> > I think we need to re-consider to use Eric's approach (although I am not the right decision maker). With cdo approach we will get different land-sea mask (compare with ncl and Python's approach). While this may not be a critical issue, we (at least me) don't really have a good understanding on how cdo did the vector remap and rotation internally as they are on diferent grid, and the output file format (here I mean the variable dimension) will not be the same as from ncl/Python, and this output will not be able to feed in as input for the step right after this. On the other side, output from Python mimic what ncl did almost perfect and the output format is identical to ncl's. In ncl/Python the output file variables dimension et. al are defined by their script, but in cdo it is controlled by cdo internally and we don't have control of it.
> 
> Eric's method will require significant rewrite and, as is, is not acceptable for merging into the global-workflow. Further, unless someone wants to invest the time in getting the ESMPy API to work within Python, CDO is the quickest solution. NCL is EOL and therefore not supported and NCO will not support xESMF (which https://github.com/HenryWinterbottom-NOAA/ufs_tools uses).
> 
> The CDO method works properly (as you demonstrated above). However, there are several steps that are required to make sure that the information that is interpolated correctly and require careful debugging. We (the global workflow team) are actively working on this.
> 
> If Eric's method is your desired solution then it will need to be rewritten to comply with both the global-workflow standards as well as properly using OOP methods/packages.
> 
> If questions remain regarding CDO, I suggest reading the documentation [here](https://www.climate-cryosphere.org/wiki/index.php?title=Regridding_with_CDO).

@jiandewang I agree that it is difficult to determine exactly how CDO is performing the vector remap and rotation, but CDO will be more efficient and easier to maintain than Python in the global-workflow. I have tested the remapping routine using ESMPy, and it is not nearly as efficient as CDO. I have also tested the remapping using f2py. Although the use of f2py significantly increases efficiency compared to ESMPy, this will be difficult to maintain and it would make the program less portable. This limitation would still hold even if we were to make the revisions necessary for it to be acceptable to merge into the global-workflow.

ush/cdo_post.sh
WalterKolczynski-NOAA: Add a shellcheck directive the line above to mute the warning. We've had to do this in other places where single-quotes are necessary (like config.com).

ush/remap.sh
WalterKolczynski-NOAA: It looks like this on my screen (the block is flush with the surrounding code):
image

Make sure you are using only spaces for indentation. It looks like this on GH sometimes if tabs are mixed in.

ush/remap.sh
WalterKolczynski-NOAA: Well, what I was getting at was: should this be `rm -f` to make sure it gets removed? And still not sure why the STDOUT is being redirected to `/dev/null`.

ush/remap.sh
WalterKolczynski-NOAA: Just put it directly in `/ush` for now.

jiandewang:
> @jiandewang Please check file `/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/test_pr.nc` on RDHPCS Hera.
> 
> Note that only the relevant momentum variables have been output.

just had a quick look on SSU, it is on regular grid now, give me some time for further checking

ush/remap.sh
WalterKolczynski-NOAA: Okay. Somehow I was unaware of this shell shortcut, I've always used `$(pwd)`. But apparently `${PWD}` is faster.

ush/remap.sh
WalterKolczynski-NOAA: Okay, just update the comment then.

ush/remap.sh
HenryRWinterbottom: Done.

ush/remap.sh
HenryRWinterbottom: Done, at least it appeared correctly in my GH terminal.

ush/remap.sh
HenryRWinterbottom: I piped the stdout/stderr to a file in in the runtime directory. See `REMAP_LOG`.

ush/remap.sh
HenryRWinterbottom: Done.

ush/remap.sh
HenryRWinterbottom: Done.

ush/remap.sh
HenryRWinterbottom: Done.

ush/remap.sh
HenryRWinterbottom: Done.

ush/remap.sh
WalterKolczynski-NOAA: Still not sure why this is even necessary. It's not like `rm` prints out a bunch of garbage output.

ush/remap.sh
WalterKolczynski-NOAA: Short description (just above this) still says "new file".

HenryRWinterbottom: Sorry. I missed this. Done.

jiandewang:
> @jiandewang Please check file `/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/test_pr.nc` on RDHPCS Hera.
> 
> Note that only the relevant momentum variables have been output.

 they are on regular grid now but diff a lot from Python's output
SSUV-Henry

HenryRWinterbottom:
> > @jiandewang Please check file `/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/test_pr.nc` on RDHPCS Hera.
> > Note that only the relevant momentum variables have been output.
> 
> they are on regular grid now but diff a lot from Python's output SSUV-Henry

@jiandewang Can you please provide me your plotting script as well as any instructions as to how to you are arriving at your solution?

EricSinsky-NOAA:
@jiandewang @HenryWinterbottom-NOAA I can help test CDO and the python programs side by side to see where the differences originate. I'll test SSU and SSV in particular.

HenryRWinterbottom:
@EricSinsky-NOAA thank you. But I'd like to see @jiandewang 's script first. At this point it appears to be a sign error.

jiandewang:
> @EricSinsky-NOAA thank you. But I'd like to see @jiandewang 's script first. At this point it appears to be a sign error.

@HenryWinterbottom-NOAA I am using GrADS to process the data and plotting, see /scratch1/NCEPDEV/climate/Jiande.Wang/working/Eric-regrid/plot-Henry-test_pr-nc/ssuv.gs
I wish my plotting script is wrong.
the data I used in the above ploting is from
/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton
431915244 Sep 26 09:03 test_pr.nc

jiandewang:
@HenryWinterbottom-NOAA one of our team memeber Lydia is going to join the task, she will do the verification/plotting independly.

jiandewang:
@HenryWinterbottom-NOAA in the output you provided (generated by cdo), it has metadata information like
        x = 1440 ;
        y = 721 ;
double lon(y, x)
double lat(y, x) 
float SSU(time, y, x)

while from Python: it is:
        lat = 721 ;
        lon = 1440 ;
       float lon(lon) ;
        float lat(lat) ;
 float SST(time, lat, lon) 

so in Python generated output, lat/lon are defined as 1D while in cdo they are defined as 2D.  The output generated from cdo is going to break global-workflow convert to grib2 step.

jiandewang:
currently we only run ocean post on non-wcoss2 machine using output generated from ncl. Eric's python mimic ncl yield identical netcdf header so it will fit in the workflow when transit to wcoss2

EricSinsky-NOAA:
For reference, here are SSU and SSV maps from both the original NCL program and Python program. The data is 0.25 degrees on a latlon grid. The results are almost exact. These maps are valid at 2013100106 (initial time is 2013100100). 
image
image

HenryRWinterbottom:
Can you please provide verification that it will break that step?

HenryRWinterbottom:
@jiandewang can you also please provide or point me to you plotting script?

jiandewang:
> @jiandewang can you also please provide or point me to you plotting script?

that *gs file is the plotting script

aerorahul:
> @HenryWinterbottom-NOAA in the output you provided (generated by cdo), it has metadata information like x = 1440 ; y = 721 ; double lon(y, x) double lat(y, x) float SSU(time, y, x)
> 
> while from Python: it is: lat = 721 ; lon = 1440 ; float lon(lon) ; float lat(lat) ; float SST(time, lat, lon)
> 
> so in Python generated output, lat/lon are defined as 1D while in cdo they are defined as 2D. The output generated from cdo is going to break global-workflow convert to grib2 step.

@jiandewang 
We should not be worried about the conversion to grib2 step in the global-workflow.
Please verify if the interpolation and remapping from CDO are accurate/as desired.

JessicaMeixner-NOAA:
@aerorahul can you explain a little more about why we should not be worried about the grib2 step in the global-workflow as it's my understanding the output from this is the input to that?

aerorahul:
> @aerorahul can you explain a little more about why we should not be worried about the grib2 step in the global-workflow it's my understanding the output from this is the input to that?

The output from the remapping is an input to the `reg2grb2` program. 
However, that program will probably need to be re-written to be able to run on WCOSS2 as we don't have `wgrib2_api` as a library on WCOSS2.  To my knowledge, it hasn't been shown to run on WCOSS2.

EricSinsky-NOAA:
@jiandewang @HenryWinterbottom-NOAA I think I found the issue regarding the large differences in SSU and SSV from the CDO-generated results (/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/test_pr.nc). It looks like SSU and SSV are identical, which shouldn't be the case.

EricSinsky-NOAA:
@aerorahul I have tested the reg2grb2 program and it seems to work normally on WCOSS2 when the Python-generated remapped data was used as input.

aerorahul:
> @aerorahul I have tested the reg2grb2 program and it seems to work normally on WCOSS2 when the Python-generated remapped data was used as input.

Thanks @EricSinsky-NOAA

JessicaMeixner-NOAA:
> > @aerorahul can you explain a little more about why we should not be worried about the grib2 step in the global-workflow it's my understanding the output from this is the input to that?
> 
> The output from the remapping is an input to the `reg2grb2` program. However, that program will probably need to be re-written to be able to run on WCOSS2 as we don't have `wgrib2_api` as a library on WCOSS2. To my knowledge, it hasn't been shown to run on WCOSS2.

Thanks @aerorahul  I was unaware that part also needed to be re-written

HenryRWinterbottom:
@EricSinsky-NOAA Can you please compare your results for `SSU` and `SSV` to those in the following file on RDHPCS Hera?

`/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/debug_uv_rotate.nc`

As suspected, there was a sign error in the rotation computation.

EricSinsky-NOAA:
@HenryWinterbottom-NOAA Sure. I have plotted the differences for SSU and SSV between the Python-generated results and the CDO-generated results. The differences are considerably smaller compared to before. 
image
image

HenryRWinterbottom:
@EricSinsky-NOAA Thank you! This was helpful.

I would expected there to be some differences based on the introduction of computations to CDO and numpy precision. 

Do you have a scale for the errors? And if so, are they within the thresholds of acceptable values?

EricSinsky-NOAA:
> @EricSinsky-NOAA Thank you! This was helpful.
> 
> I would expected there to be some differences based on the introduction of computations to CDO and numpy precision.
> 
> Do you have a scale for the errors? And if so, are they within the thresholds of acceptable values?

I am a little concerned about some of the differences around some of the coastlines and the equator. Some of these differences might be related to the differences in how the masks are handled during the remapping in CDO, but it is hard to say for sure. It might also be related to the differences in the interpolation procedure.  I have included the same plots here that I posted before, except the difference plots range from -0.1 to 0.1 m/s.   
image
image

HenryRWinterbottom:
Thank you @EricSinsky-NOAA for your help. It is appreciated.

@NeilBarton-NOAA Can comment on the validity of these results with respect to @EricSinsky-NOAA's plots?

LydiaStefanova-NOAA:
At least part of the problem is what Denise pointed out earlier: the CDO method appears not to be handling things appropriately when there are missing vector component values on the edge of a grid cell (let's say - the south edge). In the NCL/Python approach, these missing values (i.e. the equivalent of "ocean currents over land") are set to 0 and the interpolation is then between the non-missing value on the north edge and the 0 on the south edge. In the CDO approach, it's an interpolation between the non-missing value on the north edge and a missing value on the south edge, with the outcome being a missing value. This has the effect of more missing values near coastlines in the CDO-interpolated version than in the NCL/Python-interpolated version.

HenryRWinterbottom:
> At least part of the problem is what Denise pointed out earlier: the CDO method appears not to be handling things appropriately when there are missing vector component values on the edge of a grid cell (let's say - the south edge). In the NCL/Python approach, these missing values (i.e. the equivalent of "ocean currents over land") are set to 0 and the interpolation is then between the non-missing value on the north edge and the 0 on the south edge. In the CDO approach, it's an interpolation between the non-missing value on the north edge and a missing value on the south edge, with the outcome being a missing value. This has the effect of more missing values near coastlines in the CDO-interpolated version than in the NCL/Python-interpolated version.

That does not explain the differences along the equator in the EPAC, does it? I get that it will impact the coastal values but I am struggling to understand how the mid-ocean values will be impacted.

jiandewang:
> > At least part of the problem is what Denise pointed out earlier: the CDO method appears not to be handling things appropriately when there are missing vector component values on the edge of a grid cell (let's say - the south edge). In the NCL/Python approach, these missing values (i.e. the equivalent of "ocean currents over land") are set to 0 and the interpolation is then between the non-missing value on the north edge and the 0 on the south edge. In the CDO approach, it's an interpolation between the non-missing value on the north edge and a missing value on the south edge, with the outcome being a missing value. This has the effect of more missing values near coastlines in the CDO-interpolated version than in the NCL/Python-interpolated version.
> 
> That does not explain the differences along the equator in the EPAC, does it? I get that it will impact the coastal values but I am struggling to understand how the mid-ocean values will be impacted.

a typical value of SSU is ~0.2m/s in the middle of ocean. However in equator (pretty narrow band, 7S-7N)the currents are strong (>1m/s). So we see relative large difference in the plot.

HenryRWinterbottom:
@EricSinsky-NOAA Can you please do a similar evaluation/comparison for the variables in the RDHPCS Hera file `/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/reset_missing.nc`?

LydiaStefanova-NOAA:
Yes, I agree that is only part of the problem (here's a pic illustrating the coastline effect): 
image. Another thing that this image shows is the different treatment of the tripole seam between the two methods (left blank in NCL/Python version, interpolated over in CDO version). It's possible that for some purposes it may be desirable to know where it is; I don't have a strong opinion one way or another, just pointing it out as a difference

I don't know what's going on at the equator, but here's a zoomed in version of in a small (8x10deg) domain across the equator. Note that the scale of the differences is about 10% of the scale for the individual fields. 
image

DeniseWorthen:
The rotation from MOM6 tripole I-J to EW can best be seen in the cap

```
      ocz_rot(i,j) = ocean_grid%cos_rot(ig,jg)*ocz(i,j) + ocean_grid%sin_rot(ig,jg)*ocm(i,j)
      ocm_rot(i,j) = ocean_grid%cos_rot(ig,jg)*ocm(i,j) - ocean_grid%sin_rot(ig,jg)*ocz(i,j)
```
ocz = zonal, ocm=merid

I can't be sure, but it looks like the CDO script is doing 

```
"xr=${cosang}*${xvar}+${sinang}*${yvar}; yr=${cosang}*${xvar}-${sinang}*${yvar}"
```

where xr=rotated u and yr=rotated v.

The ``yr`` variable does not appear to be correct.

HenryRWinterbottom:
@DeniseWorthen Please update your branch. This error was detected and corrected.

LydiaStefanova-NOAA:
@HenryWinterbottom-NOAA 

It looks like taking care of the missing values doesn't solve the differences problem. This is what reset_missing.nc looks like for the Arctic small region: 
image
For the equatorial region, it hasn't changed from before.

HenryRWinterbottom:
@LydiaStefanova-NOAA The reason for the different treatment is an artifact of ESMF. When evaluating a grid with multiple poles, singularities can occur. These have been cited as "degenerative points" and discussed more [here](https://github.com/JiaweiZhuang/xESMF/issues/22#issuecomment-403072638). 

The remapping using CDO understands multiple types of curvilinear and unstructured grid types.

HenryRWinterbottom:
@LydiaStefanova-NOAA How do we know which solution is correct?

DeniseWorthen:
I just checked out your feature/cdo_post branch. What I see in remap.sh is:

yr=cos(${theta})*${xvar}-sin(${theta})*${yvar}

I believe it should be 

yr=cos(${theta})*${yvar}-sin(${theta})*${xvar}

to agree with the cap rotations

ocm_rot(i,j) = ocean_grid%cos_rot(ig,jg)*ocm(i,j) - ocean_grid%sin_rot(ig,jg)*ocz(i,j)

HenryRWinterbottom:
@DeniseWorthen Yes, that was the correction I made.

I apologize, I made the commit but didn't push to my branch. Thank you for confirming however.

LydiaStefanova-NOAA:
@HenryWinterbottom-NOAA If in "how do we know which is correct" you are referring to the treatment of missing values near coastlines, treating them as 0s seems more reasonable to me because that really is the speed of currents over land. But I assume you mean the bigger question - how do we know that the Python/NCL version is correct over open ocean (like around the equator)?

jiandewang:
> Yes, I agree that is only part of the problem (here's a pic illustrating the coastline effect): image. Another thing that this image shows is the different treatment of the tripole seam between the two methods (left blank in NCL/Python version, interpolated over in CDO version). It's possible that for some purposes it may be desirable to know where it is; I don't have a strong opinion one way or another, just pointing it out as a difference
> 
> I don't know what's going on at the equator, but here's a zoomed in version of in a small (8x10deg) domain across the equator. Note that the scale of the differences is about 10% of the scale for the individual fields. image

this remain me the question of "pair". How does cdo know that U/V come as a pair in interpolation ? So if you do
(1) cdo rmapbil,..........variable=SSU   
(2) cdo remapbil,.......variable=SSU,SSV
(3) cdo remapbil,.......variable=SSU,SSV,SST,SSH,SW.........
can we expect to get different results of SSU between (1) and (2) ?
I reacll last week when Henry gave me a sample regular T grid file, the difference of SST is ~E-3 to E-4. But here we are seeing 10% of difference in currents over the open ocean. This difference can't be explained by the difference in interpolation algorithm between two tools.

HenryRWinterbottom:
@LydiaStefanova-NOAA Yes, how do we know what the correct interpolation method is.

jiandewang:
> @LydiaStefanova-NOAA Yes, how do we know what the correct interpolation method is.

if one is wrong while the other is correct, how can we get so close of SST (differennce on the order of E-3 to E-4) but 10% difference in U/V ?

HenryRWinterbottom:
> > @LydiaStefanova-NOAA Yes, how do we know what the correct interpolation method is.
> 
> if one is wrong while the other is correct, how can we get so close of SST (differennce on the order of E-3 to E-4) but 10% difference in U/V ?

That is exactly my point. How do we know which vector interpolations are correct? Just because a method "works" doesn't mean that it is correct.

WalterKolczynski-NOAA:
If necessary, we should be able to have an intermediary step where we replace missing values with 0 prior to interpolation. Should be a simple NCO call.

HenryRWinterbottom:
> If necessary, we should be able to have an intermediary step where we replace missing values with 0 in a temporary file prior to interpolation. Should be a simple NCO call.

@WalterKolczynski-NOAA I have already done this. It will be included in PR #1875.

jiandewang:
> > > @LydiaStefanova-NOAA Yes, how do we know what the correct interpolation method is.
> 
> > 
> 
> > if one is wrong while the other is correct, how can we get so close of SST (differennce on the order of E-3 to E-4) but 10% difference in U/V ?
> 
> 
> 
> That is exactly my point. How do we know which vector interpolations are correct? Just because a method "works" doesn't mean that it is correct.

Here I mean on open ocean, nothing to do with rotation 
I want to know how does cdo know the two variables are a pair?

HenryRWinterbottom:
@jiandewang 
https://github.com/HenryWinterbottom-NOAA/global-workflow/blob/feature/cdo_post/ush/remap.sh#L232

LydiaStefanova-NOAA:
I think Jiande's point is that near the equator rotation is not in play

DeniseWorthen:
@LydiaStefanova-NOAA pointed me to these three files:

original tripole: /scratch1/NCEPDEV/climate/Jiande.Wang/For-others/For-Eric/ocn2013100106.01.2013100100.nc
python: /scratch1/NCEPDEV/climate/Jiande.Wang/working/Eric-regrid/plot-Henry-test_pr-nc/WC2-Eric-025-regular.nc
cdo: /scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/reset_missing.nc

I'm looking at a point which is 5N, 224W (136E).

Tripole grid, ssu and ssv using ferret's internal regrid-to-tgrid: 
```
list/d=1/x=224w/y=5n ssu[g=temp],ssv[g=temp]
I / *:    0.1028 -0.2116
```

Python script 
```
list/d=2/x=224w/y=5n/k=1 ssu,ssv
I / *:     0.1310 -0.2078
```

CDO
```
list/i=897/j=381/d=3 lat,lon,ssu,ssv
I / *:     5.000  224.0 -0.05027  0.2568
```

HenryRWinterbottom:
Please try the following file: `/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/vector_reorder.nc`

EricSinsky-NOAA:
> Please try the following file: `/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/vector_reorder.nc`

@HenryWinterbottom-NOAA After comparing the CDO-generated results (`/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/vector_reorder.nc`) to the Python-generated results, it looks like the differences have reduced considerably.  

comparison_maps_ncl_vs_python_SSU
comparison_maps_ncl_vs_python_SSV

EricSinsky-NOAA:
> > Please try the following file: `/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/vector_reorder.nc`
> 
> @HenryWinterbottom-NOAA After comparing the CDO-generated results (`/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/vector_reorder.nc`) to the Python-generated results, it looks like the differences have reduced considerably.
> 


My apologies, I accidentally compared @HenryWinterbottom-NOAA's current CDO-generated result to his previous CDO-generated result. Please disregard.

LydiaStefanova-NOAA:
Looking just at the two smaller domains I do see reduced differences for vector_reorder.nc. Only now the tripole seam shows up as a discontinuity, which was't the case with the previous versions (including reset_missing.nc)
image
image

EricSinsky-NOAA:
Thank you for checking those differences, @LydiaStefanova-NOAA. Also, after further checking, it turns out that my results actually are correctly comparing the CDO-generated results (vector_reorder.nc) to the Python-generated results.

HenryRWinterbottom:
> Looking just at the two smaller domains I do see reduced differences for vector_reorder.nc. Only now the tripole seam shows up as a discontinuity, which was't the case with the previous versions (including reset_missing.nc) image image

@LydiaStefanova-NOAA Thank you for taking a look. The issue with a tripolar projection is that different interpolation schemes will handle it differently. It appears that ESMF and CDO are a case of that.

LydiaStefanova-NOAA:
@HenryWinterbottom-NOAA For the tripole seam, currently the CDO approach appears to put 0s instead of setting missing values or interpolating -- (in addition to the small map showing the dark blue strip, see this cut across it): 
image. 
To take care of this (and the land mask), once all the interpolations are completed, the interpolated T-grid (for which the missing values should be retained to begin with) at each depth should be used to mask out the interpolated U-grid results at each depths.

HenryRWinterbottom:
@LydiaStefanova-NOAA The `zeroes` is likely due to the fact that the missing values were reset to 0.0 prior to doing the interpolation as was noted above -- although this didn't seem to make a difference with respect to the rotation issue.

Please see RDHPCS Hera file `/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/vector_reorder_nonzero.nc` for modifications.

LydiaStefanova-NOAA:
Ugh :(. This gets rid of the discontinuity but I think the rotation is back to being wrong -- using /scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/vector_reorder_nonzero.nc: 
image

HenryRWinterbottom:
@LydiaStefanova-NOAA OK, the difference was what I expected. I was previously trying to resolve the discontinuity. 

Please check the updated RDHPCS Hera file `/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/vector_reorder_nonzero.nc` and let me know what you learn.

LydiaStefanova-NOAA:
Ok - now it's without the discontinuity, but also without the points in which the original grid has one missing value at the cell edge (so we are back to there being a larger gap at continents' edges). 
image
You had one earlier version, /scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/reset_missing.nc , that was without the tripole seam discontinuity, but with interpolation to 0s at the coast (the downside of that version was that the differences wrt the python version were larger and more widespread).

HenryRWinterbottom:
> Ok - now it's without the discontinuity, but also without the points in which the original grid has one missing value at the cell edge (so we are back to there being a larger gap at continents' edges). image You had one earlier version, /scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/reset_missing.nc , that was without the tripole seam discontinuity, but with interpolation to 0s at the coast (the downside of that version was that the differences wrt the python version were larger and more widespread).

@LydiaStefanova-NOAA The "gap" is due to how the mask is treated for ESMF versus CDO.

All of the above being noted, how do you suggest that we proceed?

LydiaStefanova-NOAA:
@HenryWinterbottom-NOAA Of these choices, my vote is for the version in /scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/vector_reorder.nc which looked like this 
image, followed by applying a mask based on the interpolated T-grid (where the T-grid interpolation is done without setting any of the missing data to 0).

parm/post/mom6_interp.csv
```diff
@@ -0,0 +1,18 @@
...
+sensible remapbil MOM6_INTERP_NC 0
+SW remapbil MOM6_INTERP_NC 0
+LW remapbil MOM6_INTERP_NC 0
+evap remapbil MOM6_INTERP_NC 0
+lprec remapbil MOM6_INTERP_NC 0
+LwLatSens remapbil MOM6_INTERP_NC 0
+Heat_PmE remapbil MOM6_INTERP_NC 0
+SSU,SSV remapbil MOM6_INTERP_NC 1 cos_rot,sin_rot
+uo,vo remapbil MOM6_INTERP_NC 1 cos_rot,sin_rot
+taux,tauy remapbil MOM6_INTERP_NC 1 cos_rot,sin_rot
```
EricSinsky-NOAA: To be consistent with the NCL/Python remapping programs, The following variables should use conservative remapping (referred to as remapcon in CDO):

1. latent
2. sensible
3. SW
4. LW
5. evap 
6. lprec
7. fprec
8. LwLatSens
9. Heat_PmE
10. taux,tauy

ush/remap.sh
```diff
@@ -0,0 +1,327 @@
...
+        echo "Remapping variables ${varname} with rotation."
+        cdo_rotate "${varname}" "${srcgrid}" "${interp_type}" "${angle}"
+	
+    else
+        echo "FATAL ERROR: Rotation option ${rotate} not recognized. Aborting!!!"
+        exit 101
+    fi
+    
+done < "${variable_file}"
+
```
EricSinsky-NOAA, line 328: To fit into the rest of the global workflow, it is suggested that the final remapped netcdf output follows the format of the following NetCDF file on Hera in terms of the global attributes, variable attributes, and coordinate names: `/scratch2/NCEPDEV/ensemble/noscrub/Eric.Sinsky/MOM6/ocnr2013100106.1.2013100100_0p25x0p25_MOM6.nc`.  This output is needed as input for the reg2grb2 program.

EricSinsky-NOAA (changes requested):

ush/remap.sh
HenryRWinterbottom: @EricSinsky-NOAA OK, this is going to need to be a step after application of `remap.sh`. This application, #1871,  is agnostic to the forecast model. Similar to PR #1875, another step will be need to update the attributes accordingly.

ush/remap.sh
```diff
@@ -0,0 +1,329 @@
...
+        cosang="${out_string}"
+	cdo_remap "${cosang}" "${varfile}" "${interp_type}" "${nc_output_path}" 
+        strip_whitespace "${angle_array[1]}"
+        sinang="${out_string}"
+	cdo_remap "${sinang}" "${varfile}" "${interp_type}" "${nc_output_path}" 	
+	cdo -expr,"xr=${cosang}*${xvar}+${sinang}*${yvar}; yr=${cosang}*${yvar}-${sinang}*${xvar}" -selname,"${xvar}","${yvar}","${cosang}","${sinang}" "${nc_output_path}" "${var_rotate_path}" >> "${REMAP_LOG}" 2>&1 
+	varname_update "xr" "${xvar}" "${var_rotate_path}"
+	cdo_remap "${xvar}" "${var_rotate_path}" "${interp_type}" "${output_path}"
+	varname_update "yr" "${yvar}" "${var_rotate_path}"
+	cdo_remap "${yvar}" "${var_rotate_path}" "${interp_type}" "${output_path}"
```
EricSinsky-NOAA, line 250: For the vector remapping, it is suggested to do the following so that there is more consistency with the NCL/Python remapping programs:

1. Replace missing values with 0.0 before any interpolation or vector rotation. 
2. On the interpolated u and v fields, for grid points where the interpolated mask equals 1 (non-ocean points), set those values on the interpolated u and v fields to a netcdf FillValue. This mask can be derived from a 2D and 3D scalar variable (e.g. SST for the 2D variable and Temperature for the 3D variable).  

I think once the above 2 items are added, the CDO remapping will be in great shape.

EricSinsky-NOAA (changes requested):

ush/remap.sh
EricSinsky-NOAA: Thank you, @HenryWinterbottom-NOAA. If there is any more information you need that could help, please let me know.

ush/remap.sh
HenryRWinterbottom: @EricSinsky-NOAA:

Replace missing values with 0.0 before any interpolation or vector rotation.

This was explored and produced a discontinuity due to the tripolar remapping as noted by @LydiaStefanova-NOAA . Therefore this does not seem like a usable solution.

The CDO application does not provide a way, at least that I am aware of, to handle the mask as you are suggesting. If there is a feature that I am unaware of I would ask that you fork my branch and attempt to add it. Otherwise this issue and PR need to be closed ASAP.

EricSinsky-NOAA: There is no CDO feature that I am currently aware of that can handle masks in such a way. It may require some more time and attention to add these features. However, since these items are only suggestions and not required for the CDO-based remapping to fit in the global-workflow, these items can be addressed in a separate issue and future PR only if needed, and if there are no objections from anyone. As `remap.sh` currently stands, I still think the results are considerably close to the original NCL remapping program `ush/ocnpost.ncl`.

parm/post/mom6_interp.csv
GwenChen-NOAA, line 1: 1. `speed` should not be directly interpolated using remapbil. It should be recalculated using the remapped `SSU` and `SSV`. `speed` = sqrt(`SSU`^2 + `SSV`^2)
2. `evap` is not required to output in operational forecast files and can be removed.
3. `lprec` is not required to output in operational forecast files and can be removed.

HenryRWinterbottom: @GwenChen-NOAA Thank you for your feedback. These values were provided to me by @jiandewang.

Those three variables can be removed from the comma-delimited list. Since speed can be (re)calculated it should be the job of a script/application beyond that of `remap.sh`.

jiandewang: speed is outputted internally from model so there is no need to recalculate it

ush/remap.sh
DeniseWorthen: I do not follow this repo, so I apologize if I'm out-of sync. A basic question: How (or where) does CDO obtain the land mask for the destination grid? 

In the NCL/Python versions, the destination land mask can be thought of as an "output", meaning the destination land mask is given by the points where a source grid ocean point cannot be mapped fully to a destination grid point (ie, the destination point would be represented by a combination of a land and ocean point on the source grid). This is more and more important as the bathymetry closes in since the ocean model bathymetry will not in general match something like etopo5. 

This is an entirely separate issue than the polar seam. That is the result of the fact that the last latitude on the T-cell tripole grid is at 89.64 and so no T-cell grid point can be mapped to 90N.

parm/post/mom6_interp.csv
GwenChen-NOAA: @jiandewang, `speed` in the tripolar netCDF files is outputted from model and no need to change. `speed` in the latlon netCDF or GRIB2 files needs to be recalculated, because `SSU` and `SSV` have been remapped to a new grid. This is how wind speed and direction are handled in UPP. Directly interpolating `speed` to a new grid treats `speed` as a scalar variable, which is incorrect.

@HenryWinterbottom-NOAA, speed calculation is a simple function. Should be easy to add it into `remap.sh` and append the output into the latlon netCDF files.

HenryRWinterbottom: > @jiandewang, `speed` in the tripolar netCDF files is outputted from model and no need to change. `speed` in the latlon netCDF or GRIB2 files needs to be recalculated, because `SSU` and `SSV` have been remapped to a new grid. This is how wind speed and direction are handled in UPP. Directly interpolated `speed` to a new grid treats `speed` as a scalar variable, which is incorrect.
> 
> @HenryWinterbottom-NOAA, speed calculation is a simple function. Should be easy to add it into `remap.sh` and append the output into the latlon netCDF files.

Thank you for following up @GwenChen-NOAA. The `remap.sh` script is not intended for a specific application (i.e., it is agnostic to the respective component model output to be remapped). The purpose is to generate netCDF files containing the specified/respective variables to be defined on a new destination grid. It does not care whether it is an ice or an ocean or any other component model variable fields that are being remapped. Therefore, the `speed` in this case should be computed elsewhere, not within `remap.sh`. 

Please see my branch [here](https://github.com/HenryWinterbottom-NOAA/global-workflow/tree/feature/cdo_post).

GwenChen-NOAA: Thanks, @HenryWinterbottom-NOAA! I'm new to global-workflow. Perhaps @aerorahul or @WalterKolczynski-NOAA can advise where this function should be in global-workflow if @jiandewang wants `speed` to be outputted in latlon netCDF or GRIB2 files. I personally is fine to leave it out and let users calculate it.

HenryRWinterbottom: @GwenChen-NOAA No problem at all and I hope my explanation was helpful.

ush/remap.sh
EricSinsky-NOAA: @DeniseWorthen It is unclear what CDO uses as a land mask for the destination grid by default. It does look like there is a way to create a mask file and apply it to the interpolated field such as in this example on the CDO forum: https://code.mpimet.mpg.de/boards/53/topics/10933

Here are the basics of what I gathered from this example so far:

1. Interpolate field: `cdo -remapbil,r360x180 infile.nc infile_r360x180.nc`
2.  create mask file with the same resolution as the destination grid: `cdo -f nc -remapbil,r360x180 -topo topo_r360x180.nc`
3. mask the land areas using the mask file: `cdo -expr,'topo = ((topo<0.0)) ? 1.0 : topo/0.0' topo_r360x180.nc mask_land.nc`
4. mask interpolated field with the mask file: `cdo -mul mask_land.nc infile_r360x180.nc infile_r360x180_mask_land.nc`

parm/post/mom6_interp.csv
HenryRWinterbottom: @NeilBarton-NOAA @GwenChen-NOAA Is there a consensus for how we (the global-workflow team) should proceed with this PR? We would like to close out issue #923 and PR #1871 as possible. Thank you.

NeilBarton-NOAA: @aerorahul and I talked a little about this before his A/L. I'll comment once we have a more detail conversation

GwenChen-NOAA: @HenryWinterbottom-NOAA, can you point me to a sample latlon netCDF file generated using the latest code? I need it for testing. Thanks!

HenryRWinterbottom: @GwenChen-NOAA The following contains a sample file that was used for debugging:

`/scratch1/NCEPDEV/da/Henry.Winterbottom/scratch/Neil.Barton/vector_reorder_nonzero.nc`

It contains the analysis/forecast variables `SSH`, `SSU`, and `SSV`.

GwenChen-NOAA: @HenryWinterbottom-NOAA, I looked at your sample data. The remapped SSH looks very good to me. The remapped SSU and SSV have more grid points along the coast with missing data, which may be caused by the rotation and land-sea mask handling. Since most variables are scalar, I think this PR is sufficient. Thanks for your good work!

vector_reorder_nonzero ssh
vector_reorder_nonzero ssu

GwenChen-NOAA (approved):

ush/remap.sh
DeniseWorthen: Thanks @EricSinsky-NOAA I took some time and rewrote the entire original ocnpost.ncl routine in Fortran. It reproduces the fields created by the original NCL to O~-8 in all fields: https://github.com/DeniseWorthen/ocnpost90

I wasn't asked to do this, but a fortran code is a) portable b) fast and c) debuggable.

parm/post/mom6_interp.csv
HenryRWinterbottom: @GwenChen-NOAA Thank you for the feedback. 

I will begin the PR and send to @aerorahul and @WalterKolczynski-NOAA for final approval.

ush/remap.sh
WalterKolczynski-NOAA: @DeniseWorthen Thanks for that. I suspect this will be what we ultimately go with. Once Rahul is back on Weds we will discuss the best way to proceed.

Thanks to everyone who's been working on this, particularly those who wrote stuff that may not be used, which I know can be frustrating.

EricSinsky-NOAA: @DeniseWorthen Thank you for writing your NCL code in FORTRAN! I do agree that FORTRAN is a great option too for this sort of application, particularly because of its efficiency and flexibility. At first glance, your FORTRAN ocnpost program looks very close to your original NCL ocnpost program.

Issue #1629, opened 2023-05-18
Add flexibility for ocean/ice output time-steps


JessicaMeixner-NOAA:
**Description**
Currently it's hard-coded in the diag_table that ocean output is 6 hourly: https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/parm_fv3diag/diag_table#L3 

However, @jiandewang  found that there are currently constraints in that the ocean/ice (which must be consistent because of the currently combined grib output), must also be the same as the forecast model due to these lines in the scripts: 
ush/forecast_postdet.sh: 
```
   fhrlst=$OUTPUT_FH

    for fhr in $fhrlst; do
      if [ $fhr = 'anl' ]; then  # Looking at OUTPUT_FH, this is never true. TODO: remove this block
        continue
      fi
      VDATE=$($NDATE $fhr $IDATE)
      YYYY=$(echo $VDATE | cut -c1-4)
      MM=$(echo $VDATE | cut -c5-6)
      DD=$(echo $VDATE | cut -c7-8)
      HH=$(echo $VDATE | cut -c9-10)
      SS=$((10#$HH*3600))
```

**Requirements**
Ideally we can run without hard-coding the 6-> (atm time step) and have the ocean/ice model have different time steps.   

**Acceptance Criteria (Definition of Done)**
Be able to run with 3 hourly atm time-steps and 6 hourly ocean or 3 hourly atm and ocn/ice and not have to change anything other than a config file.

aerorahul:
@JessicaMeixner-NOAA 
The lines referred to in the description are pointing to an aerosol section.  Can you please update the issue?  Thanks!

jiandewang:
see https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/forecast_postdet.sh#L773-L787
here Vdate relies on Idate, but Idate is defined on fv3 side

aerorahul:
actually, `IDATE=CDATE` on https://github.com/NOAA-EMC/global-workflow/blob/ae56219c0059bbe6d34044df490023e17225e6f8/ush/forecast_postdet.sh#L758.
`CDATE` is defined at the top as the current cycle and does not change.

aerorahul:
@jiandewang 
I don't understand the issue here.
The output frequency is controlled via the `diag_table`.  It is currently set to 6 hourly.  In addition, it seems like these are 6 hour averages, not instantaneous.
[...]

PR #1876, opened 2023-09-22
Merge S2S defaults config back into other configs

### LINKED ISSUE #1606: Meld config.defaults.s2sw back into other configs
**Description**
When the coupled system was first spinning up, a number of settings were different from legacy runs so we created the `config.defaults.s2sw` to override these settings without impacting the existing system. Now that we run S2S as a matter of course, are aligning things for v17, and have other ways to handle some of these when necessary, it is time to merge any settings that remain different into their standard location and retire the S2S defaults config file. 

**Requirements**
- S2SWA runs with the same settings as it currently does
- Atm-only still functions (possibly with a couple different settings)
- No `config.defaults.s2sw` file

Requirements are met

**Dependencies**
None

WalterKolczynski-NOAA:
# Description
Meld S2SW config back into regular files

When the prototype S2S was first added, we placed some settings in a `config.defaults.s2sw` file to avoid interference with existing settings. Now that we are moving towards v17 and run S2S regularly, it is time to merge these settings back into the other config files.

The atm output frequency (`$FHOUT_GFS`) is reduced to 6 hrs. We may elect to keep this change, but it is necessary right now for S2S experiments as the ocean output frequency is hard-coded to 6 hrs but the links from the forecast directory to COM are created based on the value of `$FHOUT_GFS`. Resolving #1629 should remove this limitation.

The wave output frequency settings are updated to the coarser frequency as that is more suitable for development. The prior (ops) settings are noted as comments.

There are a couple of settings that control ice fields within FV3. These are placed in `config.ice`, so when the ice model is on those settings are still used.

The forecast length overrides are just dropped as the existing ones have already been shortened (though not as much) to be suitable for most development.

"Gaussian" archive is updated to the former override values.

Resolves #1606
Refs #1629 

# Type of change

# Change characteristics
- Is this a breaking change (a change in existing functionality)? YES (some default settings have changed)

# How has this been tested?
- Cycled atm-only on Orion
- P8 on Orion

# Checklist


parm/config/gfs/config.base.emc.dyn
```diff
@@ -264,7 +262,7 @@ export FHMAX_GFS_12=${FHMAX_GFS_12:-120}
 export FHMAX_GFS_18=${FHMAX_GFS_18:-120}
 current_fhmax_var=FHMAX_GFS_${cyc}; declare -x FHMAX_GFS=${!current_fhmax_var}
 
-export FHOUT_GFS=${FHOUT_GFS:-3}
+export FHOUT_GFS=${FHOUT_GFS:-6} # Must be 6 for S2S until #1629 is addressed; 3 for ops
```
aerorahul: Does this need to be `FHOUT_GFS=${FHOUT_GFS:-6}`?  Can it not just be `FHOUT_GFS=6`?  Is there some higher level configuration passing `FHOUT_GFS` to `config.base`?

parm/config/gfs/config.wave
```diff
@@ -99,18 +99,18 @@ fi
...
 
 # gridded and point output rate
 export DTFLD_WAV=$(( FHOUT_HF_WAV * 3600 ))
-export DTPNT_WAV=${DTPNT_WAV:-3600}
+export DTPNT_WAV=${DTPNT_WAV:-10800} # 3600 for ops
 export FHINCP_WAV=$(( DTPNT_WAV / 3600 ))
 
 # Selected output parameters (gridded)
-export OUTPARS_WAV=${OUTPARS_WAV:-"WND HS FP DP PHS PTP PDIR"}
+export OUTPARS_WAV=${OUTPARS_WAV:-"WND CUR ICE HS T01 T02 DIR FP DP PHS PTP PDIR CHA"}
```
aerorahul: Same question as in `config.base` for `FHOUT_GFS`.
Infact, there are several variables in here that are defined as `var=${var:-default_value}`.  Do they need to be this way now that `config.defaults.s2sw` is no longer being sourced to provide those?

aerorahul (changes requested):
looks ok.  would like to get some clarifications on the questions below.

parm/config/gfs/config.wave
WalterKolczynski-NOAA: Probably not. Can remove if you'd like.

WalterKolczynski-NOAA: Updated.

parm/config/gfs/config.ice
```diff
@@ -2,4 +2,8 @@
 
 echo "BEGIN: config.ice"
 
+# Override atm-only FV3 settings when ice model is on
+export min_seaice="1.0e-6"
```
JessicaMeixner-NOAA, line 6: Both of these are variables set in input.nml .  Do the tests that were run confirm that there are no answer changes?

parm/config/gfs/config.wave
```diff
@@ -99,18 +99,18 @@ fi
...
-export FHMAX_HF_WAV=${FHMAX_HF_WAV:-120}
+export FHMAX_HF_WAV=${FHMAX_HF_WAV:-0} # 120 for ops
 export FHOUT_HF_WAV=${FHOUT_HF_WAV:-1}
 export FHMAX_WAV_IBP=180
 if (( FHMAX_WAV < FHMAX_WAV_IBP )); then export FHMAX_WAV_IBP=${FHMAX_GFS} ; fi
 
 # gridded and point output rate
 export DTFLD_WAV=$(( FHOUT_HF_WAV * 3600 ))
-export DTPNT_WAV=${DTPNT_WAV:-3600}
+export DTPNT_WAV=${DTPNT_WAV:-10800} # 3600 for ops
```
JessicaMeixner-NOAA: Can we leave this as the ops setting?

parm/config/gfs/config.wave
```diff
@@ -99,18 +99,18 @@ fi
 export WAVHINDH=${WAVHINDH:-0}
 export FHMIN_WAV=${FHMIN_WAV:-0}
 export FHOUT_WAV=${FHOUT_WAV:-3}
-export FHMAX_HF_WAV=${FHMAX_HF_WAV:-120}
+export FHMAX_HF_WAV=${FHMAX_HF_WAV:-0} # 120 for ops
```
JessicaMeixner-NOAA: Can we leave this with the ops setting?

parm/config/gfs/config.wave
```diff
@@ -99,18 +99,18 @@ fi
...
 
 # gridded and point output rate
 export DTFLD_WAV=$(( FHOUT_HF_WAV * 3600 ))
-export DTPNT_WAV=${DTPNT_WAV:-3600}
+export DTPNT_WAV=${DTPNT_WAV:-10800} # 3600 for ops
 export FHINCP_WAV=$(( DTPNT_WAV / 3600 ))
 
 # Selected output parameters (gridded)
-export OUTPARS_WAV=${OUTPARS_WAV:-"WND HS FP DP PHS PTP PDIR"}
+export OUTPARS_WAV="WND CUR ICE HS T01 T02 DIR FP DP PHS PTP PDIR CHA"
```
JessicaMeixner-NOAA, line 113: Can we at least make a comment here about what the ops settings for v16 were?  This is not critical, but would be nice.

JessicaMeixner-NOAA (changes requested):
The requested changes are for the updates to config.ice, I'm pretty sure these should be in something that's for input.nml//atmos model, not the ice model even though these are "ice" related variables.

parm/config/gfs/config.ice
WalterKolczynski-NOAA: I can confirm the namelists use these settings. I have not checked the output for bitwise indenticality.

parm/config/gfs/config.wave
WalterKolczynski-NOAA: Done

parm/config/gfs/config.wave
WalterKolczynski-NOAA: Done.

parm/config/gfs/config.wave
WalterKolczynski-NOAA: The v16 ops settings can always be referred to from one of the v16 tags on GitHub.  Retaining v16 information in develop for all components could lead to comment clutter.

JessicaMeixner-NOAA: It would have been nice to have had a setting saying what the previous ops settings were as we used the defaults.s2s to overwrite those for development options.  But you are correct that they do indeed exist somewhere else, so this can be skipped.  That being said, comments can be very useful and "comment clutter" is very subjective to the person who finds the comment useful.

aerorahul (approved):
LGTM

parm/config/gfs/config.ice
JessicaMeixner-NOAA: Okay, it'd might be good to check for bitwise identity too and to make sure this isn't being sourced for atm-only case as mentioned in the comments.  Right now I can't seem to find where the sourcing of this config is happening - but perhaps the logic has just changed and I haven't had time to understand but I don't see where this config is sourced for the forecast job if coupled.    I'll try to test this tomorrow unless you have saved output that we can compare results with.

WalterKolczynski-NOAA: It gets sourced by config.fcst based on the DO_<COMPONENT> switches:

https://github.com/NOAA-EMC/global-workflow/blob/527aa16d8e003ddb65e6ed4e1b5b47fb79fa5921/parm/config/gfs/config.fcst#L23-L29

JessicaMeixner-NOAA: Thanks for pointing this out!  If the output is all the same as before and has been confirmed, this all looks good to me.  Although I'm still not 100% sure I'd leave this as an "ice" configuration variables as these are really atm variables that are associated with  ice  not ice variables.   Will people know to look for these here? This isn't where I'd look personally.

WalterKolczynski-NOAA: Was finally able to check the files and they are bitwise identical.

JessicaMeixner-NOAA: Thanks for doing this check @WalterKolczynski-NOAA !!!!

JessicaMeixner-NOAA (approved):

PR #1899, opened 2023-09-29
CI driver_weekly bugfix deleting new branch if is already on repo

TerrenceMcGuinness-NOAA:
# Description
Bugfix and updates to weekly CI test that deletes the new modfied branch  if it is already on the repo but now also makes sure that the intermediary branch that creates the PR is coming from **emcbot**.

# Type of chang
- Bugfix (CI weekly test can fail if ran twice)
- Does not push branch to NOAA-EMC global-workflow branch but comes from emcbot instead

# How has this been tested?
- Ran the script in cron on Orion (didn't work on Hera because gh is to old there)

### DIFF
--- a/ci/scripts/driver_weekly.sh
+++ b/ci/scripts/driver_weekly.sh
@@ -75,13 +75,14 @@ git checkout -b "${branch}"
 # and push new branch for PR weekly CI tests to GitHub
 REPO_OWNER="emcbot"
 REPO_NAME="global-workflow"
+REMOTE_NAME="${REPO_OWNER}"
 
 rm -Rf ci/cases/pr
 mv ci/cases/weekly ci/cases/pr
 git add ci/cases
 git commit -m "Moved weekly cases files into pr for high resolution testing"
 
-git remote add upstream "git@github.com:${REPO_OWNER}/${REPO_NAME}.git"
+git remote add "${REMOTE_NAME}" "git@github.com:${REPO_OWNER}/${REPO_NAME}.git"
 
 set +e
 # Delete the branch if it exists
--- a/ci/scripts/driver_weekly.sh
+++ b/ci/scripts/driver_weekly.sh
@@ -86,7 +86,7 @@ git remote add "${REMOTE_NAME}" "git@github.com:${REPO_OWNER}/${REPO_NAME}.git"
 
 set +e
 # Delete the branch if it exists
-git ls-remote --exit-code upstream "${branch}"
+git ls-remote --exit-code "${REMOTE_NAME}" "${branch}"
 ci_status=$?
 if [[ "${ci_status}" == '0' ]]; then
     git push "${REMOTE_NAME}" --delete "${branch}"
--- a/ci/scripts/driver_weekly.sh
+++ b/ci/scripts/driver_weekly.sh
@@ -90,7 +90,7 @@ git push --set-upstream origin "${branch}"
 # Create Pull Request using GitHub CLI and add labels for CI testing
 ####################################################################
 
-REPO_OWNER="NOAA-EMC"
+REPO_OWNER="TerrenceMcGuinness-NOAA"
 REPO_NAME="global-workflow"
 BASE_BRANCH="develop"
 HEAD_BRANCH="${branch}"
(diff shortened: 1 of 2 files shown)


ci/scripts/driver_weekly.sh
```diff
@@ -22,7 +22,7 @@ set -eux
 # TODO using static build for GitHub CLI until fixed in HPC-Stack
 #################################################################
 export GH=${HOME}/bin/gh
-export REPO_URL=${REPO_URL:-"https://github.com/NOAA-EMC/global-workflow.git"}
+export REPO_URL="ssh://git@ssh.github.com:443/NOAA-EMC/global-workflow.git"
```
aerorahul, line 25: Can I ask why `ssh`?

aerorahul (commented):
Looks good other than the `ssh` bit.

ci/scripts/driver_weekly.sh
TerrenceMcGuinness-NOAA: more secure since this checkout needs to push the changes to the EMC Reop with the updated cases directory  ..  also this change is more about **not** getting REPO from env.

WalterKolczynski-NOAA: I'm still nervous about making/deleting branches in the auth repo. But I agree to not having these in a personal account either. Can emc-bot make a clone of global-workflow where the temp branches could be made?

TerrenceMcGuinness-NOAA: Yes it was doing that before during development.  I will make this change in this PR.

ci/scripts/driver_weekly.sh
```diff
@@ -84,7 +90,7 @@ git push --set-upstream origin "${branch}"
 # Create Pull Request using GitHub CLI and add labels for CI testing
 ####################################################################
 
-REPO_OWNER="NOAA-EMC"
+REPO_OWNER="TerrenceMcGuinness-NOAA"
```
WalterKolczynski-NOAA: Can we use emc-bot for this?

TerrenceMcGuinness-NOAA: Yes updated the script so it now uses **emcbot**

ci/scripts/driver_weekly.sh
```diff
@@ -73,26 +73,38 @@ git checkout -b "${branch}"
...
+REPO_OWNER="emcbot"
+REPO_NAME="global-workflow"
 
 rm -Rf ci/cases/pr
 mv ci/cases/weekly ci/cases/pr
 git add ci/cases
 git commit -m "Moved weekly cases files into pr for high resolution testing"
-git push --set-upstream origin "${branch}"
+
+git remote add upstream "git@github.com:${REPO_OWNER}/${REPO_NAME}.git"
```
WalterKolczynski-NOAA: Recommend using `emcbot` (or maybe just `${REPO_NAME}`) instead of `upstream` for the remote name to avoid confusion since "upsteam" already has a meaning in git parlance.

WalterKolczynski-NOAA (changes requested):

WalterKolczynski-NOAA:
Other than the one quibble about the remote name, I think this is ready if it works (I see Orion failed, but not sure if it is related to these changes).

ci/scripts/driver_weekly.sh
TerrenceMcGuinness-NOAA: yes will do ...

ci/scripts/driver_weekly.sh
```diff
@@ -73,26 +73,39 @@
...
-git push --set-upstream origin "${branch}"
+
+git remote add "${REMOTE_NAME}" "git@github.com:${REPO_OWNER}/${REPO_NAME}.git"
+
+set +e
+# Delete the branch if it exists
+git ls-remote --exit-code upstream "${branch}"
+ci_status=$?
+if [[ "${ci_status}" == '0' ]]; then
+    git push "${REMMOT_NAME}" --delete "${branch}"
```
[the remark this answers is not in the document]
TerrenceMcGuinness-NOAA: oh rats!!!

TerrenceMcGuinness-NOAA (commented):
fixed and double fixed REMOTE_REPO name changes

TerrenceMcGuinness-NOAA:
> Other than the one quibble about the remote name, I think this is ready if it works (I see Orion failed, but not sure if it is related to these changes).

Yeah they can't because this script doesn't actually run the test, the test run itself is entirely separate.

TerrenceMcGuinness-NOAA (commented):
done

WalterKolczynski-NOAA (approved):

PR #1901, opened 2023-10-02
Update vrfy.sh to only run MINMON in cycled mode

### LINKED ISSUE #1076: GSI minmon script should not run in forecast-only mode
**Expected behavior**
The GSI minmon script should be turned off when in forecast-only mode. Under the current configuration the gfsvrfy job fails starting in the second cycle because there is no gsistat file.

**Current behavior**
Minmon is run for gfs CDUMP even in forecast-only mode

**Machines affected**
All

**To Reproduce**
1. Setup a forecast-only experiment of longer than one cycle
2. Witness the `gfsvrfy` job fail in every cycle after the first

**Possible Implementation**
Update [jobs/rocoto/vrfy.sh](https://github.com/NOAA-EMC/global-workflow/blob/develop/jobs/rocoto/vrfy.sh) to only run the VMINMON script if `${MODE}` is cycled.

KateFriedman-NOAA:
# Description

This PR updates the MINMON if-block in the `jobs/rocoto/vrfy.sh` script to only invoke this section when `MODE=cycled`.

Resolves #1076

# Type of change

# Change characteristics

# How has this been tested?

Forecast-only coupled test on WCOSS2.

### DIFF
--- a/jobs/rocoto/vrfy.sh
+++ b/jobs/rocoto/vrfy.sh
@@ -91,7 +91,7 @@ fi
 ###############################################################
 echo
 echo "=============== START TO RUN MINMON ==============="
-if [[ "${VRFYMINMON}" == "YES" && "${PDY}${cyc}" != "${SDATE}" ]]; then
+if [[ "${VRFYMINMON}" == "YES" && "${PDY}${cyc}" != "${SDATE}" && "${MODE}" = "cycled" ]]; then
 
     export M_TANKverfM0="${M_TANKverf}/stats/${PSLOT}/${RUN}.${PDY}/${cyc}"
     export M_TANKverfM1="${M_TANKverf}/stats/${PSLOT}/${RUN}.${PDYm1c}/${pcyc}"


WalterKolczynski-NOAA (commented):
I don't think the other GSI monitoring scripts need to be run either. I think the appropriate way to do this is split the GSI monitoring off into a separate jobs that don't run in f-o (really all these should be separate jobs).

WalterKolczynski-NOAA:
(I did not do a good job of fully spec'ing out the issue when I opened it.)

KateFriedman-NOAA:
> I don't think the other GSI monitoring scripts need to be run either. I think the appropriate way to do this is split the GSI monitoring off into a separate jobs that don't run in f-o (really all these should be separate jobs).

@WalterKolczynski-NOAA I agree...that can be tackled in a separate PR to resolve that part of issue #235. This PR will at least resolve the DEAD job in coupled mode for now until that split work can be done.

WalterKolczynski-NOAA (approved):

PR #1906, opened 2023-10-04
Split clean-up into separate job

### LINKED ISSUE #583: Split clean-up activities off from archive job
Right now the clean-up of older cycles is handled within the archive script. This should be split out into its own script and job.

### LINKED ISSUE #1872: Ensemble archive step does not clean old directories
### What is wrong?

When cycling with global-workflow develop, the enkfgdas directories in the ROTDIR are not being automatically deleted during the earc step.  

When @azadeh-gh ran a recent ATM-only C384/C192 experiment, several days of ensemble data remained in the ROTDIR.  This was observed on Hera; it's unclear if this impacts other machines.  

The log file says the target directories are missing, even though they were present at runtime: (/scratch2/NCEPDEV/stmp3/Azadeh.Gholoubi/TestExperiment/ThompsonMP2/logs/2023040800/enkfgdasearc00.log)

```
+ exgdas_enkf_earc.sh[150]: echo 'No directory /scratch2/NCEPDEV/stmp3/Azadeh.Gholoubi/TestExperiment/ThompsonMP2/enkfgdas.20230402/00/mem001
/analysis/atmos to remove files from, skiping'
```

All of the directory names seem to have a line break after the member number, but I'm not sure if this is part of the issue or not.  Note that since these jobs ran, the enkfgdas directories were deleted by hand to free up some stmp space.  

When the earc step runs, it should automatically remove ensemble data older than a few cycles. 

Hera

### Steps to reproduce

Cycle global-workflow develop ATM-only for at least 5 cycles on Hera.

### Additional information

N/A

_No response_

WalterKolczynski-NOAA:
# Description
Moves the clean-up that was previously done in the archive jobs into their own separate job. The clean-up is also streamlined considerably by using only `COM_TOP` instead of going through every template. There is also additional streamlining/corrections in the function that does the actual removing.

Some settings used by both jobs were elevated to `config.base`. Others only needed for cleanup were moved to the new config for that job.

Also corrects a small error encountered when attempting to rerun an ensemble forecast.

Resolves #583
Resolves #1872 

# Type of change
- Bugfix

# Change characteristics

# How has this been tested?
- Cycled test on Orion

# Checklist


WalterKolczynski-NOAA:
Still working on testing and cleaning things up, but wanted to get a draft up for Rahul to look at if he has time before he goes on leave.

jobs/JGLOBAL_CLEANUP
```diff
@@ -0,0 +1,17 @@
+#! /usr/bin/env bash
+
+source "${HOMEgfs}/ush/preamble.sh"
+source "${HOMEgfs}/ush/jjob_header.sh" -e "cleanup" -c "base"
```
aerorahul: Do you not need to source `config.cleanup` after `config.base`?

scripts/exglobal_cleanup.sh
```diff
@@ -0,0 +1,214 @@
...
+source "${HOMEgfs}/ush/preamble.sh"
+
+###############################################################
+# Clean up previous cycles; various depths
+# PRIOR CYCLE: Leave the prior cycle alone
+# shellcheck disable=SC2153
+GDATE=$(${NDATE} -"${assim_freq}" "${PDY}${cyc}")
+
+# PREVIOUS to the PRIOR CYCLE
+GDATE=$(${NDATE} -"${assim_freq}" "${GDATE}")
```
aerorahul: Please use `date --utc ...`

scripts/exglobal_cleanup.sh
```diff
@@ -0,0 +1,214 @@
...
+done
+
+# Remove archived gaussian files used for Fit2Obs in $VFYARC that are
+# $FHMAX_FITS plus a delta before $CDATE. Touch existing archived
+# gaussian files to prevent the files from being removed by automatic
+# scrubber present on some machines.
+
+if [[ "${RUN}" == "gfs" ]]; then
+    fhmax=$((FHMAX_FITS + 36))
+    RDATE=$(${NDATE} -"${fhmax}" "${PDY}${cyc}")
```
aerorahul: please use `date --utc ...`

jobs/rocoto/cleanup.sh
```diff
@@ -0,0 +1,19 @@
+#! /usr/bin/env bash
+
+source "${HOMEgfs}/ush/preamble.sh"
+
+###############################################################
+# Source FV3GFS workflow modules
+. "${HOMEgfs}"/ush/load_fv3gfs_modules.sh
+status=$?
+[[ ${status} -ne 0 ]] && exit "${status}"
```
aerorahul, line 9: What modules are needed for `cleanup`?

scripts/exglobal_cleanup.sh
```diff
@@ -0,0 +1,214 @@
...
+fi
+
+# Step back every assim_freq hours and remove old rotating directories
+# for successful cycles (defaults from 24h to 120h).
+# Retain files needed by Fit2Obs
+# TODO: This whole section needs to be revamped to remove marine component
+#  directories and not look at the rocoto log.
+GDATEEND=$(${NDATE} -"${RMOLDEND:-24}"  "${PDY}${cyc}")
+GDATE=$(${NDATE} -"${RMOLDSTD:-120}" "${PDY}${cyc}")
+RTOFS_DATE=$(${NDATE} -48 "${PDY}${cyc}")
```
aerorahul: please use `date --utc ...`

scripts/exglobal_cleanup.sh
```diff
@@ -0,0 +1,214 @@
...
+        if [[ -d "${COMIN}" ]] && (( GDATE < CDATE_MOS )); then rm -rf "${COMIN}" ; fi
+    fi
+
+    # Remove any empty directories
+    target_dir="${ROTDIR:?}/${RUN}.${gPDY}/${gcyc}/"
+    if [[ -d "${target_dir}" ]]; then
+        find "${target_dir}" -empty -type d -delete
+    fi
+
+    GDATE=$(${NDATE} +"${assim_freq}" "${GDATE}")
```
aerorahul: please use `date --utc ...`

scripts/exglobal_cleanup.sh
```diff
@@ -0,0 +1,214 @@
...
+        tcyc="${TDATE:8:2}"
+        TDIR="${ROTDIR}/vrfyarch/${RUN}.${tPDY}/${tcyc}"
+        [[ -d ${TDIR} ]] && touch "${TDIR}"/*
+        TDATE=$(${NDATE} +6 "${TDATE}")
+    done
+fi
+
+# Remove $RUN.$rPDY for the older of GDATE or RDATE
+GDATE=$(${NDATE} -"${RMOLDSTD:-120}" "${PDY}${cyc}")
+RDATE=$(${NDATE} -"${FHMAX_GFS}" "${PDY}${cyc}")
```
aerorahul: please use `date --utc ...`

aerorahul (changes requested):
Looks ok to me.  
1. please use `date --utc ...` instead of `$NDATE`

I assume `earc.sh` will have its own cleanup; e.g. `ecleanup.sh`

scripts/exgdas_enkf_earc.sh
```diff
@@ -133,172 +133,4 @@ if [ "${ENSGRP}" -eq 0 ]; then
...
-            readarray file_list < <(printf -- '%s\n' "${file_list_old[@]}" | grep -v "${exclude}")
-            if (( ${#file_list[@]} == 0 )); then return; fi
-        done
-
-        for file in "${file_list[@]}"; do
-            rm -f "${file}"
-        done
-        # Remove directory if empty
-        rmdir "${directory}" || true
-    }
```
HenryRWinterbottom, line 177: @WalterKolczynski-NOAA I am addressing this function while working with Catherine Thomas on issue #1872. I have asked her to test a more robust solution to `remove_files` (see [here](https://github.com/HenryWinterbottom-NOAA/global-workflow/blob/bug/gdas_enkf_scrubbing/scripts/exgdas_enkf_earc.sh#L181)).

Would you prefer that I freeze her issue for now until we determine which is the optimal path forward?

parm/config/gfs/config.base.emc.dyn
```diff
@@ -395,6 +395,9 @@ export ARCH_CYC=00           # Archive data at this cycle for warm_start capabil
 export ARCH_WARMICFREQ=4     # Archive frequency in days for warm_start capability
 export ARCH_FCSTICFREQ=1     # Archive frequency in days for gdas and gfs forecast-only capability
 
-export DELETE_COM_IN_ARCHIVE_JOB="YES"   # NO=retain ROTDIR.  YES default in arch.sh and earc.sh.
+#--online archive of nemsio files for fit2obs verification
+export FITSARC="YES"
+export FHMAX_FITS=132
+[[ "${FHMAX_FITS}" -gt "${FHMAX_GFS}" ]] && export FHMAX_FITS=${FHMAX_GFS}
```
KateFriedman-NOAA, line 400: Why move these out of `config.arch`?

parm/config/gfs/config.cleanup
```diff
@@ -0,0 +1,15 @@
+#! /usr/bin/env bash
+
+########## config.cleanup ##########
+echo "BEGIN: config.cleanup"
+
+# Get task specific resources
+source "${EXPDIR}/config.resources" cleanup
+
+export DELETE_COM_IN_ARCHIVE_JOB="YES"   # NO=retain ROTDIR.  YES default in cleanup.sh
```
KateFriedman-NOAA: Should this be `DELETE_COM="YES"` and be moved to `config.base`?

scripts/exglobal_cleanup.sh
```diff
@@ -0,0 +1,214 @@
...
+
+# Remove the TMPDIR directory
+# TODO Only prepbufr is currently using this directory, and all jobs should be
+#   cleaning up after themselves anyway
+COMIN="${DATAROOT}/${GDATE}"
+[[ -d ${COMIN} ]] && rm -rf "${COMIN}"
+
+if [[ "${DELETE_COM_IN_ARCHIVE_JOB:-YES}" == NO ]] ; then
+    exit 0
+fi
```
KateFriedman-NOAA: See other comment about whether this variable should be something like `DELETE_COM`.

I'm guessing this check is here because the cleanup job will always be in the mesh but users can turn off cleanup for whatever reason?

scripts/exglobal_cleanup.sh
KateFriedman-NOAA: Perhaps make `RTOFS_DATE` look-back configurable as well? `RMOLDRTOFS` or something similar and make 48hrs the default?

KateFriedman-NOAA (commented):
Looks great, thanks @WalterKolczynski-NOAA ! Left a few comments for now. Happy to re-review and approve when this is out of draft form.

scripts/exgdas_enkf_earc.sh
WalterKolczynski-NOAA: No, go ahead and keep working on it. The function is basically unchanged, so any solutions you find can then be applied here.

jobs/JGLOBAL_CLEANUP
WalterKolczynski-NOAA: I should have to, yes. Will fix.

WalterKolczynski-NOAA:
> I assume `earc.sh` will have its own cleanup; e.g. `ecleanup.sh`

I merged cleanup of the enkf directories into the regular cleanup, so there is no need to run a separate one for enkf.

parm/config/gfs/config.base.emc.dyn
WalterKolczynski-NOAA: `FHMAX_FITS` is being used by cleanup as well. `FITSARC` looks like it can move back though.

KateFriedman-NOAA: Wondering also if all of these variables should move to `config.fit2obs`? Would need the cleanup job to source that config as well though I think, for `FHMAX_FITS`.

CatherineThomas-NOAA:
@WalterKolczynski-NOAA 

> I merged cleanup of the enkf directories into the regular cleanup, so there is no need to run a separate one for enkf.

Should the new cleanup job also have a dependency on the completion of earc?  Looks like just arch currently.

WalterKolczynski-NOAA:
> @WalterKolczynski-NOAA
> 
> > I merged cleanup of the enkf directories into the regular cleanup, so there is no need to run a separate one for enkf.
> 
> Should the new cleanup job also have a dependency on the completion of earc? Looks like just arch currently.

Yes! Good catch.

HenryWinterbottom-NOAA:
@CatherineThomas-NOAA I have been unable to reproduce your result so I am not sure if something in the runtime environment is different or what.

However, I did build a regression test for the `remove_files` function and tried to catch any edge-cases (perhaps yours is one).

Can you please update your branch and try again? If this does not solve the problem we may have to see what PR #1906 yields.

Thank you for your patience.

jobs/JGLOBAL_CLEANUP
WalterKolczynski-NOAA: Done

scripts/exglobal_cleanup.sh
WalterKolczynski-NOAA: Done

parm/config/gfs/config.cleanup
WalterKolczynski-NOAA: Renamed, but kept it here.

scripts/exglobal_cleanup.sh
WalterKolczynski-NOAA: Yeah. Since cleanup does very little with it off, in the future we probably should just turn off the job. But I'm leaving that outside the scope of this PR.

scripts/exglobal_cleanup.sh
WalterKolczynski-NOAA: Added

WalterKolczynski-NOAA:
Completely revamped the cleanup to be less convoluted and actually work properly. This should resolve #1872 now as well.

WalterKolczynski-NOAA:
Rerunning some testing after I ran into an issue with `find`, but I think it is close enough for review now.

HenryRWinterbottom (commented):
@WalterKolczynski-NOAA This LGTM. If it is confirmed that this also resolves #1872, I approve.

@CatherineThomas-NOAA Can you confirm whether this resolves issue #1872?

CatherineThomas-NOAA:
@HenryWinterbottom-NOAA 
I started a test with Walter's fork this morning, but it needs to run a few more cycles.  I'll let you know once it's gone far enough.

HenryRWinterbottom:
Great! Thank you @CatherineThomas-NOAA.

WalterKolczynski-NOAA:
After sleeping on it, I think now that the clean-up is streamlined, it will be better to have enkf run its own cleanup after all. Will make the change today. Shouldn't invalidate any running tests.

CatherineThomas-NOAA:
I ran two days worth of cycles with the single cleanup job (before Walter's split at [bde3297](https://github.com/NOAA-EMC/global-workflow/pull/1906/commits/bde329759a14db5e4d6dd03484e54b04a7a4d3a6)).  The files were cleaned up mostly as expected, fixing the problem with issue #1872.

The only hiccup was in the exclude file list.  It looks like the exclude file list is being shared between the deterministic and ensemble:
```
+ exglobal_cleanup.sh[47]: find /scratch1/NCEPDEV/stmp4/Catherine.Thomas/ROTDIRS/cleanup/enkfgdas.20211221/00 -type l -not '(' -name '*prepbufr*' -or -name '*prepbufr*' -or -name '*cnvstat*' -or -name '*prepbufr*' -or -name '*prepbufr*' -or -name '*cnvstat*' -or -name '*atmanl.nc' ')' -delete

```
The only file left in my ensemble directory is the cnvstat ensmean file:
```
/scratch1/NCEPDEV/stmp4/Catherine.Thomas/ROTDIRS/cleanup/enkfgdas.20211221/06/ensstat/analysis/atmos
[Catherine.Thomas@hfe01 atmos]$ ls
enkfgdas.t06z.cnvstat.ensmean
```

I don't think this is the behavior that we want.  We should probably have the option of having separate exclude lists for deterministic and ensemble.

Being that I tested this fork before @WalterKolczynski-NOAA split the cleanup job into det and ens, I'll pull his new changes and try one more cycle, but it doesn't look to me that it will make a difference for this particular issue.

CatherineThomas-NOAA:
@HenryWinterbottom-NOAA @WalterKolczynski-NOAA 
I ran an additional cycle with [bde3297](https://github.com/NOAA-EMC/global-workflow/pull/1906/commits/bde329759a14db5e4d6dd03484e54b04a7a4d3a6) and it performed as expected -- it ran separate gdas and enkfgdas cleanup jobs that cleaned up all the files except for the exclude list files.  As anticipated, this change did not address the single exclude list for both the deterministic and ensemble.

WalterKolczynski-NOAA:
@CatherineThomas-NOAA I went back and checked what it was before I changed it in the COM reorg. It looks like the exclude for ENKF was `f006.ens`. If you confirm that is correct I will make the change.

HenryRWinterbottom:
@WalterKolczynski-NOAA That was the only one that I saw as well.

CatherineThomas-NOAA:
@WalterKolczynski-NOAA 
Yes, that's right. To save the F006 ensmean and enspread.

WalterKolczynski-NOAA:
@CatherineThomas-NOAA Updated to set the list in `config.cleanup` and base it on `$RUN`. Note if you test using an existing experiment, you will have to similarly update the `config.cleanup` to match.

CatherineThomas-NOAA (approved):
I merged @WalterKolczynski-NOAA's changes and ran another cycle.  The different exclude lists for gdas and enkfgdas work as intended.  It even cleaned up the older cycle’s enkf cnvstats that were erroneously saved.  Looks good to me.  

Thanks @HenryWinterbottom-NOAA and @WalterKolczynski-NOAA for your help on this.

HenryRWinterbottom:
@CatherineThomas-NOAA Thanks for following up and also for testing.

HenryRWinterbottom (approved):
Thank you @WalterKolczynski-NOAA for resolving two issues with this PR.

jobs/rocoto/cleanup.sh
WalterKolczynski-NOAA: Removed

TerrenceMcGuinness-NOAA:
This PR needs to be updated with recent changes in the CI scripts in the development branch.  This is causing the creation of the EXPDIRs to fail in the automated CI scripts.

parm/config/gfs/config.cleanup
```diff
@@ -0,0 +1,25 @@
...
+echo "BEGIN: config.cleanup"
+
+# Get task specific resources
+source "${EXPDIR}/config.resources" cleanup
+
+export CLEANUP_COM="YES"   # NO=retain ROTDIR.  YES default in cleanup.sh
+
+#--starting and ending hours of previous cycles to be removed from rotating directory
+export RMOLDSTD=144
+export RMOLDEND=24
```
KateFriedman-NOAA, line 13: Add `RMOLDRTOFS=48` here?

KateFriedman-NOAA (approved):
One comment left. Looks good otherwise, thanks @WalterKolczynski-NOAA !

jobs/JGLOBAL_CLEANUP
WalterKolczynski-NOAA: Backed this change out. I think this is what caused the failure because setpy.sh isn't available.

WalterKolczynski-NOAA:
@aerorahul Can you approve so I can merge without overriding restrictions?

workflow/rocoto/gfs_tasks.py
```diff
@@ -944,6 +944,23 @@ def arch(self):
 
         return task
 
+    # Cleanup
+    def cleanup(self):
+        deps = []
+        if 'enkf' in self.cdump:
+            dep_dict = {'type': 'metatask', 'name': 'enkfgdaseamn'}
```
aerorahul, line 951: Do we need to add a dependency on `enkfgfseamn` when the early cycle EnKF is also ON?

aerorahul (approved):
LGTM, though I have a clarifying question regarding job dependencies of cleanup when early cycle EnKF is enabled.

PR #1927, opened 2023-10-12
Update nth_eupd to 8 for Orion

TerrenceMcGuinness-NOAA:
# Description

Small update for `config.resources eupd` when running on Orion.
We updated `ntheupd=8` as is on Hera in anticipation of tomorrow's Weekly CI Tests
as they had previously failed for _enkfgdaseupd_ (`enkf.x`) in **C384C192_hybatmda**

# Type of change

# Change characteristics

# How has this been tested?
- Clone, build, and cycle tested on Orion

### DIFF
--- a/parm/config/gefs/config.resources
+++ b/parm/config/gefs/config.resources
@@ -806,7 +806,7 @@ elif [[ ${step} = "eupd" ]]; then
       fi
     elif [[ ${CASE} = "C384" ]]; then
       export npe_eupd=270
-      export nth_eupd=2
+      export nth_eupd=8
       if [[ "${machine}" = "WCOSS2" ]]; then
         export npe_eupd=315
         export nth_eupd=14


RussTreadon-NOAA:
**General comment.**  
I've encountered the same OOM issue on Orion and Hera.  While increasing the number of threads (`nth`) decreases the number of tasks per node (`ppn`), this isn't the best solution for all jobs.  It would be nice to have the option to define `ppn` directly without toggling `nth`.  Yes, there is danger in allowing this flexibility.  It's possible a user configures a job such that `ppn x nth` exceeds `npe_node_max`.   On the other hand, increasing `nth` to reduce `ppn` seems to be the wrong way to address an OOM problem.  It's not the number of threads which is causing the OOM kill.   It's the number of tasks per node which causes the OOM kill.

WalterKolczynski-NOAA:
@RussTreadon-NOAA Do you know why the GSI requires so much more memory with Intel 2022?

parm/config/gefs/config.resources
```diff
@@ -806,7 +806,7 @@ elif [[ ${step} = "eupd" ]]; then
       fi
     elif [[ ${CASE} = "C384" ]]; then
       export npe_eupd=270
-      export nth_eupd=2
+      export nth_eupd=8
```
WalterKolczynski-NOAA, line 809: Did you try a smaller increase to 4?

RussTreadon-NOAA:
> @RussTreadon-NOAA Do you know why the GSI requires so much more memory with Intel 2022?

eupd executes `enkf.x`.  global_enkf ctest did not flag a significant increase in memory when intel/2022 PR was being worked on.  We do not have a GSI code manager so I don't think anyone looked deeply at important details like this.

aerorahul (approved):
Looks good.
This is necessary to make sure the executable runs without running out of memory at this resolution.

aerorahul:
> **General comment.** I've encountered the same OOM issue on Orion and Hera. While increasing the number of threads (`nth`) decreases the number of tasks per node (`ppn`), this isn't the best solution for all jobs. It would be nice to have the option to define `ppn` directly without toggling `nth`. Yes, there is danger in allowing this flexibility. It's possible a user configures a job such that `ppn x nth` exceeds `npe_node_max`. On the other hand, increasing `nth` to reduce `ppn` seems to be the wrong way to address an OOM problem. It's not the number of threads which is causing the OOM kill. It's the number of tasks per node which causes the OOM kill.

Agreed @RussTreadon-NOAA 
An option to provide _tasks_per_node_  independent of _npe_node_max_/_nth_ and yet putting a bound on _tasks_per_node_ x _nth_ not to exceed _npe_node_max_ would be nice to have.
We can add this feature next time we do a refactor of the resource management in the global-workflow.

RussTreadon-NOAA:
> > @RussTreadon-NOAA Do you know why the GSI requires so much more memory with Intel 2022?
> 
> eupd executes `enkf.x`. global_enkf ctest did not flag a significant increase in memory when intel/2022 PR was being worked on. We do not have a GSI code manager so I don't think anyone looked deeply at important details like this.

Do we have eupd job log files from pre and post intel/2022 enkf.x automated g-w tests?  If `enkf.x` stdout is not echoed to the eupd job log file, we need to look in enkfstat.  We can check the task 0 memory usage via w3tage output.

parm/config/gefs/config.resources
TerrenceMcGuinness-NOAA: No I hadn't, let me give it a shot.

aerorahul:
This is needed to make the tests pass on Orion.
I will merge now and allow for an update for better resource allocation later.

PR #1931, opened 2023-10-13
Updates to CI to clarify HOMEgfs

aerorahul:
# Description
This PR:
- removes obfuscating use of `HOMEgfs` and `HOMEgfs_PR` in the CI driver scripts.  `HOMEgfs_PR` is renamed to `HOMEgfs`.  Previously used `HOMEgfs` is the location of the CI driver scripts that are used **not** from the PR.
- Updates case yamls with the above change in HOMEgfs_PR.
- moves `create_experiment.py` from `ci/` to `workflow` so it can be used by all (not just CI).

# Type of change

# Change characteristics

# How has this been tested?
- changes to the CI scripts need to be tested by @TerrenceMcGuinness-NOAA as the maintainer of the `ROOT_DIR` space.

# Checklist


aerorahul:
I think I might break this PR down into 2.  
One for SOCA and another for CI.

ci/cases/pr/C48_atm3DVar_ocnhybDA.yaml
```diff
@@ -0,0 +1,18 @@
...
+  mode: cycled
+
+arguments:
+  pslot: ${pslot}
+  app: S2S
+  resdet: 48
+  comrot: ${RUNTESTS}/COMROT
+  expdir: ${RUNTESTS}/EXPDIR
+  # icsdir: ${ICSDIR_ROOT}/C48cpldDA
+  icsdir: "/scratch2/NCEPDEV/ocean/Guillaume.Vernieres/data/ICSDIR/C48O500" # TODO: make platform agnostic
```
WalterKolczynski-NOAA: I think this is what we already have as C48C48mx500

aerorahul: ok sounds good.  I'll do a comparison and update.

ci/platforms/soca_gfs_defaults_ci.yaml
WalterKolczynski-NOAA, line 1: Should this file also be moved out into the other PR?

parm/config/gfs/yaml/defaults.yaml
```diff
@@ -13,13 +15,13 @@ landanl:
...
-  CASE_ANL: 'C48'
-  SOCA_OBS_LIST: ''
-  COMIN_OBS: '/scratch2/NCEPDEV/marineda/r2d2'
-  SABER_BLOCKS_YAML: ''
-  SOCA_NINNER: 50
-  R2D2_OBS_SRC: 'gdas_marine'
-  R2D2_OBS_DUMP: 's2s_v1'
+  SOCA_INPUT_FIX_DIR: "/scratch2/NCEPDEV/ocean/Guillaume.Vernieres/data/static/72x35x25/soca"  # TODO: These need to go to glopara fix space. @guillaumevernieres will open an issue
+  CASE_ANL: "C48"
+  COMIN_OBS: "/scratch2/NCEPDEV/marineda/r2d2-v2-v3"  # TODO: make platform agnostic
```
WalterKolczynski-NOAA, line 20: Wary about merging this into develop until these are on both RDHPCS platforms

WalterKolczynski-NOAA (changes requested):

parm/config/gfs/yaml/defaults.yaml
aerorahul: it doesn't matter.  There is no soca test. Nothing will break

ci/platforms/soca_gfs_defaults_ci.yaml
aerorahul: doesn't matter to me.  it is not exercised.  It will be exercised when the disabled test is enabled in the next PR.

ci/scripts/driver.sh
```diff
@@ -145,18 +144,17 @@ for pr in ${pr_list}; do
     # loop over every yaml file in the PR's ci/cases
     # and create an run directory for each one for this PR loop
     #############################################################
-    HOMEgfs_PR="${pr_dir}/global-workflow"
-    export HOMEgfs_PR
-    cd "${HOMEgfs_PR}"
+    export HOMEgfs="${pr_dir}/global-workflow"
```
TerrenceMcGuinness-NOAA: Having **HOMEgfs** exported in the shell environment may cause `create_experment.py` to fail.  Let me check into this. Not sure we need to export it here anymore because **HOMEgfs_PR** was used inside the YAML import but that is not necessary now as it should be picked up by **_top** and expressed in **config.base** as intended.

parm/config/gfs/yaml/defaults.yaml
aerorahul: See https://github.com/NOAA-EMC/GDASApp/issues/678

aerorahul:
@TerrenceMcGuinness-NOAA 
Let me know if I can help in getting these changes tested.

@aerorahul Running a CI PR from a remote imported clone of this branch with the updates in my dev cron tabs on Orion now.

ci/platforms/soca_gfs_defaults_ci.yaml
aerorahul: I removed the soca test files from this PR.

ci/cases/pr/C48_atm3DVar_ocnhybDA.yaml
aerorahul: this file is removed.

WalterKolczynski-NOAA (approved):

ci/scripts/driver.sh
TerrenceMcGuinness-NOAA: Yes this indeed fails ... @aerorahul  I send you an e-mail showing why and proposed a demonstration using your Jinja stuff that works.

WalterKolczynski-NOAA (approved):

TerrenceMcGuinness-NOAA (approved):
Tested lates update (removing export on HOMEgfs) and the CI scripts work ok

ci/scripts/driver.sh
TerrenceMcGuinness-NOAA: This is not fixed as we have eliminated the need to export **HOMEgfs**

PR #1967, opened 2023-10-24
Test tarballs for rstprod before calling chgrp

### LINKED ISSUE #1460: Detect group instead of assuming rstprod for certain tarballs
**Description**
We're currently assuming certain tarballs contain restricted data. However, on platforms that do not support rstprod, this causes a non-fatal error because the group does not exist. This is also the incorrect approach, as it is possible to use only unrestricted data.

**Requirements**
Scan the group of all the files being added to a tarball and detect if any are rstprod. If so, make the tarball rstprod; otherwise do not change the group membership.

Tarball groups that match the most restrictive group of any of the files contained therein, rather than a blanket group change for a pre-set list of tarballs.

**Dependencies**
None
See also: PR #1436

DavidHuber-NOAA:
# Description

Certain tarballs may or may not contain `rstprod` data.  For instance, the first half cycle gdas and enkfgdas tarballs will not contain `rstprod`, while future cycles likely will.  Also, some systems do not have `rstprod` on them.  This will test the contents of the tarballs first before attempting to change the group to rstprod.

Resolves #1460
# Type of change

- New feature

# Change characteristics

# How has this been tested?
Ran a cycled test on Hera.  The first half cycle `gdas.tar`, `gdas_restarta.tar`, and `enkfgdas.tar` tarballs did not change groups to `rstprod` while he first full cycle did, with the latter including the restricted `cnvstat`, `radstat`, `prepbufr`, and `nsstbufr` files.

# Checklist


scripts/exglobal_archive.sh
```diff
@@ -264,13 +264,32 @@ if [[ ${HPSSARCH} = "YES" || ${LOCALARCH} = "YES" ]]; then
...
+                # Change group to rstprod if it was found
+                if [[ "${has_rstprod}" == "YES" ]]; then
+                    ${HSICMD} chgrp rstprod "${ATARDIR}/${CDATE}/${targrp}.tar"
+                    ${HSICMD} chmod 640 "${ATARDIR}/${CDATE}/${targrp}.tar"
+                fi
                 ;;
             *) ;;
         esac
+
+        # For safety, test if the htar/tar command failed after changing groups
```
aerorahul, line 305: `status` is the return code for `${TARCMD}`, so it is not going to check on the state of `chgrp`
Should this next if-block be before the test for containment of rstprod?

DavidHuber-NOAA: @aerorahul No, I think it should be after `chgrp`.  That way, if the `htar` command fails due to something like a timeout, the `chgrp` command still executes to protect the contents.  That said, it would probably be a good idea to have a test for `chgrp` as well.  I will add that in.

aerorahul: yes, understood. thanks

DavidHuber-NOAA:
@aerorahul I added a test for `chgrp` as well and did a little cleanup.

scripts/exgdas_enkf_earc.sh
```diff
@@ -112,12 +112,40 @@ if [ "${ENSGRP}" -eq 0 ]; then
...
+        has_rstprod="NO"
+        while IFS= read -r file; do
+            if [[ -f ${file} ]]; then
+                group=$( stat -c "%G" "${file}" )
+                if [[ "${group}" == "rstprod" ]]; then
+                    has_rstprod="YES"
+                    break
+                fi
+            fi
+        done < "${ARCH_LIST}/${RUN}.txt"
```
WalterKolczynski-NOAA: Should we do this check before creating the tarball, so there is less of a gap between creation and restriction? I'm not sure how long the file list may take to run through.

DavidHuber-NOAA: @WalterKolczynski-NOAA It is very fast, << 1 second, though I do think it would help with readability, so I will go ahead and change the order.

WalterKolczynski-NOAA (approved):
Conditionally approved pending successful CI completion.

Issue #1986, opened 2023-10-26
Update for Prototype HR3 UGWP functionality


mdtoyNOAA:
### What new functionality do you need?

New updates to the Unified Gravity Wave Physics (UGWP) have been merged with the 'develop' branch of the 'ufs-community/ufs-weather-model' repo (see [PR#1923](https://github.com/ufs-community/ufs-weather-model/pull/1923)) as part of the HR3 prototype development work.  These updates require new namelist options, etc. to function properly in the global workflow.

### What are the requirements for the new functionality?

The new UGWP updates require new and modified namelist options and specifying a new CCPP suite definition file to function properly.

### Acceptance Criteria

The global workflow runs the updated UGWP properly for HR3.

PR #1933, opened 2023-10-17
Update ufs model to version from 10-12-23

JessicaMeixner-NOAA:
# Description

This updates the ufs-weather-model to the commit hash from 10-12-23 from the HR2 tag.  The diffs can be seen here: https://github.com/ufs-community/ufs-weather-model/compare/GFSv17.HR2...68050e58589a82ab509aaefaafdc2a6b90f34e48  This replaces PR https://github.com/NOAA-EMC/global-workflow/pull/1889 

Resolves https://github.com/NOAA-EMC/global-workflow/issues/1811

# Type of change

- Maintenance (code refactor, clean-up, new CI test, etc.) - routine update of model 

# Change characteristics
- Is this a breaking change (a change in existing functionality)? NO
- Does this change require a documentation update? NO

# How has this been tested?
- Free forecast for low resolution S2SWA on hera (previously tested other machines as well). 
-  This is ready for additional testing 

# Checklist
- [ ] Any dependent changes have been merged and published
- [x] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] I have made corresponding changes to the documentation if necessary

PR #1987, opened 2023-10-26
Update functionality for Prototype HR3 UGWP

mdtoyNOAA:
# Description

** Note that this PR builds off of [PR#1933](https://github.com/NOAA-EMC/global-workflow/pull/1933) once that code is merged. **

This PR adds/modifies namelist options and activates new 'ccpp' suite definition files for the updated Unified Gravity Wave Physics (UGWP) in the ufs-weather-model recently updated for HR3.  The changes to UGWP are the use of the GSL drag suite for orographic gravity wave drag (GWD) and blocking, as well as the new UGWP Version 1 non-stationary GWD.  See ufs-weather-model [PR#1923](https://github.com/ufs-community/ufs-weather-model/pull/1923) for details.

This PR resolves issue [#1986](https://github.com/NOAA-EMC/global-workflow/issues/1986).

# Type of change

# Change characteristics

# How has this been tested?
- Clone and build on Hera
- Forecast-only on Hera

# Checklist


JessicaMeixner-NOAA:
@mdtoyNOAA would it be possible for you to sync your branch with the latest develop now that https://github.com/NOAA-EMC/global-workflow/pull/1933 was merged?

mdtoyNOAA:
The change I'm proposing for '[ush/forecast_postdet.sh](https://github.com/NOAA-EMC/global-workflow/pull/1987/files#diff-356529d3a45758eae0ab358a98c7867594523b4beca8187ce87b17d74f71701d)' is to copy the files 'model_configure' and 'nems.configure' to the ${DATA} directory after 'gfsfcst' completion.  I have found these files useful to have in addition to 'input.nml' for documenting the settings used for a given run.

JessicaMeixner-NOAA:
@aerorahul I have a minor update to this that I'm testing including adding this: https://github.com/NOAA-EMC/global-workflow/issues/2012     The queue is slow today, but should have that pushed to Mike's PR here as soon as the tests are finished and we should be good for further testing on this PR later this afternoon.

JessicaMeixner-NOAA (approved):
The changes here include the physics changes for HR3.  + resolves issue https://github.com/NOAA-EMC/global-workflow/issues/2012 

Thanks @mdtoyNOAA !!!!!

WalterKolczynski-NOAA (changes requested):
I think you just need to fix the one shellcheck complaint. I'm going to start testing in the meantime.

ush/parsing_namelists_FV3.sh
```diff
@@ -59,6 +59,11 @@
   ${NCP} "${FIELD_TABLE}" field_table
 fi
 
+# set resolution dependent variables
+if [[ ${do_gsl_drag_ls_bl:-.true.} == ".true." ]]; then
+   export cdmbgwd=$cdmbgwd_gsl
```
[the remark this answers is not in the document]
JessicaMeixner-NOAA: @WalterKolczynski-NOAA so you'd like line 64 to be: export cdmbgwd=${cdmbgwd_gsl} ? 
Correct?

JessicaMeixner-NOAA:
> I think you just need to fix the one shellcheck complaint. I'm going to start testing in the meantime.

Thanks @WalterKolczynski-NOAA  - Just a heads up, the PR that was merged in before this moves UFS past where this PR is pointing to.  Not sure if that matters for testing but wanted to comment on here since there's a conflict so that everyone knows to keep the version that is in develop now instead of the version of UFS here.

JessicaMeixner-NOAA:
@mdtoyNOAA I fixed the shell check warning and merged into develop  to fix the conflicts and made a PR to your branch: https://github.com/mdtoyNOAA/global-workflow/pull/2

aerorahul:
@NeilBarton-NOAA 
We would like to make sure that the GEFS configuration in the global-workflow will not be broken after this change.  Please highlight any flags that you may deem appropriate.

NeilBarton-NOAA:
The next versions of GEFS plans on using these changes.  It looks good to me.

NeilBarton-NOAA (approved):

mdtoyNOAA:
> @mdtoyNOAA I fixed the shell check warning and merged into develop to fix the conflicts and made a PR to your branch: [mdtoyNOAA#2](https://github.com/mdtoyNOAA/global-workflow/pull/2)

@JessicaMeixner-NOAA   Done.

parm/config/gfs/config.base.emc.dyn
```diff
@@ -214,7 +214,7 @@ case "${APP}" in
     export DO_COUPLED="YES"
     export DO_OCN="YES"
     export DO_ICE="YES"
-    export CCPP_SUITE="FV3_GFS_v17_coupled_p8"  # TODO: Does this include FV3_GFS_v17_p8?  Can this be used instead of FV3_GFS_v17_p8 on L149
+    export CCPP_SUITE="FV3_GFS_v17_coupled_p8_ugwpv1"  # TODO: Does this include FV3_GFS_v17_p8?  Can this be used instead of FV3_GFS_v17_p8 on L149
```
aerorahul, line 220: Can someone answer the question on #TODO here?  Also, note the line has changed from L149 to 151.

ush/parsing_namelists_FV3.sh
```diff
@@ -59,6 +59,11 @@ else
   ${NCP} "${FIELD_TABLE}" field_table
 fi
 
+# set resolution dependent variables
+if [[ ${do_gsl_drag_ls_bl:-.true.} == ".true." ]]; then
+   export cdmbgwd=${cdmbgwd_gsl}
+fi
```
aerorahul: Is this a resolution dependent variable?  
Why can't this be elevated to `config.fcst` under the if-block of `if ((gwd_opt = 2))`?

mdtoyNOAA: @aerorahul Yes, 'cdmbgwd' is a resolution dependent variable.  It also depends on the orographic gravity wave drag (OGWD) scheme being called.  Both the 'old' GFSv16 and 'new' GSL OGWD schemes can each be called with 'gwd_opt=2'.  What switches between them is the namelist variable 'do_gsl_drag_ls_bl'.  I originally tried elevating to config.fcst, but this 'do_gsl_drag_ls_bl' is not passed into the script.  Modifying 'parsing_namelists_FV3.sh' seemed the best way to go.

aerorahul: I see in `config.fcst` when `gwd_opt == 2`, `do_gsl_drag_ls_bl` is set to `.true.`
Why can't `cdmbgwd=${cdmbgwd_gsl}` in that block?
Is `do_gsl_drag_ls_bl` going to be `.true.` when `gwd_opt != 2`?  

Yes, I understand mountain blocking and gravity wave drag coefficients are resolution dependent.  However, that choice is made in `config.ufs`, and not here. 

I don't see a technical reason why we cannot do this in `config.fcst`

parm/config/gfs/config.base.emc.dyn
JessicaMeixner-NOAA: I remember this was an open question.  I wanted to say that Moorthi would run the "cpld" CCPP suite with the uncoupled atm-only, but the details are fuzzy. Based on memory, the coupled CCPP suite files have one extra call incase you have an ice model, but otherwise are the same.  I think that what is currently here is consistent with the way things are run in regression tests.

ush/forecast_postdet.sh
```diff
@@ -515,6 +515,8 @@ FV3_out() {
   else
     # No need to copy FV3 restart files when RUN=gfs or gefs
     ${NCP} "${DATA}/input.nml" "${COM_ATMOS_HISTORY}/input.nml"
+    ${NCP} "${DATA}/model_configure" "${COM_ATMOS_HISTORY}/model_configure"
+    ${NCP} "${DATA}/nems.configure" "${COM_ATMOS_HISTORY}/nems.configure"
```
aerorahul: I think we now have `COM_CONF`.  Please copy these files (and `input.nml`) to `COM_CONF` with a prefix `ufs.`

ush/parsing_namelists_FV3.sh
mdtoyNOAA: No, 'do_gsl_drag_ls_bl' will not be .true. when gwd_opt != 2.  I can try your suggestion this evening or tomorrow when Hera is back up from maintenance.  HOWEVER, I think I tried something similar to what you suggested, but in config.fcst, the variable 'cdmbgwd_gsl', which is set in 'config.ufs', wasn't available and the script would exit with an error.

ush/forecast_postdet.sh
JessicaMeixner-NOAA: So these lines would become: 

${NCP} "${DATA}/model_configure" "${COM_CONF}/ufs.model_configure"
${NCP} "${DATA}/nems.configure" "${COM_CONF}/ufs.nems.configure"

Correct?

ush/parsing_namelists_FV3.sh
mdtoyNOAA: The current method of parsing 'cdmbgwd' into 'input.nml' is directly from 'config.ufs' to 'ush/parsing_namelists_FV3.sh' ('config.fcst' is bypassed).  The proposed method for 'cdmbgwd = cdmbgwd_gsl' follows this same approach, we only need the 'if [[ ${do_gsl_drag_ls_bl:-.true.} == ".true." ]]; then' statement in 'parsing_namelists' to achieve the same end when using the GSL drag parameterizations.  As mentioned earlier, the 'cdmbgwd_gsl' variable wasn't accessible in 'config.fcst'.

aerorahul: I don't understand this. 
`config.fcst` sources `config.ufs` on [line 21](https://github.com/NOAA-EMC/global-workflow/blob/5183c43bbbd07285869feacb49d8680afe85929a/parm/config/gfs/config.fcst#L21).
At this point, one should have both `cdmbgwd` and `cdmbgwd_gsl` available.

mdtoyNOAA: You're right.  I see this.  Once Hera is back up, I'll retry setting 'cdmbgwd=cdmbgwd_gsl' in 'config.fcst'.  Stand by.

mdtoyNOAA: @aerorahul I tested on Hera and your suggested changes work.  I remember now that I was trying to use 'do_gsl_drag_ls_bl' into 'config.ufs' to set 'cdmbgwd', but that didn't work.  In 'config.fcst' an 'if' statement is still necessary because a user could change the default 'do_gsl_drag_ls_bl' for 'gwd_opt=2' from .true. to .false. and cdmbgwd would need to be changed accordingly.  My modifications to 'config.fcst' assures this.

ush/forecast_postdet.sh
mdtoyNOAA: I tried the lines above but the test failed with the error:
/scratch1/BMC/wrfruc/mtoy/git_local/global-workflow.test/ush/forecast_postdet.sh: line 517: COM_CONF: unbound variable

@aerorahul Can you clarify what needs to be done here?  Thank you.

parm/config/gefs/config.fcst
```diff
@@ -119,24 +119,27 @@ if (( gwd_opt == 2 )); then
...
+    export do_gsl_drag_ls_bl=".true."
     export do_gsl_drag_ss=".true."
-    export do_gsl_drag_tofd=".false."
+    export do_gsl_drag_tofd=".true."
     export do_ugwp_v1_orog_only=".false."
     launch_level=$(echo "${LEVS}/2.35" |bc)
     export launch_level
+    if [[ ${do_gsl_drag_ls_bl} == ".true." ]]; then
+       export cdmbgwd=${cdmbgwd_gsl}
+    fi
```
aerorahul, line 137: Is this if-block needed?  This is *always* going to be `.true.` as it is set on line 129?
Why not simply do:
```suggestion
    export cdmbgwd=${cdmbgwd_gsl}
```

parm/config/gefs/config.fcst
```diff
@@ -148,6 +151,9 @@ if (( LEVS == 128 )) && [[ "${CDUMP}" =~ "gdas" ]]; then
    export d2_bg_k2=0.0
 fi
 
+# n_split adjustment
+export n_split=5
+
```
aerorahul: This is interesting.
Is this something that is needed for all resolutions?  I am sure it works, but seems like this is needed because the model blew up somewhere for someone at some resolution with the timestep they were using.

parm/config/gfs/config.base.emc.dyn
aerorahul: Ok. Thanks.  Can the `TODO` comment be updated to reflect the correct line number or removed if there is no need to combine and verify Moorthi's result?

ush/forecast_postdet.sh
WalterKolczynski-NOAA: `COM_CONF` needs to be added to the list of variables created using `generate_com` in the j-job before it can be used.

parm/config/gfs/config.base.emc.dyn
JessicaMeixner-NOAA: I'd recommend retesting just incase my memory is bad or in case things have since changed.

aerorahul: Great. Can someone on the modeling team verify that it runs and produces acceptable results?
It does not have to be done for this PR to be merged.  Do however, update the comment and if possible, add that testing will be done to resolve the `TODO`.

parm/config/gefs/config.fcst
mdtoyNOAA: It will not always be .true.  As I mentioned above, the user has flexibility to go back to the old GFSv16 orographic GWD if desired.  In that case, they would set 'do_ugwp_v0=.true.' and 'do_gsl_drag_ls_bl=.false.', then the 'old default' values of cdmbgwd would be needed.  The if-block is needed.

aerorahul: I respectfully disagree.
The way it is currently set up, if `gwd_opt = 2`, then the statements on line 129 will set `do_gsl_drag_ls_bl=.true.`
In which case the if block is redundant.
However, I agree that if the user comes in and manually changes the default `do_gsl_drag_ls_bl=.false.` on line 129, then they will have to also have to comment out line 135 or 136.

parm/config/gefs/config.fcst
mdtoyNOAA: Yes, this is the case.  We had gone from 'n_split=6' to 'n_split=4' (larger dynamics time step for faster runs) and started experiencing numerical instability crashes at C768.  Compromising with 'n_split=5' eliminated the crashes.

parm/config/gefs/config.fcst
mdtoyNOAA: I agree that 'do_gsl_drag_ls_bl' is the default for 'gwd_opt=2', but it is not a requirement.  The 'gwd_opt=2' option originated with the 'unified_ugwp' CCPP_suite which allows use of either the 'old' GFS orographic GWD or the new 'GSL' scheme.  This switch is made with the 'do_gsl_drag_ls_bl' variable.  The proposed scheme for HR3 is the 'GSL' scheme, which is why setting it to .true. is the default.  But don't we want the flexibility for a user to revert back to the 'old' GFS scheme if desired?  This is what I had hoped for.

parm/config/gefs/config.fcst
aerorahul: Ideally, we would have `n_split` based on resolution as part of `config.ufs`.  But since we are prescribing the same for all resolutions, I am inclined to suggest updating the default on line 384 in `ush/forecast_postdet.sh`

mdtoyNOAA: I updated the default in ush/forecast_postdet.sh and removed the explicit setting in config.fcst.

parm/config/gfs/config.base.emc.dyn
JessicaMeixner-NOAA: I went to look at the ufs-weather-model regression tests and how the workflow is now is consistent with how I saw the regression tests being run, with different CCPP suite files for coupled versus uncoupled.  While it's possible I missed something, if I did not,  is there a reason to move away from how the regression tests are run with the two separate CCPP suite files for the coupled or uncoupled cases?

aerorahul: I don't think the UFSWM RT and global-workflow are consistent in general.  Yes, they are consistent in that they both use the different CCPP suite file for coupled and uncoupled.  However, the question I asked is, is that necessary?  Should they be the same or should they be different? 
From the perspective of unification, an uncoupled configuration is a coupled model with all the components OFF.
I am not expecting a response for that question in this PR, as the answer will require further investigation.

ush/forecast_postdet.sh
mdtoyNOAA: Can I get some more specific details on which scripts need to be modified and what needs to be added?  I tried this with jobs/JGLOBAL_FORECAST:
@@ -46,7 +46,7 @@ declare -rx gcyc="${GDATE:8:2}"
 
 # Construct COM variables from templates (see config.com)
 YMD=${PDY} HH=${cyc} generate_com -rx COM_ATMOS_RESTART COM_ATMOS_INPUT COM_ATMOS_ANALYSIS \
-  COM_ATMOS_HISTORY COM_ATMOS_MASTER COM_TOP
+  COM_ATMOS_HISTORY COM_ATMOS_MASTER COM_TOP COM_CONF
 
 RUN=${rCDUMP} YMD="${gPDY}" HH="${gcyc}" generate_com -rx \
   COM_ATMOS_RESTART_PREV:COM_ATMOS_RESTART_TMPL

But it did not work.

Thank you.

WalterKolczynski-NOAA: That should've been enough. Can you point me to your log?

JessicaMeixner-NOAA: Just chatted with @aerorahul I think we also need to create the folder and he showed me where.  I'll try to add this after my current meeting.

mdtoyNOAA: @WalterKolczynski-NOAA  The log file is on Hera at:
/scratch1/BMC/wrfruc/mtoy/git_local/global-workflow.test/FV3GFSrun/GFS_v17_test.C768/logs/2019110300/gfsfcst.log
And my modified JGLOBAL_FORECAST file is at:
/scratch1/BMC/wrfruc/mtoy/git_local/global-workflow.test/jobs/JGLOBAL_FORECAST

mdtoyNOAA: @JessicaMeixner-NOAA Is the folder you discussed creating a 'conf' folder?  The input.nml, etc. files would be sent there instead of the 'history' directory?  I kind of liked having them go to the 'history' directory.  :)

JessicaMeixner-NOAA: @mdtoyNOAA apologies, I didn't get to finish this testing before leaving for the weekend.  Since configuration files are not considered model history files, these files will all go to the "conf" directory where other configuration files will also go.  Yes, the CONF directory is the directory folder where things should go.  I'll start looking at this again now.

mdtoyNOAA: @JessicaMeixner-NOAA  Sounds good.  Thank you for the update.

JessicaMeixner-NOAA:
Is it fair to say that the outstanding issues with this PR are addressed in: https://github.com/NOAA-EMC/global-workflow/pull/2054 ?  

Are there other changes that still need to be addressed or made @WalterKolczynski-NOAA ?

WalterKolczynski-NOAA:
> Is it fair to say that the outstanding issues with this PR are addressed in: #2054 ?
> 
> Are there other changes that still need to be addressed or made @WalterKolczynski-NOAA ?

from my standpoint, but I think we're waiting on @aerorahul

aerorahul (approved):
looks fine.

JessicaMeixner-NOAA:
Thanks @WalterKolczynski-NOAA and @aerorahul, in terms of next steps should @mdtoyNOAA back-out the copying on conf changes in this PR, or just  wait until #2054  is merged and then merge those changes in here?

WalterKolczynski-NOAA:
> Thanks @WalterKolczynski-NOAA and @aerorahul, in terms of next steps should @mdtoyNOAA back-out the copying on conf changes in this PR, or just wait until #2054 is merged and then merge those changes in here?

Just back them out and we'll handle it in #2054.

mdtoyNOAA:
> > Thanks @WalterKolczynski-NOAA and @aerorahul, in terms of next steps should @mdtoyNOAA back-out the copying on conf changes in this PR, or just wait until #2054 is merged and then merge those changes in here?
> 
> Just back them out and we'll handle it in #2054.

I'll get to work on it.  Thanks.

JessicaMeixner-NOAA:
> ```
> Experiment C96C48_hybatmDA_b6fc8437 Terminated: *** FAILED ***
> Experiment C96C48_hybatmDA_b6fc8437 Terminated with 1 tasks failed at Tue Nov 14 20:30:40 UTC 2023
> Error logs:
> /scratch1/NCEPDEV/global/Terry.McGuinness/GFS_CI_ROOT/PR/1987/RUNTESTS/COMROT/C96C48_hybatmDA_b6fc8437/logs/2021122100/gfspostanl.log
> ```

The error is: 
91: /scratch1/NCEPDEV/global/Terry.McGuinness/GFS_CI_ROOT/PR/1987/global-workflow/exec/upp.x: error while loading shared libraries: libpng16.so.16: cannot open shared object file: No such file or directory
Does an extra library need to be loaded or is this a different error?  The hera tests had passed on this branch before.

JessicaMeixner-NOAA:
> ```
> Experiment C48_S2SA_gefs_b6fc8437 **SUCCESS** at Tue Nov 14 15:06:19 CST 2023
> Experiment C96_atm3DVar_b6fc8437 Terminated: *** FAILED ***
> Experiment C96_atm3DVar_b6fc8437 Terminated with 1 tasks failed at Tue Nov 14 15:16:31 CST 2023
> Error logs:
> /work2/noaa/stmp/GFS_CI_ROOT/PR/1987/RUNTESTS/COMROT/C96_atm3DVar_b6fc8437/logs/2021122100/gfsanal.log
> ```

This timed out.  I think this is because this branch is behind and does not have the fix that was committed recently

PR #1954, opened 2023-10-23
Make the early cycle work with the coupled UFS configured as S2S 

guillaumevernieres:
# Description
This PR proposes a few changes to address the issues that we found while cycling the early and late cycle in conjunction with the atmos and marine DA and the S2S UFS model. 

- fixes #1949 

# Type of change
- Bug fix (fixes something broken) as defined by @guillaumevernieres 
- New feature (adds functionality) as defined by @aerorahul 

# Change characteristics
- Is this a breaking change (a change in existing functionality)? **YES** Pinging @JessicaMeixner-NOAA , this will break the forecast only case (mentioned to me by @aerorahul ).
- Does this change require a documentation update? **NO**

# How has this been tested?
Cycling the WCDA system with the S2S model on Hera 

# Checklist
- [ ] Any dependent changes have been merged and published
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] I have made corresponding changes to the documentation if necessary

PR #1998, opened 2023-10-30
Parallelize the build scripts

### LINKED ISSUE #1978: Optimize the build scripts

The whole building process takes a very long time (~2 hours with the GSI and longer with GDASApp).  It would be nice to have simultaneous builds going on with a specified max number of cores to speed up the building process.  Perhaps having the option of building on a compute node would also be an option.

Build times for the entire system are reduced significantly.

### Acceptance Criteria

All builds are successful on all systems and build failures are appropriately captured

I see two possible solutions:

1) Rewrite the build scripts completely and create a master CMakeLists.txt file in global-workflow/sorc to handle all of the builds simultaneously.  I believe this would require a common set of modulefiles to make this work, but it is certainly an elegant solution.  We would get closer to the common set of modulefiles once #1868 is completed.

2) Rewrite build_all.sh to run each of the build scripts with a specified number of cores simultaneously.  Perhaps kick off the GDASApp first in the background, then group the remaining builds something like (ufs, gsi), (upp, gfs_utils, ufs_utils), (gsi_utils, gsi_monitor).

DavidHuber-NOAA:
# Description

Optimized/parallelized build scripts.

A few notes:

1) The default number of build jobs is 20, but one could argue for using 40.  When researching this, I looked up what the SRW is compiling with.  That system uses 40 cores, which seems a little excessive, but on testing the global workflow, the actual number of cores being used at any given time rarely exceeds 16 when running with 40 cores.  This is because the builds tend to use multiple threads in the beginning when compiling low-level modules while the higher-level modules are more or less serial AND because the GDASApp takes several minutes to initialize all of its subrepositories by which time the smaller builds are complete.

2) I also updated checkout.sh so that all checkouts are simultaneous.  The CPU load for `git submodule` is quite low, so running 16 instead of 8 jobs at once is not much more expensive.

3) To make this work, I had to add `-j` options to most of the build scripts.  The only exception is build_upp, for which the build script within the UPP is hard coded to use 6 cores.

4) I fixed a few small bugs in the build scripts along the way.

5) Lastly, this reduce the total build time from ~2.5 hours for the entire system (including GDAS and GSI in the same build) to ~40 minutes when running with `-j 40`.

Resolves #1978 

# Type of change

# Change characteristics

# How has this been tested?
Tested build_all.sh with multiple options on Hera.

# Checklist


sorc/build_all.sh
```diff
@@ -105,170 +107,190 @@ ERRSCRIPT=${ERRSCRIPT:-'eval [[ $err = 0 ]]'}
...
-    err=$((err + rc))
-  fi
-else
-  echo " .... Skip building gsi utilities .... "
-fi
+# Go through all builds and adjust CPU counts down if necessary
+requested_cpus=0
+build_list=""
+for build in "${!build_jobs[@]}"; do
+   [[ -z "${build_list}" ]] && build_list="${build}" || build_list="${build_list}, ${build}"
```
WalterKolczynski-NOAA: I would prefer a real if/then/else here instead of a short-circuit.

sorc/build_all.sh
```diff
@@ -105,170 +107,190 @@ ERRSCRIPT=${ERRSCRIPT:-'eval [[ $err = 0 ]]'}
...
+   if [[ ${procs_in_use} -lt ${_build_job_max} ]]; then
+      if [[ -n "${build_jobs[gdas]+0}" && -z "${build_ids[gdas]+0}" ]]; then
+         if [[ ${_build_job_max} -ge $(( build_jobs['gdas'] + procs_in_use )) ]]; then
+            ./build_gdas.sh -j "${build_jobs[gdas]}" "${_verbose_opt}" > \
+               "${logs_dir}/build_gdas.log" 2>&1 &
+            build_ids["gdas"]=$!
+            echo "Starting build_gdas.sh"
+            procs_in_use=$(( procs_in_use + build_jobs['gdas'] ))
+         fi
+      fi
```
WalterKolczynski-NOAA: This bit of code is getting repeated over and over. Can we turn this into a loop?

sorc/build_all.sh
```diff
@@ -105,170 +107,190 @@ ERRSCRIPT=${ERRSCRIPT:-'eval [[ $err = 0 ]]'}
...
+            if [[ ${build_stat} == 0 ]]; then
+               echo "${build} completed successfully!"
+            else
+               echo "${build} failed with status ${build_stat}!"
+            fi
+
+            # Remove the completed build from the list of PIDs
+            unset 'build_ids[${build}]'
+            unset 'build_jobs[${build}]'
+            break
```
WalterKolczynski-NOAA: I'm assuming this is the reason for the break here. Would like a comment so no one else wonders why.
```suggestion
            # Removing elements from the loop array can cause issues, so just start the loop over.
            break
```

WalterKolczynski-NOAA (changes requested):

sorc/build_all.sh
DavidHuber-NOAA: Probably, but just noting that some of the build scripts take different arguments than others.  For instance, build_upp.sh does not take a `-j` and build_ufs.sh takes a `-a` argument for the `$APP` of choice.  I could create an additional array that stores the build command for each application, which would be fewer lines overall and make it easier for developers to find/adjust those build commands as needed.  I will give that a try.

WalterKolczynski-NOAA: That was my idea as well.

sorc/build_all.sh
DavidHuber-NOAA: The break was just for testing.  It seems to handle the deletion fine without it, so I have removed it.

sorc/build_all.sh
DavidHuber-NOAA: I made an implementation which seemed to be working, but it has revealed a case where the WW3 pre/post applications attempt to build before the UFS, which causes an issue as those require some modules that are only compiled when the entire UFS is compiled.  I will need to add another if block to ensure `build_ww3prepost.sh` doesn't execute until `build_ufs.sh` completes.  This also means I can throw more CPUs at WW3 since it won't be competing with the UFS.

sorc/build_all.sh
DavidHuber-NOAA: Fixed.

sorc/build_all.sh
JessicaMeixner-NOAA: @DavidHuber-NOAA  This is a bit confusing to me.  Depending on where things are built I could see compiling both UFS and WW3 pre/post at the same time would potentially be an issue but perhaps this just needs a script change.  Also FYI - I have an update coming that will add an extra build option to both UFS and WW3 for unstructured grid - I'm sure it'll be easy to implement the update no matter which goes first, just wanted you to be aware.

DavidHuber-NOAA: @JessicaMeixner-NOAA Hmm, your suspicion is correct, but this leaves me a bit perplexed.   The last time I ran the build_all.sh script, buid_ww3prepost.sh launched before build_ufs.sh and crashed with several undefined reference errors.  For instance:
```
[ 72%] Linking Fortran executable ../../bin/ww3_strt
../../lib/libww3.a(w3gdatmd.F90.o): In function `w3gdatmd_mp_w3gntx_':
/scratch1/NCEPDEV/nems/David.Huber/GW/gw_parallel_builds/sorc/ufs_model.fd/WW3/model/src/w3gdatmd.F90:2953: undefined reference to `w3gsrumd_mp_w3gsuc_ptr_r8_'
/scratch1/NCEPDEV/nems/David.Huber/GW/gw_parallel_builds/sorc/ufs_model.fd/WW3/model/src/w3gdatmd.F90:3001: undefined reference to `w3gsrumd_mp_w3cgdm_r4_'
/scratch1/NCEPDEV/nems/David.Huber/GW/gw_parallel_builds/sorc/ufs_model.fd/WW3/model/src/w3gdatmd.F90:2986: undefined reference to `w3gsrumd_mp_w3cgdm_r4_'
```
However, when I review an older build log, I see that the missing module, `w3gsrumd`, was built during the ww3prepost job.  I'm not sure how this could happen.  Any ideas?  I will save the log file here if you wish to take a look while I continue testing: `/scratch1/NCEPDEV/nems/David.Huber/GW/gw_parallel_builds/sorc/logs/build_ww3prepost.save.log`.

JessicaMeixner-NOAA: @DavidHuber-NOAA I've built using the WW3 script without ever building UFS, I'll take a look at your log file.

JessicaMeixner-NOAA: @DavidHuber-NOAA so the WW3 pre/post builds multiple executables.  Is this an issue that you are trying to build at the top level with multiple processors instead of the actual cmake command having multiple processors?  (I ask because something similar happened a long time ago).  So for example, the top level build_ww3prepost should not be called with multiple MPI processors but within it the make commands could use multiple?    If that doesn't make sense let me know and I can try to explain in person...

sorc/build_ww3prepost.sh
```diff
@@ -85,7 +86,7 @@ if (( rc != 0 )); then
   echo "Fatal error in cmake."
   exit "${rc}"
 fi
-make -j 8
+make -j "${BUILD_JOBS:-8}"
```
JessicaMeixner-NOAA, line 90: @DavidHuber-NOAA Okay I see this in the code now.   I'm not sure why there are issues...

DavidHuber-NOAA: @JessicaMeixner-NOAA I believe I found the issue.  I had killed the ww3prepost build then restarted it.  However, the build_ww3prepost.sh script does not delete the build directory.  I've added in a `[[ -d ${path_build} ]] && rm -rf ${path_build}` which solves the issue.  Typically, the ww3prepost job cleans out the build directory at the end of the build so this probably hasn't been an issue.

JessicaMeixner-NOAA: I'm glad it was a simple thing, I was struggling to think of another cause.   Thanks for these updates! It'll be exciting to have a faster build!

KateFriedman-NOAA (approved):
Approved pending successful final testing and resolution of open comments.

sorc/build_all.sh
DavidHuber-NOAA: I have implemented a new array that keeps track of build options and consolidated the build if blocks.

WalterKolczynski-NOAA (approved):

aerorahul:
The failure was because of a bug in #1954 
It is addressed in #2009

aerorahul:
please merge develop #2009 fixed the failing test.

WalterKolczynski-NOAA:
We're killing these tests after the build is successful anyway.

PR #2001, opened 2023-10-30
Allow CI case to exclude running on certain platforms

### LINKED ISSUE #1968: Feature to exclude specific systems for CI cases

We need to add a list of exclusionary systems for each of the cases in the CI framework.  This will allow us to "skip" cases in the CI test suite not to run on HPC platforms that the case can not be supported on.

An added field in the YAML specification for CI cases with an exclusion list for systems it can not support

### Acceptance Criteria

At CI test runtime the cases with excluded systems specified are not ran on that systems but all the others that are supported do

_No response_

TerrenceMcGuinness-NOAA:
# Description

Added a feature so we can skip specific platforms within a case file in the CI framework

Resolves #1968 

# Type of change

# Change characteristics

# How has this been tested?

Added the line

```
exclude:
 orion
 hera
```

to case file and passed it to the `create_experiment.py` Python script with the `--yaml` switch to demonstrate that the script simply prints in the log that the case **pslot** name is skipped on Orion and quits.

### DIFF
--- a/workflow/create_experiment.py
+++ b/workflow/create_experiment.py
@@ -30,7 +30,6 @@ import setup_xml
 
 from hosts import Host
 
-
 _here = os.path.dirname(__file__)
 _top = os.path.abspath(os.path.join(os.path.abspath(_here), '..'))
 
--- a/workflow/create_experiment.py
+++ b/workflow/create_experiment.py
@@ -79,7 +79,7 @@ if __name__ == '__main__':
 
     if 'exclude' in testconf:
         host = Host()
-        if host.machine.lower() in testconf.exclude:
+        if host.machine.lower() in [ excluded_host.lower() for excluded_host in testconf.exclude ]:
             logger.info(f'Skipping creation of case: {testconf.arguments.pslot} on {host.machine.capitalize()}')
             sys.exit(0)
 
--- a/workflow/create_experiment.py
+++ b/workflow/create_experiment.py
@@ -72,7 +72,7 @@ def input_args():
 if __name__ == '__main__':
 
     user_inputs = input_args()
-    
+
     # Create a dictionary to pass to parse_j2yaml for parsing the yaml file
     data = AttrDict(HOMEgfs=_top)
     data.update(os.environ)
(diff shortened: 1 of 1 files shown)


workflow/create_experiment.py
```diff
@@ -27,6 +28,8 @@
 import setup_expt
 import setup_xml
 
+from hosts import Host
+
```
WalterKolczynski-NOAA, line 32:
```suggestion
```

workflow/create_experiment.py
```diff
@@ -75,6 +78,12 @@ def input_args():
     data.update(os.environ)
     testconf = parse_j2yaml(path=user_inputs.yaml, data=data)
 
+    if 'exclude' in testconf:
+        host = Host()
+        if host.machine.lower() in testconf.exclude:
```
WalterKolczynski-NOAA:
```suggestion
        if host.machine.lower() in [ excluded_host.lower() for excluded_host in testconf.exclude ]:
```

WalterKolczynski-NOAA (dismissed):

TerrenceMcGuinness-NOAA (commented):
good to cover cases for caps

aerorahul (approved):

workflow/create_experiment.py
```diff
@@ -75,6 +77,12 @@ def input_args():
     data.update(os.environ)
     testconf = parse_j2yaml(path=user_inputs.yaml, data=data)
 
+    if 'exclude' in testconf:
```
aerorahul, line 80: While `exclude` is fine, IMO, it is not descriptive in its intent.
The purpose here is to skip CI on the excluded host.
Would a key such as `skip_ci_on_hosts:` be descriptive to include in the yamls?
Thoughts?

aerorahul (approved):
LGTM.

workflow/create_experiment.py
TerrenceMcGuinness-NOAA: Oh yes completely agree!

TerrenceMcGuinness-NOAA: let me make the changes in a push

Issue #1845, opened 2023-09-11
Placeholder for tracking WCOSS2 Forecast Job Hanging Issues 


JessicaMeixner-NOAA:
### What new functionality do you need?

Many forecast jobs running the HR2 tag on WCOSS2 have hung.  This issue was decreased by not eliminated by adding the following environment variables recommended by GDIT: 
export MPICH_ENV_DISPLAY=1
export MPICH_VERSION_DISPLAY=1
export FI_MR_CACHE_MONITOR=memhooks

It's also been suggested to try adding: 
export FI_MR_CACHE_MONITOR=memhooks

There is an open issue from @jiandewang with GDIT about this problem. 

Lastly, @junwang-noaa suggests updating to the latest HDF5 libraries which would require updating the hash of ufs-waether-model that's being used in global-workflow. 

This issue has been created to track the possible need of adding these variables to the forecast job for WCOSS2 (particularly when running C768, I'm unaware of issues at lower resolution).  It's not 100% clear if these are needed at this time, although they're highly recommend if you are running the HR2 tag & C768 resolution on WCOSS2. 

 

### What are the requirements for the new functionality?

C768 S2SW GFS forecast jobs do not hang. 

### Acceptance Criteria

C768 S2SW GFS forecast jobs do not hang.

KateFriedman-NOAA:
@JessicaMeixner-NOAA Have you spoken with GDIT about the hangs? I investigated a similar issue with them (late 2022 into early 2023) after hangs started with the enkf forecast jobs in both production and development tests (C384). We didn't find a root cause but the hangs seemed to stop after some machine upgrades. Some hangs were traced to hardware issues but not every one.

Interesting to see it in C768 now and that those environment variables help some. GDIT could assist to investigate further if you haven't already opened a helpdesk ticket. :)

JessicaMeixner-NOAA:
Yes, apologies I deleted that these were from GDIT as I edited the issue.  I'll add that back in.

KateFriedman-NOAA:
Great! Upgrading to the newer HDF5 (as Jun suggested) should hopefully also help (GDIT and George V suggested something similar during investigations). Look forward to the new version!

jiandewang:
with many tests, Jun and I confirm that by add "ucx" modules the hanging issue is gone.
    module load craype-network-ucx
    module load cray-mpich-ucx

KateFriedman-NOAA:
[...]

PR #2021, opened 2023-11-02
Add two ucx modules to solve C768 S2SW hanging

jiandewang:
# Description
It happened frequently that fcst job hang in wcoss2 for S2SW C768. Adding two ucx modules solve the issue.
Resolves #https://github.com/NOAA-EMC/global-workflow/issues/1845

# Type of change

# Change characteristics

# How has this been tested?
- Clone and build on WCOSS

# Checklist


WalterKolczynski-NOAA:
Happy to add it temporarily, but seems like this should really be a PR into UFS.

jiandewang:
@WalterKolczynski-NOAA the hanging issue only happened to C768 and C1152 while in UFS the max resolution is C384

WalterKolczynski-NOAA:
I would argue this is a UFS problem even if they don't have any regression tests for those resolutions.

@junwang-noaa, what do you think?

aerorahul:
> I would argue this is a UFS problem even if they don't have any regression tests for those resolutions.
> 
> @junwang-noaa, what do you think?

We might need to include it in the global-workflow while the UCX, Slingshot issues are being worked on.
My q. is, clearly this has an impact based on high resolutions.  Will this create issues when the model is run at lower resolutions; C384 for e.g.?

junwang-noaa:
The impact to low resolutions is minimal, slightly faster for C384 runs, unnoticeable for lower resolutions.

aerorahul (approved):

PR #2030, opened 2023-11-03
Enable different coupled settings for ensemble

### LINKED ISSUE #1692: Enable ensemble to run different coupled configuration
**Expected behavior**
`DO_OCN_ENKF` settings that control which coupled components run should properly set up a forecast run with the given settings, even if those settings are different from the deterministic model. (`DO_OCN` and `DO_OCN_ENKF` are used throughout this issue as a stand-in for all components.)

**Current behavior**
If the settings do not match the deterministic settings, the forecast directory will not be setup correctly.

**Machines affected**
All

**To Reproduce**
1. Create a warm-start cycled `S2S` experiment
2. Set `DO_OCN_ENKF` to `"NO"` (see Additional Information)
3. Observe the deterministic forecast succeed but the ensemble forecast fail

**Context**
`config.fcst` is setting several variables derived by `DO_OCN`, most notably the namelist variables (`cplflx`, `cplchm`, etc.). These are not then redefined after `config.efcs` modifies the `DO_OCN` variables to be the values determined by `DO_OCN_ENKF`.

Even before getting to the namelist, the derived values are used to determine which sections of the forecast script are executed, but _not_ what COM variables are set, so currently the job fails due to undefined COM variables before it even tries to run the forecast.

**Detailed Description**

**Additional Information**
As of bug submission time, all the `DO_OCN_ENKF` default to `"NO"` in `config.efcs`. However, I plan to comment out the overriding of `DO_OCN` in `config.efcs`, so the values of `config.fcst` are used until this is fixed.

Parts or all of this may be mooted by the refactor of the forecast job into python.

**Possible Implementation**
While we could just duplicate the definition of the derived variables, that seems like unnecessary code duplication. We need to rethink how we are doing this. The cleanest low-effort way would be to simply move the definitions out of the configs so they have the final settings before determining them.

Also, to update scripts to current workflow practices, the fortran booleans should not be used for logical branching within the shell scripts, instead using the `DO_OCN` variables.

WalterKolczynski-NOAA:
# Description
Adds the ability to turn off some coupled components for the ensemble.

To enable this, many settings are moved out of `config.fcst` and into `config.ufs` so they are calculated after any overrides of `DO_*` have been made. Some settings also had to be moved out of `config.base` for the same reason. A switch to turn on gocart is added to `config.ufs`, but unlike the other switches this one takes no arguments.

The name of the ufs configure template used is converted from a partial file name to the full file name.

Resolves #1692

# Type of change

# Change characteristics

# How has this been tested?
- Cycled test on Orion with S2S deterministic and atmos-only ensemble

# Checklist


WalterKolczynski-NOAA:
NB: config.ufs feels a little clunky now. The existing code there used the resolution arguments to determine which components were on. While I tried to reuse these for the most part, the loading of the component config files still uses the `DO_*` switches. Also, since there is no resolution for gocart, those settings also still use `DO_AERO`. So now it is a mixture of `DO_*` and the `skip_*` (and even between the two, the "DO" switches use generic component names, whereas the "skip" switches use specific model names (e.g. `DO_OCN` vs `skip_mom6`)

parm/config/gfs/config.fcst
```diff
@@ -303,11 +263,6 @@ elif [[ "${CDUMP}" =~ "gfs" ]] ; then # GFS cycle specific parameters
     # Write gfs restart files to rerun fcst from any break point
     export restart_interval=${restart_interval_gfs:-12}
 
-    # Choose coupling with wave
-    if [[ "${DO_WAVE}" = "YES" && "${WAVE_CDUMP}" != "gdas" ]]; then
-        export cplwav=".true."
-    fi
-
```
aerorahul, line 313: How is this `&&` condition handled? 
Apologies, but I find these `gdas` and  `gfs` logic with `WAVE_CDUMP` confusing.

aerorahul (approved):
LGTM.   Thanks for adding this capability.  
Please take a look at the comment on `WAVE_CDUMP != gdas`.

aerorahul:
Does `gefs/config.base.emc.dyn` need an update as well?

aerorahul (changes requested):
https://github.com/NOAA-EMC/global-workflow/blob/cbdc09711a83493f363a737d1e3e46c48a09719d/scripts/exglobal_forecast.sh#L91
is always going to be `atm`.
Also, 
https://github.com/NOAA-EMC/global-workflow/blob/cbdc09711a83493f363a737d1e3e46c48a09719d/scripts/exglobal_forecast.sh#L108
will always only validate for `atm`.
The [cplvalidate](https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/cplvalidate.sh) function needs to be updated with the changes in this PR.

parm/config/gfs/config.fcst
JessicaMeixner-NOAA: @aerorahul This confuses to me too.  I saw logic before where waves were turned off for enkf, which makes sense but i actually don't understand this logic here, because we do cpl with waves in the gdas forecast job.... I tried to look at the history, but it didn't provide additional insight.

aerorahul: Thank you @JessicaMeixner-NOAA 
Sometime (sooner rather than later), when we start working on wave initialization for GEFS, if we can get together to understand some of the nuances, it would help us gaining familiarity and maintaining it better.

WalterKolczynski-NOAA:
Taking this to draft until I update and test GEFS.

parm/config/gfs/config.fcst
WalterKolczynski-NOAA: `WAVECDUMP` can be one of three values: `gfs`, `gdas`, or `both`. Since this was within an `if gfs` block, waves stay on as long as the value isn't `gdas`. I removed it because it is redundant with L9-12.

JessicaMeixner-NOAA: > Thank you @JessicaMeixner-NOAA Sometime (sooner rather than later), when we start working on wave initialization for GEFS, if we can get together to understand some of the nuances, it would help us gaining familiarity and maintaining it better.

Happy to meet sometime sooner than later - we should also include @sbanihash as she's the GEFS POC for waves as well.  

@WalterKolczynski-NOAA  - thanks for cleaning this up!

WalterKolczynski-NOAA:
> https://github.com/NOAA-EMC/global-workflow/blob/cbdc09711a83493f363a737d1e3e46c48a09719d/scripts/exglobal_forecast.sh#L91
> 
> 
> is always going to be `atm`.
> Also,
> https://github.com/NOAA-EMC/global-workflow/blob/cbdc09711a83493f363a737d1e3e46c48a09719d/scripts/exglobal_forecast.sh#L108
> 
> 
> will always only validate for `atm`.
> The [cplvalidate](https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/cplvalidate.sh) function needs to be updated with the changes in this PR.

Took this (and the script) out. Now checking the template file exists.

WalterKolczynski-NOAA:
GEFS has been added.

GEFS test passes (single cylce)
Cycle test with GFS launching only at 00z blows up on the second GFS forecast after ~66h.

parm/config/gfs/config.aero
```diff
@@ -34,10 +34,17 @@ esac
 # Biomass burning emission dataset. Choose from: GBBEPx, QFED, NONE (default)
 AERO_EMIS_FIRE=QFED
 
+export AERO_DIAG_TABLE="${AERO_DIAG_TABLE:-${HOMEgfs}/parm/ufs/fv3/diag_table.aero}"
+export AERO_FIELD_TABLE="${AERO_FIELD_TABLE:-${HOMEgfs}/parm/ufs/fv3/field_table.aero}"
+AERO_EMIS_FIRE="${AERO_EMIS_FIRE:-none}"
+export AERO_EMIS_FIRE=${AERO_EMIS_FIRE,,}
+export AERO_CONFIG_DIR="${AERO_CONFIG_DIR:-${HOMEgfs}/parm/ufs/chem}"
+export AERO_INPUTS_DIR="${AERO_INPUTS_DIR:-}"
```
aerorahul: This is `config.aero` and there is no higher level where these values are being set or should be set; thus they should not have pass-through options
```suggestion
export AERO_DIAG_TABLE="${HOMEgfs}/parm/ufs/fv3/diag_table.aero"
export AERO_FIELD_TABLE="${HOMEgfs}/parm/ufs/fv3/field_table.aero"
export AERO_EMIS_FIRE="none"
export AERO_CONFIG_DIR="${HOMEgfs}/parm/ufs/chem"
export AERO_INPUTS_DIR=""
```

Can the `chem` in `AERO_CONFIG_DIR` be named `aerosols` or `aero`?

parm/config/gfs/config.ufs
```diff
@@ -430,4 +466,44 @@ if [[ "${skip_ww3}" == "false" ]]; then
   export ntasks_ww3 nthreads_ww3
 fi
 
+# GOCART specific settings
+if [[ "${DO_AERO}" = "YES" ]]; then # temporary settings for aerosol coupling
```
aerorahul: We have the `skip_` thingy going on here.  Would we want to stay with that for aerosols and define `skip_gocart` and use that toggle?

ush/parsing_namelists_FV3.sh
```diff
@@ -54,7 +54,7 @@ if [[ -n "${AERO_FIELD_TABLE:-}" ]]; then
     tail -n $(( nrec - prec )) "${FIELD_TABLE}" ; \
   } > field_table
   # add non-prognostic tracers from additional table
-  dnats=$(( dnats + dnats_aero ))
+  dnats=$(( dnats + ${dnats_aero:0} ))
```
aerorahul: Is this default required?  
This is in the block of the condition `if [[ -n ${AERO_FIELD_TABLE} ]]`.  Should the check on line 43 also have a `DO_AERO = YES`?  If `DO_AERO` is `NO`, this will never be executed.  If it is `YES`, `dnats_aero` is defined in `config.aero` and if it isn't, it should be an error here.

aerorahul (changes requested):
Moving the contents of `aerosols` into `config.aero` is a good thing.
A few resulting clarifications that were touched as part of that move.

ush/parsing_namelists_FV3.sh
WalterKolczynski-NOAA: Thought it was needed, but now I think I was just missing exports.

parm/config/gfs/config.aero
WalterKolczynski-NOAA: Changed it to `gocart`, as that is consistent with other directories `mom6` and `fv3`.

parm/config/gfs/config.ufs
WalterKolczynski-NOAA: This is going to require more work, so I'm going to vote it as out-of-scope. To eliminate the need for the `DO_*` switches:
- The argparsing needs to be updated, as there wouldn't be an argument for `--gocart`, so should only `shift` once
- Need to redo to the component config loading on L64-69, as the config files are all by component but the skip switches are all based on model name

parm/config/gefs/config.ufs
```diff
@@ -8,7 +8,7 @@
 
 echo "BEGIN: config.ufs"
 
-if [ $# -le 1 ]; then
+if (( $# <- 1 )); then
```
aerorahul:
```suggestion
if (( $# <= 1 )); then
```

WalterKolczynski-NOAA:
Testing after resolving conflicts following #2071, then I will send it through CI again.

parm/config/gfs/config.ufs
aerorahul: I don't think we can use the `shift` thing here because `DO_AERO` makes sense if there is an atmosphere.
The inputs to `config.ufs` are `--fv3 $fv3_res --mom6 $mom6_res` etc.  There is not going to be a `--gocart $gocart_res` as GOCART runs on the same grid as atmosphere (FV3).
The suggestion was to add a default `skip_gocart=true` and then set `skip_gocart=false` if `DO_AERO = YES`.
But this is fine.

parm/config/gefs/config.ufs
WalterKolczynski-NOAA: Fixed

aerorahul (approved):
looks good.

PR #2039, opened 2023-11-06
Use assim_freq to configure "offset" in dependencies

### LINKED ISSUE #2038: use assim_freq in config.base to configure "offset" in dependencies

use `assim_freq` in config.base to configure "offset" in dependencies.
This is the very first step to develop a workflow cycled more frequently than 6-hourly.

Change the hard-coded "06:00:00" to a variable whose content is determined by the "`assim_freq`" setting in config.base

### Acceptance Criteria

The generated rotoco xml files are the same between new and old versions for a 6-hourly workflow (excluding the different timestamps about when the xml files are generated)

_No response_

guoqing-noaa:
# Description
Use the assim_freq from config.base to configure "offset" in dependencies instead of hard-coding them.

Resolves #2038

# Type of change

# Change characteristics

# How has this been tested?

Tested on Hera:
Compared two rocoto xml files generated separately by the latest develop branch and this PR and they are the same except the difference in the timestamp information about when the xml files were generated.

# Checklist


workflow/rocoto/gfs_tasks.py
```diff
@@ -86,10 +86,11 @@ def prep(self):
         gfs_enkf = True if self.app_config.do_hybvar and 'gfs' in self.app_config.eupd_cdumps else False
 
         deps = []
-        dep_dict = {'type': 'metatask', 'name': 'gdaspost', 'offset': '-06:00:00'}
+        myoffset = '-{h:02n}:00:00'.format(h=self._base["assim_freq"])
```
aerorahul: I would suggest adding the following in `class Tasks` after L35 in [tasks.py](https://github.com/NOAA-EMC/global-workflow/blob/75269e4bb9764b81589efc7d703825b80e74c8f5/workflow/rocoto/tasks.py#L35)
```python3
self._base['cycling_interval'] = to_timedelta(self._base["assim_freq"])
```
(The above might need to be made a string of `HH:00:00`)
and then use `[-]self._base.cycling_interval` where necessary instead of calculating `myoffset` each time.

guoqing-noaa: Thank @aerorahul for the great suggestion! I will make this change soon.

guoqing-noaa: @aerorahul I tried wxflow.timetools.to_timedelta(), it will need something as follows:
```
from wxflow import to_timedelta
self._base['cycling_interval'] =str(to_timedelta(str(self._base['assim_freq'])+'H'))

```
This looks a little bit more complicated than using the str.format().
Hope I did not miss anything.

I will add a dictionary entry as suggested to avoid calculating `myoffset `many times but will use str.format(). Thanks!

workflow/rocoto/gfs_cycled_xml.py
```diff
@@ -14,7 +14,7 @@ def __init__(self, app_config: AppConfig, rocoto_config: Dict) -> None:
     def get_cycledefs(self):
         sdate = self._base['SDATE']
         edate = self._base['EDATE']
-        interval = self._base.get('INTERVAL', '06:00:00')
+        interval = '{h:02n}:00:00'.format(h=self._base["assim_freq"])
```
WalterKolczynski-NOAA:
```suggestion
        interval = f'{self._base["assim_freq"]:02n}:00:00'
```

workflow/rocoto/gfs_tasks.py
WalterKolczynski-NOAA: Current python prefers using f-strings over `.format()` or concatenation when possible:
```
from wxflow import to_timedelta
self._base['cycling_interval'] =str(to_timedelta(f'{self._base["assim_freq"]}H'))
```

wxflow also has a function to convert a timedelta to HH:MM:SS:
```
from wxflow import to_timedelta, timedelta_to_HMS
self._base['cycling_interval'] = timedelta_to_HMS(to_timedelta(f'{self._base["assim_freq"]}H'))
```

However, I don't think you need to even bother converting to a string (the default casting of a timedelta to a string is the proper format), so this probably works:
```
from wxflow import to_timedelta
self._base['cycling_interval'] = to_timedelta(f'{self._base["assim_freq"]}H')
```

WalterKolczynski-NOAA (changes requested):

workflow/rocoto/gfs_tasks.py
aerorahul: My apologies @guoqing-noaa 
I should have been more specific in my change request.
@WalterKolczynski-NOAA thank you for clarifying and providing the exact methods that perform the operations.

workflow/rocoto/tasks.py
```diff
@@ -33,6 +33,7 @@ def __init__(self, app_config: AppConfig, cdump: str) -> None:
         # Save dict_configs and base in the internal state (never know where it may be needed)
         self._configs = self.app_config.configs
         self._base = self._configs['base']
+        self._base['minus_cycle_interval'] = '-{h:02n}:00:00'.format(h=self._base["assim_freq"])
```
aerorahul: One should not define a specific use variable so high in the class.
```suggestion
        self._base['cycle_interval'] =  to_timedelta(f"{self._base['assim_freq']}H") 
```

workflow/rocoto/gfs_tasks.py
```diff
@@ -86,10 +86,10 @@ def prep(self):
         gfs_enkf = True if self.app_config.do_hybvar and 'gfs' in self.app_config.eupd_cdumps else False
 
         deps = []
-        dep_dict = {'type': 'metatask', 'name': 'gdaspost', 'offset': '-06:00:00'}
+        dep_dict = {'type': 'metatask', 'name': 'gdaspost', 'offset': self._base['minus_cycle_interval']}
```
aerorahul: create the `minus` here with appropriate methods.
```suggestion
        dep_dict = {'type': 'metatask', 'name': 'gdaspost', 'offset': f"-{timedelta_to_HMS(self._base['cycle_interval'])}"}
```

workflow/rocoto/gfs_cycled_xml.py
aerorahul: `interval` should be, but currently is not, a `timedelta` object, just like `sdate` and `edate` are `datetime` objects.
If you can make that change, here and promote it elsewhere in this method, it would be the proper use of the datatype.  But, I understand if you would rather leave that to the developers.

aerorahul (changes requested):
A few changes are requested in `tasks.py` and `gfs_tasks.py`.  
Not sure if corresponding to `gfs_tasks.py` changes are also needed in `gefs_tasks.py`.  @WalterKolczynski-NOAA?
Also a note on `gfs_cycled_xml.py`

guoqing-noaa:
Thanks @WalterKolczynski-NOAA and @aerorahul for the great help and detailed suggestions. 
I committed new changes based on the review.

workflow/rocoto/gfs_cycled_xml.py
guoqing-noaa: @aerorahul Great suggestion! I would hope I can do so right now. However, I am not very familiar with this part. I hope I will learn more and more about this and be able to do more in the future. Thanks for the understanding!

aerorahul (approved):
looks good to me.
thank you for incorporating the suggestions.

WalterKolczynski-NOAA (approved):

workflow/applications/applications.py
```diff
@@ -167,14 +168,14 @@ def get_task_names(self) -> Dict[str, List[str]]:
...
 
     @staticmethod
-    def get_gfs_interval(gfs_cyc: int) -> str:
+    def get_gfs_interval(gfs_cyc: int) -> timedelta:
         """
         return interval in hours based on gfs_cyc
         """
 
-        gfs_internal_map = {'0': None, '1': '24:00:00', '2': '12:00:00', '4': '06:00:00'}
+        gfs_internal_map = {'1': '24H', '2': '12H', '4': '6H'}
```
WalterKolczynski-NOAA, line 176: @aerorahul Any reason these keys are strings instead of just using ints and avoiding the cast in L179?

aerorahul: They can be `int`.

WalterKolczynski-NOAA: Also, looks like the `0` key (cycling only) was removed and needs to be added back in unless some if upstream stops this from being called.

aerorahul: I think I don't call this method at all if `gfs_cyc == 0`.  If I missed someplace, please let me know.

aerorahul: see the updated `gfs_cycled.py` line 253

guoqing-noaa: I can update 1, 2, 4 from strings to ints if that's desired. Thanks!

WalterKolczynski-NOAA:
@guoqing-noaa Can you pull in develop and handle any conflicts? Then we'll rerun the tests and merge.

guoqing-noaa:
> @guoqing-noaa Can you pull in develop and handle any conflicts? Then we'll rerun the tests and merge.

@WalterKolczynski-NOAA I merged the latest develop branch. I think it is ready for rerunning the test. Thanks!

guoqing-noaa:
@WalterKolczynski-NOAA Sorry that the bug fix was lost when merging the latest develop branch. Now I added it back. Could you restart the tests?  Thanks!

guoqing-noaa:
Thank you, @WalterKolczynski-NOAA and @aerorahul for your great help on this PR!

PR #2049, opened 2023-11-09
CI Feature to detect stalled experiments

### LINKED ISSUE #2008: Feature BASH CI detects when a dependacy isn't being met

We need an update to the **check_ci** bash script to detect that a job dependency is not being met.
Currently if a job dependency is not being met **rocotrun** does nothing and yet all jobs are SUCCESS so the dependency failure it never caught.

A stalled experiment is indeed detected and reported to the user in the GitHub messages and is Labeled FAIL

### Acceptance Criteria

Force a test case with a false value for **get_ic** that can induce a dependency failure and then see if the updated script succefully detects it and flags the user of the issue.

In the **check_ci** bash script that marshals through all the jobs after **rocotorun** is executed that checks for succesfull completion, add the additional check to see if there are no jobs that are at least RUNNING, PENDING, or QUEUED.  If that is the case then **rocotorun** did nothing, and sense the experiment hadn't concluded, then we have a depedancey failure.

TerrenceMcGuinness-NOAA:
# Description

A helper python script `rocoto_statcount.py` is added in `$HOMEgfs/ci/scrpts/utils` that checks to see if an experiment is "stalled".   This is done by running **rocotorun** and checking within all the jobs and if none are RUNNING, SUBMITTING, or QUEUED then the experiment is STALLED.

 Resolves #2008 

# Type of change

# Change characteristics

# How has this been tested?
- Ran in CI CRON development space and set up a case that had a missing file dependency and it was appropriately flagged.

# Checklist


ci/scripts/check_ci.sh
```diff
@@ -131,7 +131,7 @@ for pr in ${pr_list}; do
       "${GH}" pr edit --repo "${REPO_URL}" "${pr}" --remove-label "CI-${MACHINE_ID^}-Running" --add-label "CI-${MACHINE_ID^}-Failed"
       error_logs=$("${rocotostat}" -d "${db}" -w "${xml}" | grep -E 'FAIL|DEAD' | awk '{print "-c", $1, "-t", $2}' | xargs "${rocotocheck}" -d "${db}" -w "${xml}" | grep join | awk '{print $2}') || true
       {
-       echo "Experiment ${pslot} Terminated: *** FAILED ***"
+       echo "Experiment ${pslot} Terminated: *** FAILED *** on ${MACHIND_ID^}"
```
aerorahul: Small typo
```suggestion
       echo "Experiment ${pslot} Terminated: *** FAILED *** on ${MACHINE_ID^}"
```

ci/scripts/check_ci.sh
```diff
@@ -152,8 +152,8 @@ for pr in ${pr_list}; do
       rm -f "${output_ci_single}"
       # echo "\`\`\`" > "${output_ci_single}"
       DATE=$(date)
-      echo "Experiment ${pslot} **SUCCESS** ${DATE}" >> "${output_ci_single}"
-      echo "Experiment ${pslot} **SUCCESS** at ${DATE}" >> "${output_ci}"
+      echo "Experiment ${pslot} **SUCCESS** ${DATE +%Y%m%d} on ${MACHINE_ID^}" >> "${output_ci_single}"
+      echo "Experiment ${pslot} **SUCCESS** at ${DATE +%Y%m%d} on ${MACHIND_ID^}" >> "${output_ci}"
```
aerorahul: This cannot possibly work `${DATE +%Y%m%d}`.

ci/scripts/run_ci.sh
```diff
@@ -81,7 +83,20 @@ for pr in ${pr_list}; do
...
-    echo "Running: ${rocotorun} -v 10 -w ${xml} -d ${db}"
-    "${rocotorun}" -v 10 -w "${xml}" -d "${db}"
+    set +e
+    ${ROOT_DIR}/ci/scripts/utils/rocoto_statcount.py -d "${db}" -w "${xml}" --check_stalled
+    rc=$?
+    if [[ "${rc}" -ne 0 ]]; then
+      output_ci="${pr_dir}/output_runtime_single.log"
+      {
+        echo "${pslot} has *** STALLED **** on ${MACHINE_ID^}"
+        echo "A job in expermint ${pslot} in ${pslot_dir}"
```
aerorahul:
```suggestion
        echo "A job in experiment ${pslot} in ${pslot_dir}"
```

aerorahul (commented):
looks ok, but I spotted some minor typos.
See suggestions.
And one comment on something that probably is incorrect.

ci/scripts/utils/rocoto_statcount.py
```diff
@@ -0,0 +1,104 @@
...
+    rocotostat_output_all = rocotostat_output_all.splitlines()[1:]
+    rocotostat_output_all = [line.split()[0:4] for line in rocotostat_output_all]
+    rocotostat_output_all = [line for line in rocotostat_output_all if len(line) != 1]
+
+    rocoto_status = {
+       'Cycles' : len(rocotostat_output),
+       'Done_Cycles' :  sum([ sublist.count('Done') for sublist in rocotostat_output ])
+    }
+    
+    status_cases = [ 'SUCCEEDED', 'FAIL', 'DEAD', 'RUNNING', 'PENDING', 'QUEUED']
```
WalterKolczynski-NOAA: Is 'PENDING' a valid status, or do you mean 'SUBMITTED' (which needs one).

ci/scripts/utils/rocoto_statcount.py
```diff
@@ -0,0 +1,104 @@
...
+            rocoto_run = which("rocotorun")
+            rocoto_run.add_default_arg(['-w',args.w,'-d',args.d])
+            rocoto_run()
+            rocoto_status2 = rocoto_statcount()
+            if rocoto_status2 == rocoto_status:
+                rocoto_state = 'Stalled'
+                print(f'Rocoto State : {rocoto_state}')
+                sys.exit(-1)
+            else:
+                rocoto_state = 'Running'
```
WalterKolczynski-NOAA: I don't think this is robust. It's not guaranteed the statuses to change on a `rocotorun` (particularly if a long job is running). As we discussed before, what we should be looking for is the number of successful or pending/queued jobs. You don't even need to call `rocotorun` the second time or have this extra `check_stalled` option.
```suggestion
    if rocoto_status['Cycles'] == rocoto_status['Done_Cycles']:
        rocoto_state = 'Done'
    elif rocoto_status['RUNNING'] + rocoto_status['SUBMITTING'] + rocoto_status['QUEUED'] == 0:
        rocoto_state = 'Stalled'
    else:
        rocoto_state = 'Running'
```

WalterKolczynski-NOAA (changes requested):

ci/scripts/utils/rocoto_statcount.py
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env python3
+
+import sys
+import os
+
+from wxflow import Executable, which, Logger
```
aerorahul:
```suggestion
from wxflow import Executable, which, Logger, CommandNotFoundError
```

ci/scripts/utils/rocoto_statcount.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    Run rocotostat and process its output.
+    """
+
+    args = input_args()
+
+    rocotostat = which("rocotostat")
+    if not rocotostat:
+        logger.exception("rocotostat not found in PATH")
+        sys.exit(-1)
```
aerorahul:
```suggestion
    rocotostat = which("rocotostat")
    if not rocotostat:
        logger.exception("rocotostat not found in PATH")
        sys.exit(-1)
    try:
        rocotostat = which("rocotostat")
    except CommandNotFoundError:
        logger.exception("rocotostat not found in PATH")
        raise CommandNotFoundError("rocotostat not found in PATH")
```

ci/scripts/utils/rocoto_statcount.py
```diff
@@ -0,0 +1,104 @@
...
+
+    xml_file_path = os.path.abspath(args.w)
+    db_file_path = os.path.abspath(args.d)
+
+    rocotostat_all = which("rocotostat")
```
aerorahul, line 56: Why do this again?  You already have `rocotostat`.  You can use that any number of times with different arguments.

@aerorahul I have not been able to find a way to use `add_default_args` to "update" the argument list of an Executable Object.  Can you fine an example, could not find such a use case in the repo.

ci/scripts/utils/rocoto_statcount.py
```diff
@@ -0,0 +1,104 @@
...
+    rocotostat_all.add_default_arg(['-w',xml_file_path,'-d',db_file_path,'-a'])
+
+    rocotostat_output = rocotostat(output=str)
+    rocotostat_output = rocotostat_output.splitlines()[1:]
+    rocotostat_output = [line.split()[0:2] for line in rocotostat_output]
+
+    rocotostat_output_all = rocotostat_all(output=str)
+    rocotostat_output_all = rocotostat_output_all.splitlines()[1:]
+    rocotostat_output_all = [line.split()[0:4] for line in rocotostat_output_all]
+    rocotostat_output_all = [line for line in rocotostat_output_all if len(line) != 1]
```
aerorahul: I can show you how to reduce this and simplify it considerably

ci/scripts/utils/rocoto_statcount.py
```diff
@@ -0,0 +1,104 @@
...
+    rocoto_status = rocoto_statcount()
+    for status in rocoto_status:
+        print(f'Number of {status} : {rocoto_status[status]}')
+    rocoto_state = 'Running'
+    if rocoto_status['Cycles'] == rocoto_status['Done_Cycles']:
+        rocoto_state = 'Done'
+
+    if args.check_stalled:
+        if rocoto_state != 'Done':
+            rocoto_run = which("rocotorun")
```
aerorahul: Having seen this, I think we should have a `RocotoCommands` class and initialize it with everything from rocoto you need; e.g. `rocotostat`, `rocotorun`, etc.  While you are at it, also provide the `xml` and `database` to the class object rather than have to do this over and over.

ci/scripts/utils/rocoto_statcount.py
WalterKolczynski-NOAA: Oh, and 'UNKNOWN' is also a valid state, and should either be automatically rewound or treated like a fail. It happens when rocoto is unable to get information about a job from the scheduler (usually because it wasn't run for a while and the job aged off, like during a long downtime).

ci/scripts/utils/rocoto_statcount.py
aerorahul: `elif any(x in rocoto_status for x in ['RUNNING', 'SUBMITTING', 'QUEUED'])` should do for line 90.
@aerorahul not exactly, we need to check that they are all **_not_** RUNNING, SUBMITTING, or QUEUED as Walt has the total SUM be zero as the eval in dict has number of occurrences.

aerorahul (commented):
some suggestions for `rocoto_statcount.py`

ci/scripts/utils/rocoto_statcount.py
TerrenceMcGuinness-NOAA: Oh right @WalterKolczynski-NOAA of course, not advancing doesn't mean the all the status states don't change, its when after  running **rocotorun** there are none that are RUNNING, SUBMITTING, or QUEUED.  Really simplifies things indeed.

ci/scripts/utils/rocoto_statcount.py
TerrenceMcGuinness-NOAA: Its actually **SUBMITTING** (which you got right later in this review) but yes you are correct there is no **PENDING**

And yeah, I agree we should add **UNKOWN** to **FAILED** since _Slurm_ would be "off" in that case and the experiment at that point would be questionable.

```
tmcguinness ~/NOAA-EMC/rocoto (develop) $ find . -type f  | xargs grep -h  ":state]=" | grep SUBMIT
            next unless @active_jobs[taskname][cycle][:state]=="SUBMITTING"
          newjob[:state]="SUBMITTING"
tmcguinness ~/NOAA-EMC/rocoto (develop) $ find . -type f  | xargs grep -h  ":state]=" | grep PEND
tmcguinness ~/NOAA-EMC/rocoto (develop) $ 
```

Looks like **FAIL** should be **FAILED** too:
```
tmcguinness ~/NOAA-EMC/rocoto (develop) $ find . -type f  | xargs grep -h  "record\[\:state\]=" | grep -Ev 'if|unless' | sed 's/ //g' | cut -d"=" -f2 | sed 's/"//g' | sed "s/'//g" | sort | uniq 
FAILED
QUEUED
RUNNING
SUCCEEDED
UNKNOWN
USERHOLD
```

ci/scripts/utils/rocoto_statcount.py
TerrenceMcGuinness-NOAA: Could not find a way to "change" arguments once the are set on the Executable object.

TerrenceMcGuinness-NOAA (commented):
Done with my review

ci/scripts/utils/rocoto_statcount.py
TerrenceMcGuinness-NOAA: Sure that would be great, you can jot down pseudo code it you like

ci/scripts/run_ci.sh
```diff
@@ -81,7 +83,22 @@ for pr in ${pr_list}; do
     pslot=$(basename "${pslot_dir}")
     xml="${pslot_dir}/${pslot}.xml"
     db="${pslot_dir}/${pslot}.db"
-    echo "Running: ${rocotorun} -v 10 -w ${xml} -d ${db}"
     "${rocotorun}" -v 10 -w "${xml}" -d "${db}"
+    set +e
+    "${ROOT_DIR}"/ci/scripts/utils/rocoto_statcount.py -d "${db}" -w "${xml}"
```
WalterKolczynski-NOAA:
```suggestion
    "${ROOT_DIR}/ci/scripts/utils/rocoto_statcount.py" -d "${db}" -w "${xml}"
```

ci/scripts/run_ci.sh
```diff
@@ -81,7 +83,22 @@ for pr in ${pr_list}; do
...
+    rc=$?
+    
+    if [[ "${rc}" -ne 0 ]]; then
+      output_ci="${pr_dir}"/output_runtime_single.log
```
WalterKolczynski-NOAA:
```suggestion
      output_ci="${pr_dir}/output_runtime_single.log"
```

ci/scripts/utils/rocoto_statcount.py
```diff
@@ -0,0 +1,99 @@
...
+    args : Namespace
+        The parsed command-line arguments.
+    """
+
+    description = """
+        Using rocotostat to get the status of all jobs this scripts
+        determines rocoto_state: if all cycles are done, then rocoto_state is Done.
+        Assuming rocotorun had just been run, and the rocoto_state is not Done, then
+        rocoto_state is Stalled if there are no jobs that are RUNNING, SUBMITTING, or QUEUED.
+        """
```
WalterKolczynski-NOAA, line 27: This seems more like a description for the whole script or `rocoto_statcount` than this function.

WalterKolczynski-NOAA:
@TerryMcGuinness-NOAA still waiting on you to make the suggestions I recommended

PR #2074, opened 2023-11-16
Fix some rocoto dependencies

### LINKED ISSUE #2075: Incorrect offset string for intervals of 24 hrs
### What is wrong?

When the forecast interval is 24, the generated XML contains an offset in the forecast job that is unrecognized by rocoto:
```<not><cycleexistdep cycle_offset="-1 day, 0:00:00"/></not>```

Generated XML offsets should be in `HH:MM:SS` format even when the interval is a day or more.

All or N/A

### Steps to reproduce

1. Create an experiment with gfs_cyc=1
2. Examine the gfsfcst job dependencies in the XML

### Additional information

None

Call `timedelta_to_HMS()` to make sure the string is in the correct format.

### PR DESCRIPTION
**Pull request by WalterKolczynski-NOAA**
# Description
A few small errors/cleanup have crept into the dependencies with recent changes , mostly in fcst:

- Fixed the offset, as it would yield a string rocoto doesn't recognize when the interval was 24+ hrs
- Removed the duplicate wave dependency. When `waveprep` is included, it depends on `waveinit` so there is no need to add a `waveinit` dependency.
- Converted an interval in the archive job that was missed in #2039
Resolves #2075

# Type of change

# Change characteristics

# How has this been tested?
- XML generation on Orion

# Checklist

### DIFF
--- a/workflow/rocoto/gfs_tasks.py
+++ b/workflow/rocoto/gfs_tasks.py
@@ -156,7 +156,7 @@ class GFSTasks(Tasks):
             interval = self._base['INTERVAL_GFS']
         elif self.cdump in ['gdas']:
             interval = self._base['INTERVAL']
-        offset = f'-{timedelta_to_HMS(interval)}'
+        offset = timedelta_to_HMS(-interval)
 
         # Files from previous cycle
         files = [f'@Y@m@d.@H0000.fv_core.res.nc'] + \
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by KateFriedman-NOAA — APPROVED** [review]
Looks good, thanks @WalterKolczynski-NOAA !

File: workflow/rocoto/gfs_tasks.py [hunk 1]
```diff
@@ -156,7 +156,7 @@ def aerosol_init(self):
             interval = self._base['INTERVAL_GFS']
         elif self.cdump in ['gdas']:
             interval = self._base['INTERVAL']
-        offset = f'-{timedelta_to_HMS(interval)}'
+        offset = timedelta_to_HMS(-interval)
```
aerorahul, line 159: so if L1013 works, why does this not?

WalterKolczynski-NOAA: This worked, but I changed this to be consistent with L502 (it is the cleaner way to do it).

aerorahul: ok. thanks.  I was trying to understand what was the "bug" part of this PR.

aerorahul (approved):
looks good.

workflow/rocoto/gfs_tasks.py
WalterKolczynski-NOAA: Bug was the one on L502 that wasn't using `timedelta_to_HMS()`.

HenryRWinterbottom (approved):
LGTM.

Issue #1996, opened 2023-10-30
GSI-based analysis jobs dying after hitting wallclock limit on Orion


WalterKolczynski-NOAA:
### What is wrong?

The GSI-based analysis jobs are running out of walltime before completing on Orion.

### What should have happened?

Analysis jobs should complete

### What machines are impacted?

Orion

### Steps to reproduce

1. Set up a cycled experiment on Orion
2. Observe workflow failure for analysis jobs

### Additional information

Unsure what conditions changed on Orion to cause this issue.

RussTreadon-NOAA:
@WalterKolczynski-NOAA , gfsanal completes but gdasanal does not.  Looking at the log files gdasanal has finished two outer loops.  It is writing obs-anl information to diagnostic files via ncdiag.  `gsi.x` appears to be hanging during this process.   

I copied the C96C48_hybatmDA setup to my directory.  Submit gdasanal with the obs-anl diagnostics turned off via the addition of 
```
##TEST
export SETUP='write_diag(3)=.false.,'
##TEST
```
to `config.anal`.    gdasanal ran to completion with a total elapsed time of 01:19:51 (hh:mm:ss).   If I turn on the obs-anl diagnostics (ie, revert to `write_diag(3)=.true.`), I expect `gsi.x` will hang.

@DavidHuber-NOAA found some [memory issues](https://github.com/NOAA-EMC/GSI-ncdiag/issues/7) with ncdiag.  Not sure if what he found is related to the `gsi.x` hangs in gdasanal.

The GSI hash cloned with g-w dates back to 9/25/2023.  When did gdasanal start failing due to exceeding the wall clock limit?   Knowing when might point to system and/or library changes.

aerorahul:
@RussTreadon-NOAA 
As you noted, the GSI hash was last updated 9/25/2023.  There were no test timeouts in the GSI analysis steps until last week.  
[This](https://github.com/NOAA-EMC/global-workflow/pull/1966#issuecomment-1778189340) comment from a PR last week (10/24/2023) shows no timeouts in the GSI analysis jobs on Orion.
The weekly test on 10/23/2023 also had no issues. See [this](https://github.com/NOAA-EMC/global-workflow/pull/1958) comment.
These timeouts started after 10/24/2023.  Incidentally, Orion had PM on 10/23/2023.

RussTreadon-NOAA:
[...]

PR #2077, opened 2023-11-17
Add --cpus-per-task to srun commands for Hera and Orion

### LINKED ISSUE #2044: Add --cpus-per-task to APRUN commands for Hera and Orion

After update to SLURM version on Orion we need to now set the `--cpus-per-task` flag with the srun command.

Also add flag for srun commands on Hera for consistency even though workaround was implemented by sysadmins there.

Issue #1996 documents a slowdown in the analysis job (`gsi.x`), troubleshooting with Orion helpdesk, and resulting solution tested by @RussTreadon-NOAA . A hotfix is being made to add the flag for the analysis job. Now investigate adding the flag for all job srun commands on Orion.

Comments from helpdesk:
```
Russ, I've been digging into this, and I spotted something this time. You and another user who's been having a 
similar slowdown issue with a completely different package have the same sbatch directive:

#SBATCH --cpus-per-task=10

Per the slurm release page:

"NOTE: Beginning with 22.05, srun will not inherit the --cpus-per-task value requested by salloc or sbatch. It 
must be requested again with the call to srun or set with the SRUN_CPUS_PER_TASK environment variable if 
desired for the task(s). "

I have actually just been testing using

#SBATCH --ntasks=400

since this test is nodes=10, with 40 tasks per node, and it appears to be MUCH faster.

For your submission, you should use the same I believe, --ntasks=400.
```
```
The following change was implemented on NOAA's R&D HPC systems after upgrading
to version 22.05.6.

There has been a change of behavior in Slurm going from 21.08.8-2 to 22.05.6.
In 21, when srun was run from an sbatch-submitted job script, it inherited the
cpus-per-task from the sbatch environment. In 22, srun seems to default to
cpus-per-task=2 instead of inheriting from the sbatch environment. This causes
massive slowdowns in jobs which are expecting the 21.x inheritance behavior.

We are adding the following line to Hera's task prolog:
echo export SRUN_CPUS_PER_TASK=$SLURM_CPUS_PER_TASK

This line sets the contents of the SRUN_CPUS_PER_TASK variable (used by srun)
to whatever the existing contents of the SLURM_CPUS_PER_TASK were, thus
emulating the 21.x behavior.
```

The new flag works on Hera and Orion.

### Acceptance Criteria

Jobs run at same speed (or faster). Evaluate speed.

Add `--cpus-per-task $THREAD_VAR` to each job in `HERA.env` and `ORION.env`.

KateFriedman-NOAA:
# Description

This PR adds the `--cpus-per-task` flag to non-CFP `APRUN`/srun commands in the env files for Orion and Hera. This is needed as a result of an upgrade to SLURM on Orion that no longer exported the thread value to the environment. Issue #1996 was a hotfix to add this flag for the analysis job that was suddenly running slowly after the SLURM upgrade. A similar SLURM upgrade was done on Hera but the sysadmins there implemented a workaround so this issue did not appear there.

This PR adds the `--cpus-per-task` flag for both machines...needed on Orion and as a safety measure on Hera.

Also included in this PR is a fix to indentations in the `scripts/exglobal_diag.sh` script that was noticed during work for this PR.

Timing differences were not observed for other jobs after adding this srun flag. Job timings were of a similar variation between control and test runs. Any potential additional speedup was not observed. The test runs also reproduced the control runs with `develop`.

Resolves #2044

# Type of change

# Change characteristics

# How has this been tested?

- Cycled tests on Hera and Orion against controls
- Coupled forecast-only test on Orion against control


aerorahul (approved):
looks good.

aerorahul:
Adding the JEDI developers to review the changes to their specific tasks.

RussTreadon-NOAA (approved):
@KateFriedman-NOAA reports no change in analysis, forecast, or downstream results & no impact on job run times.

`atmanalrun` and `atmensanlrun` changes look correct.  Did not test to confirm they work as intended.

Approve pending successful completion of CI.

WalterKolczynski-NOAA (approved):

guillaumevernieres (approved):
Thanks @KateFriedman-NOAA .

jiaruidong2017 (approved):
Looks good to me.

DavidHuber-NOAA (approved):
Since ESMF handles threading, this doesn't need to be added to the forecast jobs, correct?

KateFriedman-NOAA:
> Since ESMF handles threading, this doesn't need to be added to the forecast jobs, correct?

@DavidHuber-NOAA As I understand it, yes.

env/HERA.env
```diff
@@ -43,7 +43,7 @@ if [[ "${step}" = "prep" ]] || [[ "${step}" = "prepbufr" ]]; then
 
 elif [[ "${step}" = "preplandobs" ]]; then
 
-    export APRUN_CALCFIMS="${launcher} -n 1"
+    export APRUN_CALCFIMS="${launcher} -n 1" # TODO: Does this need threading (--cpus-per-task flag)?
```
CoryMartin-NOAA: I don't think so, for this or the other land job

KateFriedman-NOAA: Thanks @CoryMartin-NOAA ! I have removed those comments.

CoryMartin-NOAA (approved):

PR #2089, opened 2023-11-28
Add wave init to GEFS

### LINKED ISSUE #2017: add Wave init job to GEFS free forecast

Add the wave prep job that creates the mod def file (inp2moddef) to the GEFS free forecasting g-w configuration 

a new task dependency in the GEFS g-w configuration 

### Acceptance Criteria

wave mod def file will be created at the beginning of a g-w GEFS suite to be used for the free forecasts

- Hopefully existing scripts for GFS g-w can be re-used.

WalterKolczynski-NOAA:
# Description
Adds the wave init job to GEFS. Since the output would be the same for every member, the job is only run once for the control, then links are created in each of the member directories pointing to that copy.

New ww3_multi and ww3_shell templates for GEFS are copied from the GFS versions.

The GEFS restart frequency is changed to not be zero, and the comment about not setting it to zero is copied from the gfs config.base. The restart interval for the GEFS perturbed members is changed to equal that of the control.

Also fixes a bug where `RUNwave` was not defined for ensemble forecasts, and another bug in the GEFS stage_ic dependency.

Resolves #2017

# Type of change

# Change characteristics

# How has this been tested?
- GEFS test on Orion
- Cycled GFS test on Orion
- F-O GFS test on Orion

# Checklist


aerorahul (approved):
Looks good to me, though I am not familiar with wave initialization needs.

parm/wave/ww3_multi.gefs.inp.tmpl
```diff
@@ -0,0 +1,38 @@
+$ WAVEWATCH III multi-grid input file
```
JessicaMeixner-NOAA: At this point this is not needed for GEFS, so I'm not sure if it's worth creating this file?

scripts/exgfs_wave_init.sh
```diff
@@ -203,6 +203,17 @@ source "${HOMEgfs}/ush/preamble.sh"
     fi
   done
 
+# Copy to other members if needed
```
JessicaMeixner-NOAA, line 206: Does this need to be linked this many times or can we simply just grab the single mod_def file in the forecast job?

JessicaMeixner-NOAA (commented):
From what I see everything here is fine, although I made a couple of comments for your consideration of potential alternatives.   For the GEFS orion test, is there output by any chance I could take a quick look at for S2SW or S2SWA?

JessicaMeixner-NOAA:
Tagging @sbanihash as she is the GEFS wave POC and should also take a look at this PR.

WalterKolczynski-NOAA:
@JessicaMeixner-NOAA @sbanihash  Output can be found at `/work2/noaa/global/wkolczyn/noscrub/global-workflow/wave_init_gefs`

scripts/exgfs_wave_init.sh
WalterKolczynski-NOAA: When the forecast runs (and other jobs?), they don't really know how to find the deterministic member. Plus the GFS and GEFS (currently) label the deterministic differently (GEFS uses `mem000`, whereas GFS is in an entirely different `RUN` from EnKF).

JessicaMeixner-NOAA: Thanks for the explanation

parm/wave/ww3_multi.gefs.inp.tmpl
WalterKolczynski-NOAA: Maybe not, but even the GFS isn't being used currently.

Is there any reason these files would be different for GFS and GEFS? We could just use symlinks for these two (`ww3_multi` and `ww3_shel`) if not (or just take the `$RUN` out of the filename).

JessicaMeixner-NOAA: Fair enough. 

These should be configurable enough to handle GFS or GEFS and allow for taking $RUN out of the filename.

WalterKolczynski-NOAA: Consolidated these into single files without the `$RUN`.

WalterKolczynski-NOAA:
Added @TerrenceMcGuinness-NOAA since I'm now updating a CI test.

JessicaMeixner-NOAA (approved):

PR #2093, opened 2023-11-28
Robust CI Restarts

### LINKED ISSUE #2060: bug/fix on false positive for link phase fail
### What is wrong?

On a user rebuild from label change in CI a corner case emerges when the link step fails and a false positive occurs because the driver scripts doesn't issue the fail to stop.

The driver script calling the build script in CI should have processed the error code correctly rather than simply exiting when a rebuild had also been issued by the user.

All or N/A

### Steps to reproduce

This only occurs when there is a non-zero error code returned from the link phase on a re-build request from the use from a label change.

### Additional information

Needs to be fixed so it a link failure stops the CI test so it can be addressed.

Update the driver script to interrogate the return code to check if a link failure occurred to cover for the above corner case.

### LINKED ISSUE #2067: Add Kill Switch (Label) to CI

Add a new Label MACINE-Kill that Removes PR from the CI System entirely and does not restart it.

NOTE: All the functionality for this feature is already in place with [PR 2093](https://github.com/NOAA-EMC/global-workflow/pull/2093) and should not take much programming effort.

When ever a previously issued Read label PR CI test is in progress this Label update by the user will:
- Have it removed from the test suite (and data base)
- Kill the all the build scripts the CI is currently building when the kill switch is issued
- Purge the test directories from the file system
- Remove all jobs in scheduler all running from that PR

### Acceptance Criteria

Test in CI cron development environment and above capabilities can be demonstrated.

Update the check CI bash  script to read the db state record and include a time stamp to react to this case.
This can also be used for when ever PR is merged and there are labels other than **PASSED**

TerrenceMcGuinness-NOAA:
# Description

Improved CI robustness for reverting back to **CI-Ready** from any given state

New Features:

- Improved `scancel` routine (refactored into bash "subroutine")
- Improved messaging (see below) when ever a user changes state
- Any and all previous build scripts and running experiments are killed as a result of reset to **Ready**

  Resolves #2060

- Bug fix (Resetting CI tests had lingering builds simultaneously running)
- New feature (Robust Resetting with messages)
- Maintenance (code refactor - use of BASH subroutines in main drivers)

Example of output when user updates state and request a restart:

```
CI Update on Hera at 11/28/23 09:33:11 PM
=================================================
PR:201 Reset to Hera-Ready by user and is now restarting CI tests
Driver PID: 302744 on hfe12 is no longer running this test
Driver PID: has restarted as 163669 on hfe12
No current experiments to cancel in PR: 201 on Hera
```

# Change characteristics

# How has this been tested?
Reset at all steps in dev Cron CI space with appropriate kills taking place on Hera and Orion

# Checklist


ci/scripts/check_ci.sh
```diff
@@ -33,6 +33,7 @@ case ${MACHINE_ID} in
 esac
 set +x
 source "${ROOT_DIR}/ush/module-setup.sh"
+source "${ROOT_DIR}/ci/scripts/utils/ci_utils.h"
```
WalterKolczynski-NOAA:
```suggestion
source "${ROOT_DIR}/ci/scripts/utils/ci_utils.sh"
```

ci/scripts/driver.sh
```diff
@@ -68,24 +69,54 @@ pr_list=$(${GH} pr list --repo "${REPO_URL}" --label "CI-${MACHINE_ID^}-Ready" -
...
+    rm -f "${output_ci_single}"
+    {
+      echo "CI Update on ${MACHINE_ID^} at $(date +'%D %r')" || true
+      echo "================================================="
+      echo "PR:${pr} Reset to ${MACHINE_ID^}-Ready by user and is now restarting CI tests" || true
+    } >> "${output_ci_single}"
+    if [[ "${driver_PID}" -ne 0 ]]; then
+      echo "Driver PID: ${driver_PID} no longer running this build having it killed"
+      if [[ "${driver_HOST}" == "${host_name}"  ]]; then
+        pstree -A -p "${driver_PID}" | grep -Eow "[0-9]+" | xargs kill || true
```
WalterKolczynski-NOAA: It's more robust to include look-behind and look-ahead to make sure it is the right bit. Also I don't think we want to swallow any errors with `|| true`.
```suggestion
        pstree -A -p "${driver_PID}" | grep -Pow "(?<=\()[0-9]+(?=\))" | xargs kill
```

ci/scripts/driver.sh
```diff
@@ -68,24 +69,54 @@ pr_list=$(${GH} pr list --repo "${REPO_URL}" --label "CI-${MACHINE_ID^}-Ready" -
...
+        sleep 30
```
WalterKolczynski-NOAA: Not sure why we need to sleep.

ci/scripts/driver.sh
```diff
@@ -68,24 +69,54 @@ pr_list=$(${GH} pr list --repo "${REPO_URL}" --label "CI-${MACHINE_ID^}-Ready" -
...
+      else
+        ssh "${driver_HOST}" 'pstree -A -p "${driver_PID}" | grep -Eow "[0-9]+" | xargs kill'
+        sleep 30
       fi
-      pslot=$(basename "${cases}")
-      sacct --format=jobid,jobname%35,WorkDir%100,stat | grep "${pslot}" | grep "PR\/${pr}\/RUNTESTS" |  awk '{print $1}' | xargs scancel || true
-    done
-    rm -Rf "${pr_dir}"
+      {
+        echo "Driver PID: ${driver_PID} on ${driver_HOST} is no longer running this test"
```
WalterKolczynski-NOAA: Message adjustment since we can't guarantee the jobs have actually been killed (unless we check again).
```suggestion
        echo "Driver PID: Requested termination of ${driver_PID} and children on ${driver_HOST}"
```

ci/scripts/utils/ci_utils.sh
```diff
@@ -0,0 +1,16 @@
...
+
+function cancel_slurm_jobs() {
+
+  local substring=$1
+  local job_ids
+  job_ids=$(squeue -u "${USER}" -h -o "%i")
+
+  for job_id in ${job_ids}; do
+    job_name=$(sacct -j "${job_id}" --format=JobName%100 | head -3 | tail -1 | sed -r 's/\s+//g') || true
+    if [[ "${job_name}" == *"${substring}"* ]]; then
```
WalterKolczynski-NOAA:
```suggestion
    if [[ "${job_name}" =~ "${substring}" ]]; then
```

WalterKolczynski-NOAA (changes requested):

ci/scripts/driver.sh
```diff
@@ -68,24 +69,54 @@ pr_list=$(${GH} pr list --repo "${REPO_URL}" --label "CI-${MACHINE_ID^}-Ready" -
...
+        echo "Driver PID: has restarted as $$ on ${host_name}"
+      } >> "${output_ci_single}"
+    fi
+
+    experiments=$(find "${pr_dir}/RUNTESTS/EXPDIR" -mindepth 1 -maxdepth 1 -type d) || true
+    if [[ -z "${experiments}" ]]; then
+       echo "No current experiments to cancel in PR: ${pr} on ${MACHINE_ID^}" >> "${output_ci_single}"
+    else
+      for cases in ${experiments}; do
```
DavidHuber-NOAA: Nitpick: should probably be `case` instead of `cases`.

ci/scripts/check_ci.sh
TerrenceMcGuinness-NOAA: yeah just getting to that so my jobs can now finish this morning, thanks!

ci/scripts/driver.sh
TerrenceMcGuinness-NOAA: Ok that should be fine to do a better job on the grep filter, and that darn true is for the shellnorm complaints and I need it on one line for the ssh part, I'll add an ignore exception for that then.

ci/scripts/driver.sh
TerrenceMcGuinness-NOAA: agreed, I was chasing down race conditions when I put that in there.

ci/scripts/utils/ci_utils.sh
TerrenceMcGuinness-NOAA: didn't know about this construct, it will still pick up substrings with this operator?

ci/scripts/driver.sh
```diff
@@ -68,24 +69,54 @@ pr_list=$(${GH} pr list --repo "${REPO_URL}" --label "CI-${MACHINE_ID^}-Ready" -
...
+        cancel_slurm_jobs "${case_name}"
+        {
+          echo "Canceled all jobs for experiment ${case_name} in PR:${pr} on ${MACHINE_ID^}"
+        } >> "${output_ci_single}"
+      done
+    fi
+    sed -i "1 i\`\`\`" "${output_ci_single}"
+    "${GH}" pr comment "${pr}" --repo "${REPO_URL}" --body-file "${output_ci_single}"
+    db_list=$("${ROOT_DIR}/ci/scripts/pr_list_database.py" --remove_pr "${pr}" --dbfile "${pr_list_dbfile}")
+    db_list=$("${ROOT_DIR}/ci/scripts/pr_list_database.py" --add_pr "${pr}" --dbfile "${pr_list_dbfile}")
```
DavidHuber-NOAA: I'm not sure that assigning this output to `db_list` is necessary unless some error checking needs to happen.  Should there be?

DavidHuber-NOAA (commented):
Looks good, just a couple nitpicks.

ci/scripts/driver.sh
TerrenceMcGuinness-NOAA: Oh right, that was a cut and paste oversight.

ci/scripts/utils/ci_utils.sh
WalterKolczynski-NOAA: Yes, it is a regex matcher.

TerrenceMcGuinness-NOAA (commented):
Doing extra commit to make sure reviews are in.

ci/scripts/driver.sh
```diff
@@ -89,14 +89,14 @@ for pr in ${pr_list}; do
     if [[ "${driver_PID}" -ne 0 ]]; then
       echo "Driver PID: ${driver_PID} no longer running this build having it killed"
       if [[ "${driver_HOST}" == "${host_name}"  ]]; then
-        pstree -A -p "${driver_PID}" | grep -Eow "[0-9]+" | xargs kill || true
-        sleep 30
+        #shellcheck disable=SC2312,SC2312
```
WalterKolczynski-NOAA:
```suggestion
        # shellcheck disable=SC2312
```

WalterKolczynski-NOAA (changes requested):

ci/scripts/driver.sh
TerrenceMcGuinness-NOAA: I was doubling that trying to get it to stop complaining still.

ci/scripts/utils/ci_utils.sh
```diff
@@ -0,0 +1,16 @@
...
+
+  local substring=$1
+  local job_ids
+  job_ids=$(squeue -u "${USER}" -h -o "%i")
+
+  for job_id in ${job_ids}; do
+    job_name=$(sacct -j "${job_id}" --format=JobName%100 | head -3 | tail -1 | sed -r 's/\s+//g') || true
+    if [[ "${job_name}" =~ ${substring} ]]; then
+      echo "Canceling Slurm Job ${job_name} with: scancel ${job_id}"
+      scancel "${job_id}"
```
aerorahul, line 20: You could add a `break` here since the match is made and there is nothing more to do.  Otherwise, it will just keep looping till the end of all jobs.

ci/scripts/driver.sh
```diff
@@ -186,7 +227,7 @@ for pr in ${pr_list}; do
     done
 
     "${GH}" pr edit --repo "${REPO_URL}" "${pr}" --remove-label "CI-${MACHINE_ID^}-Building" --add-label "CI-${MACHINE_ID^}-Running"
-    "${ROOT_DIR}/ci/scripts/pr_list_database.py" --dbfile "${pr_list_dbfile}" --update_pr "${pr}" Open Running
+    "${ROOT_DIR}/ci/scripts/pr_list_database.py" --dbfile "${pr_list_dbfile}" --update_pr "${pr}" Open Running "0:0"
```
aerorahul, line 233: What is "0:0"?

ci/scripts/driver.sh
```diff
@@ -174,7 +215,7 @@ for pr in ${pr_list}; do
         } >> "${output_ci}"
       else
         {
-          echo "*** Failed *** to create experiment: ${pslot}"
+          echo "*** Failed *** to create experiment: ${pslot} on ${MACINE_ID^}"
```
aerorahul: typo
```suggestion
          echo "*** Failed *** to create experiment: ${pslot} on ${MACHINE_ID^}"
```

ci/scripts/driver.sh
```diff
@@ -110,34 +141,44 @@ for pr in ${pr_list}; do
...
   fi
-  "${GH}" pr edit --repo "${REPO_URL}" "${pr}" --remove-label "CI-${MACHINE_ID^}-Ready" --add-label "CI-${MACHINE_ID^}-Building"
-  "${ROOT_DIR}/ci/scripts/pr_list_database.py" --dbfile "${pr_list_dbfile}" --update_pr "${pr}" Open Building
-  echo "Processing Pull Request #${pr}"
+  id=$("${GH}" pr view "${pr}" --repo "${REPO_URL}" --json id --jq '.id')
   pr_dir="${GFS_CI_ROOT}/PR/${pr}"
+  output_ci="${pr_dir}/output_ci_${id}"
+  output_ci_single="${GFS_CI_ROOT}/PR/${pr}/output_single.log"
+  driver_build_PID=$$
+  driver_build_HOST=$(hostname -s)
```
aerorahul, line 152: Isn't this already `MACHINE_ID`?

aerorahul (commented):
There is a lot of dense logic being introduced here that will make it very difficult to follow (except for the developer ofcourse).
This is fine for now, but once the system is running as expected, please make a note to make it lean and clearer.

ci/scripts/driver.sh
TerrenceMcGuinness-NOAA: No, MACHINE_ID is say Hera, hostname is the specific head node that the PID for driver.sh is running on.

ci/scripts/driver.sh
TerrenceMcGuinness-NOAA: first field is PID second is `hostname`

ci/scripts/utils/ci_utils.sh
TerrenceMcGuinness-NOAA: oh yes, so is should be a `continue` so we can go on to the next one because it needs to kill all the jobs associated to that specific running Case name as it has a tag too.

ci/scripts/driver.sh
aerorahul: This is the exact front-end node the build is being performed.

ci/scripts/driver.sh
TerrenceMcGuinness-NOAA: updated documentation

ci/scripts/clone-build_ci.sh
```diff
@@ -72,16 +72,17 @@ cd sorc || exit 1
 set +e
 ./checkout.sh -c -g -u >> log.checkout 2>&1
 checkout_status=$?
+DATE=$(date +'%D %r')
 if [[ ${checkout_status} != 0 ]]; then
   {
     echo "Checkout: *** FAILED ***"
-    echo "Checkout: Failed at $(date)" || true
+    echo "Checkout: Failed at ${DATE}" || true
```
aerorahul: Minor quib here: `DATE` here will be the `date` when `DATE` was set and not the actual date "now".

ci/scripts/driver.sh
```diff
@@ -68,24 +69,57 @@ pr_list=$(${GH} pr list --repo "${REPO_URL}" --label "CI-${MACHINE_ID^}-Ready" -
...
+      echo "PR:${pr} Reset to ${MACHINE_ID^}-Ready by user and is now restarting CI tests" || true
+    } >> "${output_ci_single}"
+    if [[ "${driver_PID}" -ne 0 ]]; then
+      echo "Driver PID: ${driver_PID} no longer running this build having it killed"
+      if [[ "${driver_HOST}" == "${host_name}"  ]]; then
+        # shellcheck disable=SC2312
+        pstree -A -p "${driver_PID}" | grep -Pow "(?<=\()[0-9]+(?=\))" | xargs kill
+      else
+        # shellcheck disable=SC2312
+        ssh "${driver_HOST}" 'pstree -A -p "${driver_PID}" | grep -Eow "[0-9]+" | xargs kill'
```
aerorahul, line 99: passwordless ssh in an automated script to kill jobs -- what could possibly go wrong?

TerrenceMcGuinness-NOAA (commented):
updated doc and added `continue` for look for next slurm job id

ci/scripts/driver.sh
TerrenceMcGuinness-NOAA: I know right

TerrenceMcGuinness-NOAA: My ssh-password-less was working fine on Hera and Orion is always on Login Node 1 so it will use the first clause in the IF.  When we go nems account we need to make sure ssh-passwordless-ness is working on Hera.

ci/scripts/clone-build_ci.sh
TerrenceMcGuinness-NOAA: It's set on the line above and helps with shellnorms, I should take out the || true now.

TerrenceMcGuinness-NOAA (commented):
updated all changes

ci/scripts/clone-build_ci.sh
TerrenceMcGuinness-NOAA: is in the line above and is set in to places on the condition a billionth of second later (this was to appease shell norms).

aerorahul (approved):
I haven't looked in detail with the changes, but as long as it works this is fine. 
We should clean this up in the future to make it clearer.

ci/scripts/utils/ci_utils.sh
WalterKolczynski-NOAA: Did this change get undone at some point? It is showing as `==` again.

WalterKolczynski-NOAA: Updated suggestion, since shellcheck was complaining:
```suggestion
    if [[ "${job_name}" =~ ${substring} ]]; then
```

TerrenceMcGuinness-NOAA (commented):
committed missed regex on BASH substring if statement.

WalterKolczynski-NOAA (approved):

Issue #1588, opened 2023-05-09
Port to Hercules


KateFriedman-NOAA:
**Description**

Update the system to support the new MSU Hercules.

**Dependencies**

- [x] EPIC hpc-stack/spack-stack available
- [x] Staged fix files in place (if not accessible from Orion side) - update `link_workflow.sh` to add Hercules on Orion line
- [x] Staged ICs in place (if not accessible from Orion side) - update `config.coupled_ic` for Hercules (same as Orion)
- [x] Test GDA in place (if not accessible from Orion side) - add Orion `DMPDIR` to Hercules hosts file
- [x] Workflow updates (e.g. env)
- [x] ufs-weather-model support: https://github.com/ufs-community/ufs-weather-model/pull/1733
- [x] UPP support https://github.com/NOAA-EMC/UPP/pull/760
- [x] UFS_UTILS support https://github.com/ufs-community/UFS_UTILS/pull/880
- [x] gfs-utils support
- [x] GSI-EnKF support https://github.com/NOAA-EMC/GSI/issues/574
- [x] GSI-Utils support
- [ ] GSI-Monitor support
- [ ] GDASApp support https://github.com/NOAA-EMC/GDASApp/pull/774
- [ ] EMC_verif-global support
- [x] OBSPROC packages support (obsproc/prepobs) https://github.com/NOAA-EMC/prepobs/issues/21
- [x] TC_tracker support https://github.com/NOAA-EMC/TC_tracker/issues/3
- [x] Fit2Obs support ~https://github.com/NOAA-EMC/Fit2Obs/issues/15~ Not necessary to port.  The Orion executables work fine on Hercules.

spack-stack is available on hercules via:
```sh
module purge
module use /work/noaa/epic-ps/role-epic-ps/spack-stack/modulefiles
```

Information from admin announcement:

A new computing resource is now available for the NOAA R&D HPC user community at Mississippi State University (MSU).
The new system has been named “Hercules” and as with Orion, is owned and managed by MSU. As this is a brand new
system, with a brand new software stack, we would like your help in flushing out any issues before we place the system
into full production. So we are asking our current Orion users to assist us with the pre-operational testing of this new
system. We would greatly appreciate it if you could try compiling your models, running your models, testing your
workflows, and then provide us with feedback by emailing “rdhpcs.orion.help@noaa.gov”. Please also email any questions
or issues to the same email address.  If everything goes well with the testing then we hope to announce full production
in early May. Thank you for all your help!

[...]

PR #2128, opened 2023-12-06
Enable forecast-only experiments on Hercules

DavidHuber-NOAA:
# Description
This add forecast-only support for Hercules to the global workflow. Partially satisfies #1588.

Refs #1588 

# Type of change

# Change characteristics

# How has this been tested?
Forecast-only ATM and ATMA experiments on Hercules.

# Checklist


parm/config/gfs/config.aero
```diff
@@ -11,7 +11,7 @@ case ${machine} in
   "HERA")
     AERO_INPUTS_DIR="/scratch1/NCEPDEV/global/glopara/data/gocart_emissions"
     ;;
-  "ORION")
+  "ORION"|"HERCULES")
```
WalterKolczynski-NOAA:
```suggestion
  "ORION" | "HERCULES")
```

ush/detect_machine.sh
```diff
@@ -26,6 +26,9 @@ case $(hostname -f) in
 
   Orion-login-[1-4].HPC.MsState.Edu) MACHINE_ID=orion ;; ### orion1-4
 
+  hercules-login-[1-4].hpc.msstate.edu) MACHINE_ID=hercules ;; ### hercules1-4
+  Hercules-login-[1-4].HPC.MsState.Edu) MACHINE_ID=hercules ;; ### hercules1-4
```
WalterKolczynski-NOAA: are both of these needed?

versions/run.hercules.ver
```diff
@@ -0,0 +1,10 @@
+export stack_intel_ver=2021.9.0
+export stack_impi_ver=2021.9.0
+
+export ncl_ver=6.6.2
+
+source "${HOMEgfs:-}/versions/run.spack.ver"
+
+# wgrib2 and cdo are different on Hercules from all the other systems
+export wgrib2_ver=3.1.1
+export cdo_ver=2.2.0
```
WalterKolczynski-NOAA, line 10: Is it possible to update the versions for the other machines so they match again?

WalterKolczynski-NOAA (dismissed):
Looks pretty good. There are some things that match what is already there but probably shouldn't be what we are doing that led me to opening a couple new issues to be handled separately.

versions/run.hercules.ver
DavidHuber-NOAA: Unfortunately not.  These versions are not available on the other machines under spack-stack.  It looks like the standard versions (`cdo/2.0.5` and `wgrib2/2.0.8`) [don't compile on Hercules](https://github.com/JCSDA/spack-stack/blob/8780bcdff613e970e2cf881301587b72bd7f3ef8/configs/sites/hercules/packages.yaml#L24-L30).  There are plans to upgrade to these versions for all systems in future releases (https://github.com/JCSDA/spack-stack/pull/884 and https://github.com/JCSDA/spack-stack/issues/859).

ush/detect_machine.sh
DavidHuber-NOAA: Probably not.  Hercules' nodes are all undercased, but Orion's start with uppercase.  This switched on S4 once after an update, so I put both in here for the just-in-case scenario.  I'm fine with taking it out if that's preferred.

KateFriedman-NOAA: FYI Orion once changed case on part of the system as well, which broke our builds, We complained to the Orion helpdesk and got them to change it back (it was an unintentional change IIRC). So we can likely go with what it is now on Hercules.

KateFriedman-NOAA (approved):
Looks great! No comments. Thanks for this @DavidHuber-NOAA !

ush/detect_machine.sh
DavidHuber-NOAA: Thanks @KateFriedman-NOAA!  I'll just go with lower case then.

aerorahul (approved):
Looks good.
Thanks for this first step on porting to Hercules.

DavidHuber-NOAA:
Starting CI testing.

junwang-noaa:
@DavidHuber-NOAA @aerorahul Just let you know we found some issues on HDF5/nccmp on Hercules, see ufs WM issue https://github.com/ufs-community/ufs-weather-model/issues/2015

DavidHuber-NOAA:
Thanks for the heads up @junwang-noaa!

Issue #289, opened 2021-03-10
[NCO Bug] Remove usage of compath.py for COMIN variables


WalterKolczynski-NOAA:
The current GFS package uses compath.py to determine the default $COMIN. However, compath.py should only be used for determining the path of upstream models and is not intended for references to the current model.

NCO requests this be corrected in the next major upgrade.

```
JGDAS_ATMOS_GEMPAK:export COMIN=${COMIN:-$(compath.py ${NET}/${envir}/${RUN}.${PDY})/${cyc}/$COMPONENT}
JGDAS_ATMOS_GEMPAK_META_NCDC:export COMIN=${COMIN:-$(compath.py ${NET}/${envir}/${RUN}.${PDY})/${cyc}/$COMPONENT/gempak}
JGFS_ATMOS_AWIPS_20KM_1P0DEG:export COMIN=${COMIN:-$(compath.py ${NET}/${envir}/${RUN}.${PDY})/${cyc}/$COMPONENT}
JGFS_ATMOS_GEMPAK_PGRB2_SPEC:export COMIN=${COMIN:-$(compath.py ${NET}/${envir}/${NET}.${PDY})/${cyc}/$COMPONENT}
JGFS_ATMOS_PGRB2_SPEC_NPOESS:export COMIN=${COMIN:-$(compath.py ${NET}/${envir}/${RUN}.${PDY})/${cyc}/$COMPONENT}
JGLOBAL_ATMOS_EMCSFC_SFC_PREP:export COMINgfs=${COMINgfs:-$(compath.py $NET/$envir/$RUN.$PDY)/$cyc/$COMPONENT}
JGLOBAL_WAVE_GEMPAK:export COMIN=${COMIN:-$(compath.py ${NET}/${envir}/${RUN}.${PDY})/${cyc}/$COMPONENT}
JGLOBAL_WAVE_PRDGEN_BULLS:export COMIN=${COMIN:-$(compath.py ${NET}/${envir}/${RUN}.${PDY})/${cyc}/$COMPONENT}
JGLOBAL_WAVE_PRDGEN_GRIDDED:export COMIN=${COMIN:-$(compath.py ${NET}/${envir}/${RUN}.${PDY})/${cyc}/$COMPONENT}
```

Refs: [Bugzilla #1210](http://www2.spa.ncep.noaa.gov/bugzilla/show_bug.cgi?id=1210)

WalterKolczynski-NOAA:
[...]

PR #2147, opened 2023-12-12
Stop Gempak pgrb2 spec from modifying $DATA

### LINKED ISSUE #1224: Gempak pgrb2 spec should not modify DATA during execution
**Description**
The current `JGFS_ATMOS_GEMPAK_PGRB2_SPEC` is redefining `$DATA` between each ex-script call.

**Requirements**
`$DATA` should not be modified during execution.

Updated scripts that do not change `$DATA` without causing file collisions that could cause unexpected behavior.

**Dependencies**
Can likely be accomplished at the same time as #1223

HenryRWinterbottom:
# Description

This PR addresses issue #1224. The following is accomplished:

- Instances where `DATA` is modified, corrections are made to use a locally defined variable such that `DATA` is no longer modified at run-time;
- Shell linter exceptions are addressed accordingly;
- The respective GEMPAK tasks is added to the Rocoto mesh.

Resolves #1224

# Type of change

# Change characteristics

# How has this been tested?

This needs to be tested using the CI and evaluated by the relevant parties.

# Checklist


jobs/JGFS_ATMOS_GEMPAK_PGRB2_SPEC
```diff
@@ -33,25 +31,21 @@ export EXT=""
...
-export COMIN=${COMIN:-$(compath.py ${envir}/${NET}/${gfs_ver})/${RUN}.${PDY}/${cyc}/${COMPONENT}}
-export COMOUT=${COMOUT:-$(compath.py -o ${NET}/${gfs_ver}/${NET}.${PDY})/${cyc}/${COMPONENT}/gempak}
-
-export SENDDBN=${SENDDBN:-NO}
-export DBNROOT=${DBNROOT:-${UTILROOT}/fakedbn}
-
-mkdir -m 775 -p ${COMOUT}
-
+export COMIN="${COMIN:-$(compath.py "${envir}"/"${NET}"/"${gfs_ver}")/${RUN}.${PDY}/${cyc}/${COMPONENT}}"
+export COMOUT="${COMOUT:-$(compath.py -o "${NET}"/"${gfs_ver}"/"${NET}"."${PDY}")/${cyc}/${COMPONENT}/gempak}"
```
aerorahul, line 35: This is not introduced in this PR, but IIRC, we are not supposed to use `compath.py` for locating data under the current application.  Use of `compath.py` should be to find data in upstream applications.
@WalterKolczynski-NOAA Do we have an issue for this?

KateFriedman-NOAA: I can answer that...we have https://github.com/NOAA-EMC/global-workflow/issues/289 to address this.

> Use of compath.py should be to find data in upstream applications.

Correct. Upstream or outside applications/models.

jobs/JGFS_ATMOS_GEMPAK_PGRB2_SPEC
```diff
@@ -33,25 +31,21 @@ export EXT=""
...
 
 #################################################################
 # Execute the script for the regular grib
 #################################################################
-export DATA=${DATA_HOLD}/SPECIAL
-mkdir -p ${DATA}
-cd ${DATA}
+export DATA_SPECIAL="${DATA}/SPECIAL"
+mkdir -p "${DATA_SPECIAL}"
+# shellcheck disable=SC2164
```
KateFriedman-NOAA: Don't need to disable SC2164 here...can address that shellcheck code by doing what is suggested here: https://www.shellcheck.net/wiki/SC2164

jobs/JGFS_ATMOS_GEMPAK_PGRB2_SPEC
```diff
@@ -65,14 +59,15 @@ echo "RUNS the Program"
...
 
 #################################################################
 # Execute the script for the 221 grib
 
-export DATA=${DATA_HOLD}/SPECIAL221
-mkdir -p ${DATA}
-cd ${DATA}
+export DATA_SPECIAL221="${DATA}/SPECIAL221"
+mkdir -p "${DATA_SPECIAL221}"
+# shellcheck disable=SC2164
```
KateFriedman-NOAA: Same comment about addressing SC2164.

jobs/JGFS_ATMOS_GEMPAK_PGRB2_SPEC
```diff
@@ -86,27 +81,28 @@ echo "RUNS the Program"
...
 ########################################################
 # Execute the script.
-${SRCgfs}/exgfs_atmos_goes_nawips.sh
+"${SRCgfs}/exgfs_atmos_goes_nawips.sh"
 export err=$?; err_chk
 ########################################################
 
 echo "end of program"
-cd ${DATA_HOLD}
+# shellcheck disable=SC2164
```
KateFriedman-NOAA: Same comment about addressing shellcheck code.

KateFriedman-NOAA (changes requested):
Update looks good...was it tested to make sure the underlying `scripts/exgfs_atmos_goes_nawips.sh` script doesn't fail because of the directory change? I ask this not 100% sure that the jobs work and are even testable so let me know if not.

Left a few comments, mostly regarding resolving shellcheck warnings instead of disabling.

jobs/JGFS_ATMOS_GEMPAK_PGRB2_SPEC
```diff
@@ -33,25 +31,21 @@ export EXT=""
...
-export DATA_HOLD=${DATA}
+mkdir -m 775 -p "${COMOUT}"
 
 #################################################################
 # Execute the script for the regular grib
 #################################################################
-export DATA=${DATA_HOLD}/SPECIAL
-mkdir -p ${DATA}
-cd ${DATA}
+export DATA_SPECIAL="${DATA}/SPECIAL"
```
aerorahul, line 45: I am not sure what the `exscript` here is doing or how it is handling data (and `DATA`).
All told, I don't think this is per EE2 either way.

aerorahul: FYI, it seems to be bigger than just what the title says.

jobs/JGFS_ATMOS_GEMPAK_PGRB2_SPEC
HenryRWinterbottom: @KateFriedman-NOAA @aerorahul Is this an oversight on my end or something that should be addressed in another issue/PR?

HenryRWinterbottom:
> Update looks good...was it tested to make sure the underlying `scripts/exgfs_atmos_goes_nawips.sh` script doesn't fail because of the directory change? I ask this not 100% sure that the jobs work and are even testable so let me know if not.
> 
> Left a few comments, mostly regarding resolving shellcheck warnings instead of disabling.

@KateFriedman-NOAA This was not tested. My understanding was that the testing of the verification scripts is up to the respective verification team.

jobs/JGFS_ATMOS_GEMPAK_PGRB2_SPEC
HenryRWinterbottom: @aerorahul ^^^ I apologize, I don't follow the last comment.

jobs/JGFS_ATMOS_GEMPAK_PGRB2_SPEC
aerorahul: another PR.

jobs/JGFS_ATMOS_GEMPAK_PGRB2_SPEC
HenryRWinterbottom: @KateFriedman-NOAA I didn't muck with resolving all of the linter complaints here. I don't have a way to check whether they will break the code. I am happy to look into this further, but I need a known configuration against which I can test such that any modifications I make do not break this. Similar to other instances, if this PR clears I am happy to open another issue/PR to address them once I can test things further.

KateFriedman-NOAA: Totally understand the concern but I don't think adding `|| exit` will break anything. If it does happen to break things when eventually tested then the developer who comes in to get these jobs working can resolve it. Would prefer to make small shellcheck fixes like this now than put in disables that could get missed by a follow-up developer.

Please make the changes to satisfy SC2164 (here and in other spots I noted). Thanks!

HenryRWinterbottom: @KateFriedman-NOAA I apologize, I missed some of these probably because I am trying to wrangle too PRs (and need coffee).

They should be resolved. Thank you for catching them.

jobs/JGFS_ATMOS_GEMPAK_PGRB2_SPEC
WalterKolczynski-NOAA: This also won't work anyway, as it still uses pre-COM-refactor paths.

jobs/JGFS_ATMOS_GEMPAK_PGRB2_SPEC
```diff
@@ -33,25 +31,20 @@ export EXT=""
...
+mkdir -p "${DATA_SPECIAL}"
+cd "${DATA_SPECIAL}" || exit
```
WalterKolczynski-NOAA: If the directory change fails, the script should fail
```suggestion
cd "${DATA_SPECIAL}" || exit 1
```

jobs/JGFS_ATMOS_GEMPAK_PGRB2_SPEC
```diff
@@ -65,14 +58,14 @@ echo "RUNS the Program"
...
 
 #################################################################
 # Execute the script for the 221 grib
 
-export DATA=${DATA_HOLD}/SPECIAL221
-mkdir -p ${DATA}
-cd ${DATA}
+export DATA_SPECIAL221="${DATA}/SPECIAL221"
+mkdir -p "${DATA_SPECIAL221}"
+cd "${DATA_SPECIAL221}" || exit
```
WalterKolczynski-NOAA:
```suggestion
cd "${DATA_SPECIAL221}" || exit 1
```

jobs/JGFS_ATMOS_GEMPAK_PGRB2_SPEC
```diff
@@ -86,27 +79,27 @@ echo "RUNS the Program"
...
 ########################################################
 # Execute the script.
-${SRCgfs}/exgfs_atmos_goes_nawips.sh
+"${SRCgfs}/exgfs_atmos_goes_nawips.sh"
 export err=$?; err_chk
 ########################################################
 
 echo "end of program"
-cd ${DATA_HOLD}
+cd "${DATA}" || exit
```
WalterKolczynski-NOAA:
```suggestion
cd "${DATA}" || exit 1
```

jobs/JGFS_ATMOS_GEMPAK_PGRB2_SPEC
WalterKolczynski-NOAA: After this PR, I'll update this job for the COM refactor.

WalterKolczynski-NOAA (changes requested):

WalterKolczynski-NOAA (approved):

Issue #2135, opened 2023-12-07
Set shared or exclusive for all jobs on WCOSS2


KateFriedman-NOAA:
### What new functionality do you need?

All jobs on WCOSS2 should have a setting for either shared or exclusive. An upcoming enforced requirement on WCOSS2 will need every job to set either shared or exclusive. No longer set just exclusive and not set shared (assume shared). Additionally, WCOSS2 now defaults to exclusive when not shared.

This came out of information given during updated WCOSS2 training.

### What are the requirements for the new functionality?

The PBS statements for rocoto (via the `<native>` tag) and ecflow should always set either "shared" or "exclusive" for each job on WCOSS2.

### Acceptance Criteria

Generated rocoto xmls and ecflow scripts always set either shared or exclusive for each job.

### To Do

- [x] Update GFSv17 development system - PR #2156
~~- [ ] Update GFSv16 operational system~~ Will update GFSv16 ops system if requested. Updated GFSv17 is sufficient for now.

KateFriedman-NOAA:
Generated rocoto xml on WCOSS2 now shows either exclusive (`exclhost`) or shared (`shared`) for every job:
```
kate.friedman@dlogin07> grep place testexcl.xml 
        <native>-l debug=true,place=vscatter:exclhost</native>
        <native>-l debug=true,place=vscatter:exclhost</native>
        <native>-l debug=true,place=vscatter:exclhost</native>
        <native>-l debug=true,place=vscatter:exclhost</native>
        <native>-l debug=true,place=vscatter:shared</native>
        <native>-l debug=true,place=vscatter:exclhost</native>
        <native>-l debug=true,place=vscatter:exclhost</native>
        <native>-l debug=true,place=vscatter:exclhost</native>
                <native>-l debug=true,place=vscatter:exclhost</native>
        <native>-l debug=true,place=vscatter:shared</native>
        <native>-l debug=true,place=vscatter:shared</native>
        <native>-l debug=true,place=vscatter:shared</native>
        <native>-l debug=true,place=vscatter:shared</native>
        <native>-l debug=true,place=vscatter:shared</native>
        <native>-l debug=true,place=vscatter:shared</native>
        <native>-l debug=true,place=vscatter:exclhost</native>
        <native>-l debug=true,place=vscatter:exclhost</native>
        <native>-l debug=true,place=vscatter:shared</native>
        <native>-l debug=true,place=vscatter:shared</native>
[...]

PR #2156, opened 2023-12-14
Set place=shared for shared jobs on WCOSS2 (develop)

KateFriedman-NOAA:
# Description

This PR updates the rocoto xml setup to set `place=shared` (add `:shared` to place setting) via the `<native>` tag when a job is shared (not exclusive). Previously only the `:exclhost` setting was added to PBS `place` when a job was to be run exclusively and all other jobs were assumed to run shared. PBS on WCOSS2 no longer assumes shared when exclusive is not set, it actually assumes the opposite now, that a job is exclusive if not specified. We now need to set either exclusive (`exclhost`) or shared (`shared`) for all jobs run on WCOSS2.

Have also updated all ecf scripts in `develop` to include `:shared` when not already set to `:exclhost`. The ecf scripts need updates for GFSv17 but for now they will be correct in term of exclusive vs shared until a developer sweeps through for the GFSv17 updates.

This change came from knowledge obtained during recent updated WCOSS2 training.

Addresses #2135

# Type of change

# Change characteristics

# How has this been tested?

1.5 cycle atmos-only test on WCOSS2-Dogwood.

# Checklist

### DIFF
--- a/ecf/scripts/enkfgdas/analysis/create/jenkfgdas_diag.ecf
+++ b/ecf/scripts/enkfgdas/analysis/create/jenkfgdas_diag.ecf
@@ -5,7 +5,7 @@
 #PBS -A %PROJ%-%PROJENVIR%
 #PBS -l walltime=00:06:00
 #PBS -l select=1:mpiprocs=48:ompthreads=1:ncpus=48:mem=24GB
-#PBS -l place=vscatter
+#PBS -l place=vscatter:shared
 #PBS -l debug=true
 
 model=gfs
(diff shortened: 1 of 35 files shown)


WalterKolczynski-NOAA (approved):
You probably didn't need to waste your time updating all the ecFlow definitions. They are all horribly out-of-date for develop anyway.

HenryRWinterbottom (approved):
This is only a suggestion moving forward, specifically with the reorganization/modernization, it might be worthwhile specifying and environment variable (e.g., `${VSCATTER}` or something similar) to reduce the redundancy -- but that is beyond the scope of the PR. Just a thought...

KateFriedman-NOAA:
> This is only a suggestion moving forward, specifically with the reorganization/modernization, it might be worthwhile specifying and environment variable (e.g., `${VSCATTER}` or something similar) to reduce the redundancy -- but that is beyond the scope of the PR. Just a thought...

Something to consider! I think every job on WCOSS2 uses that setting currently but that doesn't mean that'll be the way in the future.

PR #2162, opened 2023-12-15
[NO NOT MERGE] CI Self-Test on Hercules (FF Only)

TerrenceMcGuinness-NOAA:
CI Self Test on Hercules (Free Forecast only)


TerrenceMcGuinness-NOAA:
We need to have the rest of `detect_machine.sh` scripts in the other repos updated ie .`gfs_utils.fd/ush/detect_machine.sh` to include **Hercules**.

PR #2170, opened 2023-12-19
Fix handling of excluded hosts list

TerrenceMcGuinness-NOAA:
# Description

This PR fixes the issue with `create_experment.py` so that the yaml configuration file can have a string of list of strings when specifying the host name to exclude from the CI tests.

# Type of Change

# Change characteristics

# How has this been tested?

- Ran updated create_experiment.py with several use cases for specifying excluded hosts.


workflow/create_experiment.py
```diff
@@ -79,7 +79,7 @@ def input_args():
 
     if 'skip_ci_on_hosts' in testconf:
         host = Host()
-        if host.machine.lower() in testconf.skip_ci_on_hosts.lower():
+        if host.machine.lower() in ''.join([x.lower() for x in testconf.skip_ci_on_hosts]):
```
WalterKolczynski-NOAA:
```suggestion
        if host.machine.lower() in [ machine.lower() for machine in testconf.skip_ci_on_hosts ]:
```

WalterKolczynski-NOAA (changes requested):

workflow/create_experiment.py
TerrenceMcGuinness-NOAA: That join is so that the use can use a "-" or not in the YAML.  If you take that out it will fail. Oh wait you also changed the iterator, let me test it.

TerrenceMcGuinness-NOAA: Ok that works when we use a dash in the YAML.  If there is no dash it will not work but not fail.

WalterKolczynski-NOAA (approved):

workflow/create_experiment.py
TerrenceMcGuinness-NOAA: This is how list of things are supposed to be designated in YAML and I agree it is better to be consistent.

workflow/create_experiment.py
```diff
@@ -79,7 +79,7 @@ def input_args():
 
     if 'skip_ci_on_hosts' in testconf:
         host = Host()
-        if host.machine.lower() in testconf.skip_ci_on_hosts.lower():
+        if host.machine.lower() in [ machine.lower() for machine in testconf.skip_ci_on_hosts ]:
```
WalterKolczynski-NOAA:
```suggestion
        if host.machine.lower() in [machine.lower() for machine in testconf.skip_ci_on_hosts]:
```

WalterKolczynski-NOAA (changes requested):

workflow/create_experiment.py
WalterKolczynski-NOAA: Sorry, this was my fault in the previous suggestion.

TerrenceMcGuinness-NOAA: fixed

WalterKolczynski-NOAA (approved):

PR #2172, opened 2023-12-19
Stop calling ufs-utils link script

WalterKolczynski-NOAA:
# Description
The `link_workflow.sh` script was calling the ufs_utils version of the same. That step is only necessary if a user is going to run one of the stand-alone ufs-utils, which is rare.

This change will also sidestep the issue of the script not yet having been updated for Hercules.

# Type of change

# Change characteristics

# How has this been tested?

# Checklist

### DIFF
--- a/sorc/link_workflow.sh
+++ b/sorc/link_workflow.sh
@@ -122,12 +122,6 @@ do
 done
 
 
-if [[ -d "${HOMEgfs}/sorc/ufs_utils.fd" ]]; then
-  cd "${HOMEgfs}/sorc/ufs_utils.fd/fix" || exit 1
-  ./link_fixdirs.sh "${RUN_ENVIR}" "${machine}" 2> /dev/null
-fi
-
-
 #---------------------------------------
 #--add files from external repositories
 #---------------------------------------


KateFriedman-NOAA (approved):

DavidHuber-NOAA (approved):
I'm good with this given that the docs have instructions for generating ICs based on a clean checkout of UFS_Utils.

emcbot (commented):
make it so

emcbot (approved):

TerrenceMcGuinness-NOAA (approved):
oops did that as emcbot on accident

WalterKolczynski-NOAA:
> I'm good with this given that the docs have instructions for generating ICs based on a clean checkout of UFS_Utils.

It also could still be run even using the gw copy, just separately.

DavidHuber-NOAA:
@WalterKolczynski-NOAA True, my only concern was that perhaps we needed to update the docs to explicitly state that link_fixdirs.sh now needs to be run before running gdas_init, but since it is already stated in the docs, I think our bases are covered.

WalterKolczynski-NOAA:
I'm holding this until the CI fix gets in (#2170) so I can merge that in and run CI here, just in case.

aerorahul:
The GSI and cycled part is not ported to hercules yet.  We should disable building with the GSI (remove the `-g` and `-u` options) from the build on Hercules.
This is expected.

TerrenceMcGuinness-NOAA:
@aerorahul oh yes! that makes sense, maybe wee need to make a conitional on this some how

CoryMartin-NOAA:
FYI this is not updated in the documentation, and I disagree that users "rarely" will generate ICs...

aerorahul:
> FYI this is not updated in the documentation, and I disagree that users "rarely" will generate ICs...

@CoryMartin-NOAA 
Users/developers are free to generate ICs. As it stands currently, the workflow assumes those ICs have been generated offline before kicking off the workflow.
We should restore the capability of generating ICs as part of the workflow (where appropriate), but that will take some time as generating the ICs e.g. for the coupled model is not yet a streamlined process and requires several intermediate steps.

JessicaMeixner-NOAA:
I thought the process of streamlining things for the coupled model/fractional grid was complete? @GeorgeGayno-NOAA can you confirm?

JessicaMeixner-NOAA:
It does state https://global-workflow.readthedocs.io/en/latest/init.html#manual-generation that users can use the build from the g-w, so perhaps at minimum that should be updated?  I'm actually in the process of trying to make some ICs today.

WalterKolczynski-NOAA:
> It does state https://global-workflow.readthedocs.io/en/latest/init.html#manual-generation that users can use the build from the g-w, so perhaps at minimum that should be updated? I'm actually in the process of trying to make some ICs today.

Yes, it should. When I made this PR, I was told we were recommending a stand-alone ufs_utils clone for creating ICs, but I foolishly didn't confirm that myself.

Issue #827, opened 2022-06-01
Add ensemble wave post


WalterKolczynski-NOAA:
**Description**
As we move towards the merger of GEFS into the GFS system, global workflow needs to be updated to produce ensemble wave products.

**Requirements**
Need to be able to take ensemble WW3 model output and produce the wave products that are currently produced by GEFS:
- [ ] Member lat-lon grib2 files
- [ ] Mean/spread/prob grib2 files
- [ ] Station and bulletin files (tarred together)
- [ ] Member gempak files

This will likely be a combination of modifying existing scripts to accommodate ensembles and porting scripts over from GEFS.

**Acceptance Criteria (Definition of Done)**
A series of jobs that are able to create all the same ensemble wave products that are currently produced by GEFS.

**Dependencies**
Add atm-wave ensemble free-forecast #826

WalterKolczynski-NOAA:
@XianwuXue-NOAA is this essentially complete since GEFS is staying as a separate workflow (i.e. can GEFS produce wave products using current GFS develop)?

XianwuXue-NOAA:
@WalterKolczynski-NOAA I am still working on atmos forecast, when I move to atm+wave, I will do a test.

Issue #823, opened 2022-06-01
Add atm ensemble product generation


WalterKolczynski-NOAA:
**Description**
As we move towards the merger of GEFS into the GFS system, global workflow needs to be updated to produce the same products the current GEFS system is creating.

**Requirements**
Some post scripts will need to be updated to accommodate ensemble products, while others may need to be added based on GEFS scripts. The end result will need to be able to produce both GFS and GEFS products, depending on the member number.

Products:
- [ ] Member lat-lon grib files
- [ ] Mean and spread grib files
- [ ] Gempak files

Production of ensstat files can be skipped for now.

**Acceptance Criteria (Definition of Done)**
Given master grib files from an ensemble forecast, be able to run a series of post-processing jobs to produce the above products as GEFS does.

**Dependencies**
Reorganize COM (https://github.com/NOAA-EMC/global-workflow/issues/761)
Update workflow setup scripts (https://github.com/NOAA-EMC/global-workflow/issues/789) - Closed
Add ensemble free-forecast (#822) - Done 
Move post scripts from UPP under global-workflow (#769) - Done
Create flexible product script to replace old post script (#270) [may be concurrent]

PR #2189, opened 2023-12-22
Refactor rocoto task XML creation

WalterKolczynski-NOAA:
# Description
Refactors the rocoto task generation to be recursive. This will allow nested metatasks to loop over multiple variables, which is needed for GEFS product generation.

As part of this refactor, there is no longer separate arguments to designate metatasks. Instead, task dicts can include a nested 'task_dict' as well as a 'var_dict' containing the variables to loop over. The nested task dict can then either have another layer, or be the innermost task.

To accommodate the new recursive nature, some defaults that were previously defined in `create_wf_task()` had to be pushed down into the function that creates the innermost task. Also, former keywords have been absorbed by the task dict.

Refs #823
Refs #827

# Type of change

# Change characteristics

# How has this been tested?
- Cycled test on Orion
- Forecast-only test on Orion
- GEFS test on Orion

# Checklist


WalterKolczynski-NOAA:
Not quite working yet, still need to make some tweaks.

WalterKolczynski-NOAA:
Okay, think I have the bug worked out.

HenryRWinterbottom (commented):
This LGTM. The docstrings are helpful and making the dictionaries more readable is easy on the eyes. This is beyond the scope of this PR, but with some time investment I think some of the redundant code blocks can be parsed out and stuffed in wrapper functions (e.g., decorators) to make this code even more readable. But, just my $0.02.

WalterKolczynski-NOAA:
Orion tests are randomly failing due to stalling on module load.

workflow/rocoto/rocoto.py
```diff
@@ -76,7 +123,7 @@ def create_task(task_dict: Dict[str, Any]) -> List[str]:
     nodes = resources_dict.get('nodes', 1)
     ppn = resources_dict.get('ppn', 1)
     threads = resources_dict.get('threads', 1)
-    log = task_dict.get('log', 'demo.log')
+    log = task_dict.get('log', f'&ROTDIR;/logs/@Y@m@d@H/{taskname}.log')
```
aerorahul: I think this needs to be reverted.

aerorahul (commented):
mostly looks good.  I think one tiny thing slipped through.

workflow/rocoto/rocoto.py
WalterKolczynski-NOAA: Oops. Fixed.

aerorahul (approved):

PR #2220, opened 2024-01-12
[DO NOT MERGE] Weekly CI Tests Friday Jan 12, 2024

emcbot:
[DO NOT MERGE] Weekly CI Tests Friday Jan 12, 2024


DavidHuber-NOAA:
The errors on both Orion and Hera occur during staging of `S2SWA` ICs with an unsupported ocean resolution of 0.5 degrees:

https://github.com/NOAA-EMC/global-workflow/blob/c59047614c29f6ec15a4d4a6fa1855929287debd/scripts/exglobal_stage_ic.sh#L86-L105

I am guessing that the ocean resolution needs to be specified via the new `setup_expt.py --resdetocean` flag.  I'm not sure if 0.25, or 1.00 degrees would be most appropriate for this test.

DavidHuber-NOAA:
Do we know what resolution this test was run at before `--resdetocean` was added?

DavidHuber-NOAA:
Per conversation with @aerorahul, closing while the ocean resolution is fixed for the `S2SWA` test.

PR #2221, opened 2024-01-14
Apply fixes to archiving jobs

### LINKED ISSUE #2076: Archive jobs failing to find sbuv2_n19 ozone data
### What is wrong?

The `gdasarch` job looks for a few ozone monitor outputs when `DO_VERFOZN="YES"` based on a hard-coded list in hpssarch_gen.sh, including `sbuv2_n19`.  However, SBUV/2 data on NOAA 19 stopped producing data in 2013, so there is no data available to verify/monitor leading to `gdasarch` to fail because the `htar` command cannot find it.

The archiving job should be able to look for only the instruments available.  There will be times when other instruments fail or are offline briefly and the job should continue to archive what is available without error.

All or N/A

### Steps to reproduce

Run a full analysis cycle and look for htar error reported in gdasarch.log.

### Additional information

Found during spack-stack testing.

The htar command to archive gdas.tar (where the monitor files are pushed to) should either be split up into a baseline set of data (the data that should always be produced by the workflow) and an auxiliary set (including the monitoring data).  Archive the baseline set first, then append to the archive with the auxiliary set.  Errors should be produced for members not present in the baseline set while warnings should be produced for data not present in the auxiliary.

DavidHuber-NOAA:
# Description
Add checks for potentially missing log/data files and update log names for archiving.

Missing sflux index files are also added to the products script.

Resolves #2076 

# Type of change

# Change characteristics

# How has this been tested?
Cycled test on Orion

# Checklist


WalterKolczynski-NOAA (approved):

DavidHuber-NOAA:
I went ahead and added the 4 lines of code to generate sflux index files since it was easier/quicker than writing another issue/PR.  Tested on Orion.

ush/hpssarch_gen.sh
```diff
@@ -368,13 +368,16 @@ if [[ ${type} == "gdas" ]]; then
...
+    if [[ "${WRITE_DOPOST}" == ".false." ]]; then
+       echo "./logs/${PDY}${cyc}/gdas*upp*.log"
+    fi
 
     fh=0
     while [[ ${fh} -le 9 ]]; do
       fhr=$(printf %03i "${fh}")
       echo "${COM_ATMOS_MASTER/${ROTDIR}\//}/${head}sfluxgrbf${fhr}.grib2"
-      echo "${COM_ATMOS_MASTER/${ROTDIR}\//}/${head}sfluxgrbf${fhr}.grib2.idx"
+      #echo "${COM_ATMOS_MASTER/${ROTDIR}\//}/${head}sfluxgrbf${fhr}.grib2.idx"
```
WalterKolczynski-NOAA: This can be uncommented now that the indexes are being created.

WalterKolczynski-NOAA (changes requested):

ush/hpssarch_gen.sh
DavidHuber-NOAA: Yep, good call.  Fixed and tested on Orion.

WalterKolczynski-NOAA (approved):

PR #2199, opened 2024-01-05
Add a yaml for snow DA testing.

aerorahul:
# Description
This PR:
- adds a new test yaml C96_atmsnowDA.yaml for 3DVar atmosphere with GSI and Land (Snow) DA with JEDI
- moves a few yamls from `platforms/` to `yamls/`
- adds ability to overwrite a previously created experiment as an addition to user input.

Part of #1821 

# Type of change
- New feature (adds functionality)

# Change characteristics
- Is this a breaking change (a change in existing functionality)? NO
- Does this change require a documentation update? NO

# How has this been tested?

# Checklist
- [ ] Any dependent changes have been merged and published
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] I have made corresponding changes to the documentation if necessary

PR #2232, opened 2024-01-17
Move the IMS snow process from the preplandobs job into the landanl job

jiaruidong2017:
# Description
This PR is to re-organize the snow DA related jobs by moving the IMS snow process from the preplandobs job into the landanl job, that way the preprocessed file specific to each member only lives in $DATA and is not retained (only the input file is retained in $COM_OBS). This PR contributes to the test case for adding JEDI-based snow DA in the global-workflow CI test (PR #2199).

# Change characteristics
Is this a breaking change (a change in existing functionality)? NO
Does this change require a documentation update? NO

# How has this been tested?

# Checklist

- [ ]  Any dependent changes have been merged and published
- [ ]  My code follows the style guidelines of this project
- [ ]  I have performed a self-review of my own code
- [ ]  I have commented my code, particularly in hard-to-understand areas
- [ ]  My changes generate no new warnings
- [ ]  New and existing tests pass with my changes
- [ ]  I have made corresponding changes to the documentation if necessary


ush/python/pygfs/task/land_analysis.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, config):
                 'npz': self.config.LEVS - 1,
                 'LAND_WINDOW_BEGIN': _window_begin,
                 'LAND_WINDOW_LENGTH': f"PT{self.config['assim_freq']}H",
-                'OPREFIX': f"{self.runtime_config.RUN}.t{self.runtime_config.cyc:02d}z.",
+                'OPREFIX': f"{self.runtime_config.CDUMP}.t{self.runtime_config.cyc:02d}z.",
```
aerorahul: This should be `RUN` and not `CDUMP`.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -187,6 +187,7 @@ def prepare_IMS(self) -> None:
         os.symlink(exe_src, exe_dest)
 
         # execute CALCFIMSEXE to calculate IMS snowdepth
+        os.chdir(localconf.DATA)
```
aerorahul: Shouldn't you already be in `DATA`?

ush/python/pygfs/task/land_analysis.py
```diff
@@ -369,6 +379,39 @@ def finalize(self) -> None:
...
+        for ob in observers:
+            obfile = ob['obs space']['obsdatain']['engine']['obsfile']
+            basename = os.path.basename(obfile)
+            if 'ims_snow' not in obfile:
+                copylist.append([os.path.join(self.task_config['COM_OBS'], basename), obfile])
+        obs_dict = {
+            'mkdir': [os.path.join(self.runtime_config['DATA'], 'obs')],
+            'copy': copylist
+        }
+        return obs_dict
```
aerorahul: This needs to be configured so that it conforms with the generic method in `class Analysis`.

aerorahul (changes requested):

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: To use `CDUMP` here is for either `gdas` or `enkfgdas`. However, `RUN` is the only for `gdas`. We will use the same `land_analysis.py` code for the later hybrid runs.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: This is for the enkfgdas run, and the DATA are different for the different ensemble members.

aerorahul: We will have `RUN=enkfgdas` for the ensemble component.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: @CoryMartin-NOAA suggested to create this new `get_obs_dict()` function in the `class LandAnalysis`. The only difference for this new `get_obs_dict()` compared with the generic method in `class Analysis` is to skip putting the IMS observations into the copylist.

CoryMartin-NOAA: @aerorahul can you clarify what you mean by your comment here?

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: @aerorahul Your this comment is for the above conversation, correct? If `RUN=enkfgdas` is available, I can change `CDUMP` back to `RUN`.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -307,6 +313,10 @@ def execute(self) -> None:
         for key in keys:
             localconf[key] = self.task_config[key]
 
+        logger.info("Generating IMS snow OBS")
+        if f"{ self.runtime_config.cyc }" == '18':
+            self.prepare_IMS()
```
aerorahul, line 313: This is preparing the observation.  IIRC, this was part of the preplandobs job

jiaruidong2017: @aerorahul From our email conversations on Nov 30, 2023 as below, we have agreement to bring this IMS snow processing into `landanl` job. 

```
Jiarui/Mike,

Rahul and I just had a quick chat, and he is fine with combining the obs preprocessing with the OI job, 
especially since we will have 1 high res OI job and a set of meta tasks for multiple ensemble OI jobs.

Jiarui, please proceed and let me know if you have any questions or how I can help.

-Cory
```

WalterKolczynski-NOAA:
@jiaruidong2017 Would you please update the title and description to reflect the work being done without users needing to jump to other PRs or issues

jiaruidong2017:
> @jiaruidong2017 Would you please update the title and description to reflect the work being done without users needing to jump to other PRs or issues

Updated. Thanks.

ush/python/pygfs/task/land_analysis.py
aerorahul: yes. That is true and is fine.
However, this is not complete and I don't think is necessary for the snowDA test to work.
Besides, if this is going to be added here, then the counterpart in [exglobal_prep_land_obs.py](https://github.com/NOAA-EMC/global-workflow/blob/7759163c668eb6ccc11dc1ecd39c0dc5d433cdc1/scripts/exglobal_prep_land_obs.py#L24) should be removed.

Moving of `prepare_IMS` from the prep job to the inline analysis job should be its own issue and PR after the test is added.

ush/python/pygfs/task/land_analysis.py
```diff
@@ -244,7 +245,12 @@ def initialize(self) -> None:
             Instance of the LandAnalysis object
         """
 
-        super().initialize()
```
aerorahul: Undoing this, eliminates the hierarchical class and works against what the design here is supposed to accomplish.

jiaruidong2017: Yes, I understand. However, if `super().initialize()` is left here, we need to add a special logic for the IMS observation type in the `class Analysis`. Do you want me to do this way or do you have any other suggestions? Thanks.

aerorahul: `super.initialize()` is generic and agnostic to land (or any other component).
It should be the first thing to do followed by anything that the land-specific implementation needs.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: The IMS process in the `scripts/exglobal_prep_land_obs.py` was removed. 

The support to move IMS processing from the prep job to the landanl job was already merged into the GDASApp develop. If you don't want to include this change in this PR, we need to revert the changes made in the GDASApp. Please let me know what you think. Thanks.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: Removed the changes.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: Removed `os.chdir(localconf.DATA)` for now.

ush/python/pygfs/task/land_analysis.py
jiaruidong2017: I made small changes in `class Analysis` to skip putting `ims_snow` files into the copylist. This approach can solve the our issues, and keep to use `super().initialize()`. What do you think about this way?

CoryMartin-NOAA:
This PR is going to be closed. We will keep a separate `preplandobs` job but will write out to `$DATA`. `$RUNDIRS/gdaslandanl` will only have one ob file for IMS, but `$RUNDIRS/enkfgdaslananl` will have `memNNN` directories analogous to the ensemble fcst job which has a root `$DATA` and then subdirectories.

PR #1151, opened 2022-12-07
Support for the GDASApp testing in containers

guillaumevernieres:
**Description**

Features proposed to be added:
-  Support for running part of the global-workflow in a container (only works with the GDASApp for now).
~- Moves a few default values from `config.ocnanal` to the new `defaults.yaml` recently added by @aerorahul~

`CONTAINER.env` is quite minimalist at this point, it could also certainly be cleaned up, but since this is just a draft PR ...

Issues fixed:
- Fixes #1234

**Type of change**

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**
In containers, running the 3 ctests related to the g-w
  
**Checklist**

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes need updates to the documentation. I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] Any dependent changes have been merged and published

PR #2258, opened 2024-01-24
Add settings for containers to config.resources

### LINKED ISSUE #2257: CONTAINER case is missing
### What is wrong?

running in a container does not work anympore

...

All or N/A

### Steps to reproduce

build in a container

### Additional information

n/a

yes

guillaumevernieres:
I'm not sure when this happened, but the container case in `config.resources` is missing.
- fixes #2257

### DIFF
--- a/parm/config/gfs/config.resources
+++ b/parm/config/gfs/config.resources
@@ -60,6 +60,9 @@ case ${machine} in
     export PARTITION_BATCH="compute"
     npe_node_max=40
     ;;
+  "CONTAINER")
+    npe_node_max=1
+    ;;
   *)
     echo "FATAL ERROR: Unknown machine encountered by ${BASH_SOURCE[0]}"
     exit 2


WalterKolczynski-NOAA:
Near as I can tell, `config.resources` never had an entry for containers. Your original PR adding containers (#1151) didn't add one, and we don't really support them. Git log didn't find any change in the number of `CONTAINER`s since your PR.

WalterKolczynski-NOAA (approved):
Should be fine as it won't impact any other machines, though I'm sure some stuff will break if someone tried actually running with a max of one.

guillaumevernieres:
> Near as I can tell, `config.resources` never had an entry for containers. Your original PR adding containers (#1151) didn't add one, and we don't really support them. Git log didn't find any change in the number of `CONTAINER`s since your PR.

Yes, no idea what happened, it worked until it didn't.

aerorahul (approved):

guillaumevernieres:
@aerorahul , @WalterKolczynski-NOAA :
While that fixed my issue, it looks like the problem is not just when we build in container. The build of gdas.cd is currently broken.  
Could a `CI` test that simply build the gdasapp be added (suggested by @CoryMartin-NOAA )?

WalterKolczynski-NOAA:
We already build with gdas.cd as part of our CI tests.

PR #2309, opened 2024-02-13
Add JEDI 3DEnVar atmosphere only CI test

### LINKED ISSUE #2294: JEDI-DA atmosphere only CI case

Add JEDI-DA atmosphere only case to `ci`.  This is a requested test case listed in issue #1821.

Cycle g-w in atmosphere only mode using UFS-DA applications for the variational and local ensemble analyses.

### Acceptance Criteria

JEDI-DA atmosphere only CI test passes

_No response_

RussTreadon-NOAA:
# Description
Add a JEDI-based 3DEnVar atmosphere only CI test to ensure g-w updates do not break the JEDIATMVAR and JEDIATMENS cycling capability.    

The JEDI 3DEnVar case cycles a C48L127 deterministic state with a two member C48L127 ensemble over the period 2021032312 through 2021032400.   The test starts from 20210312312 start with a `gdas_half` cycle.  The `gdas_half` cycle begins with `gdasstage_ic` to populate `ROTDIR` with initial conditions required by the test.
 
Part of #2294 

# Type of change
- Maintenance (new CI test)

# Change characteristics
- Is this a breaking change (a change in existing functionality)? **YES**
- Does this change require a documentation update? **NO**

# How has this been tested?
- Clone, build, and cycle test on Hera

# Checklist


aerorahul (commented):
@RussTreadon-NOAA 
Thank you for adding a test for the JEDI Atm-DA cases.

The updates needed for JEDI based Atm-DA configuration looks ok (`DONST` etc).

I am however concerned about the staging of the initial conditions and necessary bias,  radstat files.
I appreciate expanding `stage_ic` to include the cycled mode, however, that splits the logic of creating and staging of ics in 2 places; `setup_expt.py` and `exglobal_stage_ic.sh`.
This will make it difficult to manage, maintain, and adds confusion as to why there are 2 places for staging initial experiment files.
As you also notice, in the forecast-only, we are copying ICs from their staged location into COM, while the logic for cycling is linking.  We want to ensure that we copy when we stage this data to make sure the COM is complete and consistent. 

While this "works", this will need to be cast in a way that either `setup_expt.py` does the staging for all cases (include coupled model for WCDA) or `exglobal_stage_ic.sh` is updated to stage for all cycled configurations.  

I suggest we split this PR into the non-staging updates and we can work to extend `exglobal_stage_ic.sh` or its variant to perform the function of copying ICs for cycled DA that includes WCDA as well and deprecate the `fill_ROTDIR_cycled` method from `setup_expt.py` as was done for the `forecast-only` mode.

RussTreadon-NOAA:
Agreed.  I don't like how I tweaked `exglobal_stage_ic.sh` to work.  It's functional but not optimal.  What you propose @aerorahul works for me.   FYI, the ICs for the C48C48L127 JEDI 3DEnVar atmosphere only CI test are on Hera in `/scratch1/NCEPDEV/stmp2/Russ.Treadon/ICS/C48C48`.

Is the following list of files to revert back to `develop` sufficient
- `parm/config/gfs/config.stage_ic`
- `scripts/exglobal_stage_ic.sh`
- `workflow/applications/gfs_cycled.py`
- `workflow/rocoto/gfs_tasks.py`

RussTreadon-NOAA:
What needs to be done on my side to move this PR forward?

aerorahul:
@RussTreadon-NOAA 
The changes look good to me.
I do have a clarifying question re. the test.
As it is currently coded, the test will be kicked off on Hera.  Is my understanding correct that that test will fail because of the changes we held back for staging the initial conditions etc.?
If so, we should add `hera` to the list of `skip_ci_on_hosts:` section of the test yaml.  
That will skip the test on `hera` and when the initial conditions and other files staging is ready, we can remove that.
Does that work for you?

RussTreadon-NOAA:
@aerorahul :  Correct, the test will fail on Hera for the reason you state.  Recommended change committed at [`a78ac19`](https://github.com/RussTreadon-NOAA/global-workflow/commit/a78ac19a5c1ed24e907d5c12353ee0e4d5c30c16).

aerorahul (approved):
This sets up the global-workflow to run the JEDI-atm tests as soon as the pieces for DA diagnostics and initial conditions are in place.

RussTreadon-NOAA:
Thank you @aerorahul

PR #2334, opened 2024-02-22
Jenkins updates

TerrenceMcGuinness-NOAA:
# Description

Jenkins Updates Resolving final kinks:
- Removed all `git` shell commands and now exclusively use Software Control Manger (**scm**) plugin.
- Add feature for skipping hosts per configuration specified in case yaml files.
- Solved and tested false positive builds and experiments.
- Tested archiving of task error log on case fail
- First case fail quits pipeline and cancels all pending scheduled jobs
- Duel build per yaml configuration arguments supported
- All designated case files in PR directory pass on intended host (fully tested on Hera)

Remaining updates:
- Fist build fail short circuit when building sub-modules and archiving build error log.
- Re-build/no-build built in logic for Replay and Rerunning previously failed experiments.

# Type of change
- Bug fix (fixes something broken): shorter characters in path to workspace solved datapath size constraint issue.
- New feature (adds functionality): skip hosts now supported

# Change characteristics

# How has this been tested?
Full test run-through conducted on Hera with [PR 251 in my repo](https://github.com/TerrenceMcGuinness-NOAA/global-workflow/pull/251) on all cases designated to do so in `$HOMEgfs/ci/cases/pr` which can be [viewed](https://jenkins.epic.oarcloud.noaa.gov/blue/organizations/jenkins/global-workflow%2FEMC-Pipelines/detail/PR-251/1/pipeline) on the Jenkins Controler portal.

Here is the Blue Ocean view of a duel system build with skipped host cases ran on Hera: 

image


ci/scripts/utils/get_host_case_list.py
```diff
@@ -0,0 +1,28 @@
...
+from wxflow import AttrDict
+from workflow import hosts
+
+_here = os.path.dirname(__file__)
+_top = os.path.abspath(os.path.join(os.path.abspath(_here), '../../..'))
+
+if __name__ == '__main__':
+
+    case_list = []
+    host = hosts.Host()
```
aerorahul: Instead of detecting the host here, why not take an input argument for the host name.  The Jenkinsfile when calling this script already knows what the host is.
This would eliminate the need to link the `workflow` directory into this directory.
Just a suggestion.

TerrenceMcGuinness-NOAA: sure why not, I can make that change.

TerrenceMcGuinness-NOAA: Sure, that can work.  I'll make the change and I can get ride of the soft link and extra import.

Jenkinsfile
```diff
@@ -99,6 +98,9 @@ pipeline {
                                        }
                                        pullRequest.addLabel("CI-${Machine}-Running")
                                     }
+                                    if (system == 'gfs') {
```
WalterKolczynski-NOAA, line 101: Why only GFS? Or is this just a way to only do it once?

WalterKolczynski-NOAA (approved):
LGTM

PR #2367, opened 2024-03-01
gefs build is not supported on bash orion

TerrenceMcGuinness-NOAA:
Orion is running CI Bash which does not support gefs builds so skiphost Orion is being added to C48_S2SWA_gefs


aerorahul:
should we just jump to Jenkins on Orion as well?
Thoughts @TerrenceMcGuinness-NOAA

PR #2376, opened 2024-03-05
Rewrote pr_list_database.py to use wxflow's SQLiteDB Class

TerrenceMcGuinness-NOAA:
# Description

This PR updates the `pr_list_database.py` code to use **wxflow** SQLiteDB Class

- Improved code's readability
- Uses better code style matching project's software culture 
- Better docstring standards

# Type of change
- Maintenance (code refactor, clean-up)

# Change characteristics

# How has this been tested?
Tested with running PR [258](https://github.com/TerrenceMcGuinness-NOAA/global-workflow/pull/258) in Bash dev space:
- Interrupted build and let CI Bash restart build with messaging to test database updates using new code.
- Used new code to update database state and restart after gefs case failed
- 
# Checklist


ci/scripts/pr_list_database.py
```diff
@@ -2,154 +2,121 @@
...
     ----------
-    filename : Path
-        Full path to a sqlite3 database file
-
-    Returns
-    -------
-    sqlite3.Connection
-        Sqlite3 Connection object for updating table
-
+    ci_database : SQLiteDB
```
aerorahul:
```suggestion
    db : SQLiteDB
```

ci/scripts/pr_list_database.py
```diff
@@ -2,154 +2,121 @@
...
+        The database to add the pull request to.
+    pr : str
+        The pull request to add.
     """
+    rows = db.fetch_data('pr_list')
+    for row in rows:
+        if str(row[0]) == str(args.add_pr[0]):
+            print(f"pr {row[0]} already is in list: nothing added")
+            sys.exit(0)
```
aerorahul: I think there is a limitation in this logic.
What happens if a `row` in the list of `rows` is not in the db after a `row` that is in the list?
The logic here will return without checking beyond the current `row`.

ci/scripts/pr_list_database.py
```diff
@@ -174,42 +141,19 @@ def input_args():
...
-                if int(args.display[0]) == int(row[0]):
-                    print(' '.join(map(str, row)))
-        else:
-            for row in rows:
-                print(' '.join(map(str, row)))
-
-    con.commit()
-    con.close()
+        for rows in display(ci_database, args):
+            print(rows)
```
aerorahul, line 161: if you are going to use `wxflow`, also use `logger` instead of `print`.

aerorahul (commented):
looks good.
some suggestions and possibly spotted an issue.

ci/scripts/pr_list_database.py
TerrenceMcGuinness-NOAA: Well this output a specific artifact of the the arguments and is used and scrubbed by the bash script so I would think not.

ci/scripts/pr_list_database.py
TerrenceMcGuinness-NOAA: Yes but this is data base is not really a database but a state machine.  The way it works is that you can only have one and only one PR running at a time and the PR number is unique.  It is this fact that makes the CI work for when some one updates a label of a PR that is already running, it then checks if it should be quit or be restarted.  On the other hand, given that you posed this question,  I can reconsider this approach and and use the STATE column in the logic of the bash script and impose uniqueness into the database schema.  I will have to look into how that is done and update the SQLiteDB Class to support it.

aerorahul: makes sense.  thanks for explaining.

ci/scripts/pr_list_database.py
aerorahul: Ok.

aerorahul (approved):
looks good.

aerorahul:
this PR will need to update the hash of wxflow that includes the `sqlitedb` interface.

ci/scripts/pr_list_database.py
```diff
@@ -2,154 +2,122 @@
...
+    pr : str
+        The pull request to add.
     """
-
-    obj.execute("CREATE TABLE processing(pr integer PRIMARY KEY, state text, status text, reset_id integer, cases text)")
+    rows = db.fetch_data('pr_list')
+    entities = (args.add_pr[0], 'Open', 'Ready', 0, 'ci_repo')
+    try:
+        db.insert_data('pr_list', entities)
+    except (SQLiteDB.Error.IntegrityError) as e:
```
aerorahul:
```suggestion
    except (SQLiteDBError.IntegrityError) as e:
```

ci/scripts/pr_list_database.py
```diff
@@ -2,154 +2,122 @@
 
 import sys
 import os
-from pathlib import Path
+from wxflow import SQLiteDB
```
aerorahul:
```suggestion
from wxflow import SQLiteDB, SQLiteDBError
```

aerorahul (approved):
a couple of comments.

ci/scripts/pr_list_database.py
TerrenceMcGuinness-NOAA: Ok, done, but now we have to iterate on wxflow PR before I can update the gitmodule for this PR

TerrenceMcGuinness-NOAA: I had to add **SQLiteDBError** in `__init__.py` and remove **Error** from **SQLiteDB** in **wxflow**

aerorahul (approved):
Looks good.

ci/scripts/pr_list_database.py
```diff
@@ -2,154 +2,108 @@
...
 import sys
 import os
-from pathlib import Path
-from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter, REMAINDER, ZERO_OR_MORE
-import sqlite3
+from wxflow import SQLiteDB, SQLiteDBError
+from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter, REMAINDER
 
 
 def full_path(string):
```
WalterKolczynski-NOAA, line 9: Why was the docstring deleted?

ci/scripts/pr_list_database.py
```diff
@@ -2,154 +2,108 @@
...
-    """
-
     if os.path.isfile(string) or os.path.isdir(os.path.dirname(string)):
         return os.path.abspath(string)
     else:
         raise NotADirectoryError(string)
 
 
-def sql_connection(filename: os.path) -> sqlite3.Connection:
+def create(db: SQLiteDB):
```
WalterKolczynski-NOAA:
```suggestion
def create_table(db: SQLiteDB):
```

ci/scripts/pr_list_database.py
```diff
@@ -2,154 +2,108 @@
...
     """
-    Returns an Sqlite3 Cursor object from a given path to a sqlite3 database file
+    Create a new database.
```
WalterKolczynski-NOAA:
```suggestion
    Create a new table in a database.
```

ci/scripts/pr_list_database.py
```diff
@@ -2,154 +2,108 @@
...
 
-def sql_fetch(obj: sqlite3.Cursor) -> list:
-    """ Gets list of all rows in table
-
-    Parameters
-    ----------
-    obj : sqlite.sql_connection
-        sqlite3 Cursor Object
-
+def display(db, display):
```
WalterKolczynski-NOAA:
```suggestion
def display_db(db, display):
```

ci/scripts/pr_list_database.py
```diff
@@ -2,154 +2,108 @@
...
     Parameters
     ----------
-    obj : sqlite.sql_connection
-        sqlite3 Connection Object
-    pr : str
-        pr number acting as key for removing the row with in it
+    ci_database : SQLiteDB
+        The database to display.
+    args : argparse.Namespace
+        The command line arguments.
```
WalterKolczynski-NOAA: These do not match the function arguments.

WalterKolczynski-NOAA (changes requested):

ci/scripts/pr_list_database.py
```diff
@@ -2,154 +2,126 @@
...
-    pr : str
-        The given pr number to update in the table
-    updates : dict
-        Dictionary of values to update for a given PR to include by postion
-        state, The new value for the state (Open, Closed)
-        status, The new value for the status (Ready, Running, Failed)
-        reset_id, The value for number of times reset_id to Ready
-        cases, Information regarding which cases are used (i.e. self PR)
-
+    ci_database : SQLiteDB
```
WalterKolczynski-NOAA:
```suggestion
    db : SQLiteDB
```

TerrenceMcGuinness-NOAA (commented):
made update with docstring name

WalterKolczynski-NOAA (approved):

PR #2400, opened 2024-03-14
Develop CI Bash selftest on Orion [do not merge]

TerrenceMcGuinness-NOAA:
develop branch CI Bash self test on Orion


aerorahul:
The failed test is because the bash CI does not support the 2 builds (required for unstructured waves).  That caused the gefs failure.

TerrenceMcGuinness-NOAA:
Removed gefs case from test PR and restarted Bash CI without and skipped build "by hand".
When BASH CI starts messaging here that they are passing we know BASH CI on Orion is good.

PR #2425, opened 2024-03-21
hotfix to update full path to error logs on CI case fail

TerrenceMcGuinness-NOAA:
# Description

This hotfix PR prepends the full path to the error logs on disk to be communitated correctly to the GitHub message to the PR that is being processed when there is a failure in from a case.

# Type of change

# Change characteristics

# How has this been tested?
Only on the command line.


aerorahul (approved):
looks good.

PR #2429, opened 2024-03-22
Uploading error logs to GitHub from Jenkins CI Runs

TerrenceMcGuinness-NOAA:
# Description

This PR enhances the user experience within GitHub for when errors occur during build and running of CI cases from within the PR messages.  This is done by uploading the error logs to GitHub Gists and then publishing the links to them along with the full paths of the logs on disk.

This PR adds the python Class **GitHubPR** in `${HOMEgfs}/ci/scripts/utils/githubpr.py`  by inhearting the GitHub Class from **pyGitHub**.  We use this module to introduce a helper python utility that can publish a list of log files into a GitHub Gist and/or the designated branch **error_logs** in the **emcbot**  repo **ci-global-workflows** for storing error log files for review from any git configured terminal.  This upload feature will also create persistence of errors over time.

Also the `build_all.sh -k` script has been updated to support a "quick kill" feature (thanks David) that stops the parallel builds whenever one fails and creates an error_log file that has the paths to the error files that are also uploaded and published in the PR messages in GitHub.

# Type of change
- Maintenance (code refactor, adding a GitHub Class to use GItHub CLI using python)

# Change characteristics

# How has this been tested?
This PR has been run within Jenkins and has reported out the results [correctly to the messages](https://github.com/NOAA-EMC/global-workflow/pull/2422#issuecomment-2015436622).


ci/scripts/utils/githubpr.py
```diff
@@ -0,0 +1,125 @@
...
+
+        Parameters
+        ----------
+        repo_url : Repository URL
+            The GitHub repository.
+        """
+        if repo_url is None:
+            repo_url = os.environ.get("REPO_URL")
+        match = re.search(r"github\.com/(.+)", repo_url)
+        repo_identifier = match.group(1)[:-4]
```
WalterKolczynski-NOAA, line 80: Why just the first four characters?

ci/scripts/utils/githubpr.py
```diff
@@ -0,0 +1,125 @@
...
+
+        Returns
+        -------
+        list
+            A list of all pull request numbers.
+        """
+        pr_list = []
+        for pull in self.pulls:
+            pr_list.append(pull.number)
+        return pr_list
```
WalterKolczynski-NOAA: This can be one line using list notation (does it even need to be a method?)
```suggestion
        return [ pull.number for pull in self.pulls ]
```

ci/scripts/utils/githubpr.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        pr_list = []
+        for pull in self.pulls:
+            labels = pull.get_labels()
+            ci_labels = [s for s in labels if 'CI' in s.name]
+            for label in ci_labels:
+                if state in label.name:
+                    if host is not None:
+                        if host.lower() in label.name.lower():
+                            pr_list.append(pull.number)
```
WalterKolczynski-NOAA, line 118:
```suggestion
                            pr_list.append(pull.number)
                            break
```

ci/scripts/utils/githubpr.py
```diff
@@ -0,0 +1,125 @@
...
+                    else:
+                        pr_list.append(pull.number)
```
WalterKolczynski-NOAA, line 121:
```suggestion
                        pr_list.append(pull.number)
                        break
```

ci/scripts/utils/publish_logs.py
```diff
@@ -0,0 +1,107 @@
...
+
+import os
+import secrets
+import re
+import string
+from githubpr import GitHubPR, GitHubDBError
+from argparse import ArgumentParser, ArgumentTypeError, FileType
+
+
+def is_file(path):
```
WalterKolczynski-NOAA: This function name doesn't really match what it is doing. `is_file` sound like it should return a bool about whether the file exists (as `os.path.isfile` does). It shouldn't be raising an exception if the file doesn't exist.

WalterKolczynski-NOAA: Is this even being used?

ci/scripts/utils/publish_logs.py
```diff
@@ -0,0 +1,107 @@
...
+    argparse.ArgumentTypeError
+        If the path is not an existing file.
+    """
+
+    if not os.path.isfile(path):
+        raise ArgumentTypeError(f"{path} does not exist or is not a file.")
+    return path
+
+
+def input_args():
```
WalterKolczynski-NOAA:
```suggestion
def parse_args():
```

ci/scripts/utils/publish_logs.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+
+    description = """Arguments for creating and updating error log files
+    """
+    parser = ArgumentParser(description=description)
+
+    parser.add_argument('--file', help='path to file for uploading to GitHub', required=False, type=FileType('r'), nargs='+')
+    parser.add_argument('--gist', help='create a gist of the file', nargs=1, metavar='identifier_string', required=False)
+    parser.add_argument('--repo', help='create a file in a repo', nargs=1, metavar='path_header', required=False)
+    parser.add_argument('--delete_gists', help='deletes all the gits from authenticated user', action='store_true', default=False, required=False)
```
WalterKolczynski-NOAA: Delete *all* of them?

ci/scripts/utils/publish_logs.py
```diff
@@ -0,0 +1,107 @@
...
+    args = parser.parse_args()
+    if not args.delete_gists and not args.file:
+        parser.error("--file is required when --delete_gists is not used")
+    return args
+
+
+def random_string() -> str:
+    characters = string.ascii_letters + string.digits + string.punctuation
+    return re.sub('[^A-Za-z0-9 ]+', '', ''.join(secrets.choice(characters) for _ in range(6)))
```
WalterKolczynski-NOAA: I don't think you need `re` here. Also, is there a reason it needs to be use secrets?

```suggestion
def random_string(length=6) -> str:
    characters = string.ascii_letters + string.digits + string.punctuation
    return ''.join(secrets.choice(characters) for _ in range(length))
```

or

```suggestion
def random_string(length=6) -> str:
    characters = string.ascii_letters + string.digits + string.punctuation
    return ''.join(random.choices(characters, k=length))
```

ci/scripts/utils/publish_logs.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+if __name__ == '__main__':
+
+    args = input_args()
+    emcbot_ci_url = "https://github.com/emcbot/ci-global-workflows.git"
+    emcbot_gh = GitHubPR(repo_url=emcbot_ci_url)
+
+    if args.gist:  # Add error logs to a gist in GitHub emcbot's account
```
WalterKolczynski-NOAA: Each of these ifs should be a function.

ci/scripts/utils/publish_logs.py
WalterKolczynski-NOAA: Are these mutually exclusive? Should this be modal like setup_exp?

ci/scripts/utils/publish_logs.py
```diff
@@ -0,0 +1,107 @@
...
+
+        gist_files = {}
+        for file in args.file:
+            file_content = file.read()
+            gist_files[os.path.basename(file.name)] = emcbot_gh.InputFileContent(file_content)
+
+        gist = emcbot_gh.user.create_gist(public=True, files=gist_files, description=f"error log file from CI run {args.gist[0]}")
+        print(gist.html_url)
+
+    if args.repo:  # Upload error logs to emcbot's ci-global-workflows error_logs branch
```
WalterKolczynski-NOAA: Are we even going to use this option? I thought we had settled on gists.

sorc/build_all.sh
```diff
@@ -228,11 +257,31 @@ while [[ ${builds_started} -lt ${#build_jobs[@]} ]]; do
       fi
    done
 
+   # If requested, check if any build has failed and exit if so
+   if [[ "${_quick_kill}" == "YES" ]]; then
+      check_builds
+      build_stat=$?
+      if [[ ${build_stat} != 0 ]]; then
```
WalterKolczynski-NOAA:
```suggestion
      if (( build_stat != 0 )); then
```

sorc/build_all.sh
```diff
@@ -193,6 +197,31 @@ fi
 procs_in_use=0
 declare -A build_ids
 
+check_builds()
```
WalterKolczynski-NOAA, line 200:
```suggestion
kill_if_failure()
```

sorc/build_all.sh
```diff
@@ -228,11 +257,31 @@ while [[ ${builds_started} -lt ${#build_jobs[@]} ]]; do
...
+
    sleep 5s
 done
 
+
 # Wait for all jobs to complete and check return statuses
-while [[ ${#build_jobs[@]} -gt 0 ]]; do
+while [[ "${#build_jobs[@]}" -gt 0 ]]; do
+
+   # If requested, check if any build has failed and exit if so
```
WalterKolczynski-NOAA, line 276: I'm not sure we need this in both places. @DavidHuber-NOAA can you take a look at this script's changes?

WalterKolczynski-NOAA (changes requested):

ci/scripts/utils/githubpr.py
TerrenceMcGuinness-NOAA: Completely arbitrary.  Intend to be update to increment numerally in next iteration.

TerrenceMcGuinness-NOAA: It was to remove the `.com` portion after the match as it converts from full path URL to **GitHub**'s user/repo format.

ci/scripts/utils/githubpr.py
TerrenceMcGuinness-NOAA: How does this help?

ci/scripts/utils/githubpr.py
WalterKolczynski-NOAA: Sorry, you're right. My manual test of slicing wasn't great.

ci/scripts/utils/githubpr.py
WalterKolczynski-NOAA: Otherwise you continue looping over labels until you run out of them (and may even add a PR to the list twice).

ci/scripts/utils/publish_logs.py
TerrenceMcGuinness-NOAA: Oh right, it was OBE when I updated the code to allow --file to be list and used FileType instead.  I'll remove it.

ci/scripts/utils/publish_logs.py
TerrenceMcGuinness-NOAA: They are not.  You can use and/or with `--gist` and `--repo`.

ci/scripts/utils/publish_logs.py
TerrenceMcGuinness-NOAA: Yes I am going to it for persistence.  We can remove from the reporting if you like.

WalterKolczynski-NOAA: Okay, that's fine.

ci/scripts/utils/githubpr.py
TerrenceMcGuinness-NOAA: The list results needs to be exhaustive.  We need to check all PRs and continue to find ones that match such labels.

ci/scripts/utils/publish_logs.py
TerrenceMcGuinness-NOAA: I think it's a good idea to only use one though and update this too.

ci/scripts/utils/publish_logs.py
TerrenceMcGuinness-NOAA: oh I suppose.

TerrenceMcGuinness-NOAA:
Manually remove Build Label because messaging to **GitHub** and label updates failed due to _Job's SCM is not GitHub_ error.

TerrenceMcGuinness-NOAA:
@WalterKolczynski-NOAA The **Run Tests** _Failed to build stage_ was a forced error I administered to test the error files being [published](https://gist.github.com/emcbot/23d6f114ad4b8b69b2e72ac415a796d4).

I added the gist link here by hand because Jenkins had one of those authentication errors from the Pipeline GitHub SCM Plugin.

ci/scripts/utils/githubpr.py
WalterKolczynski-NOAA: This should only break the innermost loop, which is over labels, not the PRs, and the PR currently being looked at is already added.

TerrenceMcGuinness-NOAA: oh ok, let me take a closer look.

TerrenceMcGuinness-NOAA: Oh I see, host is fixed, so its first occurrence is sufficient, and Kill would also be a unique criteria once - got it.

ci/scripts/utils/publish_logs.py
```diff
@@ -50,58 +22,89 @@ def input_args():
...
-    parser.add_argument('--delete_gists', help='deletes all the gits from authenticated user', action='store_true', default=False, required=False)
     args = parser.parse_args()
-    if not args.delete_gists and not args.file:
-        parser.error("--file is required when --delete_gists is not used")
+    # if not args.delete_gists and not args.file:
+    #    parser.error("--file is required when --delete_gists is not used")
+    if not args.gist and not args.repo:  # At least one of the two is required
+        parser.error("--gist or --repo is required")
+    if args.gist and args.repo:  # can not use booth
+        parser.error("only use --gist or --repo")
```
WalterKolczynski-NOAA: Use `!=`?
```suggestion
    if bool(args.gist) != bool(args.repo):  # Exactly one of the two is required
        parser.error("Exactly one of --gist and --repo is required")
```

WalterKolczynski-NOAA (changes requested):

ci/scripts/utils/publish_logs.py
WalterKolczynski-NOAA: Is there a reason not to allow both?

TerrenceMcGuinness-NOAA: clever

WalterKolczynski-NOAA (approved):

PR #2442, opened 2024-03-28
Turn GEFS CI test on for Hera

TerrenceMcGuinness-NOAA:
# Description

Re-enabling gefs case to test gefs system builds in Jenkins on Hera

# Type of change
- Maintenance (CI Case configuration)

# Change characteristics

# How has this been tested?
It hasn't this is a trial and previously tested use case
Performing self-test no with in Jenkins CI


WalterKolczynski-NOAA (approved):
Conditionally approved pending successful completion of CI on Hera.

PR #2447, opened 2024-04-01
Move bash utility functions out of preamble

### LINKED ISSUE #2346: Relocate bash functions out of preamble

The preamble is getting a bit cluttered as we add more bash functions, so move those out to a separate bash script that is sourced by the preamble.

Move bash functions (except maybe the `postamble` and `set_strict`/`set_trace`) to a separate script.

### Acceptance Criteria

- [ ] Bash functions moved to new separate script (except as noted above)
- [ ] Source new bash functions script in `preamble.sh`
- [ ] Output identical to existing

_No response_

WalterKolczynski-NOAA:
# Description
The preamble was accumulating a bunch of utility functions. These functions are now moved to a separate file that is sourced by the preamble. The only functions remaining in the preamble are those related to script control and logging (`set_trace()`, `set_strict()`, `postamble()`).

Resolves #2346

# Type of change

# Change characteristics

# How has this been tested?
- Cycled test on Orion

# Checklist


WalterKolczynski-NOAA:
We might be able to move the loading of the util script to the j-job header and avoid sourcing it anew every script, but more testing would be needed for that.

ush/bash_utils.sh
```diff
@@ -0,0 +1,126 @@
...
+        fi
+        value=$(echo "${!template}" | envsubst)
+        # shellcheck disable=SC2086
+        declare ${opts} "${com_var}"="${value}"
+        echo "generate_com :: ${com_var}=${value}"
+    done
+    set_trace
+}
+# shellcheck disable=
+declare -xf generate_com
```
aerorahul: This is fine here, but I suggest we do the `declare -xf` of all the functions at the end to collect them in one place.  It hopefully helps with code organization and readability.

WalterKolczynski-NOAA: Done.

WalterKolczynski-NOAA:
> We might be able to move the loading of the util script to the j-job header and avoid sourcing it anew every script, but more testing would be needed for that.

This actually won't work as-is because the python paths need the python version and they are set before the j-job.

KateFriedman-NOAA (approved):
LGTM

PR #2484, opened 2024-04-13
Add rocoto `sh` tag, script to check netcdf file and apply this to check ocean output

### LINKED ISSUE #2328: ocean/ice post failed occasionally
### What is wrong?

ocean and ice post failed occasionally

post job shall work all the time

All or N/A

### Steps to reproduce

clone the latest g-w 
setup C1152 and make a test run

### Additional information

this problem doesn't happen all the time but it does happen occasionally

aerorahul:
# Description
This PR:
- adds a Rocoto dependency tag that executes a shell command.  The return code of the shell expression serves as a dependency check
- adds a script that executes `ncdump` on a netCDF file.  If the file is a valid netCDF file, the return code is 0, else it is non-zero
- combines the above 2 to use as a dependency check for MOM6 output.  If the model is still in the process of writing out the ocean output, the rocoto will execute the shell script and gather the return code.

This PR also:
- changes permissions on some `ush/` scripts that did not have executable permissions.

Resolves #2328 

With the addition of the `sh`tag, the `gfsocean_prod` dependency will look like:
```xml
<dependency>
    <and>
        <datadep age="120"><cyclestr>&ROTDIR;/gfs.@Y@m@d/@H//model_data/ocean/history/gfs.ocean.t@Hz.6hr_avg.#dep#.nc</cyclestr></datadep>
        <sh shell="/bin/sh"><cyclestr>&HOMEgfs;/ush/check_netcdf.sh &ROTDIR;/gfs.@Y@m@d/@H//model_data/ocean/history/gfs.ocean.t@Hz.6hr_avg.#dep#.nc</cyclestr></sh>
    </and>
</dependency>
```

# Type of change

# Change characteristics

# How has this been tested?
This was tested at low resolution and in a standalone XML where an invalid netCDF file was purposefully placed in the path to check for rocoto status.
This should be tested in the case where the issue was first seen and reported.

# Checklist


ush/check_netcdf.sh
```diff
@@ -0,0 +1,14 @@
+#! /usr/bin/env bash
+
+readonly HOMEgfs=$(cd "$(dirname "$(readlink -f -n "${BASH_SOURCE[0]}" )" )/.." && pwd -P)
+declare -rx HOMEgfs
```
WalterKolczynski-NOAA:
```suggestion
# shellcheck disable=SC2312
declare -rx HOMEgfs=$(cd "$(dirname "$(readlink -f -n "${BASH_SOURCE[0]}" )" )/.." && pwd -P)
```

workflow/rocoto/rocoto.py
```diff
@@ -333,6 +335,31 @@ def _add_streq_tag(dep_dict: Dict[str, Any]) -> str:
...
+def _add_sh_tag(dep_dict: Dict[str, Any]) -> str:
+    """
+    create a simple shell execution tag
+    :param: dep_dict: shell command to execute
+    :type dep_dict: dict
+    :return: Rocoto simple shell execution dependency
+    :rtype: str
+    """
+
+    shell = dep_dict.get('shell', '/bin/sh')
```
WalterKolczynski-NOAA, line 347: We should make the default `bash` to match the rest of workflow

WalterKolczynski-NOAA (dismissed):

NeilBarton-NOAA:
tagging @EricSinsky-NOAA for awareness

workflow/rocoto/rocoto.py
aerorahul: Rocoto's default shell is `sh`.  When rocoto submits jobs, it does so with
`/bin/sh`. IMO, the default should not be altered.

GwenChen-NOAA:
Adding @jiandewang.

aerorahul:
Detective @CoryMartin-NOAA traced the failure to:
```python
str_date = re.search(r'\d{8}', self.filename).group()
        my_date = datetime.strptime(str_date, "%Y%m%d")
```
In the above code, a search is being performed to find 8 integers from `self.filename`.
`self.filename` in this instance is: `/scratch1/NCEPDEV/stmp2/Terry.McGuinness/RUNDIRS/C96_atmaerosnowDA_96009402/gdassnowanl_18/IMSscf.20211220.C96_oro_data.nc`
The first 8 integers in this filename are: `96009402`.  Incidentally, that is the 96009402 of the test commit.  `datetime` is having a difficult time (pun intended) to convert `96009402` into a valid date.  In the year 9600, there is no month 94 (at least on this planet).

I will merge `develop` and resubmit the test.

WalterKolczynski-NOAA:
That explains it. Let me know if you need regex voodoo to find the right string.

CoryMartin-NOAA:
I've opened an issue. It's in a JCSDA repository so it may not be fixed immediately. In the meantime, we shall hope for alphanumeric hashes only!

aerorahul:
> That explains it. Let me know if you need regex voodoo to find the right string.

@CoryMartin-NOAA has opened an issue in the JCSDA-internal/ioda-converters repo.
The new hash should not suffer from the same issue.

CoryMartin-NOAA:
Please also note though that GDASApp will not currently build due to the RZDM meltdown.

WalterKolczynski-NOAA:
@CoryMartin-NOAA This revised regex should only get an eight-digit string if it is between two periods:
```
str_date = re.search(r'\.(\d{8})\.', self.filename).group(1)
```

Test:
```
>>> import re
>>> test_string="abc12345678_def.87654321.xyz"
>>> print( re.search(r'\.(\d{8})\.', test_string).group(1) )                                                                                             
87654321
```

jiandewang:
@aerorahul my test run is successful on ocean post

WalterKolczynski-NOAA (approved):

PR #2488, opened 2024-04-15
CI Refactoring and STALLED case detection

### LINKED ISSUE #2008: Feature BASH CI detects when a dependacy isn't being met

TerrenceMcGuinness-NOAA:
# Description

These updates to the CI Framework does some bash refactoring and adds python tools in order to effectively create the feature for detecting when a CI Case has an experiment that is in a state where it can not advance such as missing a requisite dependency:
- Added separate python script for checking the status of Rocoto driven cases and integrated its use into the bash CI drivers and Jenkins having state logic done in one place.
- Added log publishing python utilities into the bash CI drivers as part of refactoring and consolidations of functionalities
- Update Jenkins behavior while incorporating the above python codes for Rocoto state checking:
    - polling on PR works with one update away from including multiple labels as well
    - Label updates to FAIL as soon as first case fails and continues other cases until completes or is killed by user

  Resolves #2008

# Type of change
- New feature (adds functionality to detect stalled CI cases)
- Maintenance (code refactor, clean-up, new CI functionalities and behavior)

# Change characteristics

# How has this been tested?
BASH cases tested for [stall](https://github.com/TerrenceMcGuinness-NOAA/global-workflow/pull/266#issuecomment-2030822876) and log failing reporting in dev bash cron
Jenkins tested in development mulit-branch project with [fail tests](https://github.com/TerrenceMcGuinness-NOAA/global-workflow/pull/269#issuecomment-2053277298) and full-end-to end [success path](https://github.com/TerrenceMcGuinness-NOAA/global-workflow/pull/270)

# Checklist


ci/Jenkinsfile
```diff
@@ -46,10 +49,8 @@ pipeline {
                         properties([parameters([[$class: 'NodeParameterDefinition', allowedSlaves: ['built-in', 'Hera-EMC', 'Orion-EMC'], defaultSlaves: ['built-in'], name: '', nodeEligibility: [$class: 'AllNodeEligibility'], triggerIfResult: 'allCases']])])
                         HOME = "${WORKSPACE}"
                         sh(script: "mkdir -p ${HOME}/RUNTESTS;rm -Rf ${HOME}/RUNTESTS/error.logs")
-                        pullRequest.addLabel("CI-${Machine}-Building")
-                        if (pullRequest.labels.any { value -> value.matches("CI-${Machine}-Ready") }) {
-                            pullRequest.removeLabel("CI-${Machine}-Ready")
-                        }
+                        sh(script: """${GH} pr edit ${env.CHANGE_ID} --repo ${repo_url} --add-label "CI-${Machine}-Building" """)
+                        sh(script: """${GH} pr edit ${env.CHANGE_ID} --repo ${repo_url} --remove-label "CI-${Machine}-Ready" """)
```
WalterKolczynski-NOAA: Can't adding and removing the label be done in a single GH API call?
```suggestion
                        sh(script: """${GH} pr edit ${env.CHANGE_ID} --repo ${repo_url} --add-label "CI-${Machine}-Building" --remove-label "CI-${Machine}-Ready" """)
```

ci/Jenkinsfile
```diff
@@ -108,26 +109,24 @@ pipeline {
...
                                                         }
                                                     }
-                                                    repo_url=sh(script: "${HOMEgfs}/ci/scripts/utils/publish_logs.py --file ${error_logs} --repo PR_BUILD_${env.CHANGE_ID}", returnStdout: true).trim()
-                                                    gist_url=sh(script: "${HOMEgfs}/ci/scripts/utils/publish_logs.py --file ${error_logs} --gist PR_BUILD_${env.CHANGE_ID}", returnStdout: true).trim()
                                                     try {
-                                                        pullRequest.comment("Build failed on **${Machine}** with error logs:${error_logs_message}\n\nFollow link here to view the contents of the above file(s): [(link)](${gist_url})")
+                                                        sh(script: "${HOMEgfs}/ci/scripts/utils/publish_logs.py --file ${error_logs} --repo PR_BUILD_${env.CHANGE_ID}")
+                                                        gist_url=sh(script: "${HOMEgfs}/ci/scripts/utils/publish_logs.py --file ${error_logs} --gist PR_BUILD_${env.CHANGE_ID}", returnStdout: true).trim()
```
WalterKolczynski-NOAA, line 113: Not sure I understand why publish_logs is getting callled twice, one with a repo switch and one with a gist switch. Shouldn't we only be using one of these?

ci/Jenkinsfile
```diff
@@ -108,26 +109,24 @@ pipeline {
...
                                         }
                                     }
                                     if (env.CHANGE_ID && system == 'gfs') {
                                         try {
-                                            if (pullRequest.labels.any { value -> value.matches("CI-${Machine}-Building") }) {
-                                               pullRequest.removeLabel("CI-${Machine}-Building")
-                                            }
-                                            pullRequest.addLabel("CI-${Machine}-Running")
+                                            sh(script: """${GH} pr edit ${env.CHANGE_ID} --repo ${repo_url} --add-label "CI-${Machine}-Running" """)
+                                            sh(script: """${GH} pr edit ${env.CHANGE_ID} --repo ${repo_url} --remove-label "CI-${Machine}-Building" """)
```
WalterKolczynski-NOAA: Same as before. adding and removing can be done in a single API call.

ci/Jenkinsfile
```diff
@@ -175,19 +175,21 @@ pipeline {
                         when {
                             expression { return caseList.contains(Case) }
                         }
+                        failFast false
```
WalterKolczynski-NOAA, line 176: Does this actually need to be set again?

ci/Jenkinsfile
```diff
@@ -201,15 +203,22 @@ pipeline {
...
                                                 echo "Failed to comment on PR: ${error_comment.getMessage()}"
                                             }
                                         } else {
-                                            echo "No error logs found for failed cases in $HOME/RUNTESTS/error.logs"
+                                            echo "No error logs found for failed cases in $HOME/RUNTESTS/${pslot}_error.logs"
+                                        }
+                                        STATUS = 'Failed'
+                                        try {
+                                            sh(script: """${GH} pr edit ${env.CHANGE_ID} --repo ${repo_url} --remove-label "CI-${Machine}-Running" """, returnStatus: true)
+                                            sh(script: """${GH} pr edit ${env.CHANGE_ID} --repo ${repo_url} --add-label "CI-${Machine}-${STATUS}" """, returnStatus: true)
```
WalterKolczynski-NOAA: These can be combined as well.

ci/Jenkinsfile
```diff
@@ -220,49 +229,25 @@ pipeline {
...
+            agent { label "${machine}-emc" }
+            steps {
+                script {
                     try {
-                        pullRequest.addLabel("CI-${Machine}-Failed")
-                        def timestamp = new Date().format('MM dd HH:mm:ss', TimeZone.getTimeZone('America/New_York'))
-                        pullRequest.comment("**CI FAILED** ${Machine} at ${timestamp}<br>Built and ran in directory `${HOME}`")
+                        sh(script: """${GH} pr edit ${env.CHANGE_ID} --repo ${repo_url} --remove-label "CI-${Machine}-Running" """, returnStatus: true)
+                        sh(script: """${GH} pr edit ${env.CHANGE_ID} --repo ${repo_url} --remove-label "CI-${Machine}-Building" """, returnStatus: true)
+                        sh(script: """${GH} pr edit ${env.CHANGE_ID} --repo ${repo_url} --add-label "CI-${Machine}-${STATUS}" """, returnStatus: true)
```
WalterKolczynski-NOAA: All three of these can be a single API call.

ci/scripts/utils/rocotostat.py
```diff
@@ -0,0 +1,118 @@
...
+
+    return args
+
+
+def rocoto_statcount():
+    """
+    Run rocotostat and process its output.
+    """
+
+    args = input_args()
```
WalterKolczynski-NOAA, line 47: Why is this function calling `input_args()`? `__main__` should be calling it (and already is) and then any relevant ones passed in as arguments to this function.

ci/scripts/utils/rocotostat.py
```diff
@@ -0,0 +1,118 @@
...
+    parser.add_argument('--verbose', action='store_true', help='List the states and the number of jobs that are in each', required=False)
+    parser.add_argument('-v', action='store_true', help='List the states and the number of jobs that are in each', required=False)
+    parser.add_argument('--export', action='store_true', help='create and export list of the status values for bash', required=False)
+
+    args = parser.parse_args()
+
+    return args
+
+
+def rocoto_statcount():
```
WalterKolczynski-NOAA, line 42: This function is doing two different things: a `rocotostat` to count the completed cycles and a different `rocotostat` to determine the number of jobs in each state. These should be in separate functions.

ci/scripts/check_ci.sh
```diff
@@ -123,31 +123,33 @@ for pr in ${pr_list}; do
...
-         pslot=$(basename "${kill_cases}")
-         cancel_slurm_jobs "${pslot}"
-      done
-      break
+
+    set +e
+    rocoto_state="$("${HOMEgfs}/ci/scripts/utils/rocotostat.py" -w "${xml}" -d "${db}")"
+    rocoto_error=$?
+    rm -f "${output_ci_single}"
+    if [[ "${rocoto_error}" -ne 0 ]]; then
```
WalterKolczynski-NOAA, line 141: I think `STALLED` may need to be treated differently, in which case `rocoto_error` may be redundant and this can just be an `elif` on the `rocoto_state`. For stalled workflows, there will be no error logs, but we should ideally be printing out a list of the jobs that haven't been queued (even better if we can make a gist with the `rocotocheck` for each). For a MVP, just having the PR comment reflect that it stalled is sufficient. Then we can add more functionality later.

ci/scripts/run-check_ci.sh
```diff
@@ -4,11 +4,12 @@ set -eu
 
 #####################################################################################
 # Script description: script to check the status of an experiment as reported
-#                     by Rocoto
+#                     by Rocoto
```
WalterKolczynski-NOAA: Don't add a trailing space here.

ci/scripts/run-check_ci.sh
```diff
@@ -23,7 +24,7 @@ pslot=${2:-${pslot:-?}}        # Name of the experiment being tested by this scr
 #         └── ${pslot}
 # Two system build directories created at build time gfs, and gdas
 # TODO: Make this configurable (for now all scripts run from gfs for CI at runtime)
-HOMEgfs="${TEST_DIR}/gfs"
+HOMEgfs="${TEST_DIR}/${SYSTEM_BUILD_DIR:-global-workflow}"
```
WalterKolczynski-NOAA: This default will never be used. L12 already errors if `$SYSTEM_BUILD_DIR` is not set.

ci/scripts/run-check_ci.sh
```diff
@@ -48,54 +49,56 @@ fi
...
   echo "Gather Rocoto statistics"
-  rocotostat_output=$(rocotostat -w "${xml}" -d "${db}" -s | grep -v CYCLE) || true
-  num_cycles=$(echo "${rocotostat_output}" | wc -l) || true
-  num_done=$(echo "${rocotostat_output}" | grep -c Done) || true
-  num_succeeded=$(rocotostat -w "${xml}" -d "${db}" -a | grep -c SUCCEEDED) || true
-  num_failed=$(rocotostat -w "${xml}" -d "${db}" -a | grep -c -E 'FAIL|DEAD') || true
+  # shellcheck disable=SC2312 # We want to use the exit code of the command
+  eval=$("${HOMEgfs}/ci/scripts/utils/rocotostat.py" -w "${xml}" -d "${db}" --export)
+  error_stat=$?
+  eval "${eval}"
```
WalterKolczynski-NOAA: Please find a way to do this without using `eval`. `eval` should nearly always be avoided.

Possible alternatives (from worst to best):
- Have the python script write a script to be sourced (this might be just as bad though)
- Have the python script write the variables to the environment
- Serialize the dict (maybe as a JSON) to either the output or a separate file and then parse it in bash. If you have one key-pair per line, you could use grep to get the right line and then just split on the separator.

ci/scripts/utils/ci_utils.sh
```diff
@@ -126,3 +126,30 @@ function create_experiment () {
   "${HOMEgfs}/${system}/workflow/create_experiment.py" --overwrite --yaml "${yaml_config}"
 
 }
+
+function publish_logs() {
+# publish_logs function
+# This function takes a directory path and a list of files as arguments.
+# It calls the publish_logs.py script to publish the logs and returns its gist URL.
+# Usage: publish_logs <ID> <dir_path> <file1> <file2> ... <fileN>
```
WalterKolczynski-NOAA: The docblock indentation should match the function that follows.

ci/scripts/utils/pr_list_database.py
```diff
@@ -3,28 +3,27 @@
...
 import os
 from wxflow import SQLiteDB, SQLiteDBError
+from githubpr import GitHubPR
 from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter, REMAINDER
 
 
 def full_path(string):
     """
     full_path Get the absolute path of a file or directory.
-
```
WalterKolczynski-NOAA, line 12: Why did you remove these spacer lines? Seems like it would make the docblock more difficult to read.

ci/scripts/utils/pr_list_database.py
```diff
@@ -83,16 +84,16 @@ def update_pr(db: SQLiteDB, args):
         db.update_data('pr_list', update, value, 'pr', args.update_pr[0])
 
 
-def display_db(db, display):
+def display_db(db, display) -> list:
```
WalterKolczynski-NOAA: Why add a type hint for the return but not the arguments?

ci/scripts/utils/pr_list_database.py
```diff
@@ -83,16 +84,16 @@ def update_pr(db: SQLiteDB, args):
...
     Display the database.
 
     Parameters
     ----------
-    ci_database : SQLiteDB
+    db : SQLiteDB
         The database to display.
-    display : list
-        The command line argument values.
+    args : display
```
WalterKolczynski-NOAA: This doesn't match the function signature.

ci/scripts/utils/pr_list_database.py
```diff
@@ -101,15 +102,43 @@ def display_db(db, display):
...
         values.append(' '.join(map(str, row)))
 
     return values
 
 
+def update_database(db: SQLiteDB) -> list:
+    """
+    Update the database from the GitHub PRs
+    - only PRs from host machine are added to the database
+    - if the PR is already in the database it its added to the kill list
```
WalterKolczynski-NOAA, line 131: I think I know why, but this is a bit counter-intuitive.

WalterKolczynski-NOAA (changes requested):

ci/Jenkinsfile
TerrenceMcGuinness-NOAA: The repo one is for persistence and analysis.

ci/Jenkinsfile
TerrenceMcGuinness-NOAA: ok, I'll test that, makes sense.

ci/scripts/check_ci.sh
TerrenceMcGuinness-NOAA: Sounds like a solid plan.

ci/scripts/run-check_ci.sh
TerrenceMcGuinness-NOAA: L12?  If it is not set it will be **global-workflow**, right?

TerrenceMcGuinness-NOAA: Oh line 12, looking.

TerrenceMcGuinness-NOAA: But wait, if L12 is null it does not error, it gets populated by L27 as intended, right?

ci/scripts/run-check_ci.sh
TerrenceMcGuinness-NOAA: Again, solid.  Shall we do this for this PR?

ci/scripts/utils/rocotostat.py
TerrenceMcGuinness-NOAA: oops, looks like a cut-and-paste typo

ci/scripts/check_ci.sh
TerrenceMcGuinness-NOAA: Ok, we still have been returning error code non-zero on stalled cases but now it does not look for error logs but instead just reports that it stalled and we can make those gist with the outputs from `rocotocheck` as a new feature.

ci/scripts/utils/rocotostat.py
TerrenceMcGuinness-NOAA: Not sure what that means functionally in terms of updating the  **rocoto_status** dictionary which is the only thing this tool does.

TerrenceMcGuinness-NOAA (commented):
Made changes, a couple didn't have a functional change that was apperent

ci/scripts/check_ci.sh
```diff
@@ -140,6 +140,13 @@ for pr in ${pr_list}; do
     rm -f "${output_ci_single}"
     if [[ "${rocoto_error}" -ne 0 ]]; then
         "${GH}" pr edit --repo "${REPO_URL}" "${pr}" --remove-label "CI-${MACHINE_ID^}-Running" --add-label "CI-${MACHINE_ID^}-Failed"
+        if [[ "${rocoto_state}" == "STALLED" ]]; then
+          # shellcheck disable=SC2312
+          "${GH}" pr comment "${pr}" --repo "${REPO_URL}" --body "Experiment ${pslot} **${rocoto_state}** on ${MACHINE_ID^} at $(date +'%D %r')"
+          "${HOMEgfs}/ci/scripts/utils/pr_list_database.py" --remove_pr "${pr}" --dbfile "${pr_list_dbfile}"
+          cancel_all_batch_jobs "${pr_dir}/RUNTESTS"
```
WalterKolczynski-NOAA, line 147: If it's stalled, there shouldn't be any batch jobs.

ci/scripts/run-check_ci.sh
```diff
@@ -69,9 +69,15 @@ while true; do
   # Get job statistics
   echo "Gather Rocoto statistics"
   # shellcheck disable=SC2312 # We want to use the exit code of the command
-  eval=$("${HOMEgfs}/ci/scripts/utils/rocotostat.py" -w "${xml}" -d "${db}" --export)
+  eval=$("${HOMEgfs}/ci/scripts/utils/rocotostat.py" -w "${xml}" -d "${db}" -v)
```
WalterKolczynski-NOAA:
```suggestion
  full_state=$("${HOMEgfs}/ci/scripts/utils/rocotostat.py" -w "${xml}" -d "${db}" -v)
```

ci/scripts/run-check_ci.sh
```diff
@@ -69,9 +69,15 @@ while true; do
...
   error_stat=$?
-  eval "${eval}"
+
+  CYCLES_TOTAL=$(echo "${eval}" | grep CYCLES_TOTAL | cut -d: -f2) || true
+  CYCLES_DONE=$(echo "${eval}" | grep CYCLES_DONE | cut -d: -f2) || true
+  SUCCEEDED=$(echo "${eval}" | grep SUCCEEDED | cut -d: -f2) || true
+  FAIL=$(echo "${eval}" | grep FAIL | cut -d: -f2) || true
+  DEAD=$(echo "${eval}" | grep DEAD | cut -d: -f2) || true
+  ROCOTO_STATE=$(echo "${eval}" | tail -1) || true
```
WalterKolczynski-NOAA:
```suggestion
  for state in CYCLES_TOTAL CYCLES_DONE SUCCEEDED FAIL DEAD; do
    declare ${!state}=$(echo "${full_state}" | grep "${state}" | cut -d: -f2) || true
  done
  ROCOTO_STATE=$(echo "${full_state}" | tail -1) || exit 1
```

ci/Jenkinsfile
WalterKolczynski-NOAA: Unresolving, since there was no response on whether this is really needed.

WalterKolczynski-NOAA (changes requested):
Please make sure you've also addressed all of the comments from the original review.

ci/scripts/run-check_ci.sh
TerrenceMcGuinness-NOAA: Nice!

ci/scripts/check_ci.sh
TerrenceMcGuinness-NOAA: That will cancel all the cases that are remaining when one is stalled.

ci/Jenkinsfile
TerrenceMcGuinness-NOAA: Not sure, I will have to test that using **Replay**

ci/scripts/run-check_ci.sh
```diff
@@ -48,54 +49,60 @@
...
-  num_done=$(echo "${rocotostat_output}" | grep -c Done) || true
-  num_succeeded=$(rocotostat -w "${xml}" -d "${db}" -a | grep -c SUCCEEDED) || true
-  num_failed=$(rocotostat -w "${xml}" -d "${db}" -a | grep -c -E 'FAIL|DEAD') || true
+  # shellcheck disable=SC2312 # We want to use the exit code of the command
+  eval=$("${HOMEgfs}/ci/scripts/utils/rocotostat.py" -w "${xml}" -d "${db}" -v)
+  error_stat=$?
 
-  echo "${pslot} Total Cycles: ${num_cycles} number done: ${num_done}"
+  for state in CYCLES_TOTAL CYCLES_DONE SUCCEEDED FAIL DEAD; do
+    declare ${!state}=$(echo "${full_state}" | grep "${state}" | cut -d: -f2) || true
```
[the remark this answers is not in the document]
WalterKolczynski-NOAA: Oops, somehow submitted something from an earlier version. Remove the `!`,

ci/Jenkinsfile
TerrenceMcGuinness-NOAA: Ok I looked at the Pipeline for a bit a remembered why I had failFast in two places like that.  We needed failFast to support the new behavior for setting the fail label for when there there is a failure without a post section after the main pipeline.  Before having a global fail fast simply went to post.  Now the fine grain failFast is in each parallel section, one for Create Experiment and the other for Run Experiment.  If any one element of each parallel section fails the Failed label is set and the user in signaled that CI as not passed.

ci/scripts/run-check_ci.sh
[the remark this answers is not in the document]
TerrenceMcGuinness-NOAA: ok

TerrenceMcGuinness-NOAA (commented):
made changes

WalterKolczynski-NOAA:
@TerrenceMcGuinness-NOAA there are still three comments on `pr_list_database.py` from the original review that have not been addressed.

@WalterKolczynski-NOAA Finally understood the "spacer" comment and updated.

ci/scripts/utils/pr_list_database.py
TerrenceMcGuinness-NOAA: I'm not sure that's something I need to "fix".

ci/scripts/utils/pr_list_database.py
WalterKolczynski-NOAA: ...

ci/scripts/utils/pr_list_database.py
WalterKolczynski-NOAA: ...

WalterKolczynski-NOAA (approved):

Issue #2503, opened 2024-04-18
GFSv16.3.? - GLDAS updates


KateFriedman-NOAA:
### Description

NCO/SPA Justin Cooke is opening several bugzillas related to updated needed to the GLDAS job in operations. This came out of recent issues in operations related to missing CPC gauge data.

Justin's comments:
```
1) We saw references in the code to an h2 emc space for this input data too:

export CPCGAUGE=${CPCGAUGE:-/lfs/h2/emc/global/noscrub/emc.global/dump}

Production jobs should not reference emc disk spaces. We'll be opening up a Bugzilla ticket for that. 

2) If the CPC data is missing the job will fail, the warning message needs to be changed, currently it is:

if [ ! -s $cpc ]; then
 echo "WARNING: GLDAS MISSING $cpc, WILL NOT RUN."
 exit 3
fi

It needs to be: 

if [ ! -s $cpc ]; then
 echo "FATAL ERROR: GLDAS MISSING $cpc, WILL NOT RUN."
 exit 3
fi

3) This job also runs at 06, 12, 18Z, but at those times it just reports this message:
0.319 + echo 'GLDAS only runs for 00 cycle; Skip GLDAS step for cycle 18'
GLDAS only runs for 00 cycle; Skip GLDAS step for cycle 18

Why does the job exist for those cycles?
```

### Target version

v16.3.?? (TBD)

### Expected workflow changes

Initial suggested changes:
1) So that we can retain the ability for EMC developers to still run outside of ops and use the dump data that we store in the global dump archive (that default emc space), we should set `export CPCGAUGE=/lfs/h2/emc/global/noscrub/emc.global/dump` in the dev-only config.base (config.base.emc.dyn). I am fine with Jiarui's suggestion to change the default in the script to something like `/lfs/h1/ops/prod/com/gfs/v16.3`. We can pass our dump archive path via the override and our config.base setting.
2) I agree with updating the error message. Let's get that changed to Justin's suggestion.
3) For ecflow in ops we can just remove the GLDAS job from the 06/12/18 job families and adjust job dependencies for those cycles to not wait for that job (the analysis job triggers).

KateFriedman-NOAA:
Additional email from SPA Justin:

```
EMC,

Overnight the 00Z run of today's (April 17 2024) atmos/init/jgdas_atmos_gldas job failed due to a missing input file produced by CPC. 

[...]

PR #2508, opened 2024-04-19
Workflow updates for GLDAS fixes

KateFriedman-NOAA:
# Description

This PR brings in the workflow updates to address GLDAS enhancements requested by NCO. GLDAS script updates come in via a new GLDAS repo tag [gldas_gfsv16_release.v.2.1.1](https://github.com/NOAA-EMC/GLDAS/releases/tag/gldas_gfsv16_release.v.2.1.1).

Workflow side updates:

- New GLDAS tag `gldas_gfsv16_release.v.2.1.1`
- Update ecFlow definition file to remove the `jgdas_atmos_gldas` job from the 06z, 12z, and 18z cycles; includes removing it from the forecast job triggers for those cycles as well
- Created release notes
- Updated GFS version to v16.3.17 (tentative, will adjust after receiving confirmation from NCO)

Refs #2503

# Type of change

Production update

# Change characteristics

# How has this been tested?

Visual inspection, not yet tested on WCOSS2.

### DIFF
--- a/ecf/defs/gfs_v16_3.def
+++ b/ecf/defs/gfs_v16_3.def
@@ -4866,10 +4866,6 @@ suite gfs_v16_3
                     trigger /gfs_v16_3/primary/06/obsproc/v1.0/gdas/atmos/dump/jobsproc_gdas_atmos_dump:release_sfcprep
                 endfamily
               endfamily
-              family init
-                task jgdas_atmos_gldas
-                  trigger ../analysis/jgdas_atmos_analysis == complete
-              endfamily
               family analysis
                 task jgdas_atmos_analysis
                   trigger /gfs_v16_3/primary/06/obsproc/v1.0/gdas/atmos/prep/jobsproc_gdas_atmos_prep == complete and ../obsproc/prep/jgdas_atmos_emcsfc_sfc_prep == complete
@@ -4998,7 +4994,7 @@ suite gfs_v16_3
               endfamily
             endfamily
             task jgdas_forecast
-              trigger ./atmos/analysis/jgdas_atmos_analysis:release_fcst and ./wave/prep/jgdas_wave_prep == complete and ./atmos/init/jgdas_atmos_gldas == complete
+              trigger ./atmos/analysis/jgdas_atmos_analysis:release_fcst and ./wave/prep/jgdas_wave_prep == complete
           endfamily
           family enkfgdas
             edit RUN 'gdas'
@@ -7493,10 +7489,6 @@ suite gfs_v16_3
                     trigger /gfs_v16_3/primary/12/obsproc/v1.0/gdas/atmos/dump/jobsproc_gdas_atmos_dump:release_sfcprep
                 endfamily
               endfamily
-              family init
-                task jgdas_atmos_gldas
-                  trigger ../analysis/jgdas_atmos_analysis == complete
-              endfamily
               family analysis
                 task jgdas_atmos_analysis
                   trigger /gfs_v16_3/primary/12/obsproc/v1.0/gdas/atmos/prep/jobsproc_gdas_atmos_prep == complete and ../obsproc/prep/jgdas_atmos_emcsfc_sfc_prep == complete
@@ -7625,7 +7617,7 @@ suite gfs_v16_3
               endfamily
             endfamily
             task jgdas_forecast
-              trigger ./atmos/analysis/jgdas_atmos_analysis:release_fcst and ./wave/prep/jgdas_wave_prep == complete and ./atmos/init/jgdas_atmos_gldas == complete
+              trigger ./atmos/analysis/jgdas_atmos_analysis:release_fcst and ./wave/prep/jgdas_wave_prep == complete
           endfamily
           family enkfgdas
             edit RUN 'gdas'
@@ -10120,10 +10112,6 @@ suite gfs_v16_3
                     trigger /gfs_v16_3/primary/18/obsproc/v1.0/gdas/atmos/dump/jobsproc_gdas_atmos_dump:release_sfcprep
                 endfamily
               endfamily
-              family init
-                task jgdas_atmos_gldas
-                  trigger ../analysis/jgdas_atmos_analysis == complete
-              endfamily
               family analysis
                 task jgdas_atmos_analysis
                   trigger /gfs_v16_3/primary/18/obsproc/v1.0/gdas/atmos/prep/jobsproc_gdas_atmos_prep == complete and ../obsproc/prep/jgdas_atmos_emcsfc_sfc_prep == complete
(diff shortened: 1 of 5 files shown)


HelinWei-NOAA:
@KateFriedman-NOAA The other jobs can still move on even if the 00cycle gldas fails? Or we need to set the gldas job complete manually to move other jobs. If the latter is true, probably we should skip the gldas if CPC gauge data is missing.

aerorahul (approved):
looks good.
thanks.

KateFriedman-NOAA:
> @KateFriedman-NOAA The other jobs can still move on even if the 00cycle gldas fails? Or we need to set the gldas job complete manually to move other jobs. If the latter is true, probably we should skip the gldas if CPC gauge data is missing.

@HelinWei-NOAA I defer to you on this but will give you my thoughts: If the job isn't critical to the GFS running, then I would not exit with a non-zero number. From NCO's perspective I imagine they'd prefer the job to not fail if data is missing, print a warning that the data is missing and job will not run, and exit with 0. The alternative is to have the job fail but if the missing data is not recoverable then the job shouldn't fail and hold up production. It should essentially be skipped.

> probably we should skip the gldas if CPC gauge data is missing

I tend to agree. The job should still run but if the data is missing it should print the warning and exit gracefully.

Based on what we're talking about, [these lines](https://github.com/NOAA-EMC/GLDAS/blob/release/gfs.v16.3_gldas/ush/gldas_forcing.sh#L55:L56) of `ush/gldas_forcing.sh` should be changed to something like this:
```
cpc=$pathp1/$cpc_precip
if [ ! -s $cpc ]; then cpc=$pathp2/$cpc_precip ; fi
if [ $RUN_ENVIR = "nco" ]; then cpc=$pathp2/$cpc_precip ; fi
if [ ! -s $cpc ]; then 
 echo "WARNING: GLDAS MISSING $cpc, JOB WILL NOT RUN."
 exit 0
fi
```

@aerorahul your thoughts?

HelinWei-NOAA:
The sole factor influencing GLDAS to generate different initial land states than gdas is CPC gauge data. Running GLDAS without this data is unnecessary. Therefore, I propose implementing this restriction at a higher level, such as in gfs_v16_3.def. Without the data, the task jgdas_atmos_gldas should not be initiated.

KateFriedman-NOAA:
@HelinWei-NOAA Understood, however, ecFlow does not have data dependencies (like rocoto does). We can adjust triggers based on the status of other jobs but can't (to my knowledge) fire off and/or set a job to complete based on data availability. We need to handle the data availability within the job. 99.9% of the time the data is available so I think it would be fine to always let the job run but just exit with a warning and `exit 0` if the data is missing.

ecf/defs/gfs_v16_3.def
HelinWei-NOAA, line 1: IF CPC GAUGE DATA IS MISSING AND THE GLDAS JOB EXIT WITH 2, ./atmos/init/jgdas_atmos_gldas == complete IN LINE 2374 IS STILL SET TO BE TRUE?

HelinWei-NOAA:
Sounds good. I have another question. If CPC gauge data is missing and the gldas job exit with 2, ./atmos/init/jgdas_atmos_gldas  in line 2374 of gfs_v16_3.def is still set to complete?
> @HelinWei-NOAA Understood, however, ecFlow does not have data dependencies (like rocoto does). We can adjust triggers based on the status of other jobs but can't (to my knowledge) fire off and/or set a job to complete based on data availability. We need to handle the data availability within the job. 99.9% of the time the data is available so I think it would be fine to always let the job run but just exit with a warning and `exit 0` if the data is missing.

KateFriedman-NOAA:
> I have another question. If CPC gauge data is missing and the gldas job exit with 2, ./atmos/init/jgdas_atmos_gldas in line 2374 of gfs_v16_3.def is still set to complete?

Only `exit 0` will lead to a "complete" job. Any exit # other than 0 will fire off an `ecflow_client --abort` signal and message that there was an error running the job. See this section of the GLDAS ecf script:

```
${HOMEgfs}/jobs/JGDAS_ATMOS_GLDAS
if [ $? -ne 0 ]; then
   ecflow_client --msg="***JOB ${ECF_NAME} ERROR RUNNING J-SCRIPT ***"
   ecflow_client --abort
   exit
fi
```
https://github.com/NOAA-EMC/global-workflow/blob/dev/gfs.v16/ecf/scripts/gdas/atmos/init/jgdas_atmos_gldas.ecf#L48:L53

HelinWei-NOAA:
@KateFriedman-NOAA The gldas task (/ecf/scripts/gdas/atmos/init/jgdas_atmos_gldas.ecf) uses ${HOMEgfs}/jobs/JGDAS_ATMOS_GLDAS. But I can't find this job script in this branch.

KateFriedman-NOAA:
> @KateFriedman-NOAA The gldas task (/ecf/scripts/gdas/atmos/init/jgdas_atmos_gldas.ecf) uses ${HOMEgfs}/jobs/JGDAS_ATMOS_GLDAS. But I can't find this job script in this branch.

@HelinWei-NOAA That script lives in the GLDAS repo for GFSv16 still and isn't viewable under a global-workflow clone until we run the link setup step which makes symlinks to the scripts within the component `sorc/*.fd`. Run the `sorc/link_fv3gfs.sh` script after running checkout and build and you'll then see the script under the `jobs/` folder.

Note, for GFSv17 we've absorbed all of the scripts so we won't link to the component checkouts anymore with that version.

HelinWei-NOAA:
@KateFriedman-NOAA @aerorahul @yangfanglin So I propose that if cpc gauge data is missing, we just skip the gldas step. 

Change the section in gldas_forcing.sh:

if [ ! -s $cpc ]; then
 echo "FATAL ERROR: GLDAS MISSING $cpc, WILL NOT RUN."
 exit 2
fi

to 
if [ ! -s $cpc ]; then
 echo "GLDAS only runs when CPC gauge data is available "
 exit 0
fi

KateFriedman-NOAA:
@HelinWei-NOAA I'm good with this but suggest adding `WARNING` to the message and being more verbose about why the job isn't running. Like this:

```
if [ ! -s $cpc ]; then
  echo "WARNING: GLDAS MISSING ${cpc} and can only run when CPC gauge data is available. Will not run. Exiting."
  exit 0
fi
```

KateFriedman-NOAA:
@HelinWei-NOAA Checking in...what are your thoughts on my last suggestion? Do you have additional thoughts or feedback or changes? Thanks!

HelinWei-NOAA:
@KateFriedman-NOAA Your suggestion is good. I will make that change.

KateFriedman-NOAA:
> Your suggestion is good. I will make that change.

Great, please let me know when you've made the change and recut the GLDAS tag (you can use the same name). Thanks!

HelinWei-NOAA:
Done. Same tag #
> > Your suggestion is good. I will make that change.
> 
> Great, please let me know when you've made the change and recut the GLDAS tag (you can use the same name). Thanks!

KateFriedman-NOAA:
> Done. Same tag #

Thanks @HelinWei-NOAA ! Please review this PR and, if happy with it, please approve it and I will merge these updates into the auth repo release branch.

HelinWei-NOAA (approved):

ecf/defs/gfs_v16_3.def
KateFriedman-NOAA: The job will not exit with 2, it will exit with 0 if the data is missing...so the trigger on line 2374 will always be complete regardless of the CPC gauge data being missing or not.

Issue #1934, opened 2023-10-17
Create links for GOCART output instead of copying after forecast


JessicaMeixner-NOAA:
### What new functionality do you need?

Would like to be able to link gocart diagnostic files to COM directory instead of copying them at the end like https://github.com/NOAA-EMC/global-workflow/pull/1933 implements.   

### What are the requirements for the new functionality?

The forecast for S2SWA runs even if you link the gocart*.nc files instead of copying at the end. 

### Acceptance Criteria

The aerosol team has the diagnostic output they need. 

### Suggest a solution (optional)

I tried to implement suggestions from @rmontuoro and @bbakernoaa here: https://github.com/jessicameixner-noaa/global-workflow/tree/trygocartfix but the forecast still crashed.   The setting updates are in this commit: https://github.com/JessicaMeixner-NOAA/global-workflow/commit/f766fd972c5cecf2159047ffd398fe27026af821

WalterKolczynski-NOAA:
Was a UFS and/or GOCART issue created that can be linked here? Barring information to the contrary from the developers, it seems like the issue needs to be solved there.

JessicaMeixner-NOAA:
There was an older GOCART Issues that hopefully @lipan-NOAA or @rmontuoro can point you to - apologies Raffaele I lost the issue that you previously sent.

JessicaMeixner-NOAA:
UFS issue https://github.com/ufs-community/ufs-weather-model/issues/1955 was created

HenryRWinterbottom:
> ### What new functionality do you need?
> 
> Would like to be able to link gocart diagnostic files to COM directory instead of copying them at the end like [#1933](https://github.com/NOAA-EMC/global-workflow/pull/1933) implements.
> ### What are the requirements for the new functionality?
> 
> The forecast for S2SWA runs even if you link the gocart*.nc files instead of copying at the end.
> ### Acceptance Criteria
> 
> The aerosol team has the diagnostic output they need.
> ### Suggest a solution (optional)
> 
[...]

Issue #2349, opened 2024-02-27
Use cpfs for copies that should be atomic


WalterKolczynski-NOAA:
### What new functionality do you need?

Using the standard `cp` can result in downstream programs using incomplete copies, particularly for large files. `prod_util` provides a utility, `cpfs`, that does an atomic copy by copying to a temp file and then renaming the temp file when the copy is complete. This is an NCO requirement (see Implementation Standards v11).

There are also places where we are doing this type of copy manually. These should be replaced with using `cpfs` to simplify scripts.

### What are the requirements for the new functionality?

All copies that should be atomic (e.g. most or all copies to COM) should use `cpfs` or `cpreq` instead of `cp`. Should use `cpreq` when files are required to be able to run, and `cpfs` when output is being copied to `COMOUT`.

`cpfs` does not allow globbing, so some modification may be needed.

### Acceptance Criteria

- [ ] No usage of `cp` when the copy could be used before completion (replaced with `cpfs`)
- [ ] Replace instances where we are manually doing an atmoic copy with `cpfs`
- [ ] Output identical to previous

### Suggest a solution (optional)

Should we leverage `NCP`? There are still a lot of places we are directly using `cp` instead of `NCP`.

WalterKolczynski-NOAA:
Related: #712

DavidHuber-NOAA:
A thorough search of the code shows only a handful of calls to `cp` rather than `cpfs` or `cpreq`.  Of these, only the following lines copy into COM (where `cpfs` should be used):

https://github.com/NOAA-EMC/global-workflow/blob/f294f5e39437b8bddbe08f4b0238b43d2431e046/ush/tropcy_relocate.sh#L351-L352
https://github.com/NOAA-EMC/global-workflow/blob/f294f5e39437b8bddbe08f4b0238b43d2431e046/ush/tropcy_relocate.sh#L394
https://github.com/NOAA-EMC/global-workflow/blob/f294f5e39437b8bddbe08f4b0238b43d2431e046/ush/tropcy_relocate.sh#L573-L574
https://github.com/NOAA-EMC/global-workflow/blob/f294f5e39437b8bddbe08f4b0238b43d2431e046/ush/syndat_qctropcy.sh#L357-L365
https://github.com/NOAA-EMC/global-workflow/blob/f294f5e39437b8bddbe08f4b0238b43d2431e046/ush/wave_prnc_ice.sh#L142

PR #2510, opened 2024-04-19
Add restart on failure capability for the forecast executable

### LINKED ISSUE #2273: Extend mid-forecast restart detection and resumption for coupled

Ability to detect when restart files for a coupled forecast already exist and resume the forecast from that point. This already exists for atmos, but needs to be extended to other components (and re-validated for the atmosphere).

Forecast job detects when there are existing restart files for the current cycle and resumes from that point.

### Acceptance Criteria

A forecast that is aborted and then rerun detects restart files from the failed attempt and resumes from that point without rerunning already completed forecast hours.

Update `forecast_det.sh` to also detect restarts for the coupled components when they are turned on.

Can be done one of two ways: the existing `FV3_det()` can be renamed and updated to handle all the detection, or the skeleton `*_det()` can be updated and combined to use only the lowest value.

aerorahul:
# Description
This PR:
- enables restart capability of the forecast executable from a previous failure.
- saves restarts during the run in a new `DATA` structure.  The current `DATA` structure:
current `DATA`
is being replaced by:
Screenshot 2024-04-19 at 12 55 44 PM
where, the colored boxes are described as:
Screenshot 2024-04-19 at 12 56 14 PM
- saves model output from `MOM6` and `CICE` within `MOM6_OUTPUT/` and `CICE_OUTPUT/` sub-directories.  This is done to keep the run directory clean and easily identify component output.

This PR also:
- replaces link with copy.  This enables the creation of a `DATA` directory that is self-contained and can be used to diagnose issues during failures.  This is a NCO EE2 requirement and addresses part of an outstanding bugzilla.

In the process of enabling the restart capability, functionality from `forecast_postdet.sh` is moved to `forecast_predet.sh` that does not depend on the outcome of `forecast_det.sh`.  `forecast_det.sh` determines where the initial conditions will come from; `COM` in the case of a clean run or `DATArestart` in the case of a `RERUN`.
This should make it easier to separate  **static**  configuration and data (fix files, etc) from **runtime** configuration (namelists, etc) and data (initial conditions)

Additionally, this PR:
- adds 3 utility shell scripts in `test/`.  
  - 'nccmp.sh` - compare netCDF files using `nccmp`
  - `g2cmp.sh` - compare grib2 files using `wgrib2`
  - `f90nmlcmp.sh` - compare Fortran90 nml files using `f90nml` (Requires modulefiles to load `py-f90nml` module on RDHPCS platforms)
They are not used in the workflow, but are useful for users to compare files.

Resolves #2273

# Type of change
- New feature
- Maintenance

# Change characteristics

# How has this been tested?
Summary of testing:
- GFS:  
  - C48_ATM, C48_S2S, C48_S2SW tests were conducted.  The wallclock was cut down to run half to simulate a forecast failure.   This caused the job to time out and resubmit.  On resubmission, the rerun was successfully executed.  Restarting for S2SW needs some work in the waves component.  Aerosol restart capability is tested in the GEFS tests.
  - C96 3DVar, C96C48 hybrid4DEnVar tests were run to test the IAU feature.  Restart on failure is disabled for enkfgdas and enkfgfs forecast.  If necessary, rerun capability will need to be configured such that the restart can only start **after** the assimilation window has passed when IAU is used.
- GEFS: Same as above for the `gfsfcst`, the wallclock was reduced to simulate a forecast failure.  Upon resubmission, the forecast was restarted from the last available restart set.  The ensemble members (including `mem000`), restarted as expected (note the waves test exception below).
  - C48_S2SWA_gefs had similar issues as C48_S2SW with waves.
  - C48_S2SA_gefs test was able to restart aerosols.

Restarting with aerosols exposed a failure in the script after the model finished integration.  The gocart instantaneous files cannot be linked to `COM` as is documented in #1934.  Due to this, when a restart occurs, the AOD instantaneous files **before** the failure are in the `DATA` directory of the failure.  However, **after** restart, only the AOD files from the date of the restart are in the resubmitted job's `DATA` directory.  When the model finishes, the script tries to copy the gocart AOD files back to `COM`.  It is at this time, the AOD files before the restart are not available and the job fails.
Note that this is not an issue if the model succeeds on the first attempt.

It is therefore urgent to be able to write the GOCART AOD instantaneous files to a directory outside of the `DATA` directory.  The issue #1934 must be resolved if aerosols are going to be used in GEFS forecasts. 
 
# Checklist


parm/config/gefs/config.wave
```diff
@@ -121,15 +118,15 @@ export RSTIOFF_WAV=0                   # first restart file offset relative to m
...
-if [[ ${RUNMEM} = -1 ]]; then
+if (( RUNMEM == -1 )); then
 # No suffix added to model ID in case of deterministic run
-  export waveMEMB=
+  export waveMEMB=""
 else
 # Extract member number only
-  export waveMEMB="${RUNMEM: -2}"
+  export waveMEMB="${RUNMEM}"
 fi
```
WalterKolczynski-NOAA, line 130: Do we need this section at all then, if we are just making it `-1`?

parm/ufs/fv3/diag_table
```diff
@@ -1,80 +1,80 @@
 "fv3_history",    0,  "hours",  1,  "hours",  "time"
 "fv3_history2d",  0,  "hours",  1,  "hours",  "time"
-"ocn%4yr%2mo%2dy%2hr", @[FHOUT_OCNICE],  "hours",  1,  "hours",  "time",  @[FHOUT_OCNICE],  "hours",  "@[SYEAR] @[SMONTH] @[SDAY] @[CHOUR] 0 0"
-"ocn_daily%4yr%2mo%2dy",  1,  "days",   1,  "days",   "time",  1,  "days",   "@[SYEAR] @[SMONTH] @[SDAY] @[CHOUR] 0 0"
+"@[MOM6_OUTPUT_DIR]/ocn%4yr%2mo%2dy%2hr", @[FHOUT_OCNICE],  "hours",  1,  "hours",  "time",  @[FHOUT_OCNICE],  "hours",  "@[SYEAR] @[SMONTH] @[SDAY] @[CHOUR] 0 0"
```
WalterKolczynski-NOAA, line 3: Can we add this to the atmospheric fields as well? Both for uniformity and so we can introduce some more order to the run directory.

ush/forecast_det.sh
```diff
@@ -1,122 +1,123 @@
...
+    # So we have to continue checking for restarts
   fi
 
-  #-------------------------------------------------------
-  # determine if restart IC exists to continue from a previous forecast run attempt
+  # Lets assume this is was not run before and hence this is not a RERUN
+  RERUN="NO"
+
+  # RERUN is only available for RUN=gfs|gefs It is not available for RUN=enkfgdas|enkfgfs
+  if [[ "${RUN}" == "enkfgdas" ]] || [[ "${RUN}" == "enkfgfs" ]]; then
```
WalterKolczynski-NOAA:
```suggestion
  # RERUN is only available for RUN=gfs|gefs It is not available for RUN=gdas|enkfgdas|enkfgfs
  if [[ "${RUN}" =~ "gdas" ]] || [[ "${RUN}" == "enkfgfs" ]]; then
```

ush/parsing_namelists_FV3.sh
```diff
@@ -434,7 +428,7 @@ if [[ ${DO_CA:-"NO"} = "YES" ]]; then
 EOF
 fi
 
-if [[ ${DO_LAND_PERT:-"NO"} = "YES" ]]; then
+if [[ "${DO_LAND_PERT:-}" == "YES" ]]; then
```
WalterKolczynski-NOAA: This change should be undone. If `DO_LAND_PERT` is undefined, this will have the correct logic but print a syntax error in the log about `==` not being a unary operator because there is nothing on the left side of the operator.

ush/forecast_postdet.sh
```diff
@@ -1,738 +1,414 @@
...
-      ${NCP} "${COM_WAVE_PREP}/${COMPONENTwave}.mod_def.${wavGRD}" "${DATA}/mod_def.${wavGRD}"
+  # Copy restarts in the assimilation window for RUN=gdas|enkfgdas|enkfgfs
+  if [[ "${RUN}" =~ "gdas" || "${RUN}" == "enkfgfs" ]]; then
+    local restart_date
+    restart_date=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} + ${restart_interval} hours" +%Y%m%d%H)
+    while (( restart_date < forecast_end_cycle )); do
+      echo "Copying FV3 restarts for 'RUN=${RUN}' at ${restart_date}"
+      for fv3_restart_file in "${fv3_restart_files[@]}"; do
+        restart_file="${restart_date:0:8}.${restart_date:8:2}0000.${fv3_restart_file}"
+        ${NCP} "${DATArestart}/FV3_RESTART/${restart_file}" \
```
WalterKolczynski-NOAA, line 235: We should use `cpfs` to copy these to COM (see #2349). Maybe later we'll changes `$NCP` to that, but for now:
```suggestion
        cpfs "${DATArestart}/FV3_RESTART/${restart_file}" \
```

WalterKolczynski-NOAA (changes requested):
I still need to go through all of predet & postdet, but here are a few things.

parm/config/gefs/config.wave
aerorahul: I couldn't find where `RUNMEM` is set before line 124. 
If `RUNMEM` will always be `-1` for `gefs` then I agree this can be consolidated.

parm/ufs/fv3/diag_table
aerorahul: No can't do.
I have opened an issue in fv3atm repository to add that feature. See https://github.com/NOAA-EMC/fv3atm/issues/822

ush/parsing_namelists_FV3.sh
aerorahul:
```suggestion
if [[ "${DO_LAND_PERT:-NO}" == "YES" ]]; then
```

ush/forecast_postdet.sh
```diff
@@ -1,738 +1,414 @@
...
-
-    FHINC=${FHINCP_WAV}
-    fhr=$((fhr+FHINC))
+    fhr=$((fhr + fhinc))
   done
 }
 
 WW3_nml() {
   echo "SUB ${FUNCNAME[0]}: Copying input files for WW3"
-  WAV_MOD_TAG=${RUN}wave${waveMEMB}
```
JessicaMeixner-NOAA, line 654: @aerorahul was this moved somewhere else?

aerorahul: `forecast_predet.sh` in the function `WW3_predet`, lines 512-526

JessicaMeixner-NOAA: Thanks, sorry I couldn't find them before!

ush/forecast_postdet.sh
```diff
@@ -827,27 +469,71 @@ MOM6_out() {
...
+    echo "Copying MOM6 restarts for 'RUN=${RUN}' at ${forecast_end_cycle}"
+    for mom6_restart_file in "${mom6_restart_files[@]}"; do
+      restart_file="${forecast_end_cycle:0:8}.${forecast_end_cycle:8:2}0000.${mom6_restart_file}"
+      ${NCP} "${DATArestart}/MOM6_RESTART/${restart_file}" \
+             "${COM_OCEAN_RESTART}/${restart_file}"
+    done
+  fi
+
+  # Copy restarts at the beginning/middle of the next assimilation cycle to COM for RUN=gdas|enkfgdas|enkfgfs
+  if [[ "${RUN}" =~ "gdas" || "${RUN}" == "enkfgfs" ]]; then
```
JessicaMeixner-NOAA, line 499: @aerorahul so this is a bit forward looking and we can take care of this seperately if neede be.  But I have an update to this section of code from the original section: 

image


Which was for the case when we start with a cold half-cycle w/out IAU and then use IAU the next full cycle.  To do this, we need to copy more restarts here.  I put the changes I made above (I haven't made a PR b/c the WCDA test has not worked in develop, but was planning to soon when I could re-test my changes).  I believe I copied what the atm model was doing in this case, but I see the atm model has changed.   So I'm wondering if this block should also change, but match closer to what the atm model is doing, which might be a few less restarts than the code updates  I have here.

JessicaMeixner-NOAA (commented):
@aerorahul I've done my best to look over things.  Thanks for your answers to your questions.  I have no objections as long as tests pass.  I have some work to do on the wave-side to get additional updates ready for waves.

ush/forecast_postdet.sh
aerorahul: If you don't mind, I would like to defer this to a PR dedicated to this in itself.

ush/parsing_namelists_FV3.sh
aerorahul: done.

parm/config/gefs/config.wave
aerorahul: @JessicaMeixner-NOAA may remember why this was in the first place.

ush/forecast_postdet.sh
aerorahul: What I did in this PR is collected the `restart_date` and `restart_dir` based on `RERUN` condition, composed a list of initial conditions (and increments) based on `DOIAU`, then copy those files as part of `postdet`.

I think we can address this in a separate issue when the AOWCDA test is back up and running.  This will make it easier to insert the additional files in the right place.

ush/forecast_det.sh
```diff
@@ -1,122 +1,124 @@
...
+    # So we have to continue checking for restarts
   fi
 
-  #-------------------------------------------------------
-  # determine if restart IC exists to continue from a previous forecast run attempt
+  # Lets assume this is was not run before and hence this is not a RERUN
+  RERUN="NO"
+
+  # RERUN is only available for RUN=gfs|gefs It is not available for RUN=enkfgdas|enkfgfs
+  # RERUN is only available for RUN=gfs|gefs It is not available for RUN=gdas|enkfgdas|enkfgfs
```
WalterKolczynski-NOAA:
```suggestion
  # RERUN is only available for RUN=gfs|gefs It is not available for RUN=gdas|enkfgdas|enkfgfs
```

WalterKolczynski-NOAA (approved):

parm/config/gefs/config.wave
JessicaMeixner-NOAA: I'm assuming this was here as the original attempt to unify GEFSv12 and GFSv16 scripts for waves.  They were separated fairly early on in the process but hopefully will get unified as much as possible with GEFSv13 and GFSv17.

PR #2523, opened 2024-04-23
Hotfix to disable STALLED in CI as an error

TerrenceMcGuinness-NOAA:
# Description

This **_hotfix_** to the last CI updates disables the STALLED feature by not flagging it as an error until a more succinct algorithm  is determined (i.e. the STALLED state will be at par with RUNNING).

A second minor bug is in the event that no error log is produced when a Case fails, the error of the fail is still reported to the general user in GitHub.  (Jenkins enabled users can always see the full comprehensive state of system by following the links at the bottom of the PR).

# Type of change

# Change characteristics

# How has this been tested?
Being this is a CI update the test is running the CI on itself.


aerorahul (approved):
looks good.

PR #2536, opened 2024-04-25
Hot fix for bash CI on WCOSS2

TerrenceMcGuinness-NOAA:
# Description

A couple of minor hotfixes need for bash CI to work on WCOSS2.
PR tested in bash on WCOSS with these changes


WalterKolczynski-NOAA (approved):
Hopefully we can figure out why `-j` isn't working, but fine as a hotfix.

PR #2548, opened 2024-04-27
Launch Multiple Platforms to Jenkins with polling

TerrenceMcGuinness-NOAA:
# Description

When launching Jenkins CI Tests all requested RDHPCS machines can now be selected via the Ready label at once.
This is all that is required sense polling also works now on the controller end.

Killing the jobs still needs to be done directly on the Jenkins Controller.
Do not try to update the CI process using Labels.  It is for launching only.
Jenkins will update the labels as the states change.

NOTE: When a case fails the label will be update to a **Fail** for that system but will continue to report failures of subsequent cases until otherwise stopped in the Jenkins Controller directly.

Do not update any labels to **Ready** until a push has been made to the repo of the PR, and
The job has been completed or stopped directly on the controller first.
You can do this at anytime without having to wait until the other machines are completed.

# Type of change

# Change characteristics

# How has this been tested?
This PR will be self tested by running Jenkins on it with three labels at once


WalterKolczynski-NOAA (approved):
Excellent

ci/Jenkinsfile
```diff
@@ -229,7 +258,7 @@ pipeline {
                 }
             }
         }
-        stage( 'FINALIZE' ) {
+        stage( '5. FINALIZE' ) {
```
aerorahul, line 261:
```suggestion
        stage( '5. Finalize' ) {
```

aerorahul (approved):
looks good.

WalterKolczynski-NOAA:
Seem like this PR is successfully doing what it set out to do and any remaining issues on MSU machines are unrelated.

PR #2559, opened 2024-05-01
Add ability to use GEFS replay ICs

NeilBarton-NOAA:
# Description
The PR allows the use of ICs from PSL's replay analysis. These replay ICs will be used for GEFS reforecasting and SFS. Two main changes are associated with these updates: (1) replay ICs being valid at 3Z, and (2) the use of warm starts. 

This addresses https://github.com/NOAA-EMC/global-workflow/issues/1838 

# Type of change

- New feature 

# Change characteristics
- Is this a breaking change (a change in existing functionality)? NO
- Does this change require a documentation update? NO

# How has this been tested?
- GEFS Forecast-only on Hera using replay ICs
- GEFS Forecast-only on Hera using warm start ICs
- GEFS Forecast-only on Hera using cold start ICs
- GFS Forecast-only on Hera CI test

# Checklist
- [X] Any dependent changes have been merged and published
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] I have made corresponding changes to the documentation if necessary

Issue #1776, opened 2023-08-04
Enable IAU across components in the coupled model


CatherineThomas-NOAA:
**Description**
IAU is currently used in the operational GFS as a form of initialization which incorporates increments gradually.  Current testing with the coupled model does not include IAU due to difficulties with defining consistent model times and restart intervals (e.g. [PR#1700](https://github.com/NOAA-EMC/global-workflow/pull/1700)) among components using varying definitions of IAU (4D vs 3D vs none).  The desired configuration (taking into consideration current component model capabilities) would have 4DIAU for the atmosphere, 3DIAU for the ocean, and no IAU for the sea ice and waves.  

This issue is a catch-all place to capture various configurations and details needed to enable IAU across components.

**Requirements**
4DIAU is currently operational in GFSv16 for the atmosphere and is a requirement.
3DIAU for the ocean is greatly desired, but not a strict requirement depending on the level of effort.

CatherineThomas-NOAA:
@junwang-noaa:  
@guillaumevernieres has offered to provide settings and dummy increments for the ocean IAU.

junwang-noaa:
@CatherineThomas-NOAA Thanks. @guillaumevernieres please let me know if you have the ocean IAU dummy increments ready. We can try to set up a coupled case with IAU turned on for both atm and ocean.

aerorahul:
> @CatherineThomas-NOAA Thanks. @guillaumevernieres please let me know if you have the ocean IAU dummy increments ready. We can try to set up a coupled case with IAU turned on for both atm and ocean.

@junwang-noaa I have a zero increment file that @guillaumevernieres provided at 5 degree ocean resolution.  He may have 0.25 degree increment file to contribute.

guillaumevernieres:
> @CatherineThomas-NOAA Thanks. @guillaumevernieres please let me know if you have the ocean IAU dummy increments ready. We can try to set up a coupled case with IAU turned on for both atm and ocean.

What resolution do you need @junwang-noaa ? We have 5deg and 1/4 deg increments post processed to be read by mom6's iau.
The settings to use mom6 iau are in the workflow.

guillaumevernieres:
@junwang-noaa :
5deg and 1/4 deg increments saved for posterity here: `/scratch2/NCEPDEV/ocean/Guillaume.Vernieres/incr4jun`

junwang-noaa:
@guillaumevernieres The coupled IAU test with error:

 960:
[...]

PR #2560, opened 2024-05-01
Updates for cold start half cycle, then continuing with IAU for WCDA

### LINKED ISSUE #2546: Be able to run S2S cold start w/out IAU for first half-cycle, warm start+IAU afterwards

We are working to set-up some C384mx025 S2S cycled experiments.  We plan to cold-start the first half cycle, which will not have IAU, and then use IAU after that.    

Right now, if DOIAU is no, when running the cold start half cycle for C384, the forecast failes in the final stages because there are only 3 mom_res${num}.nc restart files, and the scripts assume 4: https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/forecast_postdet.sh#L382-L384 

Moreover, the first half cycle only copies over 1 restart file for MOM6, and then you do not have restarts to start with IAU for the next cycle.  Changes similar to those made here: https://github.com/JessicaMeixner-NOAA/global-workflow/commit/94e7fc7033c88f6b51b45f4a02fe4fb3d69bc87a except will need to be refactored to match recent updates in the develop branch. 

Be able to run C384mx025 S2S cycled experiment with first half cycle being a cold start without IAU and then continue running with IAU. 

### Acceptance Criteria

Be able to run C384mx025 S2S cycled experiment with first half cycle being a cold start without IAU and then continue running with IAU. 

Need to figure out why there are only 3 MOM6 restarts for the run I'm working on currently and not 4, or add flexibility in the scripts to grab any of the restarts if we do not a-priori know how many & copy additional restarts to COM in the first half cycle, in the same way the atm model does.     There's been lots of updates since I've last successfully done this while waiting for features to be merged, so there might be other issues as well.

JessicaMeixner-NOAA:
# Description

This PR allows us to run C384 S2S with IAU, but starting with the first half-cycle as a cold-start.   This will be necessary for cycled testing as we build towards the full system for GFSv17.  This updates the copying of the restarts for RUN=gdas for both ocean and ice copying what the atm model is doing.  It also reduced the amount of restart files from 4 to 3. 

Refs #2546

Other updates: 
* Add DOJEDI ocean triggers for archiving certain files update from:  @CatherineThomas-NOAA 
* Adds COPY_FINAL_RESTARTS option to turn on/off copying the last restart file to COM.  Defaults to off... 
* Defines model_start_date_current_cycle & model_start_date_next_cycle to help with knowing which IC to grab. 

# Type of change

# Change characteristics

# How has this been tested?

This was tested by running C384 WCDA test on hera which can be found here: 
/scratch1/NCEPDEV/climate/Jessica.Meixner/cycling/iau_04/C384iaucold01 
which includes yaml case files and then in cold01 has COM/EXPDIRs

# Checklist


JessicaMeixner-NOAA:
So I still think this PR should move forward as is, but I'm going to change resolve to ref for the issue, because I've noticed that the gfsfcst is actually failing when copying restarts after the forecast is over.  The scripts are expecting a forecast hour that MOM6 is not producing.  That issue is tracked in #1776

ush/forecast_postdet.sh
```diff
@@ -500,22 +500,20 @@ MOM6_out() {
...
-             "${COM_OCEAN_RESTART}/${restart_file}"
+    restart_date=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} + ${restart_interval} hours" +%Y%m%d%H)
+    while (( restart_date < forecast_end_cycle )); do
+      echo "Copying MOM6 restarts for 'RUN=${RUN}' at ${restart_date}"
+      for mom6_restart_file in "${mom6_restart_files[@]}"; do
+	restart_file="${restart_date:0:8}.${restart_date:8:2}0000.${mom6_restart_file}"
+        ${NCP} "${DATArestart}/MOM6_RESTART/${restart_file}" \
+               "${COM_OCEAN_RESTART}/${restart_file}"
+      done
+      restart_date=$(date --utc -d "${restart_date:0:8} ${restart_date:8:2} + ${restart_interval} hours" +%Y%m%d%H)
```
aerorahul: If I understand this correctly, `COM` will contain 2 pairs of restarts; one at the beginning and another at the end of the assimilation window regardless of `DOIAU`.
Is there a way to separate that, such that only the neccessary restart is copied out to `COM`.
If `DOIAU=NO`, it will always be the middle of the window.
If `DOIAU=YES`, it will always be the beginning of the window.

I can see the previous logic was deficient in handling `DOIAU=YES` for the experiment that has a cold-start half cycle.

JessicaMeixner-NOAA: I'm not sure what another solution would be.  Is there a variable that exists that indicates that it's the first half cycle?  What is here is using what the atmosphere model is doing for gdas, but if there's another solution I can implement that here instead.

aerorahul: JEDI-based atmosphere DA requires restarts in the assimilation window, and that is why they are being copied to `COM`. 

A variable to indicate the first half cycle is a good idea to explore.

JessicaMeixner-NOAA: What about: 

https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/forecast_det.sh#L30-L32 These lines: 
```
    # Since warm start is false, we cannot do IAU
    DOIAU="NO"
    IAU_OFFSET=0
```
to 
```
    # Since warm start is false, we cannot do IAU - but set a variable to know the next cycle will have IAU
    if [[ "${DOIAU}" == "YES" ]]; then 
          NEXTCYCIAU=YES
    else 
          NEXTCYCIAU=NO 
    fi 
    DOIAU="NO"
    IAU_OFFSET=0
```
And then we can use "NEXTCYCIAU" in the copying to know which restart to copy?

JessicaMeixner-NOAA: I slightly updated this, but am using a "NEXTCYCIAU trigger for which restarts to copy for CICE, MOM6 and CMEPS (someday can be used for ww3 too).    This would reduce the amount of restarts copied.  I'll push updates after testing.  I've marked the PR as draft to resolve this issue.

JessicaMeixner-NOAA: @aerorahul trying something to see if it works.  Let me know what you think, we can try something else.  This was inspired by your ideas you shared offline, but might not have been exactly what you were originally thinking so we can tweak as needed.

ush/forecast_predet.sh
```diff
@@ -52,9 +52,13 @@ common_predet(){
   previous_cycle=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} - ${assim_freq} hours" +%Y%m%d%H)
   next_cycle=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} + ${assim_freq} hours" +%Y%m%d%H)
   current_cycle_begin=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} - ${half_window} hours" +%Y%m%d%H)
-  current_cycle_end=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} + ${half_window} hours" +%Y%m%d%H)
-  next_cycle_begin=$(date --utc -d "${next_cycle:0:8} ${next_cycle:8:2} - ${half_window} hours" +%Y%m%d%H)
-  next_cycle_end=$(date --utc -d "${next_cycle:0:8} ${next_cycle:8:2} + ${half_window} hours" +%Y%m%d%H)
+  if [[ "${DOIAU:-}" == "YES" ]]; then
```
JessicaMeixner-NOAA, line 58: @aerorahul my test is still going, but this seems to be working as expected with this fix.  If current_cycle_begin and next_cycle_begin should not change definitions, we could add a variable here perhaps startdate_current_cycle startdate_next_cycle that could be defined and updated as current_cycle_begin/next_cycle_begin are being defined here?

aerorahul: I am working on a patch that will:
1. have the ability to control if we want to save last time restarts
2. save restarts at the appropriate time of next cycle when doing IAU, but the experiments start  with a cold start

JessicaMeixner-NOAA: What are the issues with the solution (or alternative) shared here?

aerorahul: I don't believe this is correct.
Regardless of DOIAU, current_cycle_begin is the time of the beginning of the assimilation window. If it makes it more clear to call this current_cycle_window_begin, then we can rename w/ `window` in the word.
current_cycle and next_cycle are CDATE and CDATE+6h.  Those are fixed.

JessicaMeixner-NOAA: So what I understand is that you would like for the variables  current_cycle_begin to not be changed.   In this case, I can add the new variables: startdate_current_cycle startdate_next_cycle or a name you would prefer that would allow for current_cycle_begin/next_cycle_begin to remain unchanged.    Is startdate_current_cycle startdate_next_cycle  acceptable names?

JessicaMeixner-NOAA: I've made an update that hopefully addresses your concerns.  Please let me know if there are other outstanding concerns I can address.

aerorahul: Yes. Thank you.   
`sdate_current_cycle` is the `model_start_date_current_cycle` and `sdate_next_cycle` is the `model_start_date_next_cycle`. And this is captured here rather than the many places where `DOIAU` is relevant.

JessicaMeixner-NOAA: Awesome!  Would you prefer the names be model_start_date_current_cycle and model_start_date_next_cycle? If so I can update that on Monday!

aerorahul: I will leave it to you; a descriptive name is useful.

ush/forecast_predet.sh
```diff
@@ -54,7 +54,15 @@ common_predet(){
   current_cycle_begin=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} - ${half_window} hours" +%Y%m%d%H)
   current_cycle_end=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} + ${half_window} hours" +%Y%m%d%H)
   next_cycle_begin=$(date --utc -d "${next_cycle:0:8} ${next_cycle:8:2} - ${half_window} hours" +%Y%m%d%H)
-  next_cycle_end=$(date --utc -d "${next_cycle:0:8} ${next_cycle:8:2} + ${half_window} hours" +%Y%m%d%H)
+  forecast_end_cycle=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} + ${FHMAX} hours" +%Y%m%d%H)
+  #Define sdate for current_cycle and next_cycle as the time the forecast will start for this cycle and next cycle
+  if [[ "${DOIAU:-}" == "YES" ]]; then
+    sdate_current_cycle=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} - ${half_window} hours" +%Y%m%d%H)
+    sdate_next_cycle=$(date --utc -d "${next_cycle:0:8} ${next_cycle:8:2} - ${half_window} hours" +%Y%m%d%H)
```
aerorahul: These are calculated on the few lines above and it makes casts `model_start_date_current_cycle` and `model_start_date_next_cycle` in relation to the anchors above.
```suggestion
    sdate_current_cycle="${current_cycle_begin}"
    sdate_next_cycle="${next_cycle_begin}"
```

ush/forecast_predet.sh
JessicaMeixner-NOAA: I updated the name so it's more descriptive.

ush/forecast_predet.sh
```diff
@@ -54,7 +54,15 @@ common_predet(){
   current_cycle_begin=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} - ${half_window} hours" +%Y%m%d%H)
   current_cycle_end=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} + ${half_window} hours" +%Y%m%d%H)
   next_cycle_begin=$(date --utc -d "${next_cycle:0:8} ${next_cycle:8:2} - ${half_window} hours" +%Y%m%d%H)
-  next_cycle_end=$(date --utc -d "${next_cycle:0:8} ${next_cycle:8:2} + ${half_window} hours" +%Y%m%d%H)
+  forecast_end_cycle=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} + ${FHMAX} hours" +%Y%m%d%H)
```
aerorahul:
```suggestion
```
Repeated on line 66.

ush/forecast_predet.sh
aerorahul: Thanks for cleaning this up.  It looks great.

aerorahul (approved):
This looks great. Thank you!

JessicaMeixner-NOAA:
@aerorahul I have marked this as ready to review.  We'll do a separate PR to add in some updates @guillaumevernieres is working which is one reason this was still review.

JessicaMeixner-NOAA:
@aerorahul I just updated the description and remembered that i hadn't added the variable for GEFS, so also added that.

aerorahul:
@NeilBarton-NOAA 
Some of these updates will be relevant for the PR #2559

scripts/exglobal_archive.sh
```diff
@@ -237,7 +237,10 @@ if [[ ${HPSSARCH} = "YES" || ${LOCALARCH} = "YES" ]]; then
 
         #gdasocean
         if [ "${DO_OCN}" = "YES" ]; then
-            targrp_list="${targrp_list} gdasocean gdasocean_analysis"
+            targrp_list="${targrp_list} gdasocean"	  
+            if [ "${DO_JEDIOCNVAR}" = "YES" ]; then
```
aerorahul:
```suggestion
            if [[ "${DO_JEDIOCNVAR}" == "YES" ]]; then
```

aerorahul: This should resolve the shellnorm failure.

JessicaMeixner-NOAA:
@aerorahul I just wanted to chime in here to say that my test of the  C384mx025 3DVarAOWCDA with a fresh clone is running smoothly as well.

guillaumevernieres (approved):
Thanks @JessicaMeixner-NOAA . Tested over a couple of days of cycling.

Issue #1878, opened 2023-09-22
Product generation for GEFSv13 reforecast 


EricSinsky-NOAA:
### What new functionality do you need?

There are a subset of variables that need to be saved on WCOSS2 for the GEFSv13 reforecast.  These variables are needed by GEFSv13 reforecast major stakeholders. Functionalities that are needed include variable extraction from netcdf and grib2 files, daily-averaging, and data compression. 

### What are the requirements for the new functionality?

The wgrib2 utility is needed for data extraction from grib2 files, and to perform daily-averaging for a few GEFS upper atmosphere variables. The ncks utility is needed for data extraction from netcdf files. The bzip2 tool is needed to perform the data compression. For GEFS atmos, pgrb data in 0.25, 0.5 and 1.0 degree are needed as inputs. For GEFS ocean and ice, either the native resolution data or remapped 1 degree data are needed as inputs. GEFS wave and aerosol inputs will be needed as well. 

### Acceptance Criteria

This issue is considered complete when all stakeholder-requested variables for the GEFSv13 reforecast can be extracted, daily-averaged and compressed in a reasonable amount of time, and saved on WCOSS2.  

### Suggest a solution (optional)

In an effort to address this issue, a preliminary ex-script has been created to extract the stakeholder-requested variables from the atmosphere, ocean and ice model output. This preliminary script is similar to the script that was used for the the GEFSv12 reforecast. In this preliminary ex-script, extraction is performed using the wgrib2 utility for grib2 files and ncks for netcdf files. Furthermore, the daily-averaging in this script is performed using the wgrib2 utility, and compression is performed using bzip2.

DavidHuber-NOAA:
@EricSinsky-NOAA Was this resolved by https://github.com/NOAA-EMC/global-workflow/pull/2680 or is there still more work to be done?

EricSinsky-NOAA:
This task has not been tested successfully yet when `REPLAY_ICS=YES`. There are PRs that will allow this task to run successfully when  `REPLAY_ICS=YES` (PR #2778, PR #2755). There are issues already associated with those PRs, so I think this issue can be closed as complete.

DavidHuber-NOAA:
Sounds good, thanks @EricSinsky-NOAA.  Closing.

PR #2491, opened 2024-04-15
Refactored archiving

DavidHuber-NOAA:
# Description
This provides a new pygfs task, archive.py, that provides all of the tools necessary to archive data to the local (`ARCDIR`) and backup (`ATARDIR`) archive directories.  YAML-Jinja2 templates are provided to define the file to be archived or tarred to replace the `hpssarch_gen.sh`, `exglobal_earc`, and `exglobal_archive.sh` scripts and make it easier to add new data and explicitly handle optional and required files.

For `ATARDIR` archiving, a master jinja template is provided for each `RUN` (i.e. master_gdas.yaml.j2, master_gfs.yaml.j2, master_enkf.yaml.j2).  The master_enkf.yaml.j2 template is used for both `enkfgdas` and `enkfgfs` `RUN`s.  These templates then include the appropriate `RUN`-specific jinja templates (e.g. gdas.yaml.j2) based on experiment, cycle, and coupled parameters.  Each of these templates corresponds to a single tarball to populate and are tabbed 4 spaces so they are defined within the master `datasets` dictionary.

Future developers should not have to make modifications to archive.py unless archiving is being enabled for a new `RUN` (e.g. `gefs`) and then only a single `elif` needs to be added to the configure method to specify the master `ATARDIR` template to archive (e.g. `master_gefs.yaml.j2`).  If a new component is coming online that needs to be archived to `ATARDIR` (e.g. SNOW), then create a new template for each `RUN` that it needs to be archived for (e.g. `gdassnow.yaml.j2`) and reference the template in the appropriate master templates, e.g.

`master_gdas.yaml:`
```jinja
{% if DO_SNOW %}
{% include "gdassnow.yaml.j2" %}
{% endif %}
```

A few other issues were addressed along the way:
1. Aerosols have been reenabled.  Aerosol forecasts should only be performed during gdas cycles, but analyses can be performed for both gfs and gdas cycles.  This was accomplished by setting separate `AERO_<job>_CDUMP` variables to parse on for both `ANL` and `FCST` jobs.
2. Fixed the name of the `cice6_rst_ok` variable in `forecast_det.sh`.  This prevented restarts from being used for cice-enable experiments.  This feature was not tested.
3. Create a temporary fix for the `wgrib` utility.  For spack-stack 1.6.0, the `grib-util` module file does not declare `WGRIB`.  An issue is open (https://github.com/JCSDA/spack-stack/issues/1097) to fix this in existing installations.  Once complete, this temporary fix should be removed.
[...]

Issue #2274, opened 2024-01-29
Need capability to run multiple forecast segments


WalterKolczynski-NOAA:
### What new functionality do you need?

Longer forecasts (such as those for GEFS S2S and SFS) are too long to run in a single job within wallclock limits and/or fit in different, heterogenous chunks in operations.

Depends on #2273
Depends on #2674

### What are the requirements for the new functionality?

The ability to run a forecast in multiple segments, which each segment continuing from the end of the previous one. At least two are needed, but allowing an arbitrary number is best (and may be necessary for SFS).

Stretch scope: allow different resources for each segment. This can be handled as a separate issue unless it makes sense to incorporate into the design here.

### Acceptance Criteria

- [ ] Ability to run a forecast in at least two segments.
- [ ] A new or updated CI test that runs the forecast in segments.

PR #2561, opened 2024-05-02
Add ability to process ocean/ice products specific to GEFS

EricSinsky-NOAA:
# Description

This PR begins to add the capability to produce GEFSv13 ocean and ice products in the global-workflow according to stakeholder requirements. The following features are added.

- An oceanice prod yaml file has been added to address the ocean and ice products specific to GEFSv13. 
- The rocoto dependencies and config.base for GEFS have also been modified to allow for 24-hour averaged ocean and ice output.
- Various scripts have been modified to allow for ocean and ice output frequencies of 24 hours. 
-  `FHOUT_OCNICE` has been split into two variables called `FHOUT_OCN` and `FHOUT_ICE`. The same has been done for `FHOUT_OCNICE_GFS`. 

Refs #1878 

# Type of change

# Change characteristics

# How has this been tested?

These changes have been tested using the EP5d ICs on WCOSS2. Additional testing will be performed using the replay ICs from the changes in PR #2559. 

# Checklist


parm/post/oceanice_products_gefs.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+ocean:
+  namelist:
+    ftype: "ocean"
+    maskvar: "temp"
+    sinvar: "sin_rot"
+    cosvar: "cos_rot"
+    angvar: ""
+    {% if model_grid == 'mx025' or model_grid == 'mx050' or model_grid == 'mx100' %}
+    ocean_levels: [5, 15, 25, 35, 45, 55, 65, 75, 85, 95, 105, 115, 125, 135, 145, 155, 165, 175, 185, 195, 205, 215, 225.86945, 241.06255, 266.5239, 308.7874, 373.9288, 467.3998, 593.87915, 757.1453, 959.97325, 1204.059, 1489.9735, 1817.1455, 2183.879, 2587.3995, 3023.9285, 3488.7875, 3976.524, 4481.0625]
```
GwenChen-NOAA: @EricSinsky-NOAA, ocean_levels will need to round up to integer. Users of grib2 files reported that they have problems plotting decimal depths in grads and plot_data_plane. So, I'm going to update it for GFS. Please change it to:

ocean_levels: [5, 15, 25, 35, 45, 55, 65, 75, 85, 95, 105, 115, 125, 135, 145, 155, 165, 175, 185, 195, 205, 215, 226, 241, 267, 309, 374, 467, 594, 757, 960, 1204, 1490, 1817, 2184, 2587, 3024, 3489, 3977, 4481]

parm/post/oceanice_products_gefs.yaml
```diff
@@ -0,0 +1,73 @@
...
+    {% elif model_grid == 'mx500' %}
+    ocean_levels: [5, 15, 25, 35, 45, 55, 65, 75, 85, 95, 105, 115, 125, 135, 145, 155, 165, 175, 185, 195, 205, 215, 225.86945, 241.06255, 266.5239]
```
GwenChen-NOAA: Same here, please use:

ocean_levels: [5, 15, 25, 35, 45, 55, 65, 75, 85, 95, 105, 115, 125, 135, 145, 155, 165, 175, 185, 195, 205, 215, 226, 241, 267]

EricSinsky-NOAA: @GwenChen-NOAA Thank you for letting me know about this issue regarding the decimal depths. I will make these changes to `parm/post/oceanice_products_gefs.yaml`.

WalterKolczynski-NOAA:
Do we really need to wait for #2559 for this? Seems independent.

EricSinsky-NOAA:
> Do we really need to wait for #2559 for this? Seems independent.
 
@WalterKolczynski-NOAA Waiting for PR [#2559](https://github.com/NOAA-EMC/global-workflow/pull/2559) is not required for this PR. However, further tests may be needed to make sure these changes are compatible with an f003 start. The current changes in this PR have been successfully tested with the EP5d ICs (f000 start). Should this PR be set to "ready for review", and then we can add any necessary changes in future PRs?

WalterKolczynski-NOAA:
> > Do we really need to wait for #2559 for this? Seems independent.
> 
> @WalterKolczynski-NOAA Waiting for PR [#2559](https://github.com/NOAA-EMC/global-workflow/pull/2559) is not required for this PR. However, further tests may be needed to make sure these changes are compatible with an f003 start. The current changes in this PR have been successfully tested with the EP5d ICs (f000 start). Should this PR be set to "ready for review", and then we can add any necessary changes in future PRs?

Yes please.

EricSinsky-NOAA:
Marking this PR as ready for review. Some scripts may need to be modified when PR #[2559](https://github.com/NOAA-EMC/global-workflow/pull/2559) is merged, but these changes can be addressed in future PRs.

WalterKolczynski-NOAA (approved):

aerorahul (approved):
thanks!

TerrenceMcGuinness-NOAA:
Case **C48_S2SWA** on **Orion** is stalled running in

`/work2/noaa/stmp/CI/ORION/2561 C48_S2SWA_gefs_c6c24a90`

```
(C48_S2SWA_gefs_c6c24a90 on Orion)

	Total Cycles: 1
	Number Cycles done: 0
	State: STALLED
```

WalterKolczynski-NOAA:
Experiments are stalling because the dependencies of the marine product are not being met:
```
hercules-login-2 C48_S2SWA_gefs_c6c24a90/>rocotostat -d C48_S2SWA_gefs_c6c24a90.db -w C48_S2SWA_gefs_c6c24a90.xml 
       CYCLE                    TASK                       JOBID               STATE         EXIT STATUS     TRIES      DURATION
================================================================================================================================
202103231200                stage_ic                    17867447           SUCCEEDED                   0         1          21.0
202103231200               wave_init                    17867448           SUCCEEDED                   0         1          35.0
202103231200             fcst_mem000                    17867510           SUCCEEDED                   0         1        3526.0
202103231200             fcst_mem001                    17867511           SUCCEEDED                   0         1        3577.0
202103231200             fcst_mem002                    17867512           SUCCEEDED                   0         1        3539.0
202103231200    atmos_prod_mem000_f000                    17867631           SUCCEEDED                   0         1          93.0
202103231200    atmos_prod_mem000_f006                    17867632           SUCCEEDED                   0         1          98.0
202103231200    atmos_prod_mem000_f012                    17867708           SUCCEEDED                   0         1          93.0
202103231200    atmos_prod_mem000_f018                    17867709           SUCCEEDED                   0         1          95.0
202103231200    atmos_prod_mem000_f024                    17867850           SUCCEEDED                   0         1          94.0
202103231200    atmos_prod_mem000_f030                    17867851           SUCCEEDED                   0         1          93.0
202103231200    atmos_prod_mem000_f036                    17867939           SUCCEEDED                   0         1          92.0
202103231200    atmos_prod_mem000_f042                    17867940           SUCCEEDED                   0         1          95.0
202103231200    atmos_prod_mem000_f048                    17868073           SUCCEEDED                   0         1          92.0
202103231200    atmos_prod_mem000_f054                    17868074           SUCCEEDED                   0         1          95.0
202103231200    atmos_prod_mem000_f060                    17868188           SUCCEEDED                   0         1          96.0
202103231200    atmos_prod_mem000_f066                    17868189           SUCCEEDED                   0         1          68.0
202103231200    atmos_prod_mem000_f072                    17868256           SUCCEEDED                   0         1          97.0
202103231200    atmos_prod_mem000_f078                    17868257           SUCCEEDED                   0         1          93.0
202103231200    atmos_prod_mem000_f084                    17868333           SUCCEEDED                   0         1          95.0
202103231200    atmos_prod_mem000_f090                    17868334           SUCCEEDED                   0         1          93.0
202103231200    atmos_prod_mem000_f096                    17868403           SUCCEEDED                   0         1          94.0
202103231200    atmos_prod_mem000_f102                    17868404           SUCCEEDED                   0         1          93.0
202103231200    atmos_prod_mem000_f108                    17868483           SUCCEEDED                   0         1          72.0
202103231200    atmos_prod_mem000_f114                    17868484           SUCCEEDED                   0         1          69.0
202103231200    atmos_prod_mem000_f120                    17868533           SUCCEEDED                   0         1          98.0
202103231200    atmos_prod_mem001_f000                    17867633           SUCCEEDED                   0         1          95.0
202103231200    atmos_prod_mem001_f006                    17867634           SUCCEEDED                   0         1          74.0
202103231200    atmos_prod_mem001_f012                    17867710           SUCCEEDED                   0         1          93.0
202103231200    atmos_prod_mem001_f018                    17867711           SUCCEEDED                   0         1          94.0
202103231200    atmos_prod_mem001_f024                    17867852           SUCCEEDED                   0         1          93.0
202103231200    atmos_prod_mem001_f030                    17867853           SUCCEEDED                   0         1          93.0
202103231200    atmos_prod_mem001_f036                    17867941           SUCCEEDED                   0         1          94.0
202103231200    atmos_prod_mem001_f042                    17867942           SUCCEEDED                   0         1          97.0
202103231200    atmos_prod_mem001_f048                    17868075           SUCCEEDED                   0         1          67.0
202103231200    atmos_prod_mem001_f054                    17868076           SUCCEEDED                   0         1          74.0
202103231200    atmos_prod_mem001_f060                    17868190           SUCCEEDED                   0         1          96.0
202103231200    atmos_prod_mem001_f066                    17868258           SUCCEEDED                   0         1          93.0
202103231200    atmos_prod_mem001_f072                    17868259           SUCCEEDED                   0         1          93.0
202103231200    atmos_prod_mem001_f078                    17868335           SUCCEEDED                   0         1          97.0
202103231200    atmos_prod_mem001_f084                    17868336           SUCCEEDED                   0         1          94.0
202103231200    atmos_prod_mem001_f090                    17868405           SUCCEEDED                   0         1          93.0
202103231200    atmos_prod_mem001_f096                    17868406           SUCCEEDED                   0         1          93.0
202103231200    atmos_prod_mem001_f102                    17868485           SUCCEEDED                   0         1          97.0
202103231200    atmos_prod_mem001_f108                    17868486           SUCCEEDED                   0         1          97.0
202103231200    atmos_prod_mem001_f114                    17868534           SUCCEEDED                   0         1          68.0
202103231200    atmos_prod_mem001_f120                    17868535           SUCCEEDED                   0         1          96.0
202103231200    atmos_prod_mem002_f000                    17867635           SUCCEEDED                   0         1          92.0
202103231200    atmos_prod_mem002_f006                    17867636           SUCCEEDED                   0         1          99.0
202103231200    atmos_prod_mem002_f012                    17867712           SUCCEEDED                   0         1          92.0
202103231200    atmos_prod_mem002_f018                    17867713           SUCCEEDED                   0         1          94.0
202103231200    atmos_prod_mem002_f024                    17867854           SUCCEEDED                   0         1          67.0
202103231200    atmos_prod_mem002_f030                    17867855           SUCCEEDED                   0         1          95.0
202103231200    atmos_prod_mem002_f036                    17867943           SUCCEEDED                   0         1          95.0
202103231200    atmos_prod_mem002_f042                    17867944           SUCCEEDED                   0         1          96.0
202103231200    atmos_prod_mem002_f048                    17868077           SUCCEEDED                   0         1          90.0
202103231200    atmos_prod_mem002_f054                    17868078           SUCCEEDED                   0         1          67.0
202103231200    atmos_prod_mem002_f060                    17868191           SUCCEEDED                   0         1          93.0
202103231200    atmos_prod_mem002_f066                    17868192           SUCCEEDED                   0         1          96.0
202103231200    atmos_prod_mem002_f072                    17868260           SUCCEEDED                   0         1          91.0
202103231200    atmos_prod_mem002_f078                    17868261           SUCCEEDED                   0         1          70.0
202103231200    atmos_prod_mem002_f084                    17868337           SUCCEEDED                   0         1          92.0
202103231200    atmos_prod_mem002_f090                    17868338           SUCCEEDED                   0         1          87.0
202103231200    atmos_prod_mem002_f096                    17868408           SUCCEEDED                   0         1          94.0
202103231200    atmos_prod_mem002_f102                    17868487           SUCCEEDED                   0         1          98.0
202103231200    atmos_prod_mem002_f108                    17868488           SUCCEEDED                   0         1          98.0
202103231200    atmos_prod_mem002_f114                    17868536           SUCCEEDED                   0         1          98.0
202103231200    atmos_prod_mem002_f120                    17868537           SUCCEEDED                   0         1          69.0
202103231200      atmos_ensstat_f000                    17867714           SUCCEEDED                   0         1          21.0
202103231200      atmos_ensstat_f006                    17867715           SUCCEEDED                   0         1          26.0
202103231200      atmos_ensstat_f012                    17867856           SUCCEEDED                   0         1          17.0
202103231200      atmos_ensstat_f018                    17867857           SUCCEEDED                   0         1          26.0
202103231200      atmos_ensstat_f024                    17867945           SUCCEEDED                   0         1          23.0
202103231200      atmos_ensstat_f030                    17867946           SUCCEEDED                   0         1          23.0
202103231200      atmos_ensstat_f036                    17868079           SUCCEEDED                   0         1          25.0
202103231200      atmos_ensstat_f042                    17868080           SUCCEEDED                   0         1          22.0
202103231200      atmos_ensstat_f048                    17868193           SUCCEEDED                   0         1          20.0
202103231200      atmos_ensstat_f054                    17868194           SUCCEEDED                   0         1          20.0
202103231200      atmos_ensstat_f060                    17868262           SUCCEEDED                   0         1          22.0
202103231200      atmos_ensstat_f066                    17868339           SUCCEEDED                   0         1          18.0
202103231200      atmos_ensstat_f072                    17868340           SUCCEEDED                   0         1          18.0
202103231200      atmos_ensstat_f078                    17868409           SUCCEEDED                   0         1          17.0
202103231200      atmos_ensstat_f084                    17868410           SUCCEEDED                   0         1          24.0
202103231200      atmos_ensstat_f090                    17868489           SUCCEEDED                   0         1          17.0
202103231200      atmos_ensstat_f096                    17868490           SUCCEEDED                   0         1          17.0
202103231200      atmos_ensstat_f102                    17868538           SUCCEEDED                   0         1          28.0
202103231200      atmos_ensstat_f108                    17868539           SUCCEEDED                   0         1          28.0
202103231200      atmos_ensstat_f114                    17868618           SUCCEEDED                   0         1          18.0
202103231200      atmos_ensstat_f120                    17868619           SUCCEEDED                   0         1          18.0
202103231200    ocean_prod_mem000_f006                           -                   -                   -         -             -
202103231200    ocean_prod_mem000_f012                           -                   -                   -         -             -
202103231200    ocean_prod_mem000_f018                           -                   -                   -         -             -
202103231200    ocean_prod_mem000_f024                    17867952           SUCCEEDED                   0         1          17.0
202103231200    ocean_prod_mem000_f030                           -                   -                   -         -             -
202103231200    ocean_prod_mem000_f036                           -                   -                   -         -             -
202103231200    ocean_prod_mem000_f042                           -                   -                   -         -             -
202103231200    ocean_prod_mem000_f048                    17868263           SUCCEEDED                   0         1          14.0
202103231200    ocean_prod_mem000_f054                           -                   -                   -         -             -
202103231200    ocean_prod_mem000_f060                           -                   -                   -         -             -
202103231200    ocean_prod_mem000_f066                           -                   -                   -         -             -
202103231200    ocean_prod_mem000_f072                    17868411           SUCCEEDED                   0         1          21.0
202103231200    ocean_prod_mem000_f078                           -                   -                   -         -             -
202103231200    ocean_prod_mem000_f084                           -                   -                   -         -             -
202103231200    ocean_prod_mem000_f090                           -                   -                   -         -             -
202103231200    ocean_prod_mem000_f096                    17868542           SUCCEEDED                   0         1          20.0
202103231200    ocean_prod_mem000_f102                           -                   -                   -         -             -
202103231200    ocean_prod_mem000_f108                           -                   -                   -         -             -
202103231200    ocean_prod_mem000_f114                           -                   -                   -         -             -
202103231200    ocean_prod_mem000_f120                    17868547           SUCCEEDED                   0         1          19.0
202103231200    ocean_prod_mem001_f006                           -                   -                   -         -             -
202103231200    ocean_prod_mem001_f012                           -                   -                   -         -             -
202103231200    ocean_prod_mem001_f018                           -                   -                   -         -             -
202103231200    ocean_prod_mem001_f024                    17868082           SUCCEEDED                   0         1          30.0
202103231200    ocean_prod_mem001_f030                           -                   -                   -         -             -
202103231200    ocean_prod_mem001_f036                           -                   -                   -         -             -
202103231200    ocean_prod_mem001_f042                           -                   -                   -         -             -
202103231200    ocean_prod_mem001_f048                    17868265           SUCCEEDED                   0         1          47.0
202103231200    ocean_prod_mem001_f054                           -                   -                   -         -             -
202103231200    ocean_prod_mem001_f060                           -                   -                   -         -             -
202103231200    ocean_prod_mem001_f066                           -                   -                   -         -             -
202103231200    ocean_prod_mem001_f072                    17868412           SUCCEEDED                   0         1          17.0
202103231200    ocean_prod_mem001_f078                           -                   -                   -         -             -
202103231200    ocean_prod_mem001_f084                           -                   -                   -         -             -
202103231200    ocean_prod_mem001_f090                           -                   -                   -         -             -
202103231200    ocean_prod_mem001_f096                    17868548           SUCCEEDED                   0         1          15.0
202103231200    ocean_prod_mem001_f102                           -                   -                   -         -             -
202103231200    ocean_prod_mem001_f108                           -                   -                   -         -             -
202103231200    ocean_prod_mem001_f114                           -                   -                   -         -             -
202103231200    ocean_prod_mem001_f120                    17868549           SUCCEEDED                   0         1          15.0
202103231200    ocean_prod_mem002_f006                           -                   -                   -         -             -
202103231200    ocean_prod_mem002_f012                           -                   -                   -         -             -
202103231200    ocean_prod_mem002_f018                           -                   -                   -         -             -
202103231200    ocean_prod_mem002_f024                    17867958           SUCCEEDED                   0         1          16.0
202103231200    ocean_prod_mem002_f030                           -                   -                   -         -             -
202103231200    ocean_prod_mem002_f036                           -                   -                   -         -             -
202103231200    ocean_prod_mem002_f042                           -                   -                   -         -             -
202103231200    ocean_prod_mem002_f048                    17868266           SUCCEEDED                   0         1          16.0
202103231200    ocean_prod_mem002_f054                           -                   -                   -         -             -
202103231200    ocean_prod_mem002_f060                           -                   -                   -         -             -
202103231200    ocean_prod_mem002_f066                           -                   -                   -         -             -
202103231200    ocean_prod_mem002_f072                    17868413           SUCCEEDED                   0         1          14.0
202103231200    ocean_prod_mem002_f078                           -                   -                   -         -             -
202103231200    ocean_prod_mem002_f084                           -                   -                   -         -             -
202103231200    ocean_prod_mem002_f090                           -                   -                   -         -             -
202103231200    ocean_prod_mem002_f096                    17868550           SUCCEEDED                   0         1          17.0
202103231200    ocean_prod_mem002_f102                           -                   -                   -         -             -
202103231200    ocean_prod_mem002_f108                           -                   -                   -         -             -
202103231200    ocean_prod_mem002_f114                           -                   -                   -         -             -
202103231200    ocean_prod_mem002_f120                    17868553           SUCCEEDED                   0         1          13.0
202103231200    ice_prod_mem000_f006                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f012                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f018                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f024                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f030                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f036                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f042                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f048                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f054                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f060                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f066                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f072                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f078                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f084                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f090                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f096                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f102                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f108                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f114                           -                   -                   -         -             -
202103231200    ice_prod_mem000_f120                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f006                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f012                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f018                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f024                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f030                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f036                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f042                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f048                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f054                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f060                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f066                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f072                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f078                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f084                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f090                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f096                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f102                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f108                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f114                           -                   -                   -         -             -
202103231200    ice_prod_mem001_f120                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f006                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f012                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f018                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f024                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f030                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f036                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f042                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f048                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f054                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f060                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f066                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f072                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f078                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f084                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f090                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f096                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f102                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f108                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f114                           -                   -                   -         -             -
202103231200    ice_prod_mem002_f120                           -                   -                   -         -             -
202103231200    wave_post_grid_mem000                    17867559           SUCCEEDED                   0         1        3245.0
202103231200    wave_post_grid_mem001                    17867560           SUCCEEDED                   0         1        3294.0
202103231200    wave_post_grid_mem002                    17867561           SUCCEEDED                   0         1        3255.0
202103231200    wave_post_pnt_mem000                    17868554           SUCCEEDED                   0         1        3091.0
202103231200    wave_post_pnt_mem001                    17868555           SUCCEEDED                   0         1        2902.0
202103231200    wave_post_pnt_mem002                    17868556           SUCCEEDED                   0         1        2865.0
```
```
hercules-login-2 C48_S2SWA_gefs_c6c24a90/>rocotocheck -d C48_S2SWA_gefs_c6c24a90.db -w C48_S2SWA_gefs_c6c24a90.xml -c all -t ice_prod_mem000_f024

Task: ice_prod_mem000_f024
  account: nems
  command: /work2/noaa/stmp/CI/ORION/2561/gefs/jobs/rocoto/oceanice_products.sh
  cores: 1
  cycledefs: gefs
  final: false
  jobname: C48_S2SWA_gefs_c6c24a90_ice_prod_mem000_f024_12
  join: /work2/noaa/stmp/CI/ORION/2561/RUNTESTS/COMROOT/C48_S2SWA_gefs_c6c24a90/logs/2021032312/ice_prod_mem000_f024.log
  maxtries: 2
  memory: 96GB
  metatasks: ice_prod,ice_prod_000
  name: ice_prod_mem000_f024
  nodes: 1:ppn=1:tpp=1
  partition: orion
  queue: batch
  seqnum: 1,4
  throttle: 9999999
  walltime: 00:15:00
  environment
    CDATE ==> 2021032312
    CDUMP ==> gefs
    COMPONENT ==> ice
    COMROOT ==> /work2/noaa/stmp/CI/ORION/2561/RUNTESTS/COMROOT
    DATAROOT ==> /work/noaa/stmp/mterry/RUNDIRS/C48_S2SWA_gefs_c6c24a90
    ENSMEM ==> 000
    EXPDIR ==> /work2/noaa/stmp/CI/ORION/2561/RUNTESTS/EXPDIR/C48_S2SWA_gefs_c6c24a90
    FHRLST ==> 024
    HOMEgfs ==> /work2/noaa/stmp/CI/ORION/2561/gefs
    MEMDIR ==> mem000
    NET ==> gefs
    PDY ==> 20210323
    RUN ==> gefs
    RUN_ENVIR ==> emc
    cyc ==> 12
  dependencies
    /work2/noaa/stmp/CI/ORION/2561/RUNTESTS/COMROOT/C48_S2SWA_gefs_c6c24a90/gefs.20210323/12/mem000/model_data/ice/history/gefs.ice.t12z.24hr_avg.f024.nc does not exist

Cycle: 202103231200
  Valid for this task: YES
  State: active
  Activated: 2024-05-09 00:53:15 UTC
  Completed: -
  Expired: -

Job: This task has not been submitted for this cycle

Task can not be submitted because:
  Dependencies are not satisfied
```
```
hercules-login-2 C48_S2SWA_gefs_c6c24a90/>rocotocheck -d C48_S2SWA_gefs_c6c24a90.db -w C48_S2SWA_gefs_c6c24a90.xml -c all -t ocean_prod_mem000_f036

Task: ocean_prod_mem000_f036
  account: nems
  command: /work2/noaa/stmp/CI/ORION/2561/gefs/jobs/rocoto/oceanice_products.sh
  cores: 1
  cycledefs: gefs
  final: false
  jobname: C48_S2SWA_gefs_c6c24a90_ocean_prod_mem000_f036_12
  join: /work2/noaa/stmp/CI/ORION/2561/RUNTESTS/COMROOT/C48_S2SWA_gefs_c6c24a90/logs/2021032312/ocean_prod_mem000_f036.log
  maxtries: 2
  memory: 96GB
  metatasks: ocean_prod,ocean_prod_000
  name: ocean_prod_mem000_f036
  nodes: 1:ppn=1:tpp=1
  partition: orion
  queue: batch
  seqnum: 1,6
  throttle: 9999999
  walltime: 00:15:00
  environment
    CDATE ==> 2021032312
    CDUMP ==> gefs
    COMPONENT ==> ocean
    COMROOT ==> /work2/noaa/stmp/CI/ORION/2561/RUNTESTS/COMROOT
    DATAROOT ==> /work/noaa/stmp/mterry/RUNDIRS/C48_S2SWA_gefs_c6c24a90
    ENSMEM ==> 000
    EXPDIR ==> /work2/noaa/stmp/CI/ORION/2561/RUNTESTS/EXPDIR/C48_S2SWA_gefs_c6c24a90
    FHRLST ==> 036
    HOMEgfs ==> /work2/noaa/stmp/CI/ORION/2561/gefs
    MEMDIR ==> mem000
    NET ==> gefs
    PDY ==> 20210323
    RUN ==> gefs
    RUN_ENVIR ==> emc
    cyc ==> 12
  dependencies
    AND is not satisfied
      /work2/noaa/stmp/CI/ORION/2561/RUNTESTS/COMROOT/C48_S2SWA_gefs_c6c24a90/gefs.20210323/12/mem000/model_data/ocean/history/gefs.ocean.t12z.24hr_avg.f036.nc does not exist

Cycle: 202103231200
  Valid for this task: YES
  State: active
  Activated: 2024-05-09 00:53:15 UTC
  Completed: -
  Expired: -

Job: This task has not been submitted for this cycle

Task can not be submitted because:
  Dependencies are not satisfied
```
If we are only producing 24-hr averages, shouldn't there only be jobs every 24-hr instead of every 6?

EricSinsky-NOAA:
> If we are only producing 24-hr averages, shouldn't there only be jobs every 24-hr instead of every 6

@WalterKolczynski-NOAA  Yes, for the ocean and ice prod jobs, there should be jobs every 24 hours instead of every 6. Atmos prod jobs are every 6 hours. To make the ocean and ice jobs every 24 hours in rocoto, is `workflow/rocoto/gefs_tasks.py` the correct file to modify?

EricSinsky-NOAA:
If variable `FHOUT_GFS` is adjusted from 6 to 24 in `parm/config/gefs/config.base`, the ocean and prod jobs are every 24 hours in rocoto. However, the atmos prod job also changes to every 24 hours, which we do not want. I am trying to figure out a way for variable `FHOUT_OCNICE_GFS` to make the ocean/ice prod rocoto task every 24 hours when this variable is set to 24.

EricSinsky-NOAA:
The last 3 commits should allow for the ocean and ice prod jobs to be every 24 hours instead of every 6 hours in rocoto, while keeping the atmos prod jobs every 6 hours. @WalterKolczynski-NOAA @aerorahul, when you get a chance, can you review these last 3 commits to make sure they may not cause any unforeseen issues in the workflow? Thank you.

workflow/rocoto/tasks.py
```diff
@@ -151,10 +150,6 @@ def _get_forecast_hours(cdump, config, component='atmos') -> List[str]:
             fhrs_hf = range(fhmin, fhmax_hf + fhout_hf, fhout_hf)
             fhrs = list(fhrs_hf) + list(range(fhrs_hf[-1] + fhout, fhmax + fhout, fhout))
 
-        # ocean/ice components do not have fhr 0 as they are averaged output
-        if component in ['ocean', 'ice']:
-            fhrs.remove(0)
-
```
WalterKolczynski-NOAA:
```suggestion
        # ocean/ice components do not have fhr 0 as they are averaged output
        if component in ['ocean', 'ice'] and 0 in fhrs:
            fhrs.remove(0)
```

workflow/rocoto/gefs_tasks.py
```diff
@@ -218,7 +218,7 @@ def _atmosoceaniceprod(self, component: str):
                      'log': f'{self.rotdir}/logs/@Y@m@d@H/{task_name}.log',
                      'maxtries': '&MAXTRIES;'}
 
-        fhrs = self._get_forecast_hours('gefs', self._configs[config])
+        fhrs = self._get_forecast_hours('gefs', self._configs[config], str(component))
```
WalterKolczynski-NOAA: Component is already a str:
```suggestion
        fhrs = self._get_forecast_hours('gefs', self._configs[config], component)
```

workflow/rocoto/tasks.py
```diff
@@ -133,7 +133,6 @@ def _get_forecast_hours(cdump, config, component='atmos') -> List[str]:
             local_config['FHMAX_HF_GFS'] = config['FHMAX_GFS']
             local_config['FHOUT_HF_GFS'] = config['FHOUT_OCNICE_GFS']
             local_config['FHOUT_GFS'] = config['FHOUT_OCNICE_GFS']
-            local_config['FHOUT'] = config['FHOUT_OCNICE']
```
WalterKolczynski-NOAA, line 136: This may cause issues with GDAS. Recommend defining an `FHOUT_OCNICE` to equal `FHOUT_OCNICE_GFS` like we do with `FHMAX`.

WalterKolczynski-NOAA (changes requested):

WalterKolczynski-NOAA:
After making the changes, please test the workflow generation on your own and make sure it is correct before we run it through CI again.

EricSinsky-NOAA:
A test case (gefs, C48, 2021032312, 120 lead hours, forecast only, cold start) is running now on WCOSS2 with the new commits. The rocoto jobs for ocean and ice prod are correctly every 24 hours instead of every 6 hours. The main issue that remains is that the ice_prod jobs are not running because the ice hist files from the forecast job are not being created. This is only an issue when forecast is not initialized at 00z, and setting FHOUT_OCNICE_GFS to 24.

workflow/rocoto/tasks.py
EricSinsky-NOAA: Done. I also added this condition to the same if statement in gefs_tasks.py.

workflow/rocoto/gefs_tasks.py
EricSinsky-NOAA: Done.

workflow/rocoto/tasks.py
EricSinsky-NOAA: I defined `FHOUT_OCNICE` and set it equal to `FHOUT_OCNICE_GFS`. For GDAS cases, I set `fhout` equal to `FHOUT_OCNICE`. This should avoid any issues with GDAS.

EricSinsky-NOAA:
@WalterKolczynski-NOAA Would it be appropriate to add a variable in `config.base` such as `DO_REFORECAST`, which can be set to `NO` by default? If `DO_REFORECAST` is set to `YES`, then `FHOUT_HF_GFS` can be set to `24` by default. If `DO_REFORECAST` is set to `NO`, then `FHOUT_HF_GFS` can be set to `6` by default. This daily-averaging requirement for ocean and ice is only needed for the reforecast, and the reforecast will never run at 12Z (at least for GEFSv13). The rocoto ice prod jobs stall only when FHOUT_HF_GFS is set to 24 and forecast initialization is not 00Z. Doing this will also avoid a CI test failure in the 12Z test cases.

GwenChen-NOAA:
> @WalterKolczynski-NOAA Would it be appropriate to add a variable in `config.base` such as `DO_REFORECAST`, which can be set to `NO` by default? If `DO_REFORECAST` is set to `YES`, then `FHOUT_HF_GFS` can be set to `24` by default. If `DO_REFORECAST` is set to `NO`, then `FHOUT_HF_GFS` can be set to `6` by default. This daily-averaging requirement for ocean and ice is only needed for the reforecast, and the reforecast will never run at 12Z (at least for GEFSv13). The rocoto ice prod jobs stall only when FHOUT_HF_GFS is set to 24 and forecast initialization is not 00Z. Doing this will also avoid a CI test failure in the 12Z test cases.

I think this is a good idea. Ocean and ice operational forecast files will be generated for 6-hr averages. The DO_REFORECAST option will allow distinction between generating reforecast and real-time forecast.

WalterKolczynski-NOAA:
> @WalterKolczynski-NOAA Would it be appropriate to add a variable in `config.base` such as `DO_REFORECAST`, which can be set to `NO` by default? If `DO_REFORECAST` is set to `YES`, then `FHOUT_HF_GFS` can be set to `24` by default. If `DO_REFORECAST` is set to `NO`, then `FHOUT_HF_GFS` can be set to `6` by default. This daily-averaging requirement for ocean and ice is only needed for the reforecast, and the reforecast will never run at 12Z (at least for GEFSv13). The rocoto ice prod jobs stall only when FHOUT_HF_GFS is set to 24 and forecast initialization is not 00Z. Doing this will also avoid a CI test failure in the 12Z test cases.

What are the actual requirements for (a) the reforecast (b) GEFS v13? Why are the daily average files that are already produced for GFS not sufficient for this (particularly if you are only worried about 00z initializations)?

EricSinsky-NOAA:
> Why are the daily average files that are already produced for GFS not sufficient for this (particularly if you are only worried about 00z initializations)?
- The daily-averaged files that are already produced for GFS would be sufficient. I know MOM6 already produces daily-averaged files even when `FHOUT_OCNICE_GFS` is set to 6, but I do not see daily-averaged files for CICE at the current configuration. Although 12Z is not needed for the reforecast, it does cause a CI test failure. If the CICE and MOM6 daily-averaged files are successfully produced, only the ocnice_prod yaml file will need to be modified for GEFS. 
- Also, if we only have oceanice prod jobs every 24 hours instead of every 6 hours, it would reduce the number of ocnice prod rocoto tasks in the reforecast, which would be especially beneficial for 48 day runs. 

> What are the actual requirements for (a) the reforecast (b) GEFS v13? 
- To my understanding, the requirements are not finalized for the operational GEFSv13 yet. Details on the reforecast requirements can be found in issue #1878. These specific reforecast requirements may increase the need for a `DO_REFORECAST` variable. 

Adding @bingfu-NOAA here for his awareness of this PR.

jiandewang:
@EricSinsky-NOAA "**MOM6 already produces daily-averaged files even when FHOUT_OCNICE_GFS is set to** 6": this ocean daily file only save several surface variables for quick sanity check purpose and is not being used in ocean post process. The ocean master files are the one that are being controlled by FHOUT_OCNICE_GFS for its output frequency and for post process.

aerorahul:
I think we need to define `FHOUT_OCEAN` and `FHOUT_ICE` and their `_GFS` counterparts to control ocean and ice output frequency independently.  We've been relying on `FHOUT` and `FHOUT_GFS` for all components and that needs to be changed.  The wave component in their development did the right thing in defining and using `FHOUT_WAV`.
Hopefully, this will allow the flexibility that the users need.

EricSinsky-NOAA:
> @EricSinsky-NOAA "**MOM6 already produces daily-averaged files even when FHOUT_OCNICE_GFS is set to** 6": this ocean daily file only save several surface variables for quick sanity check purpose and is not being used in ocean post process. The ocean master files are the one that are being controlled by FHOUT_OCNICE_GFS for its output frequency and for post process.

@jiandewang Thank you for this information on the ocean_daily files. 

> I think we need to define `FHOUT_OCEAN` and `FHOUT_ICE` and their `_GFS` counterparts to control ocean and ice output frequency independently. We've been relying on `FHOUT` and `FHOUT_GFS` for all components and that needs to be changed. The wave component in their development did the right thing in defining and using `FHOUT_WAV`. Hopefully, this will allow the flexibility that the users need.

@aerorahul We can definitely work on separating the ocean and ice fhout's. I agree this will allow for more flexibility and satisfy the needs of various users.

EricSinsky-NOAA:
> I think we need to define `FHOUT_OCEAN` and `FHOUT_ICE` and their `_GFS` counterparts to control ocean and ice output frequency independently. We've been relying on `FHOUT` and `FHOUT_GFS` for all components and that needs to be changed. The wave component in their development did the right thing in defining and using `FHOUT_WAV`. Hopefully, this will allow the flexibility that the users need.

@aerorahul `FHOUT_OCNICE` and `FHOUT_OCNICE_GFS` have been separated in the most recent commits. There is a pynorms issue I am trying to locate. Also, more modifcations will need to be made to `ush/python/pygfs/task/oceanice_products.py`, but I have inserted a temporary workaround for now.

EricSinsky-NOAA:
Regarding the CICE daily averaging issue, `ush/forecast_postdet.sh` has been modified so that the soft links point to the CICE output properly (please see a snippet of `ls` below). A few more tests are running, then these changes will be committed and pushed. 
```
iceh_12h.2021-03-24-00000.nc -> /lfs/h2/emc/ptmp/eric.sinsky/GEFS/COMROOT/ci1/gefs.20210323/12/mem000/model_data/ice/history/gefs.ice.t12z.12hr_avg.f012.nc
iceh_24h.2021-03-25-00000.nc -> /lfs/h2/emc/ptmp/eric.sinsky/GEFS/COMROOT/ci1/gefs.20210323/12/mem000/model_data/ice/history/gefs.ice.t12z.24hr_avg.f036.nc
iceh_24h.2021-03-26-00000.nc -> /lfs/h2/emc/ptmp/eric.sinsky/GEFS/COMROOT/ci1/gefs.20210323/12/mem000/model_data/ice/history/gefs.ice.t12z.24hr_avg.f060.nc
iceh_24h.2021-03-27-00000.nc -> /lfs/h2/emc/ptmp/eric.sinsky/GEFS/COMROOT/ci1/gefs.20210323/12/mem000/model_data/ice/history/gefs.ice.t12z.24hr_avg.f084.nc
iceh_24h.2021-03-28-00000.nc -> /lfs/h2/emc/ptmp/eric.sinsky/GEFS/COMROOT/ci1/gefs.20210323/12/mem000/model_data/ice/history/gefs.ice.t12z.24hr_avg.f108.nc
iceh_ic.2021-03-23-43200.nc -> /lfs/h2/emc/ptmp/eric.sinsky/GEFS/COMROOT/ci1/gefs.20210323/12/mem000/model_data/ice/history/gefs.ice.t12z.ic.nc
```

aerorahul:
> Regarding the CICE daily averaging issue, `ush/forecast_postdet.sh` has been modified so that the soft links point to the CICE output properly (please see a snippet of `ls` below). A few more tests are running, then these changes will be committed and pushed.
> 
> ```
> iceh_12h.2021-03-24-00000.nc -> /lfs/h2/emc/ptmp/eric.sinsky/GEFS/COMROOT/ci1/gefs.20210323/12/mem000/model_data/ice/history/gefs.ice.t12z.12hr_avg.f012.nc
> iceh_24h.2021-03-25-00000.nc -> /lfs/h2/emc/ptmp/eric.sinsky/GEFS/COMROOT/ci1/gefs.20210323/12/mem000/model_data/ice/history/gefs.ice.t12z.24hr_avg.f036.nc
> iceh_24h.2021-03-26-00000.nc -> /lfs/h2/emc/ptmp/eric.sinsky/GEFS/COMROOT/ci1/gefs.20210323/12/mem000/model_data/ice/history/gefs.ice.t12z.24hr_avg.f060.nc
> iceh_24h.2021-03-27-00000.nc -> /lfs/h2/emc/ptmp/eric.sinsky/GEFS/COMROOT/ci1/gefs.20210323/12/mem000/model_data/ice/history/gefs.ice.t12z.24hr_avg.f084.nc
> iceh_24h.2021-03-28-00000.nc -> /lfs/h2/emc/ptmp/eric.sinsky/GEFS/COMROOT/ci1/gefs.20210323/12/mem000/model_data/ice/history/gefs.ice.t12z.24hr_avg.f108.nc
> iceh_ic.2021-03-23-43200.nc -> /lfs/h2/emc/ptmp/eric.sinsky/GEFS/COMROOT/ci1/gefs.20210323/12/mem000/model_data/ice/history/gefs.ice.t12z.ic.nc
> ```

What about the 12-hr average between `f108`-`f120` at the end of the forecast valid at `f120`?

EricSinsky-NOAA:
> What about the 12-hr average between `f108`-`f120` at the end of the forecast valid at `f120`?

I can work on adding that too, @aerorahul.

EricSinsky-NOAA:
Please note that the raw netcdf file of the first lead time coming from CICE has a misleading filename ("iceh_24h.2021-03-24-00000.nc"). Even though "24h" is stated in the filename, it is actually a 12-hour average in this case. The name that will be copied to COMROOT ("gefs.ice.t12z.12hr_avg.f012.nc") should get rid of any confusion, but this means two "interval" variables will need to be defined in postdet (The first variable is `interval` and the second is `interval_source`). 

image

EricSinsky-NOAA:
> > What about the 12-hr average between `f108`-`f120` at the end of the forecast valid at `f120`?
> 
> I can work on adding that too, @aerorahul.

@aerorahul, after completing a few test runs with the original changes to `ush/forecast_postdet.sh`, it looks like CICE does not write the 12-hour averaged data between f108-f120 to NetCDF when `cyc="12"` and when `FHOUT_ICE=24`. `iceh_24h.2021-03-28-00000.nc` (`gefs.ice.t12z.24hr_avg.f108.nc` in COMROOT) is the final file that is written.

Now that the forecast task has succeeded with the original changes, I am now making some changes to the ice_prod task to account for the filename changes in cases where `cyc="12"` and when `FHOUT_ICE=24`.

EricSinsky-NOAA:
The GEFS tests have passed. The recently modified archive task in the GFS test did not pass. This is because `FHOUT_OCNICE` needs to be changed to `FHOUT_ICE` and `FHOUT_OCN` in the new archive-related scripts (PR #2491). Addressing this now.

EricSinsky-NOAA:
`FHOUT_OCNICE` has been split into `FHOUT_OCN` and `FHOUT_ICE` in the new archive-related tasks. The GFS C48 test case has passed with these changes. 

`ush/check_ice_netcdf.sh` has been modified so that conditions are stricter for the ice prod tasks to run. After the current CI tests have completed and passed, these changes regarding `ush/check_ice_netcdf.sh` will be committed and pushed, then I believe this PR will be ready for review.

parm/config/gefs/config.base
```diff
@@ -218,6 +218,13 @@ case "${APP}" in
     ;;
 esac
 
+# Output frequency of the forecast model (for cycling)
+export FHMIN=0
+export FHMAX=9
+export FHOUT=3           # Will be changed to 1 in config.base if (DOHYBVAR set to NO and l4densvar set to false)
+export FHOUT_OCN=3
+export FHOUT_ICE=3
```
EricSinsky-NOAA, line 228: @WalterKolczynski-NOAA Is it ok to add this to `parm/config/gefs/config.base`? This is already included in `parm/config/gfs/config.base`. The reason why I have added this to `parm/config/gefs/config.base` is because `workflow/rocoto/tasks.py` is expecting `FHOUT_OCN` and `FHOUT_ICE`.

WalterKolczynski-NOAA: Yes

EricSinsky-NOAA:
@WalterKolczynski-NOAA Can you re-review this PR? As of today and yesterday, all tasks for the C48_S2SWA_gefs and C48_S2SW test cases have passed (and no tasks were left untriggered).

parm/config/gefs/config.resources
```diff
@@ -198,8 +198,8 @@ case ${step} in
         declare -x "wtime_${step}_gfs"="03:00:00"
         ;;
       "C384" | "C768" | "C1152")
-        declare -x "wtime_${step}"="06:00:00"
-        declare -x "wtime_${step}_gfs"="06:00:00"
+        declare -x "wtime_${step}"="14:00:00"
+        declare -x "wtime_${step}_gfs"="14:00:00"
```
WalterKolczynski-NOAA: I think this is longer than the max job limit on many machines.

workflow/rocoto/gefs_tasks.py
```diff
@@ -218,14 +220,21 @@ def _atmosoceaniceprod(self, component: str):
...
-        deps.append(rocoto.add_dependency(dep_dict))
         if component in ['ocean']:
+            dep_dict = {'type': 'data', 'data': data, 'age': 120}
+            deps.append(rocoto.add_dependency(dep_dict))
             command = f"{self.HOMEgfs}/ush/check_netcdf.sh {history_path}/{history_file_tmpl}"
             dep_dict = {'type': 'sh', 'command': command}
             deps.append(rocoto.add_dependency(dep_dict))
             dependencies = rocoto.create_dependency(dep=deps, dep_condition='and')
+        elif component in ['ice']:
+            command = f"{self.HOMEgfs}/ush/check_ice_netcdf.sh @Y @m @d @H #fhr# &ROTDIR; #member# {fhout_ice_gfs}"
```
WalterKolczynski-NOAA, line 231: You should use `_template_to_rocoto_cycstring()` to generate the directory path instead of prescribing it in `check_ice_netcdf.sh`. If the underlying COM template were to be changed, this would break. Follow the pattern used for `check_netcdf.sh`.

Is there a reason `check_netcdf.sh` can't be used here?

workflow/rocoto/gefs_tasks.py
```diff
@@ -218,14 +220,21 @@ def _atmosoceaniceprod(self, component: str):
         history_path = self._template_to_rocoto_cycstring(self._base[history_path_tmpl], {'MEMDIR': 'mem#member#'})
         deps = []
         data = f'{history_path}/{history_file_tmpl}'
-        dep_dict = {'type': 'data', 'data': data, 'age': 120}
-        deps.append(rocoto.add_dependency(dep_dict))
```
WalterKolczynski-NOAA, line 222: I don't think we want to get rid of this for the ice. We don't want to use the bash function unless the file already exists.

workflow/rocoto/tasks.py
```diff
@@ -131,9 +131,16 @@ def _get_forecast_hours(cdump, config, component='atmos') -> List[str]:
         # Ocean/Ice components do not have a HF output option like the atmosphere
         if component in ['ocean', 'ice']:
             local_config['FHMAX_HF_GFS'] = config['FHMAX_GFS']
```
WalterKolczynski-NOAA: Can probably just make this 0:
```suggestion
            local_config['FHMAX_HF_GFS'] = 0
```

ush/python/pygfs/task/oceanice_products.py
```diff
@@ -58,17 +58,27 @@ def __init__(self, config: Dict[str, Any]) -> None:
 
         valid_datetime = add_to_datetime(self.runtime_config.current_cycle, to_timedelta(f"{self.config.FORECAST_HOUR}H"))
 
+        if self.config.COMPONENT == 'ice' and self.runtime_config.current_cycle.strftime("%H") != '00' and self.config.FHOUT_ICE_GFS == 24:
```
WalterKolczynski-NOAA: Please add some comments describing why we are doing this.

ush/forecast_predet.sh
```diff
@@ -568,7 +568,7 @@ CICE_predet(){
 
   # CICE does not have a concept of high frequency output like FV3
   # Convert output settings into an explicit list for CICE
-  CICE_OUTPUT_FH=$(seq -s ' ' "${FHMIN}" "${FHOUT_OCNICE}" "${FHMAX}")
+  CICE_OUTPUT_FH=$(seq -s ' ' "${FHMIN}" "${FHOUT_ICE}" "${FHMAX}")
```
WalterKolczynski-NOAA: I think we can avoid the complicated changes later on by just generating the output hours correctly up-front. This is also more general than handling just the 24-h case.
```suggestion
  ((offset = ( cyc + FHMIN ) % FHOUT_ICE))
  if (( offset == 0 )); then
    CICE_OUTPUT_FH=($(seq -s ' ' "${FHMIN}" "${FHOUT_ICE}" "${FHMAX}"))
  else
    CICE_OUTPUT_FH=("${FHMIN}")
    CICE_OUTPUT_FH+=($(seq -s ' ' "$((FHMIN+offset))" "${FHOUT_ICE}" "${FHMAX}"))
    CICE_OUTPUT_FH+=("${FHMAX}")
  fi
```

ush/forecast_postdet.sh
```diff
@@ -546,29 +546,57 @@ CICE_postdet() {
   # Link CICE forecast output files from DATA/CICE_OUTPUT to COM
   local source_file dest_file
   for fhr in ${CICE_OUTPUT_FH}; do
```
WalterKolczynski-NOAA:
```suggestion
  for fhr in ${CICE_OUTPUT_FH[*]}; do
```

ush/forecast_postdet.sh
```diff
@@ -546,29 +546,57 @@ CICE_postdet() {
...
+      if [[  ${fhr} -le 24 ]]; then
+        interval=${cyc}
+        interval_source=24
+      else
+        interval=24
+        interval_source=${interval}
+      fi
+    else
+        (( interval = fhr - last_fhr ))
+    fi
```
WalterKolczynski-NOAA:
```suggestion
    fhr3=$(printf %03i "${fhr}")
    (( interval = fhr - last_fhr ))
```

ush/forecast_postdet.sh
```diff
@@ -546,29 +546,57 @@ CICE_postdet() {
...
     seconds=$(to_seconds "${vdate:8:2}0000")  # convert HHMMSS to seconds
     vdatestr="${vdate:0:4}-${vdate:4:2}-${vdate:6:2}-${seconds}"
 
     if [[ "${RUN}" =~ "gfs" || "${RUN}" =~ "gefs" ]]; then
-      source_file="iceh_$(printf "%0.2d" "${interval}")h.${vdatestr}.nc"
+      if [[ "${cyc}" != "00" ]] && (( FHOUT_ICE % 24 == 0 )); then
+        source_file="iceh_$(printf "%0.2d" "${interval_source}")h.${vdatestr}.nc"
+      else
+        source_file="iceh_$(printf "%0.2d" "${interval}")h.${vdatestr}.nc"
+      fi
```
WalterKolczynski-NOAA:
```suggestion
      source_file="iceh_$(printf "%0.2d" "${interval}")h.${vdatestr}.nc"
```

ush/forecast_postdet.sh
```diff
@@ -546,29 +546,57 @@ CICE_postdet() {
...
     fi
     ${NLN} "${COMOUT_ICE_HISTORY}/${dest_file}" "${DATA}/CICE_OUTPUT/${source_file}"
 
-    last_fhr=${fhr}
+
+     if [[ ${cyc} != "00" ]] && (( FHOUT_ICE % 24 == 0 )); then
+       (( last_fhr = fhr + cyc ))
+     else
+       last_fhr=${fhr}
+     fi
```
WalterKolczynski-NOAA:
```suggestion
     last_fhr=${fhr}
```

parm/config/gefs/config.ufs
WalterKolczynski-NOAA, line 1: Is there a reason not to use the GFS settings for decomposition? Those seem to be fine and the only additional thing GEFS does is stochastic physics.

WalterKolczynski-NOAA (changes requested):

ush/forecast_predet.sh
EricSinsky-NOAA: Yes, this seems like a less complicated and more generalized approach. These changes have just been applied locally. I am retesting to make sure there are no issues. Thanks.

workflow/rocoto/gefs_tasks.py
EricSinsky-NOAA: The reason why I got rid of this dependency is because in cases where `FHOUT_ICE_GFS = 24` and `cycle=t12z`, the lead hours in the `fhr` variable (in the xml file) in the ice_prod task do not match the lead hours in the ice hist filenames. This means the following file dependency would never be satisfied in these particular cases: gefs.ice.t@Hz.24hr_avg.f#fhr#.nc. That is why I removed that dependency. I agree it is ideal to only use the bash function only when the file exists.

workflow/rocoto/gefs_tasks.py
EricSinsky-NOAA: I think it would be preferable to use `check_netcdf.sh` here. The only reason why I created an additional dependency bash script for the ice_prod task is for the reason I mentioned in my previous comment.

workflow/rocoto/gefs_tasks.py
EricSinsky-NOAA: I would like to clarify the issue here with an example which may help explain why I removed this dependency:

Let's say that `FHOUT_ICE_GFS=24 `and `cyc=12`. For the ice_prod task, we have an fhr variable: `<var name="fhr">024 048 072 096 120</var>`
If we were to keep this filename dependency
(`<datadep age="120"><cyclestr>&ROTDIR;/gefs.@Y@m@d/@H/mem#member#/model_data/ocean/history/gefs.ocean.t@Hz.24hr_avg.f#fhr#.nc</cyclestr></datadep>`), the dependency would never be satisfied because these lead hours in the fhr variable are not written in the ice hist filenames. The ice hist filenames would have lead hours 036, 060, etc.

EricSinsky-NOAA: Since the fhr variable needs to be the same for t00z, t06z, t12z, and t18z, I couldn't immediately think of a better way to do this.

ush/forecast_predet.sh
EricSinsky-NOAA: @WalterKolczynski-NOAA I just finished a test with this suggestion and it works great for the most part. The only issue is that the first lead time (f012) is not being picked up by the global-workflow because the CICE output is using "24h" in its filename, even though it is actually a 12 hour average. Please see this screenshot of an `ls` in `${DATA}/CICE_OUTPUT`.  
Screenshot from 2024-06-04 18-43-50

ush/forecast_postdet.sh
WalterKolczynski-NOAA: This gets around the "UFS always using 24h in the filename" problem.
```suggestion
      source_file="iceh_$(printf "%0.2d" "${FHOUT_ICE}")h.${vdatestr}.nc"
```

ush/forecast_predet.sh
WalterKolczynski-NOAA: See below suggestion in postdet to fix this.

workflow/rocoto/tasks.py
EricSinsky-NOAA: Sure, modification has been made.

ush/forecast_postdet.sh
EricSinsky-NOAA: Ok, I see. I forgot to replace `${interval}` with `${FHOUT_ICE}` in my original test. I am running another quick test now. I agree this should solve the "UFS always using 24h in the filename" problem.

parm/config/gefs/config.resources
EricSinsky-NOAA: Ok. I'll revert the changes I made to config.resources. Instead of increasing `wtime`, it may be better to split the forecast job into segments to avoid timeouts. Fortunately, it looks like this is already marked as in issue (#2274)

EricSinsky-NOAA: It may also be useful to allow for different wall times on different machines? However, I understand this may complicate the `config.resources` file.

ush/forecast_predet.sh
```diff
@@ -568,7 +568,14 @@
 
   # CICE does not have a concept of high frequency output like FV3
   # Convert output settings into an explicit list for CICE
-  CICE_OUTPUT_FH=$(seq -s ' ' "${FHMIN}" "${FHOUT_OCNICE}" "${FHMAX}")
+  ((offset = ( cyc + FHMIN ) % FHOUT_ICE))
+  if (( offset == 0 )); then
+    CICE_OUTPUT_FH=($(seq -s ' ' "${FHMIN}" "${FHOUT_ICE}" "${FHMAX}"))
```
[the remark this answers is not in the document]
EricSinsky-NOAA: @WalterKolczynski-NOAA Should we use `read -a` or can this spellcheck be ignored?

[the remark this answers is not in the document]
WalterKolczynski-NOAA: Go ahead and try fixing it.

ush/forecast_postdet.sh
```diff
@@ -545,22 +545,22 @@
 
   # Link CICE forecast output files from DATA/CICE_OUTPUT to COM
   local source_file dest_file
-  for fhr in ${CICE_OUTPUT_FH}; do
-    fhr3=$(printf %03i "${fhr}")
+  for fhr in ${CICE_OUTPUT_FH[*]}; do
```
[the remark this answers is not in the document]
WalterKolczynski-NOAA:
```suggestion
  for fhr in "${CICE_OUTPUT_FH[@]"}; do
```

workflow/rocoto/gefs_tasks.py
WalterKolczynski-NOAA: `_get_forecast_hours()` can be updated in the same manner we are updating the `CICE_OUTPUT_FH` generation in `predet`. This would also provide the correct forecast hours for the products job, which I think are needed as well anyway.

ush/forecast_postdet.sh
WalterKolczynski-NOAA: You didn't miss it, I added after you said it failed.

ush/forecast_postdet.sh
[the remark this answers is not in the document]
EricSinsky-NOAA: @WalterKolczynski-NOAA There were two things I needed to do to make these quotes work in postdet and use `mapfile` in predet:

1.  Add `fhr=$((10#${fhr}))` to line 549 in postdet to avoid invalid number errors when `fhr` has an 8. 
2.  Add `-d ' '` to `mapfile` in predet to make sure space is treated as a delimiter. 

I am testing and committing these changes. Hopefully this will resolve the spellcheck errors.

workflow/rocoto/gefs_tasks.py
EricSinsky-NOAA: Ok, I'll try to accomplish this by updating `_get_forecast_hours()` and remove the need for `check_ice_netcdf.sh`.

workflow/rocoto/tasks.py
```diff
@@ -148,11 +155,19 @@ def _get_forecast_hours(cdump, config, component='atmos') -> List[str]:
...
+
+            sdate = config['SDATE']
+            cyc = sdate.hour
+            offset = (cyc + fhmin) % fhout
+            if component in 'ice' and offset != 0:
+                fhrs_hf = range(fhmin + 12, fhmax_hf + fhout_hf, fhout_hf)
+                fhrs = list(fhrs_hf) + list(range(fhrs_hf[-1] + fhout, fhmax, fhout))
+            else:
+                fhrs_hf = range(fhmin, fhmax_hf + fhout_hf, fhout_hf)
+                fhrs = list(fhrs_hf) + list(range(fhrs_hf[-1] + fhout, fhmax + fhout, fhout))
```
EricSinsky-NOAA: @WalterKolczynski-NOAA A routine has been added to adjust `fhrs` in cases where the `component` is `ice` and the `offset `is not `0`. The main challenge here is to find a `cyc` variable. The best I could find is `sdate.hour`, but this may be problematic if there are multiple different `cyc ` values between `sdate `and `edate` that are different from `sdate.hour`'s.

ush/python/pygfs/task/oceanice_products.py
EricSinsky-NOAA: The changes made to `fhrs` in in the xml may remove the need for this if-statement in `oceanice_products.py`. I'll remove this if-statement after a couple tests.

workflow/rocoto/tasks.py
WalterKolczynski-NOAA: Yeah, this isn't going to be as clean as I would've liked. Let's go back to the way you had it so we can get this PR in and then maybe revisit it later.

EricSinsky-NOAA: Sure. I brought back the changes I had before. I think we can figure out a better way to do this in a separate PR.

EricSinsky-NOAA:
Please hold on any CI testing until I run a couple more cases.

EricSinsky-NOAA:
Update: When the `CICE_predet` function is executed, an exit 1 status returns only when `offset` is equal to 0 (e.g. when `FHOUT_ICE=6`). However, an exit 0 status returns when preamble.sh is not sourced in `exglobal_forecast.sh`. Currently testing alternate forms of syntax to perform the arithmetic to calculate `offset`. It is important to note that this issue does not occur when `offset` is not 0.

EricSinsky-NOAA:
> Update: When the `CICE_predet` function is executed, an exit 1 status returns only when `offset` is equal to 0 (e.g. when `FHOUT_ICE=6`). However, an exit 0 status returns when preamble.sh is not sourced in `exglobal_forecast.sh`. Currently testing alternate forms of syntax to perform the arithmetic to calculate `offset`. It is important to note that this issue does not occur when `offset` is not 0.

This issue has been fixed in commit https://github.com/NOAA-EMC/global-workflow/pull/2561/commits/82ee90bebe2d985446620d7d51dd9821c550a56b. C48_S2SWA_gefs and C48_S2SW test cases have passed when `FHOUT_OCN=6` and `FHOUT_ICE=6`.

ush/forecast_postdet.sh
```diff
@@ -545,22 +545,23 @@ CICE_postdet() {
 
   # Link CICE forecast output files from DATA/CICE_OUTPUT to COM
   local source_file dest_file
-  for fhr in ${CICE_OUTPUT_FH}; do
-    fhr3=$(printf %03i "${fhr}")
+  for fhr in "${CICE_OUTPUT_FH[@]}"; do
 
+    fhr=$((10#${fhr}))
```
WalterKolczynski-NOAA: This shouldn't be necessary, since CICE_OUTPUT_FH should already lack leading zeros.

ush/forecast_predet.sh
```diff
@@ -568,7 +568,13 @@ CICE_predet(){
 
   # CICE does not have a concept of high frequency output like FV3
   # Convert output settings into an explicit list for CICE
-  CICE_OUTPUT_FH=$(seq -s ' ' "${FHMIN}" "${FHOUT_OCNICE}" "${FHMAX}")
+  if (( $(( ( cyc + FHMIN ) % FHOUT_ICE )) == 0 )); then
+    mapfile -t -d ' ' CICE_OUTPUT_FH < <(seq -s ' ' "${FHMIN}" "${FHOUT_ICE}" "${FHMAX}") || true
```
WalterKolczynski-NOAA: I don't think these should pass if the list generation fails
```suggestion
    mapfile -t -d ' ' CICE_OUTPUT_FH < <(seq -s ' ' "${FHMIN}" "${FHOUT_ICE}" "${FHMAX}") || exit 10
```

ush/forecast_predet.sh
```diff
@@ -568,7 +568,13 @@ CICE_predet(){
...
+  else
+    CICE_OUTPUT_FH=("${FHMIN}")
+    mapfile -t -d ' ' -O "${#CICE_OUTPUT_FH[@]}"  CICE_OUTPUT_FH < <(seq -s ' ' "$(( FHMIN + $(( ( cyc + FHMIN ) % FHOUT_ICE )) ))" "${FHOUT_ICE}" "${FHMAX}") || true
```
WalterKolczynski-NOAA:
```suggestion
    mapfile -t -d ' ' -O "${#CICE_OUTPUT_FH[@]}"  CICE_OUTPUT_FH < <(seq -s ' ' "$(( FHMIN + $(( ( cyc + FHMIN ) % FHOUT_ICE )) ))" "${FHOUT_ICE}" "${FHMAX}") || exit 10
```

ush/python/pygfs/task/oceanice_products.py
```diff
@@ -58,17 +58,35 @@ def __init__(self, config: Dict[str, Any]) -> None:
 
         valid_datetime = add_to_datetime(self.runtime_config.current_cycle, to_timedelta(f"{self.config.FORECAST_HOUR}H"))
 
+        if self.config.COMPONENT == 'ice':
+            offset = int(self.runtime_config.current_cycle.strftime("%H")) % self.config.FHOUT_ICE_GFS
+            # For CICE cases where offset is not 0, forecast_hour needs to be adjusted based on the offset.
+            # TODO: Consider FHMIN when calculating offset.
```
WalterKolczynski-NOAA, line 64: Please open an issue for this and add the issue number to the TODO. It will need to be completed soon for the forecast segment work.

WalterKolczynski-NOAA (changes requested):
Almost there.

ush/forecast_postdet.sh
EricSinsky-NOAA: @WalterKolczynski-NOAA When I remove this line, there is an error using `printf`.
`/lfs/h2/emc/ens/noscrub/eric.sinsky/GEFS/gw_rfcst_prod2/ush/forecast_postdet.sh: line 556: printf: 120
: invalid number`
I think when mapfile is used to generate the forecast hour list, the list does not contain base10 values.

WalterKolczynski-NOAA: Looks like the last value in the sequence gets an erroneous new-line at the end:
```
$ export FHMIN=0
$ export FHOUT_ICE=24
$ export FHMAX=120
$ mapfile -t -d ' ' CICE_OUTPUT_FH < <(seq -s ' ' "${FHMIN}" "${FHOUT_ICE}" "${FHMAX}")
$ for fhr in "${CICE_OUTPUT_FH[@]}"; do echo "==>${fhr}<=="; done
==>0<==
==>24<==
==>48<==
==>72<==
==>96<==
==>120
<==
```

ush/forecast_predet.sh
```diff
@@ -568,7 +568,13 @@
 
   # CICE does not have a concept of high frequency output like FV3
   # Convert output settings into an explicit list for CICE
-  CICE_OUTPUT_FH=$(seq -s ' ' "${FHMIN}" "${FHOUT_OCNICE}" "${FHMAX}")
+  if (( $(( ( cyc + FHMIN ) % FHOUT_ICE )) == 0 )); then
+    mapfile -t -d ' ' CICE_OUTPUT_FH < <(seq -s ' ' "${FHMIN}" "${FHOUT_ICE}" "${FHMAX}") || exit 10
```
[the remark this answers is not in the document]
WalterKolczynski-NOAA: Try this to fix the extra new-line
```suggestion
    mapfile -t CICE_OUTPUT_FH < <(seq "${FHMIN}" "${FHOUT_ICE}" "${FHMAX}" || exit 10
```

ush/forecast_predet.sh
EricSinsky-NOAA: @WalterKolczynski-NOAA It looks like this triggered a shellcheck error. Is it ok to do a shellcheck disable for this (`# shellcheck disable=SC2312`)? It looks like this particular shellcheck was disabled in instances where mapfile was used in line 102 and line 544 in `forecast_predet`.

ush/forecast_predet.sh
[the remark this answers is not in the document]
WalterKolczynski-NOAA: Repeat with the next `mapfile`

ush/python/pygfs/task/oceanice_products.py
EricSinsky-NOAA: Sure, I'll open an issue.

ush/forecast_predet.sh
WalterKolczynski-NOAA: Yes, go ahead and disable

ush/forecast_predet.sh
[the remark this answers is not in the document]
EricSinsky-NOAA: Thank you @WalterKolczynski-NOAA. After a quick test, this seems to have fixed the extra new-line.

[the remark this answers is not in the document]
WalterKolczynski-NOAA: A little bit what's going on:
We were specifying a delimiter for `seq` (a hold-over from earlier when not using `mapfile`), but `seq` is still going to tack a new line at the end of the output. `mapfile` was then using the same delimiter (space), which meant the `-t` would try to remove an extra space from the end instead of a new line. Now we are using the default delimiter (new line) for both, so `-t` works as we wanted it to.

ush/python/pygfs/task/oceanice_products.py
WalterKolczynski-NOAA: This doesn't impact executed code, so I'm going to start CI now that everything else is done.

TerrenceMcGuinness-NOAA:
Connection to Hercules was lost from the controller after the build.

DavidHuber-NOAA:
@TerrenceMcGuinness-NOAA Hercules had an issue with a handful of nodes which have now been taken offline.  This may or may not help the issue you are seeing.

WalterKolczynski-NOAA (approved):

ush/check_ice_netcdf.sh
```diff
@@ -0,0 +1,38 @@
+#! /usr/bin/env bash
```
aerorahul, line 1: @EricSinsky-NOAA 
I just noticed this script was added after an approval review was made on this PR and it slipped from being reviewed.
The function implemented here is the `datadep` with `age` that Rocoto already provides.
Besides, this is completely hardwired for GEFS options.
The function is checking the age of a file.
The file happens to be a ice netcdf file.
We will need to implement this via the proper rocoto means of checking if a file exists and if it has been present for 2 mins.

PR #2588, opened 2024-05-10
Limit gfswavepostpnt to 40 PEs/node

### LINKED ISSUE #2587: The gfswavepostpnt job runs very slowly on Hercules
### What is wrong?

When gfswavepostpnt runs, it spawns 200 MPMD jobs which are wrappers around the `ush/wave_outp_spec.sh` script.  Altogether, this script is called several thousand times and has many operations on disk (e.g. `cat`, `sed`, `grep`).  These operations do not scale well on large nodes causing the job to run very slowly.

Ideally, the job would use more efficient operations to achieve its goals, but a workaround in its current state is to run it with fewer MPMD jobs per node.

Hercules

### Steps to reproduce

Run a gfswavepostpnt job on Hercules and Hera and compare the runtimes of the wave_outp_spec.sh script with Hera.

### Additional information

Discovered while testing https://github.com/NOAA-EMC/global-workflow/pull/2527.

Spread the jobs over 5 nodes on Hercules (40/node) instead of 3.

DavidHuber-NOAA:
# Description
This fixes the slow runtime of the gfswavepostpnt job on Hercules.  The job is very I/O intensive and does not scale well to large nodes, so limit the number of jobs/node to 40.

Resolves #2587 
# Type of change

# Change characteristics

# How has this been tested?
Forecast only on Hercules

# Checklist

### DIFF
--- a/parm/config/gfs/config.resources
+++ b/parm/config/gfs/config.resources
@@ -163,8 +163,14 @@ case ${step} in
     export npe_wavepostpnt=200
     export nth_wavepostpnt=1
     export npe_node_wavepostpnt=$(( npe_node_max / nth_wavepostpnt ))
-    export NTASKS=${npe_wavepostpnt}
     export is_exclusive=True
+    # This MPMD job is very I/O heavy and does not scale well. Limit jobs/node to 40.
+    # For now this affects only Hercules, but will also affect Gaea.
+    if [[ ${npe_node_wavepostpnt} -gt 40 ]]; then
+        export npe_node_wavepostpnt=40
+        export is_exclusive=False
+    fi
+    export NTASKS=${npe_wavepostpnt}
     ;;
 
   "wavegempak")
--- a/parm/config/gfs/config.resources
+++ b/parm/config/gfs/config.resources
@@ -140,13 +140,19 @@ case ${step} in
     export memory_wavepostsbs_gfs="10GB"
     ;;
 
+  # The wavepost*pnt* jobs are I/O heavy and do not scale well to large nodes.
+  # Limit the number of tasks/node to 40.
   "wavepostbndpnt")
     export wtime_wavepostbndpnt="01:00:00"
     export npe_wavepostbndpnt=240
     export nth_wavepostbndpnt=1
     export npe_node_wavepostbndpnt=$(( npe_node_max / nth_wavepostbndpnt ))
-    export NTASKS=${npe_wavepostbndpnt}
     export is_exclusive=True
+    if [[ ${npe_node_wavepostbndpnt} -gt 40 ]]; then
+        export npe_node_wavepostbndpnt=40
+        export is_exclusive=False
+    fi
+    export NTASKS=${npe_wavepostbndpnt}
     ;;
 
   "wavepostbndpntbll")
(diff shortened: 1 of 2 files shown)


parm/config/gfs/config.resources
```diff
@@ -163,8 +163,14 @@ case ${step} in
     export npe_wavepostpnt=200
     export nth_wavepostpnt=1
     export npe_node_wavepostpnt=$(( npe_node_max / nth_wavepostpnt ))
-    export NTASKS=${npe_wavepostpnt}
     export is_exclusive=True
+    # This MPMD job is very I/O heavy and does not scale well. Limit jobs/node to 40.
+    # For now this affects only Hercules, but will also affect Gaea.
+    if [[ ${npe_node_wavepostpnt} -gt 40 ]]; then
```
JessicaMeixner-NOAA, line 177: Wouldn't this also effect WCOSS2? While this job is slow and does not scale well, I do believe it scales past 40 PEs on WCOSS2. 

This should  also be added for wavepostbndpnt and wavepostbndpntbll

DavidHuber-NOAA: Right, I forgot about WCOSS2.  I'll add a check for Hercules instead and add the same for the other `pnt` jobs.  And since Gaea is built very similarly to WCOSS2, it may not have the same issue.  (I wonder if it is a Lustre limitation 🤔)

JessicaMeixner-NOAA: So if i re-read comments correctly, this just does 40 pets per node, it's not limiting the total number of tasks correct? (sorry might need more coffee...).  If that's the case, this might also help improve things for WCOSS2 as well.  It's been a while since I did a scaling/performance study on this job.   This job does cause problems on WCOSS2 as well.

DavidHuber-NOAA: OK, noted.  I'll change it back and run a test on WCOSS2 both ways to get that data point.

aerorahul (approved):
This was found to be an issue on Hercules.
Thanks for identifying the issue and resolving it.

parm/config/gfs/config.resources
DavidHuber-NOAA: On WCOSS2, the gfswavepostpnt job ran in 1.5 hours when executed across 2 nodes (probably 100 MPMD jobs each) vs 32 minutes when executed with 40 MPMD jobs per node.

JessicaMeixner-NOAA: ..... wow.... !!!!

JessicaMeixner-NOAA (approved):
Thanks @DavidHuber-NOAA !!!!!

WalterKolczynski-NOAA (approved):

JessicaMeixner-NOAA:
As I'm hour 2 of gfswavepostpnt for a C48 job on hera, this 30 minute thing sounds exciting..  I wonder if even we need to decrease this for hera too?

WalterKolczynski-NOAA:
> As I'm hour 2 of gfswavepostpnt for a C48 job on hera, this 30 minute thing sounds exciting.. I wonder if even we need to decrease this for hera too?

Hera is already smaller than 40, but maybe the limit needs to be even smaller there. Likely some exploration of the resource space would be needed to determine the optimal value for each machine.

TerrenceMcGuinness-NOAA:
CI Build on Hera failed during checkout from GitHub.  Removing **CI-Hera-Building** label manually.

PR #2613, opened 2024-05-21
STMP CI yaml over ride updates due to Orion-Hercules cross mounting

TerrenceMcGuinness-NOAA:
# Description

This hot bug fix updates how CI over rides the STMP value for Hercules and Orion cases because they are cross mounted.

Also took this moment to update SLUM to BATCH because its a misnomer.

# Type of change

We had to include the updating of STMP in the over ride yaml configs in CI because the value is designated in `$HOMEgfs/workflow/hosts/$MACHINE_ID.yaml`

# Change characteristics

# How has this been tested?
Going to run CI on Orion/Hercules

### DIFF
--- a/ci/cases/yamls/atmaerosnowDA_defaults_ci.yaml
+++ b/ci/cases/yamls/atmaerosnowDA_defaults_ci.yaml
@@ -4,3 +4,4 @@ base:
   DOIAU: "NO"
   DO_JEDISNOWDA: "YES"
   ACCOUNT: {{ 'BATCH_ACCOUNT' | getenv }}
+  STMP: {{ 'STMP' | getenv }}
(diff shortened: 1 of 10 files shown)


ci/cases/yamls/atmaerosnowDA_defaults_ci.yaml
```diff
@@ -3,4 +3,4 @@ defaults:
 base:
   DOIAU: "NO"
   DO_JEDISNOWDA: "YES"
-  ACCOUNT: {{ 'SLURM_ACCOUNT' | getenv }}
```
aerorahul, line 6: UFS-weather-model uses something named `ACCNR`.
Can we use that instead of `BATCH_ACCOUNT` to keep some semblance of consistency?

aerorahul (changes requested):
a suggested change in the name of the account

DavidHuber-NOAA (commented):
I don't think we use it anywhere, but PTMP is also being set to a common directory for Orion and Hercules.  Should that be added here as well?

TerrenceMcGuinness-NOAA:
@WalterKolczynski-NOAA Dave is right, we added both in the shell, we should add it in the over rides too, right?

DavidHuber-NOAA (approved):

PR #2634, opened 2024-05-28
Script to keep Jenkins Agent persistent from cron

### LINKED ISSUE #2633: Script from cron for persistent Jenkins Java Agent

We need a script that can run from a cron job that check the node connection to the Jenkins agent on a machine and start up the agent if is not.

That the cron script can launch an agent if it is not connected

### Acceptance Criteria

We test the system by manually killing the agent and see if the cron starts it back up on its own

write in in bash and use curl with Jenkins Remote API to check the connection.

TerrenceMcGuinness-NOAA:
# Description

This "persistent" Java Agent launch script can be ran from a cron job:

- Uses Jenkins Remote API to check the status of the Node connection using curl for a given machine.
- If it is not connected a new agent is launched for that node.

Resolves #2633

# Type of change
- New feature (adds a more robust capability)

# Change characteristics

# How has this been tested?
Script was run on Orion, Hera, and Hercules and each could determine that the nodes where up and nothing was done.


ci/scripts/utils/launch_java_agent.sh
```diff
@@ -0,0 +1,77 @@
...
+
+source "${HOMEgfs}/ush/detect_machine.sh"
+case ${MACHINE_ID} in
+  hera | orion | hercules | wcoss2)
+   echo "Launch Jenkins Java Controler on ${MACHINE_ID}"
+   ;;
+ *)
+   echo "Unsupported platform. Exiting with error."
+   exit 1
+   ;;
```
WalterKolczynski-NOAA:
```suggestion
  hera | orion | hercules | wcoss2)
    echo "Launch Jenkins Java Controler on ${MACHINE_ID}"
    ;;
  *)
    echo "Unsupported platform. Exiting with error."
    exit 1
    ;;
```

WalterKolczynski-NOAA (changes requested):
Looks good. Thanks for getting this done so quickly.

Don't think there is any point in running CI on this, so I'll merge once the indentation is corrected.

TerrenceMcGuinness-NOAA:
@WalterKolczynski-NOAA Yes this script was greatly needed and really cleans a lot of stuff up with launching by making it uniform and defined in the repo.  BTW, I'm contributing using AI that has made these one day turn around times for these solutions.

WalterKolczynski-NOAA (approved):

PR #2651, opened 2024-06-03
Convert staging job to python and yaml

### LINKED ISSUE #2475: Stage initial conditions for cycled and forecast-only via stage_ic job.

Presently, the cycled initial conditions are staged by `setup_expt.py`.
The initial conditions for forecast-only are staged by a job `stage_ic`.

This feature requests development of the `stage_ic` job to stage cycled initial conditions in addition to the ones for forecast-only.

The new `stage_ic` job should be able to handle both `cycled` and `forecast-only` modes as well as for `gfs` and `gefs` systems.  
Extra points for making the job generic to allow users to provide (files) initial conditions of their choosing from arbitrary locations and structure.

### Acceptance Criteria

Add the `stage_ic` job to the first half-cycle of the cycled mode.

### LINKED ISSUE #2650: Convert forecast-only stage_ic job to python

The existing stage_ic job in forecast-only mode needs to be converted from shell to python. This should improve and enhance the job capabilities and make it more extensible.

This is a subtask of issue #2475.

Existing job functionality should be retained but converted to python and yaml.

### Acceptance Criteria

Job still performs the same tasks.

### TODO

- [x] update `jobs/JGLOBAL_STAGE_IC` to use `.py` version of ex-script
- [x] rename `scripts/exglobal_stage_ic.sh` to `scripts/exglobal_stage_ic.py` and convert to python
- [x] create `ush/python/pygfs/task/stage.py` and populate with defs for staging
- [x] create `parm/stage` and create yaml files within for each stage mode
- [x] check `stage_ic` job setup and dependencies

### LINKED ISSUE #2686: Change model_data to model in COM templates

By NCO's request, `model_data` should be just `model` in COM paths.

`model_data` replaced by `model`.

### Acceptance Criteria

- [ ] Output that previous landed in `model_data` now lands in `model`
- [ ] Output is otherwise unchanged

_No response_

KateFriedman-NOAA:
# Description

This PR converts the staging job from shell to python and introduces the use of yaml.

Changes in this PR:

1. Rename `scripts/exglobal_stage_ic.sh` to `scripts/exglobal_stage_ic.py`.
2. Update `jobs/JGLOBAL_STAGE_IC` to use `.py` script extension. Move `COM*` variable declarations and member loop down into yaml and python respectively. Move `GDATE/gPDY/gcyc` settings up to JJOB from ex-script and replace with newer cycle variables (as done in forecast job).
3. Create `parm/stage` folder to hold newly created `stage.yaml.j2`, which both mimics forecast-only functionality in existing `scripts/exglobal_stage_ic.sh` and adds functionality for cycled mode.
4. Create `ush/python/pygfs/task/stage.py` to house staging job python functions for call from `scripts/exglobal_stage_ic.py`.
5. Remove `stage_ic` job rocoto dependencies from xml. Do not need and removes area of duplicate maintenance.
6. Add cycled staging jobs for gdas and enkf suites.
7. Rename `model_data` to `model` for issue #2686

There will now be distinct `stage_ic` jobs for each `RUN`: `gdasstage_ic`, `gfsstage_ic`, `enkfgdasstage_ic`, `stage_ic` (for gefs).

Related work was done to set up new symlink folder structure under supported platform `ICSDIR` folder for use by updated staging job.

Resolves #2475
Resolves #2650
Resolves #2686

# Type of change

# Change characteristics

# How has this been tested?

Ran all of the CI tests (either just staging jobs or end-to-end) on WCOSS2. Ran some of the CI tests on Hera and Hercules.

Ran some control CI tests from `develop` and compared outputs against test CI from staging branch. Outputs reproduced between control and test.


ush/python/pygfs/task/stage.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+
+        stage_parm = os.path.join(stage_dict.PARMgfs, "stage")
+
+        if not os.path.isdir(stage_dict.ROTDIR):
+            raise FileNotFoundError(f"FATAL ERROR: The ROTDIR ({stage_dict.ROTDIR}) does not exist!")
+
+        for set_yaml in stage_sets:
+
+            stage_set = parse_j2yaml(os.path.join(stage_parm, set_yaml), stage_dict)
```
DavidHuber-NOAA: If all variables need to be defined in the jinja file, then you might consider adding `allow_missing=False` to the `parse_j2yaml` call.  This will raise an exception if a variable referenced in the template is not defined.

parm/stage/fv3_nest.yaml.j2
```diff
@@ -0,0 +1,17 @@
+{% set cycle_HH = current_cycle | strftime("%H") %}
+{% set cycle_YMD = current_cycle | to_YMD %}
+{% set cycle_YMDH = current_cycle | to_YMDH %}
+{% set ntile = 7 %}
+fv3_nest:
+    mkdir:
+        - "{{ COM_ATMOS_RESTART_PREV }}"
+    copy:
+        {% if EXP_WARM_START %}
```
DavidHuber-NOAA: If you want to allow for missing variables, then I would suggest logic to be explicit.  e.g.
```jinja
        {% if EXP_WARM_START == True %}
```

parm/stage/fv3_cold.yaml.j2
```diff
@@ -0,0 +1,11 @@
...
+fv3_cold:
+    mkdir:
+        - "{{ COM_ATMOS_INPUT }}"
+    copy:
+        - ["{{ BASE_CPLIC }}/{{ CPL_ATMIC }}/{{ cycle_YMDH }}/{{ MEMDIR }}/atmos/gfs_ctrl.nc", "{{ COM_ATMOS_INPUT }}"]
+        {% for ftype in ["gfs_data", "sfc_data"] %}
+        {% for ntile in range(1, ntiles + 1) %}
+        - ["{{ BASE_CPLIC }}/{{ CPL_ATMIC }}/{{ cycle_YMDH }}/{{ MEMDIR }}/atmos/{{ ftype }}.tile{{ ntile }}.nc", "{{ COM_ATMOS_INPUT }}"]
+        {% endfor %} # ntile
+        {% endfor %} # ftype
```
DavidHuber-NOAA: You have the option to tab Jinja blocks.  The YAML still has to be tabbed correctly, though.  So you could do
```jinja,yaml
        {% for ftype in ["gfs_data", "sfc_data"] %}
            {% for ntile in range(1, ntiles + 1) %}
        - ["{{ BASE_CPLIC }}/{{ CPL_ATMIC }}/{{ cycle_YMDH }}/{{ MEMDIR }}/atmos/{{ ftype }}.tile{{ ntile }}.nc", "{{ COM_ATMOS_INPUT }}"]
            {% endfor %} # ntile
        {% endfor %} # ftype
```

or
```jinja
{% for ftype in ["gfs_data", "sfc_data"] %}
    {% for ntile in range(1, ntiles + 1) %}
        - ["{{ BASE_CPLIC }}/{{ CPL_ATMIC }}/{{ cycle_YMDH }}/{{ MEMDIR }}/atmos/{{ ftype }}.tile{{ ntile }}.nc", "{{ COM_ATMOS_INPUT }}"]
    {% endfor %} # ntile
{% endfor %} # ftype
```
The latter option doesn't always work out so nicely

KateFriedman-NOAA: Good to know, thanks! I'll indent those secondary `for` loops to make it more readable in the yamls. Done @ 3f08f035

parm/stage/fv3_nest.yaml.j2
KateFriedman-NOAA: Good suggestion, thanks @DavidHuber-NOAA ! Done @ 3f08f035

KateFriedman-NOAA:
Got some good feedback from reviewers (offline and online). Will ponder feedback and work on updates. Will leave this draft PR open while working on it. Thank you all so far!

KateFriedman-NOAA:
Will be making some iterative commits. Please do not rereview until rerequested.

scripts/exglobal_stage_ic.py
```diff
@@ -0,0 +1,44 @@
...
+    for key in keys:
+        stage_dict[key] = stage.task_config[key]
+
+    # Also import all COM* directory and template variables
+    for key in stage.task_config.keys():
+        if key.startswith("COM"):
+            stage_dict[key] = stage.task_config[key]
+
+    # Add the os.path.exists function to the dict for yaml parsing
+    stage_dict['path_exists'] = os.path.exists
```
aerorahul: Check if this is used, if not, remove.

parm/config/gfs/config.stage_ic
```diff
@@ -7,39 +7,24 @@ echo "BEGIN: config.stage_ic"
...
-    ;;
-esac
-
-if [[ "${DO_NEST:-NO}" == "YES" ]] ; then
-  export CPL_ATMIC="GLOBAL-NEST_${CASE}"
+export STAGE_IC_YAML_TMPL="${PARMgfs}/stage/stage.yaml.j2"
+
+# Set ICSDIR
+
+if [[ -z "${ICSDIR}" ]] ; then
```
aerorahul:
```suggestion
# Set ICSDIR (if not defined)
if [[ -z "${ICSDIR}" ]] ; then
```

parm/stage/stage.yaml.j2
```diff
@@ -0,0 +1,99 @@
+{% set cycle_HH = current_cycle | strftime("%H") %}
```
aerorahul: Some documentation up-top will be nice.
Something like:
```suggestion
# This yaml is intended to be of the following structure:
key1:
    mkdir:
        - "COM directory to create"
    copy:
        - ["source_file", "destination_file"]
key2:
    mkdir:
        - "COM directory to create"
    copy:
        - ["source_file", "destination_file"]
# Any number of keys with nested mkdir and copy are permitted
# Jinja is permitted in this yaml, as long as the keys are:
# - COM_
# - DO_ATM, DO_OCN, DO_ICE
# For a full list see exglobal_stage_ic.py        

{% set cycle_HH = current_cycle | strftime("%H") %}
```

parm/config/gfs/config.stage_ic
KateFriedman-NOAA: Committed, thanks!

parm/stage/stage.yaml.j2
KateFriedman-NOAA: Great suggestion, thanks a bunch!

ush/python/pygfs/task/stage.py
KateFriedman-NOAA: Done, committed to my branch @ 68c121f, thanks!

scripts/exglobal_stage_ic.py
KateFriedman-NOAA: Yep, I need it and now use it for the optional ICs.

jobs/JGLOBAL_STAGE_IC
```diff
@@ -5,13 +5,99 @@ source "${HOMEgfs}/ush/jjob_header.sh" -e "stage_ic" -c "base stage_ic"
 
 # Restart conditions for GFS cycle come from GDAS
 # shellcheck disable=SC2153
-rCDUMP=${RUN}
+rRUN=${RUN}
 # shellcheck disable=SC2153
-[[ ${RUN} = "gfs" ]] && export rCDUMP="gdas"
-export rCDUMP
+[[ ${RUN} = "gfs" ]] && export rRUN="gdas"
```
WalterKolczynski-NOAA: `rRUN` is only being exported when `$RUN == 'gfs'`. Either it does not need to be exported, or it should always be exported.

WalterKolczynski-NOAA:
```suggestion
[[ ${RUN} == "gfs" ]] && rRUN="gdas"
```

jobs/JGLOBAL_STAGE_IC
```diff
@@ -5,13 +5,99 @@ source "${HOMEgfs}/ush/jjob_header.sh" -e "stage_ic" -c "base stage_ic"
...
 
-# Execute the Script
-"${SCRgfs}/exglobal_stage_ic.sh"
+# Define significant cycles
+# Locally scoped variables and functions
```
WalterKolczynski-NOAA: But you're exporting some of them...

jobs/JGLOBAL_STAGE_IC
```diff
@@ -5,13 +5,99 @@ source "${HOMEgfs}/ush/jjob_header.sh" -e "stage_ic" -c "base stage_ic"
...
+# shellcheck disable=SC2153
+half_window=$(( assim_freq / 2 ))
+current_cycle="${PDY}${cyc}"
+previous_cycle=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} - ${assim_freq} hours" +%Y%m%d%H)
+current_cycle_begin=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} - ${half_window} hours" +%Y%m%d%H)
+current_cycle_end=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} + ${half_window} hours" +%Y%m%d%H)
```
WalterKolczynski-NOAA: Why did you combine `$PDY` and `$cyc` to then just slice them in the same manner anyway?
```suggestion
previous_cycle=$(date --utc -d "${PDY} ${cyc} - ${assim_freq} hours" +%Y%m%d%H)
current_cycle_begin=$(date --utc -d "${PDY} ${cyc} - ${half_window} hours" +%Y%m%d%H)
current_cycle_end=$(date --utc -d "${PDY} ${cyc} + ${half_window} hours" +%Y%m%d%H)
```

jobs/JGLOBAL_STAGE_IC
```diff
@@ -5,13 +5,99 @@ source "${HOMEgfs}/ush/jjob_header.sh" -e "stage_ic" -c "base stage_ic"
...
+###############################################################
+for MEMDIR in "${MEMDIR_ARRAY[@]}"; do
+
+  # Export MEMDIR; need even if empty
+  export MEMDIR
+
+  # Declare COMs
+  if [[ "${MODE}" = "cycled" && "${RUN}" = "gdas" ]]; then
+    YMD=${current_cycle:0:8} HH=${current_cycle:8:2} declare_from_tmpl COMOUT_ATMOS_ANALYSIS:COM_ATMOS_ANALYSIS_TMPL
+    export COMOUT_ATMOS_ANALYSIS
```
WalterKolczynski-NOAA: Similarly for all the other `declare_from_tmpl` that need to be exported.
```suggestion
    YMD=${current_cycle:0:8} HH=${current_cycle:8:2} declare_from_tmpl -x COMOUT_ATMOS_ANALYSIS:COM_ATMOS_ANALYSIS_TMPL
```

jobs/JGLOBAL_STAGE_IC
```diff
@@ -5,13 +5,99 @@ source "${HOMEgfs}/ush/jjob_header.sh" -e "stage_ic" -c "base stage_ic"
...
+fi
+
+# Initialize return code
+err=0
+
+###############################################################
+for MEMDIR in "${MEMDIR_ARRAY[@]}"; do
+
+  # Export MEMDIR; need even if empty
+  export MEMDIR
```
WalterKolczynski-NOAA: Why does `MEMDIR` need to be exported? It gets incorporated into all of the COM variables.

jobs/JGLOBAL_STAGE_IC
```diff
@@ -5,13 +5,99 @@ source "${HOMEgfs}/ush/jjob_header.sh" -e "stage_ic" -c "base stage_ic"
...
+  done
+else
+  MEMDIR_ARRAY+=("")
+fi
+
+# Initialize return code
+err=0
+
+###############################################################
+for MEMDIR in "${MEMDIR_ARRAY[@]}"; do
```
WalterKolczynski-NOAA: Not sure about moving this loop into the j-job. But, this is also a non-production script, so maybe it doesn't matter.

jobs/JGLOBAL_STAGE_IC
```diff
@@ -5,13 +5,99 @@ source "${HOMEgfs}/ush/jjob_header.sh" -e "stage_ic" -c "base stage_ic"
...
+    export COMOUT_ATMOS_RESTART_PREV COMOUT_MED_RESTART_PREV
+  else
+    YMD=${current_cycle:0:8} HH=${current_cycle:8:2} declare_from_tmpl COMOUT_ATMOS_INPUT:COM_ATMOS_INPUT_TMPL
+    export COMOUT_ATMOS_INPUT
+  fi
+  if [[ "${DO_OCN:-}" = "YES" ]]; then
+    RUN=${rRUN} YMD=${previous_cycle:0:8} HH=${previous_cycle:8:2} declare_from_tmpl COMOUT_OCEAN_RESTART_PREV:COM_OCEAN_RESTART_TMPL
+    export COMOUT_OCEAN_RESTART_PREV
+  fi
+  if [[ "${REPLAY_ICS:-NO}" = "YES" ]]; then
```
WalterKolczynski-NOAA: This if-block should be inside the `DO_OCN` if.

parm/config/gefs/config.base
```diff
@@ -38,7 +38,8 @@ export FIXugwd=${FIXgfs}/ugwd
 export PACKAGEROOT="@PACKAGEROOT@"    # TODO: set via prod_envir in Ops
 export COMROOT="@COMROOT@"    # TODO: set via prod_envir in Ops
 export COMINsyn="@COMINsyn@"
-export BASE_CPLIC="@BASE_CPLIC@"
+export ICSDIR="@ICSDIR@"
```
WalterKolczynski-NOAA: This is being set in `config.stage_ic`. Why is it here, and templated to boot?

parm/config/gefs/config.stage_ic
```diff
@@ -7,32 +7,24 @@ echo "BEGIN: config.stage_ic"
...
-esac
+export STAGE_IC_YAML_TMPL="${PARMgfs}/stage/stage.yaml.j2"
+
+# Set ICSDIR
+
+if [[ -z "${ICSDIR}" ]] ; then
+
+  IC_VER="20240610"
+
+  if [[ "${CASE_ENS}" != "@CASEENS@" ]] ; then
```
WalterKolczynski-NOAA:
```suggestion
  if (( NMEM_ENS > 0 )) ; then
```

parm/config/gefs/config.stage_ic
```diff
@@ -7,32 +7,24 @@ echo "BEGIN: config.stage_ic"
...
-    exit 1
-    ;;
-esac
+export STAGE_IC_YAML_TMPL="${PARMgfs}/stage/stage.yaml.j2"
+
+# Set ICSDIR
+
+if [[ -z "${ICSDIR}" ]] ; then
+
+  IC_VER="20240610"
```
WalterKolczynski-NOAA: Please make the non-exported variables lowercase.

ush/python/pygfs/task/stage.py
```diff
@@ -0,0 +1,55 @@
...
+        super().__init__(config)
+
+        rotdir = self.task_config.ROTDIR + os.sep
+
+        self.task_config = AttrDict(**self.task_config)
+
+    @logit(logger)
+    def execute_stage(self, stage_dict: Dict[str, Any]) -> None:
+        """Perform local staging of initial condition files.
+        """
```
WalterKolczynski-NOAA: Missing parameter and return lists. in the doc-block.

WalterKolczynski-NOAA (changes requested):

jobs/rocoto/stage_ic.sh
```diff
@@ -7,6 +7,11 @@ source "${HOMEgfs}/ush/preamble.sh"
 status=$?
 [[ "${status}" -ne 0 ]] && exit "${status}"
 
+###############################################################
+# setup python path for workflow utilities and tasks
+PYTHONPATH="${PYTHONPATH:+${PYTHONPATH}:}${HOMEgfs}/ush/python"
+export PYTHONPATH
+
```
DavidHuber-NOAA: This isn't necessary as ${HOMEgfs}/ush/python is already in the PYTHONPATH.

```suggestion
```

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: Thanks for the suggestion, committed.

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: I pulled the cycle variables in from one of the forecast job ush scripts but you're right, I should have made them all use PDYcyc, thanks for the suggestion!

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: Thanks for this! Will make the similar updates for the other declares.

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: Moved inside the `DO_OCN` if @ 3fee34c, thanks!

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: I spoke with Rahul about this, gonna leave it in the JJOB for now. The issue is needing to re-declare the COM variables for each member inside the python/yaml/jinja. Can come back and update this in a later PR.

parm/config/gefs/config.stage_ic
KateFriedman-NOAA: Thanks for this!

jobs/rocoto/stage_ic.sh
KateFriedman-NOAA: Good to know, thanks for the suggestion!

parm/config/gefs/config.stage_ic
KateFriedman-NOAA: Done @ 00d0a8c, thanks!

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: This was a leftover from using `MEMDIR` within the python. I just removed it but then ran into an issue getting `MEMDIR` into the `COM` variables for enkfgdas. A retest of a gefs staging job does not run into this issue but an enkfgdas staging job in a different CI test doesn't get the `MEMDIR` through to the declare command. See this snippet of a log (which I added an echo of `MEMDIR` to so I could see it was set) and see line 770 does not include `mem001`:
```
758 + JGLOBAL_STAGE_IC[54]: for MEMDIR in "${MEMDIR_ARRAY[@]}"
759 + JGLOBAL_STAGE_IC[56]: echo mem001
760 mem001
761 + JGLOBAL_STAGE_IC[59]: [[ cycled = \c\y\c\l\e\d ]]
762 + JGLOBAL_STAGE_IC[59]: [[ enkfgdas = \g\d\a\s ]]
763 + JGLOBAL_STAGE_IC[63]: [[ .true. = \.\t\r\u\e\. ]]
764 + JGLOBAL_STAGE_IC[64]: RUN=enkfgdas
765 + JGLOBAL_STAGE_IC[64]: YMD=20240223
766 + JGLOBAL_STAGE_IC[64]: HH=12
767 + JGLOBAL_STAGE_IC[64]: declare_from_tmpl -x COMOUT_ATMOS_RESTART_PREV:COM_ATMOS_RESTART_TMPL
768 + bash_utils.sh[35]: [[ NO == \N\O ]]
769 + bash_utils.sh[35]: set +x
770 declare_from_tmpl :: COMOUT_ATMOS_RESTART_PREV=/lfs/h2/emc/ptmp/kate.friedman/comrot/testcyc_C96_ATM_WARM/enkfgdas.20240223/12//model/atmos/restart
```
Log on Cactus: `/lfs/h2/emc/ptmp/kate.friedman/comrot/testcyc_C96_ATM_WARM/logs/2024022318/enkfgdasstage_ic.log.0`

Here is a similar snippet and log path for the gefs staging job that did not have issues getting `MEMDIR` into the declare command:
```
648 + JGLOBAL_STAGE_IC[44]: MEMDIR_ARRAY+=("mem$(printf "%03d" "${ii}")")
649 ++ JGLOBAL_STAGE_IC[44]: printf %03d 2
650 + JGLOBAL_STAGE_IC[43]: (( ii++ ))
651 + JGLOBAL_STAGE_IC[43]: (( ii <= 2 ))
652 + JGLOBAL_STAGE_IC[51]: err=0
653 + JGLOBAL_STAGE_IC[54]: for MEMDIR in "${MEMDIR_ARRAY[@]}"
654 + JGLOBAL_STAGE_IC[60]: [[ forecast-only = \c\y\c\l\e\d ]]
655 + JGLOBAL_STAGE_IC[64]: [[ .false. = \.\t\r\u\e\. ]]
656 + JGLOBAL_STAGE_IC[68]: YMD=20210323
657 + JGLOBAL_STAGE_IC[68]: HH=12
658 + JGLOBAL_STAGE_IC[68]: declare_from_tmpl -x COMOUT_ATMOS_INPUT:COM_ATMOS_INPUT_TMPL
659 + bash_utils.sh[35]: [[ NO == \N\O ]]
660 + bash_utils.sh[35]: set +x
661 declare_from_tmpl :: COMOUT_ATMOS_INPUT=/lfs/h2/emc/ptmp/kate.friedman/comrot/teststage_C48_S2SWA_gefs/gefs.20210323/12/mem000/model/atmos/input
```
Log on Cactus: `/lfs/h2/emc/ptmp/kate.friedman/comrot/teststage_C48_S2SWA_gefs/logs/2021032312/stage_ic.log`

If I add `MEMDIR=${MEMDIR}` to the declare command I don't need to export `MEMDIR` and the issue does not persist:
```
    MEMDIR=${MEMDIR} RUN=${rRUN} YMD=${previous_cycle:0:8} HH=${previous_cycle:8:2} declare_from_tmpl -x COMOUT_ATMOS_RESTART_PREV:COM_ATMOS_RESTART_TMPL
```
```
758 + JGLOBAL_STAGE_IC[54]: for MEMDIR in "${MEMDIR_ARRAY[@]}"
759 + JGLOBAL_STAGE_IC[56]: echo mem001
760 mem001
761 + JGLOBAL_STAGE_IC[59]: [[ cycled = \c\y\c\l\e\d ]]
762 + JGLOBAL_STAGE_IC[59]: [[ enkfgdas = \g\d\a\s ]]
763 + JGLOBAL_STAGE_IC[63]: [[ .true. = \.\t\r\u\e\. ]]
764 + JGLOBAL_STAGE_IC[64]: MEMDIR=mem001
765 + JGLOBAL_STAGE_IC[64]: RUN=enkfgdas
766 + JGLOBAL_STAGE_IC[64]: YMD=20240223
767 + JGLOBAL_STAGE_IC[64]: HH=12
768 + JGLOBAL_STAGE_IC[64]: declare_from_tmpl -x COMOUT_ATMOS_RESTART_PREV:COM_ATMOS_RESTART_TMPL
769 + bash_utils.sh[35]: [[ NO == \N\O ]]
770 + bash_utils.sh[35]: set +x
771 declare_from_tmpl :: COMOUT_ATMOS_RESTART_PREV=/lfs/h2/emc/ptmp/kate.friedman/comrot/testcyc_C96_ATM_WARM/enkfgdas.20240223/12/mem001/model/atmos/restart
```
Log on Cactus: `/lfs/h2/emc/ptmp/kate.friedman/comrot/testcyc_C96_ATM_WARM/logs/2024022318/enkfgdasstage_ic.log`

`MEMDIR=${MEMDIR}` will be empty when not using members, just ran a forecast-only CI test staging job to test.

Thoughts on why `MEMDIR` doesn't get into the declare command for the enkfgdas job but did for the gefs job?

ush/python/pygfs/task/stage.py
KateFriedman-NOAA: My bad, added @ e14e6232, thanks!

parm/config/gefs/config.base
KateFriedman-NOAA: I have now moved both `BASE_IC` and `ICSDIR` from `config.base` to `config.stage_ic`. Updated `workflow/setup_expt.py` to accommodate the variable move. The `ICSDIR` is either provided by the user via `--icsdir` or is the default `''`. If `ICSDIR` is not set by the user then it is defined in the remaining part of `config.stage_ic`.

KateFriedman-NOAA: Committed to my branch @ a0b9af07

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: Removed incorrect comment that was valid before making updates to export. Thanks for catching!

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: I committed the `MEMDIR=${MEMDIR}` update to my branch @ https://github.com/NOAA-EMC/global-workflow/pull/2651/commits/55e47c67cd26c1b12f65d5278a09b851c63d8d78

jobs/JGLOBAL_STAGE_IC
WalterKolczynski-NOAA: Well, the ex-script could've done the same thing and called the python. But I'm fine leaving it.

jobs/JGLOBAL_STAGE_IC
WalterKolczynski-NOAA: > Thoughts on why `MEMDIR` doesn't get into the declare command for the enkfgdas job but did for the gefs job?

MEMDIR is marked for export in the gefs `config.base` (but not the gfs one). While we tend to re-export every time we change a variable, "exported" is really just a condition that attaches to the variable, even if it changes.

The variable has to be exported because `envsubst` (which `declare_from_tmpl()` uses under the hood) only considers exported variables (I forgot this).

jobs/JGLOBAL_STAGE_IC
```diff
@@ -53,38 +52,28 @@ err=0
...
 for MEMDIR in "${MEMDIR_ARRAY[@]}"; do
 
-  # Export MEMDIR; need even if empty
-  export MEMDIR
-
   # Declare COMs
   if [[ "${MODE}" = "cycled" && "${RUN}" = "gdas" ]]; then
-    YMD=${current_cycle:0:8} HH=${current_cycle:8:2} declare_from_tmpl COMOUT_ATMOS_ANALYSIS:COM_ATMOS_ANALYSIS_TMPL
-    export COMOUT_ATMOS_ANALYSIS
+    YMD=${current_cycle:0:8} HH=${current_cycle:8:2} declare_from_tmpl -x COMOUT_ATMOS_ANALYSIS:COM_ATMOS_ANALYSIS_TMPL
```
WalterKolczynski-NOAA: This is different than it was before because `MEMDIR` is not being used in the path.

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: Could you provide an example of calling the `declare_from_tmpl` function from the python ex-script? I attempted and couldn't figure it out.

DavidHuber-NOAA: This can actually be done right in the Jinja template using the `replace_tmpl` filter provided by `wxflow`.  See `parm/archive/master_enkf.yaml.j2` as an example:
https://github.com/NOAA-EMC/global-workflow/blob/8fee36f0307b0c08080e3a8fa8fa6703e7da5fce/parm/archive/master_enkf.yaml.j2#L52-L71

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: > The variable has to be exported because envsubst (which declare_from_tmpl() uses under the hood) only considers exported variables (I forgot this).

Good to know!

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: `COMOUT_ATMOS_ANALYSIS` didn't use `MEMDIR` before the last change and still doesn't now. It's only declared when cycled and gdas, which creates an empty `MEMDIR_ARRAY` since it doesn't use `MEMDIR`. I'm running a low-res ATM cycled test on WCOSS2 to be sure all is well after the changes I made based on reviewer feedback. Here is the `COMOUT_ATMOS_ANALYSIS` declaration within the `gdasstage_ic` job log:
```
+ JGLOBAL_STAGE_IC[34]: MEMDIR_ARRAY=()
+ JGLOBAL_STAGE_IC[35]: [[ gdas = \e\n\k\f\g\d\a\s ]]
+ JGLOBAL_STAGE_IC[35]: [[ gdas = \g\e\f\s ]]
+ JGLOBAL_STAGE_IC[46]: MEMDIR_ARRAY+=("")
+ JGLOBAL_STAGE_IC[50]: err=0
+ JGLOBAL_STAGE_IC[53]: for MEMDIR in "${MEMDIR_ARRAY[@]}"
+ JGLOBAL_STAGE_IC[56]: [[ cycled = \c\y\c\l\e\d ]]
+ JGLOBAL_STAGE_IC[56]: [[ gdas = \g\d\a\s ]]
+ JGLOBAL_STAGE_IC[57]: YMD=20211220
+ JGLOBAL_STAGE_IC[57]: HH=18
+ JGLOBAL_STAGE_IC[57]: declare_from_tmpl -x COMOUT_ATMOS_ANALYSIS:COM_ATMOS_ANALYSIS_TMPL
+ bash_utils.sh[35]: [[ NO == \N\O ]]
+ bash_utils.sh[35]: set +x
declare_from_tmpl :: COMOUT_ATMOS_ANALYSIS=/lfs/h2/emc/ptmp/kate.friedman/comrot/testcycstage/gdas.20211220/18//analysis/atmos
```

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: Thanks @DavidHuber-NOAA ! Let me try to incorporate something similar into my staging branch...

WalterKolczynski-NOAA (dismissed):
Conditionally approved

WalterKolczynski-NOAA (dismissed):

parm/stage/stage.yaml.j2
```diff
@@ -0,0 +1,195 @@
...
+        {% for ftype in ["ca_data", "fv_core.res", "fv_srf_wnd.res", "fv_tracer.res", "phy_data", "sfc_data"] %}
+        - ["{{ ICSDIR }}/{{ COMOUT_ATMOS_RESTART_PREV | relpath(ROTDIR) }}/{{ m_prefix }}.{{ ftype }}.tile{{ ntile }}.nc", "{{ COMOUT_ATMOS_RESTART_PREV }}/{{ m_prefix }}.{{ ftype }}.nest0{{ ntile-5 }}.tile{{ ntile }}.nc"]
+        {% endfor %}
+  {% else %} # cold start
+    {% set COMOUT_ATMOS_INPUT = COM_ATMOS_INPUT_TMPL | replace_tmpl(current_dict) %}
+    mkdir:
+        - "{{ COMOUT_ATMOS_INPUT }}"
+    copy:
+        {% for ftype in ["gfs_data", "sfc_data"] %}
+        - ["{{ COMOUT_ATMOS_INPUT }}/{{ ftype }}.tile{{ ntile }}.nc", "{{ COMOUT_ATMOS_INPUT }}/{{ ftype }}.nest0{{ ntile-5 }}.tile{{ ntile }}.nc"]
```
DavidHuber-NOAA: Just checking, should the nest number really be tied to `ntile-5` or should it just be `nest02`?

DavidHuber-NOAA (dismissed):
Looks good.

parm/stage/stage.yaml.j2
KateFriedman-NOAA: Good question. So the current functionality in the staging job script is to set it based on tile-5 in the warm-start block:
```
if (( tt > 6 )) ; then
            tgt="${COM_ATMOS_RESTART_PREV}/${DTG_PREFIX}.${ftype}.nest0$((tt-5)).tile${tt}.nc"
```
https://github.com/NOAA-EMC/global-workflow/blob/develop/scripts/exglobal_stage_ic.sh#L52
...and when cold-start, it hardcodes the resulting "2":
```
if (( ntiles > 6 )); then
        ${NLN} "${COM_ATMOS_INPUT}/${ftype}.tile7.nc" "${COM_ATMOS_INPUT}/${ftype}.nest02.tile7.nc"
```
https://github.com/NOAA-EMC/global-workflow/blob/develop/scripts/exglobal_stage_ic.sh#L82

With the nest, `ntiles=7` so both ways result in the same thing right now = `nest02`.

I chose to go with `nest0{{ ntile-5 }}` to avoid hardcoding the "2" in the filename and to allow for higher `tile` numbers if nests did that. Let me double check with the folks who added the nest functionality though and make sure they didn't need that hardcoded version.

KateFriedman-NOAA: @guoqing-noaa Please see my comment above. Does it matter if the filename for the nest IC is built using `nest0{{ ntile-5 }}` or hardcoded with a 2 (`nest02`). Is there a reason to set it differently in different circumstances? Let me know, thanks!

KateFriedman-NOAA:
@weihuang-jedi @AnilKumar-NOAA for supporting the updates in this staging job refactor PR, you'll need to create some symlinks within the AWS and Gaea `ICSDIR` folders. Here is what the `ICSDIR` on Hera looks like currently in terms of resolution folders and ignoring the `prototype_ICs` folder:

```
[Kate.Friedman@hfe11 ICSDIR]$ pwd
/scratch1/NCEPDEV/global/glopara/data/ICSDIR
[Kate.Friedman@hfe11 ICSDIR]$ ll | grep C | grep -v proto
lrwxrwxrwx 1 role.glopara global   10 Aug  1 18:29 C1152 -> C1152mx025
drwxr-sr-x 3 role.glopara global 4096 Jul 22 15:10 C1152mx025
drwxr-sr-x 5 role.glopara global 4096 Jun 21 13:10 C384C192
drwxr-sr-x 4 role.glopara global 4096 Jun 24 15:19 C384mx025
lrwxrwxrwx 1 role.glopara global   11 Jul  1 15:20 C48 -> C48C48mx500
drwxr-sr-x 7 role.glopara global 4096 Jun 21 17:09 C48C48mx500
lrwxrwxrwx 1 role.glopara global   11 May  1  2023 C48mx500 -> C48C48mx500
lrwxrwxrwx 1 role.glopara global    9 Aug  1 18:29 C768 -> C768mx025
drwxr-sr-x 3 role.glopara global 4096 Jul 22 15:09 C768mx025
lrwxrwxrwx 1 role.glopara global    6 Jul 17 18:13 C96 -> C96C48
drwxr-sr-x 8 role.glopara global 4096 Jun 24 14:51 C96C48
```

You'll need to do the following to add symlinks that the refactored staging job will use:

1. create the symlinks seen in that listing above for the various resolutions
2. pull the `20240610` symlink folders under each of the C* folders (making sure to preserve the symlinks and not pull through the symlinks)

For 2, below I show the timestamp (`20240610`) folders, which contain the COM structure we currently use and symlinks that point to the various ICs that are spread out under the `ICSDIR` and `ICSDIR/prototype_ICs` folders. The timestamp folder symlinks standardize the ICs source.
```
C1152mx025/20240610
C384C192/20240610
C384mx025/20240610
C48C48mx500/20240610
C768mx025/20240610
C96C48/20240610
```

If it would be helpful, I'm happy to do a Meet with one or both of you and provide guidance on making these updates. Let me know, thanks!

Also, @weihuang-jedi, this PR will update `BASE_CPLIC` to `BASE_IC` and remove the "/prototype_ICs" text from `workflow/hosts/awspw.yaml`:
```
[Kate.Friedman@hfe10 feature-issue_2475]$ git diff workflow/hosts/awspw.yaml
diff --git a/workflow/hosts/awspw.yaml b/workflow/hosts/awspw.yaml
index f925f540..a9c70825 100644
--- a/workflow/hosts/awspw.yaml
+++ b/workflow/hosts/awspw.yaml
@@ -18,7 +18,7 @@ CHGRP_RSTPROD: 'YES'
 CHGRP_CMD: 'chgrp rstprod' # TODO: This is not yet supported.
 HPSSARCH: 'NO'
 HPSS_PROJECT: emc-global #TODO: See `ATARDIR` below.
-BASE_CPLIC: '/bucket/global-workflow-shared-data/ICSDIR/prototype_ICs'
+BASE_IC: '/bucket/global-workflow-shared-data/ICSDIR'
 LOCALARCH: 'NO'
```
Committed to my branch @ ff1246b

jobs/JGLOBAL_STAGE_IC
```diff
@@ -3,15 +3,38 @@
...
-# Restart conditions for GFS cycle come from GDAS
+# Define significant cycles
 # shellcheck disable=SC2153
-rCDUMP=${RUN}
-# shellcheck disable=SC2153
-[[ ${RUN} = "gfs" ]] && export rCDUMP="gdas"
-export rCDUMP
+half_window=$(( assim_freq / 2 ))
+current_cycle="${PDY}${cyc}"
+previous_cycle=$(date --utc -d "${PDY} ${cyc} - ${assim_freq} hours" +%Y%m%d%H)
```
aerorahul: These are created/defined in `task.py` based on `PDY` and `cyc`.

jobs/JGLOBAL_STAGE_IC
```diff
@@ -3,15 +3,38 @@
...
+if [[ "${DOIAU:-NO}" == "YES" ]] && [[ "${MODE}" == "cycled" ]] ; then
+  model_start_date_current_cycle="${current_cycle_begin}"
+else
+  if [[ "${REPLAY_ICS:-NO}" == "YES" ]]; then
+    model_start_date_current_cycle=${current_cycle_end}
+  else
+    model_start_date_current_cycle=${current_cycle}
+  fi
+fi
+export current_cycle previous_cycle model_start_date_current_cycle
```
aerorahul: These could be in the `__init__` of `stage.py`. Also, it should be called `stage_ic.py`, as that is the function.

jobs/JGLOBAL_STAGE_IC
```diff
@@ -3,15 +3,38 @@
...
+err=0
+
+# Execute staging
+"${SCRgfs}/exglobal_stage_ic.py"
 
-# Execute the Script
-"${SCRgfs}/exglobal_stage_ic.sh"
+###############################################################
+# Check for errors and exit if any of the above failed
+if [[ "${err}" -ne 0 ]]; then
```
aerorahul, line 12: When will this be not 0?  It is set to 0 on line 27 and never reset?

parm/stage/stage.yaml.j2
```diff
@@ -0,0 +1,204 @@
...
+# - DO_ATM, DO_OCN, DO_ICE, etc.
+# For a full list see scripts/exglobal_stage_ic.py
+#############################################################
+
+# Set variables used below
+{% set current_cycle_YMD = current_cycle | to_YMD %}
+{% set current_cycle_HH = current_cycle | strftime("%H") %}
+{% set previous_cycle_YMD = previous_cycle | to_YMD %}
+{% set previous_cycle_HH = previous_cycle | strftime("%H") %}
+{% set m_prefix = model_start_date_current_cycle | to_YMD + "." + model_start_date_current_cycle | strftime("%H") + "0000" %}
```
aerorahul: If `strftime` is being used to format, one can pass the `0000` as part of the format, or one can completely take over `strftime` to do exactly what is needed.  Such as:
```suggestion
{% set m_prefix = model_start_date_current_cycle | strftime("%Y%m%d.%H0000") %}
```

parm/stage/stage.yaml.j2
```diff
@@ -0,0 +1,204 @@
...
+{% set p_prefix = previous_cycle | to_YMD + "." + previous_cycle | strftime("%H") + "0000" %}
```
aerorahul: same here

parm/stage/stage.yaml.j2
```diff
@@ -0,0 +1,204 @@
...
+                         '${RUN}':RUN,
+                         '${YMD}':current_cycle_YMD,
+                         '${HH}':current_cycle_HH,
+                         '${MEMDIR}': mem_char }) %}
+{% set previous_dict = ({ '${ROTDIR}':ROTDIR,
+                          '${RUN}':rRUN,
+                          '${YMD}':previous_cycle_YMD,
+                          '${HH}':previous_cycle_HH,
+                          '${MEMDIR}': mem_char }) %}
+{% set previous_run_dict = ({ '${ROTDIR}':ROTDIR,
```
aerorahul: This is called `previous_run_dict, but the `RUN` is not `previous`.  The difference   between this and the one on 45 is ambiguous.
previous_cycle_dict, previous_cycle_and_run_dict.

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: Good to know, will remove from script, thanks!

parm/stage/stage.yaml.j2
KateFriedman-NOAA: Good point, I have made these dicts less ambiguous @ 7ae9ea49. Thanks!

parm/stage/stage.yaml.j2
KateFriedman-NOAA: Thanks for the feedback! I have incorporated this @ 7ae9ea49.

parm/stage/stage.yaml.j2
KateFriedman-NOAA: Also incorporated @ 7ae9ea49

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: You are correct, have added missing error value check after existing script. Committed to branch @ 7ae9ea49

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: Have consolidated this and other configuration blocks into a new `configure` function. Committed to branch @ 7ae9ea49. Thanks!

jobs/JGLOBAL_STAGE_IC
```diff
@@ -3,15 +3,19 @@
...
 source "${HOMEgfs}/ush/jjob_header.sh" -e "stage_ic" -c "base stage_ic"
 
-# Restart conditions for GFS cycle come from GDAS
-# shellcheck disable=SC2153
-rCDUMP=${RUN}
-# shellcheck disable=SC2153
-[[ ${RUN} = "gfs" ]] && export rCDUMP="gdas"
-export rCDUMP
+# Initialize return code
+err=0
```
aerorahul: No longer necessary.

parm/config/gefs/config.stage_ic
```diff
@@ -7,32 +7,27 @@ echo "BEGIN: config.stage_ic"
...
-    ;;
-esac
+export ICSDIR="@ICSDIR@" # User provided ICSDIR; blank if not provided
+export BASE_IC="@BASE_IC@" # Platform home for staged ICs
+
+export STAGE_IC_YAML_TMPL="${PARMgfs}/stage/stage.yaml.j2"
+
+# Set ICSDIR
+
+if [[ -z "${ICSDIR}" ]] ; then
```
aerorahul: If User does not provide and `ICSDIR` remains` @ICSDIR@`, the if statement will be executed.
What if ICSDIR and BASE_IC are both not provided.
Why do we need both?

parm/stage/stage.yaml.j2
```diff
@@ -0,0 +1,204 @@
+#############################################################
```
aerorahul: Not sure why we have a singular `stage.yaml.j2` for both GFS and GEFS.
There would be a `stage_gfs_default.yaml.j2` and `stage_gefs_default.yaml.j2`
The `config/gfs|gefs/stage_ic.sh` could point to the default ones for their respective yamls.

ush/python/pygfs/task/stage_ic.py
```diff
@@ -0,0 +1,112 @@
...
+        config : Dict[str, Any]
+            Incoming configuration for the task from the environment
+
+        Returns
+        -------
+        None
+        """
+        super().__init__(config)
+
+        self.task_config = AttrDict(**self.task_config)
```
aerorahul: This line is redundant as it assigns `self.task_config` to itself.

ush/python/pygfs/task/stage_ic.py
```diff
@@ -0,0 +1,112 @@
...
+        Returns
+        -------
+        None
+        """
+
+        if not os.path.isdir(stage_dict.ROTDIR):
+            raise FileNotFoundError(f"FATAL ERROR: The ROTDIR ({stage_dict.ROTDIR}) does not exist!")
+
+        # Loop over members
+        for mem in range(stage_dict.first_mem, stage_dict.last_mem + 1):
```
aerorahul: Looping over members means that the bias correction and deterministic ICs will be copied over and over.
Is this correct?

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: Will remove, thanks!

ush/python/pygfs/task/stage_ic.py
KateFriedman-NOAA: Will remove, thanks!

ush/python/pygfs/task/stage_ic.py
KateFriedman-NOAA: There are separate jobs based on `RUN`. I see I failed to put that in the main PR comment, I have now added that.

When a cycled experiment is set up there is a `gdasstage_ic` job and an `enkfgdasstage_ic` job. The `enkfgdasstage_ic` job won't invoke the block in `stage.yaml.j2` that copies the abias*/radstat files because that block is wrapped in a check of whether `RUN == "gdas"`. See: https://github.com/KateFriedman-NOAA/global-workflow/blob/feature/issue_2475/parm/stage/stage.yaml.j2#L58

See these two logs on Hera for examples:
```
/scratch1/NCEPDEV/stmp4/Kate.Friedman/comrot/teststage_C96C48_ATM/logs/2021122018/gdasstage_ic.log
/scratch1/NCEPDEV/stmp4/Kate.Friedman/comrot/teststage_C96C48_ATM/logs/2021122018/enkfgdasstage_ic.log
```

ush/python/pygfs/task/stage_ic.py
KateFriedman-NOAA: Removed @ 30226501

jobs/JGLOBAL_STAGE_IC
KateFriedman-NOAA: Removed @ 30226501

parm/config/gefs/config.stage_ic
KateFriedman-NOAA: `ICSDIR` will either be a path provided by the user or it will be empty (`''`):
```
parser.add_argument('--icsdir', help='full path to user initial condition directory', type=str, required=False, default='')
```
For most of my testing I do not provide an `ICSDIR` and the variable in the configs ends up empty:
```
export ICSDIR="" # User provided ICSDIR; blank if not provided
```
The check on line 16 of `config.stage_ic` then gets invoked. Here is an example from a `gdasstage_ic` job where I did not provide an `ICSDIR`:
```
+++ config.stage_ic[10]: export ICSDIR=
+++ config.stage_ic[10]: ICSDIR=
+++ config.stage_ic[11]: export BASE_IC=/scratch1/NCEPDEV/global/glopara/data/ICSDIR
+++ config.stage_ic[11]: BASE_IC=/scratch1/NCEPDEV/global/glopara/data/ICSDIR
+++ config.stage_ic[13]: export STAGE_IC_YAML_TMPL=/scratch1/NCEPDEV/global/Kate.Friedman/git/feature-issue_2475/parm/stage/stage.yaml.j2
+++ config.stage_ic[13]: STAGE_IC_YAML_TMPL=/scratch1/NCEPDEV/global/Kate.Friedman/git/feature-issue_2475/parm/stage/stage.yaml.j2
+++ config.stage_ic[16]: [[ -z '' ]]
+++ config.stage_ic[18]: ic_ver=20240610
+++ config.stage_ic[20]: ((  NMEM_ENS > 0  ))
+++ config.stage_ic[21]: ensic=C48
+++ config.stage_ic[24]: [[ NO == \Y\E\S ]]
+++ config.stage_ic[28]: export ICSDIR=/scratch1/NCEPDEV/global/glopara/data/ICSDIR/C96C48/20240610
+++ config.stage_ic[28]: ICSDIR=/scratch1/NCEPDEV/global/glopara/data/ICSDIR/C96C48/20240610
+++ config.stage_ic[32]: echo 'END: config.stage_ic'
```

`BASE_IC` comes from the host file and will (should) always be set. It is used to build `ICSDIR` when `ICSDIR` isn't provided by the user or otherwise set:
```
export ICSDIR="${BASE_IC}/${CASE}${ensic:-}${ocnic:-}/${ic_ver}"
```

parm/stage/stage.yaml.j2
KateFriedman-NOAA: The current staging job only differentiated GFS/GEFS based on what the starting member would be:
https://github.com/NOAA-EMC/global-workflow/blob/develop/scripts/exglobal_stage_ic.sh#L15

Otherwise, they overlap and allow for a single yaml. A gefs version of the stage yaml would be a copy paste from the current one for all of the sections it needs.

If you would prefer separate yamls based on GFS vs GEFS, then I can do that. Let me know.

aerorahul: I think having separate yamls makes sense, otherwise the complexity to fold everything in a single yaml becomes unmanageable. 

At setup_expt.py time, we know whether its `gfs|gefs`, and we also know whether it is `forecast-only|cycled`, it shouldn't be that difficult to put in the appropriate `stage.yaml.j2` into `STAGE_IC_YAML_TMPL`.

ush/python/pygfs/task/stage_ic.py
```diff
@@ -0,0 +1,110 @@
...
+        stage_dict['rRUN'] = rRUN
+
+        # Determine ensemble member settings
+        if self.task_config.NMEM_ENS > 0 and self.task_config.RUN in ['enkfgdas', 'gefs']:  # Ensemble RUN
+            if self.task_config.RUN == "gefs":
+                stage_dict['first_mem'] = 0
+                stage_dict['last_mem'] = self.task_config.NMEM_ENS
+            elif self.task_config.RUN == "enkfgdas":
+                stage_dict['first_mem'] = 1
+                stage_dict['last_mem'] = self.task_config.NMEM_ENS
```
aerorahul: Isn't the stage_ic job only run when `RUN=gdas` when doing `cycled` regardless of ensemble members.

aerorahul: Ok. I see the next comment that we now have RUN specific stage_ic jobs.
If thats the case, I would suggest we move this block calculation to the relevant yaml.

TerrenceMcGuinness-NOAA:
Just shut down this Jenkins `Build #3`  Looks like we need to clean up that process up a bit so it doesn't spit out all these fail messages like that on shutdown.

workflow/rocoto/gfs_tasks.py
```diff
@@ -934,8 +883,8 @@ def _fcst_cycled(self):
 
         dependencies = rocoto.create_dependency(dep_condition='and', dep=dependencies)
 
-        if self.run in ['gdas']:
-            dep_dict = {'type': 'cycleexist', 'condition': 'not', 'offset': f"-{timedelta_to_HMS(self._base['cycle_interval'])}"}
+        if self.run in ['gdas', 'enkfgdas']:
```
aerorahul: Why is `enkfgdas` a dependency here? This creates either the `gdasfcst` or the `gfsfcst` task for the cycled.

aerorahul: Also, `gdasstage_ic` will only exist for the first half cycle, so the offset to check if the previous cycle exists should stay, no?

aerorahul (dismissed):

workflow/rocoto/gfs_tasks.py
KateFriedman-NOAA: > Why is enkfgdas a dependency here? This creates either the gdasfcst or the gfsfcst task for the cycled.

Removed @ a53c97e2, thanks!

> Also, gdasstage_ic will only exist for the first half cycle, so the offset to check if the previous cycle exists should stay, no?

The check for a prior cycle is only needed for the fcst job when there aren't any jobs preceding it. Now that the staging job runs before it, we don't need the prior cycle check. The check of the staging job completion replaces it.

Should we make it possible to disable the staging job, then yes, we'd want the prior cycle check for the first half cycle fcst job again.

parm/stage/stage.yaml.j2
KateFriedman-NOAA: Have separated the main yaml into master ones for gfs and gefs, and also broke the contents out into component yamls. The master ones handle cycle/mem loop setup and then do includes of the component yamls. Have tested successfully on Hera.

aerorahul (dismissed):

workflow/setup_expt.py
```diff
@@ -294,7 +65,17 @@ def _update_defaults(dict_in: dict) -> dict:
...
-    # loop over other configs and update them
+    # Update stage config
+    stage_dict = {
+        "@ICSDIR@": inputs.icsdir
+    }
+    host_dict = get_template_dict(host.info)
+    stage_dict = dict(stage_dict, **host_dict)
+    stage_input = f'{inputs.configdir}/config.stage_ic'
+    stage_output = f'{inputs.expdir}/{inputs.pslot}/config.stage_ic'
+    edit_config(stage_input, stage_output, stage_dict)
```
aerorahul, line 76: What is going on here? Why is `config.stage_ic` handled specially?

WalterKolczynski-NOAA (dismissed):

workflow/setup_expt.py
KateFriedman-NOAA: This area of `setup_expt.py` updates templated variables in the configs and places them in the `EXPDIR`. However, other than `config.base`, the other configs handled in this section are just filled in with defaults from a default yaml. I needed to handle `config.stage_ic` like `config.base` and get the user input (or empty default) and host file provided `BASE_IC` into it. There is an entire function just to update `config.base`, that was overkill for `config.stage_ic`, so I added the lines you see above.

parm/stage/master_gfs.yaml.j2
```diff
@@ -0,0 +1,189 @@
...
+{% set COMOUT_WAVE_RESTART_PREV_MEM_list = [] %}
+
+# Construct member COM directory lists
+# ------------------------------------
+{% for mem in range(first_mem, last_mem + 1) %}
+
+    {% if mem >= 0 %}
+        {% set mem_char = 'mem%03d' | format(mem) %}
+    {% else %}
+        {% set mem_char = '' %}
```
DavidHuber-NOAA, line 84: Clever!

parm/stage/master_gfs.yaml.j2
```diff
@@ -0,0 +1,189 @@
...
+# Determine restart RUN
+# ---------------------
+{% set rRUN = RUN %}
+{% if RUN == "gfs" %}
+    {% set rRUN = "gdas" %}
+{% endif %}
+
+# Set first/last mem for loop
+# ---------------------------
+{% if NMEM_ENS > 0 and RUN == "enkfgdas" %} # Ensemble RUN
```
DavidHuber-NOAA: If `NMEM_ENS == 0` and `RUN == "enkfgdas"`, will this raise an error elsewhere?  Or should this be treated as a valid deterministic case?

ci/cases/pr/C48mx500_3DVarAOWCDA.yaml
```diff
@@ -9,7 +9,7 @@ arguments:
   resdetocean: 5.0
   comroot: {{ 'RUNTESTS' | getenv }}/COMROOT
   expdir: {{ 'RUNTESTS' | getenv }}/EXPDIR
-  icsdir: {{ 'ICSDIR_ROOT' | getenv }}/C48mx500
+  icsdir: {{ 'ICSDIR_ROOT' | getenv }}/C48mx500/20240610
```
DavidHuber-NOAA, line 12: Not for this PR, but I wonder if we could add a `ci/platforms/config.common` that specifies the IC date (e.g. `20240610`).

DavidHuber-NOAA (dismissed):
Looks good!

parm/stage/master_gfs.yaml.j2
KateFriedman-NOAA: So, the reasoning behind this line...I added the 2nd condition (`RUN == "enkfgdas"`) because when it runs for `RUN=gdas` it was going into this block since `NMEM_ENS>0` is set for the whole shebang (e.g. `NMEM_ENS=2` for the CI tests with enkf). Adding the check of enkfgdas pushes the gdas instance into the other part of the block. Let me know if I didn't answer your question.

ci/cases/pr/C48mx500_3DVarAOWCDA.yaml
KateFriedman-NOAA: I was thinking something similar, although along the lines of creating a new versions file. The versions are all the same at the moment so it would be needed if we have a new version of an existing IC set in the future. Either way, it would be a good follow-up PR to do something to make the timestamp a variable and have it set once place.

parm/stage/master_gfs.yaml.j2
DavidHuber-NOAA: I see, just wondering if this could be simplified to
```
{% if RUN == "enkfgdas" %} # Ensemble RUN
```
since `NMEM_ENS` _should_ be > 0.  As is, if a user changes `NMEM_ENS` to 0 without rerunning `setup_xml.py`, this will cause the `enkfgdasstage_ic` job to go down the deterministic branch.  An unlikely situation, but one that might be tricky to debug.  But this would be more than a `stage_ic` problem and probably the best way to handle it would be to add a sanity `RUN == enkfgdas && NMEM_ENS <= 0`.

All of that is probably overkill, but seeing the conditional here made me wonder.  I think another issue to add a sanity check script to, say, `jjob-header.sh` would be the right way to handle these kind of edge cases.

KateFriedman-NOAA: > if a user changes NMEM_ENS to 0 without rerunning `setup_xml.py`

Ah, I see your point and I like your suggestion. I will adjust this block.

KateFriedman-NOAA: I am going to wait to see how the CI tests run on Hera before committing the change. Any new commits will blow away the current approvals and I might as well wait to make sure no further fixes are needed from the CI testing.

DavidHuber-NOAA: Sounds good to me.

KateFriedman-NOAA: Committed to PR branch @ 13dfad25

DavidHuber-NOAA (approved):
Looks good!

KateFriedman-NOAA:
@NeilBarton-NOAA This PR has just been merged into `develop`. Let me know how I can help update your PR branch to accommodate the staging job refactor.

NeilBarton-NOAA:
Thanks @KateFriedman-NOAA, I'm currently testing, and I will let you know when I have questions.

guillaumevernieres:
~@KateFriedman-NOAA , It looks like this PR broke the WCDA/gfsv17 tests. Was that tested? I see the hera-ci green but I don't see what was actually tested.~

Edit: just seeing @RussTreadon-NOAA 's issue ... I suspect that's my issue as well, so ignore for now

PR #2657, opened 2024-06-04
Optimize wavepostpnt

DavidHuber-NOAA:
# Description
Optimizes the gfswavepostpnt job.

This is done by
1) reducing the number of calls to `sed`, `awk`, `grep`, and `cat` by
    - performing operations on all files at once instead of looping over each file
    - removing piped `cat` calls (e.g. `cat <file> | sed 'something'`)
    - combining `sed` and `grep` calls when possible
    - adding logic to `awk` calls instead of handling that logic in bash
2) minimizing as much as possible the amount of data on disk that has to be read in (e.g. limiting sed to read only the line numbers it needs)

# Type of change
- Maintenance (optimization)

# Change characteristics

# How has this been tested?
C48_S2SW test case on Hercules.  Outputs were identical to a non-optimized run.  The optimized run took 80 minutes while the non-optimized version took 191 minutes.

# Checklist


scripts/exgfs_wave_post_pnt.sh
```diff
@@ -156,7 +156,7 @@ source "${USHgfs}/preamble.sh"
     cp -f ${PARMgfs}/wave/wave_${NET}.buoys buoy.loc.temp
     if [ "$DOBNDPNT_WAV" = YES ]; then
       #only do boundary points
-      sed -n '/^\$.*/!p' buoy.loc.temp | grep IBP > buoy.loc || {
+      sed -n '/^\$.*IBP.*/!p' buoy.loc.temp > buoy.loc || {
```
WalterKolczynski-NOAA: Can we split the buoy lists into boundary and not and avoid this step entirely? I don't think there's any reason that this filtering needs to be done every time.

scripts/exgfs_wave_post_pnt.sh
```diff
@@ -566,9 +554,9 @@ source "${USHgfs}/preamble.sh"
...
   then
     if [ ${CFP_MP:-"NO"} = "YES" ]; then
-      ${wavempexec} -n ${wavenproc} ${wave_mpmd} cmdmprogbouy
+      ${wavempexec} -n "${wavenproc}" "${wave_mpmd}" cmdmprogbuoy
     else
-      ${wavempexec} ${wavenproc} ${wave_mpmd} cmdfile.bouy
+      ${wavempexec} "${wavenproc}" "${wave_mpmd}" cmdfile.buoy
     fi
     exit=$?
   else
```
WalterKolczynski-NOAA, line 573: Can be in a later PR, but this should be updated to use `ush/run_mpmd.sh`

scripts/exgfs_wave_post_pnt.sh
```diff
@@ -553,7 +541,7 @@ source "${USHgfs}/preamble.sh"
     done
   fi
 
-  wavenproc=$(wc -l cmdfile.bouy | awk '{print $1}')
+  wavenproc=$(wc -l cmdfile.buoy | awk '{print $1}')
```
WalterKolczynski-NOAA:
```suggestion
  wavenproc=$(wc -l < cmdfile.buoy)
```

scripts/exgfs_wave_post_pnt.sh
```diff
@@ -503,46 +497,40 @@ source "${USHgfs}/preamble.sh"
...
+    sed "s/^\(.*\)$/${USHgfs//\//\\\/}\/wave_outp_cat.sh \1 ${FHMAX_WAV_PNT} bull > ${CATOUTDIR//\//\\\/}\/bull_cat_\1.out 2>&1/" buoy_lst.txt >> cmdfile.buoy
   fi
 
   if [ ${CFP_MP:-"NO"} = "YES" ]; then
     nfile=0
     ifile=0
     iline=1
     ifirst='yes'
-    nlines=$( wc -l cmdfile.bouy | awk '{print $1}' )
+    nlines=$( wc -l cmdfile.buoy | awk '{print $1}' )
```
WalterKolczynski-NOAA:
```suggestion
    nlines=$( wc -l < cmdfile.buoy)
```

scripts/exgfs_wave_post_pnt.sh
```diff
@@ -385,19 +385,13 @@ source "${USHgfs}/preamble.sh"
     if [ "$DOSPC_WAV" = 'YES' ]
     then
       export dtspec=3600.
-      for buoy in $buoys
-      do
-        echo "${USHgfs}/wave_outp_spec.sh $buoy $ymdh spec $SPECDATA > $SPECDATA/spec_$buoy.out 2>&1" >> tmpcmdfile.$FH3
-      done
+      sed "s/^\(.*\)$/${USHgfs//\//\\\/}\/wave_outp_spec.sh \1 ${ymdh} spec ${SPECDATA//\//\\\/} > ${SPECDATA//\//\\\/}\/spec_\1.out 2>&1/" buoy_lst.txt >> "tmpcmdfile.${FH3}"
```
WalterKolczynski-NOAA: Can we do the (first wave of) substitutions ahead of time in bash to avoid the escape hell here?

WalterKolczynski-NOAA (changes requested):

scripts/exgfs_wave_post_pnt.sh
DavidHuber-NOAA: I don't think we can.  `ww3_shel` and `ww3_multi` expect this file in its entirety, at least if I'm interpreting these lines correctly:

https://github.com/NOAA-EMC/global-workflow/blob/67b833e0c7bc390865d453588b4609a1a7ede981/ush/parsing_namelists_WW3.sh#L102-L105
https://github.com/NOAA-EMC/global-workflow/blob/67b833e0c7bc390865d453588b4609a1a7ede981/ush/parsing_namelists_WW3.sh#L249
https://github.com/NOAA-EMC/global-workflow/blob/67b833e0c7bc390865d453588b4609a1a7ede981/parm/wave/ww3_shel.inp.tmpl#L26
https://github.com/NOAA-EMC/global-workflow/blob/67b833e0c7bc390865d453588b4609a1a7ede981/parm/wave/ww3_multi.inp.tmpl#L21

scripts/exgfs_wave_post_pnt.sh
JessicaMeixner-NOAA: So just as an FYI - the executable this job runs, is going to be completely re-written to actually use MPI instead of having MPI via shell-script.  So I would recommend doing the minimal amount of re-writes to get things working more efficiently while the larger task of completely rewriting this is completed.    The first step towards this is under-review in WW3 now.

DavidHuber-NOAA: Thanks for that notice @JessicaMeixner-NOAA!

@WalterKolczynski-NOAA I pre-escaped the paths and verified that the sed commands were the same.

scripts/exgfs_wave_post_pnt.sh
DavidHuber-NOAA: I think I will pass on this since the job will soon be MPI'd.

WalterKolczynski-NOAA (approved):
Conditionally approved

JessicaMeixner-NOAA:
@DavidHuber-NOAA First and foremost THANK YOU!!!!  I'm about to run some tests w/and w/out this change to compare.  Would it be helpful if I ran this on a particular machine to take a look at certain aspects, or is there output results I can just double check there instead?

DavidHuber-NOAA:
@JessicaMeixner-NOAA With the updates to the UFS coming in yesterday, I think my baseline dataset will be out of date for comparisons, so I think you will want to generate a new baseline from develop.  I believe a test on any system would be fine for comparing outputs from develop and the feature/opt_pnt branch.  But if you would like to see timing differences, then I would recommend running on Hercules since it has had the biggest issues with the gfswavepostpnt job.

JessicaMeixner-NOAA:
@DavidHuber-NOAA Happy to run a new test and thanks again for trying to make everyone's life easier while we work on the longer term solution!

JessicaMeixner-NOAA:
@DavidHuber-NOAA wanted to let you know that I added the boundary point jobs in my test and they timed out, which isn't too surprising to be honest.  I've extended the length and am re-running.  Hopefully will have results by the end of the day.  I will keep you posted!

JessicaMeixner-NOAA:
@DavidHuber-NOAA I'm still having issues with jobs failing and timing out.  I did get the develop branch tests to at least run the main point job, working on the boundary jobs. 

For the test running this PR, I got a failure in the job: 
```
   Executing the boundary point cat script at : Thu Jun  6 15:12:07 CDT 2024
   ------------------------------------

+ exgfs_wave_post_pnt.sh[562]: '[' 200 -gt 1 ']'
+ exgfs_wave_post_pnt.sh[564]: '[' YES = YES ']'
+ exgfs_wave_post_pnt.sh[565]: srun -l --export=ALL -n 200 '--multi-prog --output=mpmd.%j.%t.out' cmdmprogbuoy
srun: unrecognized option '--multi-prog --output=mpmd.%j.%t.out'
Try "srun --help" for more information
+ exgfs_wave_post_pnt.sh[1]: postamble exgfs_wave_post_pnt.sh 1717692662 255
+ preamble.sh[70]: set +x
End exgfs_wave_post_pnt.sh at 20:12:08 with error code 255 (time elapsed: 03:21:06)
+ JGLOBAL_WAVE_POST_PNT[1]: postamble JGLOBAL_WAVE_POST_PNT 1717692626 255
+ preamble.sh[70]: set +x
End JGLOBAL_WAVE_POST_PNT at 20:12:08 with error code 255 (time elapsed: 03:21:42)
+ wavepostpnt.sh[1]: postamble wavepostpnt.sh 1717692622 255
+ preamble.sh[70]: set +x
```
and another error: 

```
+ exgfs_wave_post_pnt.sh[476]: srun -l --export=ALL -n 448 --multi-prog --output=mpmd.%j.%t.out cmdmprog
srun: warning: can't honor --ntasks-per-node set to 40 which doesn't match the requested tasks 448 with the number of requested nodes 12. Ignoring --ntasks-per-node.
srun: error: hercules-01-31: task 381: Exited with exit code 1
srun: Terminating StepId=1420346.0
srun: error: hercules-01-26: tasks 189-225: Terminated
srun: error: hercules-01-31: tasks 374-380,382-410: Terminated
srun: error: hercules-01-27: tasks 226-262: Terminated
srun: error: hercules-01-28: tasks 263-299: Terminated
srun: error: hercules-01-23: tasks 76-113: Terminated
srun: error: hercules-01-29: tasks 300-336: Terminated
srun: error: hercules-01-30: tasks 337-373: Terminated
srun: error: hercules-01-24: tasks 114-151: Terminated
srun: error: hercules-01-25: tasks 152-188: Terminated
srun: error: hercules-01-21: tasks 0-37: Terminated
srun: error: hercules-01-32: tasks 411-447: Terminated
srun: error: hercules-01-22: tasks 38-75: Terminated
srun: Force Terminated StepId=1420346.0
+ exgfs_wave_post_pnt.sh[1]: postamble exgfs_wave_post_pnt.sh 1717692640 143
+ preamble.sh[70]: set +x
End exgfs_wave_post_pnt.sh at 17:21:01 with error code 143 (time elapsed: 00:30:21)
+ JGLOBAL_WAVE_POST_BNDPNTBLL[1]: postamble JGLOBAL_WAVE_POST_BNDPNTBLL 1717692632 143
+ preamble.sh[70]: set +x
```


Log files are here: /work2/noaa/marine/jmeixner/hercules/testpointupdate/update/test01/COMROOT/test01/logs/2021032312 

I'm going to look into this more.  I'm curious if this is because of changes to the line calling wavempexec.  I vaguely remember struggling a lot with the "" on that line.

DavidHuber-NOAA:
@JessicaMeixner-NOAA Thanks for testing.  I'll dig in and see what the issue is.

JessicaMeixner-NOAA:
@DavidHuber-NOAA I can help more next week if I need to as well.  Need to do a couple other things today first though.  I'll check in with you early next week.

DavidHuber-NOAA:
@JessicaMeixner-NOAA thanks!  Hopefully I will be able to get somewhere today.

I did have a couple of ideas on the timing out jobs.  Perhaps if we run a shorter forecast (say 48 hours), the pnt and bullpnt jobs would be able to finish.

If instead we should be testing out to 120 hours, then we may be able to scale back the number of tasks/node to say 20.

Which one would you prefer?  I'll test that way today.

JessicaMeixner-NOAA:
I've just been testing the C48_S2SW and then turning on the boundary points (lowring the forecast hours will help speed up testing/checking, so probably a good idea for faster turn-around).  That being said, I think we'll want whatever is supposed to be running in the CI to work.  The only job that's usually run w/the CI is the point (not the boundaries) and I think that should run okay as is.

WalterKolczynski-NOAA:
> @JessicaMeixner-NOAA thanks! Hopefully I will be able to get somewhere today.
> 
> I did have a couple of ideas on the timing out jobs. Perhaps if we run a shorter forecast (say 48 hours), the pnt and bullpnt jobs would be able to finish.
> 
> If instead we should be testing out to 120 hours, then we may be able to scale back the number of tasks/node to say 20.
> 
> Which one would you prefer? I'll test that way today.

We're testing out the full 16 days, but only on WCOSS with the C96_atm3DVar_extended test, so no wave test.

DavidHuber-NOAA:
@JessicaMeixner-NOAA I fixed a few bugs in my optimizations and all jobs have run to completion for a 24-hour forecast on Hercules.  I did have to change the layout to get the current develop *bll* jobs to run in a reasonable period of time.  I did this by limiting the number of jobs run per node.  These are the layouts and runtimes for the 24 hour forecasts of the optimized and develop versions of these jobs:

___
Develop
--
| Job Name | Task/Node | # Nodes | Runtime |
|-----|-----|-----|-----|
| gfswavepostbndpnt | 10 | 24 | 8400 | 
| gfswavepostbndpntbll | 20 | 24 | 2269 | 
| gfswavepostpnt | 40 | 5 | 6171 | 

Optimized
--
| Job Name | Task/Node | # Nodes | Runtime (s) |
|-----|-----|-----|-----|
| gfswavepostbndpnt | 40 | 6 | 6186 | 
| gfswavepostbndpntbll | 40 | 12 | 1554 | 
| gfswavepostpnt | 40 | 5 | 4177 | 
___

Based on these runtimes, I adjusted `config.resources` `wtime` upward for the `wavepostbndpnt` job upward.

I also had to manually adjust the data dependency for the `gfswavepostbndpntbll` job, which required the history file `gfs.t12z.atm.logf180.txt` and in my XMLs were changed to `...024.txt`.  I made changes to `workflow/rocoto/gfs_tasks.py` and `workflow/rocoto/gefs_tasks.py` to adjust the forecast hour to look for if the max is less than 180 hours.

The `expdir`s and `comroot`s for these test cases are located in
develop
```console
/work2/noaa/global/dhuber/para/EXPDIR/dev_bullpnt
/work2/noaa/global/dhuber/para/COMROOT/dev_bullpnt
```
feature/opt_pnt
```console
/work2/noaa/global/dhuber/para/EXPDIR/opt_bullpnt
/work2/noaa/global/dhuber/para/COMROOT/opt_bullpnt
```

The contents of the `spec`, `bull`, and `cbull` tarballs (i.e. `gfswave.t12z.bull_tar  gfswave.t12z.cbull_tar  gfswave.t12z.ibpbull_tar  gfswave.t12z.ibpcbull_tar  gfswave.t12z.ibp_tar  gfswave.t12z.spec_tar.gz` in `${COMROOT}/gfs.20210323/12/products/wave/station`) were extracted to `/work2/noaa/global/dhuber/station_data/opt` and `/work2/noaa/global/dhuber/station_data/dev` and compared file by file.  There were no differences between the experiments.

DavidHuber-NOAA:
Marking this PR ready for review.

JessicaMeixner-NOAA:
Thanks @DavidHuber-NOAA ! I'll get this reviewed today

workflow/rocoto/gefs_tasks.py
```diff
@@ -387,7 +387,14 @@ def wavepostbndpntbll(self):
...
         # Is there any reason this is 180?
-        data = f'{atmos_hist_path}/{self.cdump}.t@Hz.atm.logf180.txt'
+        # The wavepostbndpntbll job runs on forecast hours up to 180 hours
+        last_fhr = Tasks._get_forecast_hours(self.cdump,
+                                             self._configs['wavepostbndpntbll'],
+                                             component='atmos')[-1]
+        if last_fhr > 180:
+            last_fhr = 180
+
+        data = f'{atmos_hist_path}/{self.cdump}.t@Hz.atm.logf{last_fhr:03d}.txt'
```
WalterKolczynski-NOAA: This should be based on `FHMAX_WAV_IBP` from `config.wave`.

workflow/rocoto/gfs_tasks.py
```diff
@@ -1175,9 +1175,17 @@ def wavepostbndpnt(self):
         return task
 
     def wavepostbndpntbll(self):
+
+        # The wavepostbndpntbll job runs on forecast hours up to 180 hours
+        last_fhr = Tasks._get_forecast_hours(self.cdump,
+                                             self._configs['wavepostbndpntbll'],
+                                             component='atmos')[-1]
+        if last_fhr > 180:
+            last_fhr = 180
```
WalterKolczynski-NOAA: Same as above.

WalterKolczynski-NOAA (changes requested):

workflow/rocoto/gefs_tasks.py
JessicaMeixner-NOAA: Why is wavepostbndpntbll being added to GEFS? This is not a job GEFS runs.

WalterKolczynski-NOAA: That's my fault. Wasn't thinking when I added the wave post jobs to GEFS. I just copied what GFS had.

WalterKolczynski-NOAA: I'll remove the boundary point jobs from GEFS later

JessicaMeixner-NOAA: Thanks for the heads up @WalterKolczynski-NOAA , they're all turned off by default, but I was confused by this PR updating it (which I thought was adding it).

DavidHuber-NOAA: Fixed.  Thanks!

workflow/rocoto/gfs_tasks.py
DavidHuber-NOAA: Fixed.

WalterKolczynski-NOAA: `FHMAX_WAV_IBP` is already clipped to `FHMAX`, so you shouldn't need to do it again here.

workflow/rocoto/gfs_tasks.py
```diff
@@ -1175,9 +1175,20 @@ def wavepostbndpnt(self):
...
+                                             component='atmos')[-1]
+
+        fhmax_wav_ibp = wave_cfg['FHMAX_WAV_IBP']
+        if last_fhr > fhmax_wav_ibp:
+            last_fhr = fhmax_wav_ibp
+
         deps = []
         atmos_hist_path = self._template_to_rocoto_cycstring(self._base["COM_ATMOS_HISTORY_TMPL"])
-        data = f'{atmos_hist_path}/{self.cdump}.t@Hz.atm.logf180.txt'
+        data = f'{atmos_hist_path}/{self.cdump}.t@Hz.atm.logf{last_fhr:03d}.txt'
```
DavidHuber-NOAA: @WalterKolczynski-NOAA `FHMAX_GFS` and `FHMAX_WAV_IBP` can differ if `FHMAX_GFS` is greater than 180 hours.

DavidHuber-NOAA: @JessicaMeixner-NOAA It seems like `FHMAX_WAV_IBP` should default to `FHMAX_WAV` instead of `FHMAX_GFS` in
https://github.com/NOAA-EMC/global-workflow/blob/e7909af8d9e1f34140388a3f8556d8e582c58fe5/parm/config/gfs/config.wave#L100-L105
And thus change to
```bash
if [[ "${RUN}" == "gfs" ]]; then
  export FHMAX_WAV=${FHMAX_WAV_GFS}
fi
export WAVHINDH=0
export FHMAX_WAV_IBP=180
if (( FHMAX_WAV < FHMAX_WAV_IBP )); then export FHMAX_WAV_IBP=${FHMAX_WAV} ; fi
```
Do you think that is correct?

JessicaMeixner-NOAA: @DavidHuber-NOAA nice catch, yes I do think this is correct.

JessicaMeixner-NOAA:
@DavidHuber-NOAA apologies for not getting to this yesterday, trying to look at your output this morning and do not have permissions to see: /work2/noaa/global/dhuber/para/COMROOT/dev_bullpnt/gfs.20210323/12/products

DavidHuber-NOAA:
@JessicaMeixner-NOAA Thanks for taking a look.  It should be opened up now.

JessicaMeixner-NOAA:
@DavidHuber-NOAA thanks! i can see the directories now

workflow/rocoto/gefs_tasks.py
```diff
@@ -386,8 +386,18 @@ def wavepostbndpnt(self):
...
+
+        # The wavepostbndpntbll job runs on forecast hours up to FHMAX_WAV_IBP
+        wave_cfg = self._configs['wave']
+        last_fhr = Tasks._get_forecast_hours(self.cdump,
+                                             wave_cfg,
+                                             component='atmos')[-1]
+
+        fhmax_wav_ibp = wave_cfg['FHMAX_WAV_IBP']
+        if last_fhr > fhmax_wav_ibp:
+            last_fhr = fhmax_wav_ibp
```
WalterKolczynski-NOAA:
```suggestion
        last_fhr = self._configs['wave']['FHMAX_WAV_IBP']
```

workflow/rocoto/gfs_tasks.py
```diff
@@ -1175,9 +1175,20 @@ def wavepostbndpnt(self):
...
+
+        # The wavepostbndpntbll job runs on forecast hours up to FHMAX_WAV_IBP
+        wave_cfg = self._configs['wave']
+        last_fhr = Tasks._get_forecast_hours(self.cdump,
+                                             wave_cfg,
+                                             component='atmos')[-1]
+
+        fhmax_wav_ibp = wave_cfg['FHMAX_WAV_IBP']
+        if last_fhr > fhmax_wav_ibp:
+            last_fhr = fhmax_wav_ibp
```
WalterKolczynski-NOAA:
```suggestion
        last_fhr = self._configs['wave']['FHMAX_WAV_IBP']
```

WalterKolczynski-NOAA (changes requested):

JessicaMeixner-NOAA (approved):
I did not repeat @DavidHuber-NOAA test but did look at a few files and as he showed, everything is the same and looks fine assuming all the CI tests pass on the various machines.  Please let me know if you'd like me to run this myself.  Thank you for adding the boundary points for this test to ensure that capability remains working.    I'll keep you posted on any updates to completely re-engineer this job.

workflow/rocoto/gefs_tasks.py
DavidHuber-NOAA: Ah, I see now.  Thanks @WalterKolczynski-NOAA.

DavidHuber-NOAA:
Starting CI on Hera and Hercules.

DavidHuber-NOAA:
Hera failed due to stmp being full.

DavidHuber-NOAA:
`enkf.x` failed with an `integer divide by zero` error during the `eupd` job at line [724 of letkf.f90](https://github.com/NOAA-EMC/GSI/blob/8e279f9c734097f673b07e80f385b2623d13ba4a/src/enkf/letkf.f90#L724).  This error does not make sense to me at that line as it is a variable declaration without any math:
```fortran
real(r_kind), dimension(:), allocatable :: work1
```
Perhaps an earlier job produced some invalid data.

Since this PR only deals with the `wavepostpnt` job, I think this failure could be ignored since all `S2SW*` tests pass, but will defer to @aerorahul.

aerorahul:
> `enkf.x` failed with an `integer divide by zero` error during the `eupd` job at line [724 of letkf.f90](https://github.com/NOAA-EMC/GSI/blob/8e279f9c734097f673b07e80f385b2623d13ba4a/src/enkf/letkf.f90#L724). This error does not make sense to me at that line as it is a variable declaration without any math:
> 
> ```fortran
> real(r_kind), dimension(:), allocatable :: work1
> ```
> 
> Perhaps an earlier job produced some invalid data.
> 
> Since this PR only deals with the `wavepostpnt` job, I think this failure could be ignored since all `S2SW*` tests pass, but will defer to @aerorahul.

I think that is fine.
@CatherineThomas-NOAA has anyone you know experienced this on Hercules?

CatherineThomas-NOAA:
@aerorahul: I haven't heard of this problem on Hercules yet, but I also don't know of anyone really cycling over there.  We run the GSI ctests there so the enkf in general is exercised, but it's not exactly testing a wide range of cases.  Thanks for the heads up.

PR #2663, opened 2024-06-05
Update ufs-weather-model 

### LINKED ISSUE #2437: Stop using stand-alone UPP

As part of the Rocky 8 upgrade for Hera (PR #2421), we had to move to a stand-alone UPP version because the one in UFS has not yet been updated. Once the UPP version in UFS is updated to include the Rocky 8 updates, we should move back to using that version instead of checking out a separate version.

No separate UPP submodule

### Acceptance Criteria

Dependency: ufs-community/ufs-weather-model#2213

- [ ] Update UFS to hash containing UPP updates for Rocky 8
- [ ] Remove UPP submodule
- [ ] Update build_upp.sh to use `ufs_model.fd/FV3/upp/tests` instead of `upp.fd/tests`
- [ ] Restore linking of `ufs_model.fd/FV3/upp` to `sorc/upp.fd` in `link_workflow.sh`

_No response_

### LINKED ISSUE #2617: fcst temp directories are not removed
### What is wrong?

Perhaps I did not run my experiment long enough, but it appears that the gdasfcst and gfsfcst directories, while emptied are never removed from the RUNDIR space.   There's a chance I just did not run enough days for this to get cleaned out. 

gdasfcst and gfsfcst run directories should be removed from the RUNDIRS. 

All or N/A

### Steps to reproduce

Run a cycled experiment, note that the temporary directories are not removed. 

An example: 
/scratch1/NCEPDEV/climate/Jessica.Meixner/cycling/iau_06/C384iaucold03/TMP/RUNDIRS/cold03/tempdebugspace

Note the contents is getting removed as expected, but the folder itself remains. 

### Additional information

None

Not at the moment.

JessicaMeixner-NOAA:
to work while spack inconsistent bt workflow and ww3/ufs

# Description
Updates ufs-weather-model, this updates RDHPCS to the newer spack-stack allowing some temporary fixes to be reverted.
* removes upp submodule
* uses upp from the ufs-weather-model
* restores the build and link that were hacked during the Hera Rocky 8 transition to allow for UPP submodule
* Removes forecast directories in clean-up

Fixes https://github.com/NOAA-EMC/global-workflow/issues/2617 
Fixes https://github.com/NOAA-EMC/global-workflow/issues/2437

co-author: @aerorahul  

# Type of change

# Change characteristics

# How has this been tested?
In progress... 

# Checklist

### DIFF
--- a/jobs/rocoto/waveinit.sh
+++ b/jobs/rocoto/waveinit.sh
@@ -5,7 +5,8 @@ source "${HOMEgfs}/ush/preamble.sh"
 ###############################################################
 echo
 echo "=============== START TO SOURCE FV3GFS WORKFLOW MODULES ==============="
-. ${HOMEgfs}/ush/load_fv3gfs_modules.sh
+#. ${HOMEgfs}/ush/load_fv3gfs_modules.sh
+. ${HOMEgfs}/ush/load_ufswm_modules.sh
 status=$?
 [[ ${status} -ne 0 ]] && exit ${status}
 
--- a/jobs/rocoto/waveinit.sh
+++ b/jobs/rocoto/waveinit.sh
@@ -5,8 +5,7 @@ source "${HOMEgfs}/ush/preamble.sh"
 ###############################################################
 echo
 echo "=============== START TO SOURCE FV3GFS WORKFLOW MODULES ==============="
-#. ${HOMEgfs}/ush/load_fv3gfs_modules.sh
-. ${HOMEgfs}/ush/load_ufswm_modules.sh
+. ${HOMEgfs}/ush/load_fv3gfs_modules.sh
 status=$?
 [[ ${status} -ne 0 ]] && exit ${status}
 
--- a/ush/forecast_postdet.sh
+++ b/ush/forecast_postdet.sh
@@ -437,10 +437,6 @@ MOM6_postdet() {
 
       # Daily output
       if (( fhr > 0 & fhr % 24 == 0 )); then
- 
-        # For daily output, the valid date in the file is based on the beginning of avg. period
-        vdate=$(date --utc -d "${vdate:0:8} ${vdate:8:2} - 24 hours" +%Y%m%d%H)
-
         source_file="ocn_daily_${vdate:0:4}_${vdate:4:2}_${vdate:6:2}.nc"
         dest_file="${RUN}.ocean.t${cyc}z.daily.f${fhr3}.nc"
         ${NLN} "${COMOUT_OCEAN_HISTORY}/${dest_file}" "${DATA}/MOM6_OUTPUT/${source_file}"
--- a/ush/forecast_postdet.sh
+++ b/ush/forecast_postdet.sh
@@ -217,7 +217,7 @@ FV3_out() {
   echo "SUB ${FUNCNAME[0]}: copying output data for FV3"
 
   # Copy configuration files
-  if [[ "${RUN}" == "gfs" || "${RUN}" == "gefs" ]]; then
+  if [[ "${RUN}" == "gfs" || "${RUN}" == "gdas" || "${RUN}" == "gefs" ]]; then
     ${NCP} "${DATA}/input.nml" "${COMOUT_CONF}/ufs.input.nml"
     ${NCP} "${DATA}/model_configure" "${COMOUT_CONF}/ufs.model_configure"
     ${NCP} "${DATA}/ufs.configure" "${COMOUT_CONF}/ufs.ufs.configure"
--- a/ush/forecast_postdet.sh
+++ b/ush/forecast_postdet.sh
@@ -438,8 +438,8 @@ MOM6_postdet() {
 
       # Daily output
       if (( fhr > 0 & fhr % 24 == 0 )); then
-	   
-	# For daily output, the valid date in the file is based on the beginning of avg. period
+ 
+        # For daily output, the valid date in the file is based on the beginning of avg. period
         vdate=$(date --utc -d "${vdate:0:8} ${vdate:8:2} - 24 hours" +%Y%m%d%H)
 
         source_file="ocn_daily_${vdate:0:4}_${vdate:4:2}_${vdate:6:2}.nc"
--- a/ush/forecast_postdet.sh
+++ b/ush/forecast_postdet.sh
@@ -438,6 +438,10 @@ MOM6_postdet() {
 
       # Daily output
       if (( fhr > 0 & fhr % 24 == 0 )); then
+	   
+	# For daily output, the valid date in the file is based on the beginning of avg. period
+        vdate=$(date --utc -d "${vdate:0:8} ${vdate:8:2} - 24 hours" +%Y%m%d%H)
+
         source_file="ocn_daily_${vdate:0:4}_${vdate:4:2}_${vdate:6:2}.nc"
         dest_file="${RUN}.ocean.t${cyc}z.daily.f${fhr3}.nc"
         ${NLN} "${COMOUT_OCEAN_HISTORY}/${dest_file}" "${DATA}/MOM6_OUTPUT/${source_file}"
--- a/ush/forecast_postdet.sh
+++ b/ush/forecast_postdet.sh
@@ -217,7 +217,7 @@ FV3_out() {
   echo "SUB ${FUNCNAME[0]}: copying output data for FV3"
 
   # Copy configuration files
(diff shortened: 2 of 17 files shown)


aerorahul:
Thanks for this PR @JessicaMeixner-NOAA 
I have opened a PR into this PR branch https://github.com/JessicaMeixner-NOAA/global-workflow/pull/13 to remove the UPP submodule and use the UPP directly from the UFS-weather-model.  This will resolve #2437

JessicaMeixner-NOAA:
Thanks @aerorahul ! I just merged this in.  I'll take the PR off of draft but I don't know if you want to wait for running full set of tests until after I've done an offline test or not?

JessicaMeixner-NOAA:
S2SW test on hera is currently going through the wave post point.

ush/forecast_postdet.sh
```diff
@@ -217,7 +217,7 @@ FV3_out() {
   echo "SUB ${FUNCNAME[0]}: copying output data for FV3"
 
   # Copy configuration files
-  if [[ "${RUN}" == "gfs" || "${RUN}" == "gefs" ]]; then
+  if [[ "${RUN}" == "gfs" || "${RUN}" == "gdas" || "${RUN}" == "gefs" ]]; then
```
WalterKolczynski-NOAA: At this point, we're just leaving out the enkf RUNs. Maybe we should just always do this?

jobs/rocoto/waveinit.sh
```diff
@@ -5,8 +5,7 @@ source "${HOMEgfs}/ush/preamble.sh"
 ###############################################################
 echo
 echo "=============== START TO SOURCE FV3GFS WORKFLOW MODULES ==============="
-#. ${HOMEgfs}/ush/load_fv3gfs_modules.sh
-. ${HOMEgfs}/ush/load_ufswm_modules.sh
+. ${HOMEgfs}/ush/load_fv3gfs_modules.sh
```
WalterKolczynski-NOAA: We may still want to use the ufs modulefile. @aerorahul?

ush/forecast_postdet.sh
JessicaMeixner-NOAA: Good point, I had a similar thought and then didn't change it.  Before we leave out enkf, @CatherineThomas-NOAA do we want this for enkf too or no?

jobs/rocoto/waveinit.sh
JessicaMeixner-NOAA: I thought I had chatted offline with @aerorahul about reverting this back. I'm good either way, let me know if we should switch this back.

JessicaMeixner-NOAA: Also note, I did not change the fcst job, just the wave pre/post jobs.

ush/forecast_postdet.sh
CatherineThomas-NOAA: I think we should save the enkf fcst configs as well since their settings are slightly different than the deterministic.  The files are small so it shouldn't be a space issue.

jobs/rocoto/waveinit.sh
aerorahul: I am fine either way.  I will leave it to @JessicaMeixner-NOAA

ush/forecast_postdet.sh
aerorahul: The ensemble members are going to overwrite each other if we remove the wrap of conditionals.
Would just one ensemble member be enough or do all `NENS` members need to copy their configuration files?  The seeds are different based on the member number.
Let us know what the preference is, and we can address it in an future PR if it will take a little time to think.

ush/forecast_postdet.sh
```diff
@@ -438,6 +438,10 @@ MOM6_postdet() {
 
       # Daily output
       if (( fhr > 0 & fhr % 24 == 0 )); then
+	   
+	# For daily output, the valid date in the file is based on the beginning of avg. period
+        vdate=$(date --utc -d "${vdate:0:8} ${vdate:8:2} - 24 hours" +%Y%m%d%H)
```
aerorahul: Did the patch from @aerorahul misalign the indentation blocks?

jobs/rocoto/waveinit.sh
JessicaMeixner-NOAA: Pro of ufs-weather-model modules: this is what it's built with, ensures consistency as things might get out of sync while developing.  However, I thought we were all supposed to be moving towards using the same modules, which I thought was the fv3gfs_modules so I thought that was the goal?

ush/forecast_postdet.sh
aerorahul:
```suggestion
	      # For daily output, the valid date in the file is based on the beginning of avg. period
        vdate=$(date --utc -d "${vdate:0:8} ${vdate:8:2} - 24 hours" +%Y%m%d%H)
```

JessicaMeixner-NOAA: There were some tabs I unintentionally put in.  Hopefully this is fixed now.

JessicaMeixner-NOAA: I think I fixed this manually before seeing your suggested change here.

jobs/rocoto/waveinit.sh
aerorahul: It is the goal. Yes. For ops, it is a requirement.
However, these are the rocoto wrappers used in development. The flexibility to allow for asynchronous updates to the components would be desirable.

JessicaMeixner-NOAA: Okay, let me switch these back then.

ush/forecast_postdet.sh
CatherineThomas-NOAA: @aerorahul One member should be enough.

JessicaMeixner-NOAA: Right now I've added the ~=enkf logic.  @CatherineThomas-NOAA i'll work with you offline to save the enkf ones? If we can, we'll add it to this PR, if not, we'll add it to a future update.

JessicaMeixner-NOAA: COMOUT_CONF is defined with member numbers in the directory structure, so we wil not be over-writing each one.  Since these are very small files, I've just removed the if-statement.  If we want to walk this back a bit, if there's a if control variable we could only do this for the one.

aerorahul (approved):
looks good.
thank you for updating the hash, coordinating the changes, and testing them out.

WalterKolczynski-NOAA:
Hit the wallclock on Orion in one product job. Maybe a bad node?

JessicaMeixner-NOAA:
The hera archive failure is probably related to the update we made for the ocean daily files.  I'll look into this on Monday.

JessicaMeixner-NOAA:
So the behavior of the dates of the ocean daily file is different it appears with and without IAU.  I'm not sure what the best way to fix this is.  Is it adding a "if IAU" part to the ocean daily linking?  If we move to a true "daily" instead of the 24 hour average, its my understanding we'd have to start the averaging at 00z (or if 06, 12, 18z that would mean the next 00z), but that might at least be consistent?    

If we want the updates of the model ASAP, we can take out the change we made for ocean daily files here or make the archive not be required to have the daily files so it wont fail (which also might not be ideal, since that  identified an issue we need to be aware of).

WalterKolczynski-NOAA:
> So the behavior of the dates of the ocean daily file is different it appears with and without IAU. I'm not sure what the best way to fix this is. Is it adding a "if IAU" part to the ocean daily linking? If we move to a true "daily" instead of the 24 hour average, its my understanding we'd have to start the averaging at 00z (or if 06, 12, 18z that would mean the next 00z), but that might at least be consistent?
> 
> If we want the updates of the model ASAP, we can take out the change we made for ocean daily files here or make the archive not be required to have the daily files so it wont fail (which also might not be ideal, since that identified an issue we need to be aware of).

I would favor taking it out since it is separate from the model update. It can be handled in a separate PR.

JessicaMeixner-NOAA:
Sounds good @WalterKolczynski-NOAA   I'll update this PR now.

JessicaMeixner-NOAA:
Okay, i have updated the PR and PR description as we will not be addressing the last ocn daily issue.

WalterKolczynski-NOAA:
Something happened with the GDAS build on Hera. Not sure what, but executables are missing.

WalterKolczynski-NOAA:
Pretty sure unrelated to this PR. Will try again in a few minutes.

WalterKolczynski-NOAA (approved):

PR #2681, opened 2024-06-12
Update logic for MOM6 number of layers/exception values

### LINKED ISSUE #2615: zeros in ocean post grib2 files on hera
### What is wrong?

When running with the sea-ice PR that was just merged, so essentially develop as of today,  it was noticed by @SulagnaRay-NOAA that all of the ocean grib2 files are constant values (mostly zeros).  The native model output is not zeros and the ice gribs also appear to be okay.    

Investigation as to what is going on and why is ongoing. 

We should have grib2 output files that match the native model output (and have non-zero/constant values). 

Hera

### Steps to reproduce

This was discovered running a C384 test case of C384mx025_3DVarAOWCDA.  However, I suspect other test cases would expose this issue as well. 

Some example output can be found here: 
/scratch1/NCEPDEV/climate/Jessica.Meixner/cycling/iau_06/C384iaucold03/cold03/COMROOT/cold03/gfs.20210703/06/products/ocean/grib2/0p25 

Log files can be found here: /scratch1/NCEPDEV/climate/Jessica.Meixner/cycling/iau_06/C384iaucold03/cold03/COMROOT/cold03/logs/2021070306

### Additional information

@GwenChen-NOAA @jiandewang @SulagnaRay-NOAA @LydiaStefanova-NOAA @guillaumevernieres @CatherineThomas-NOAA  FYI  - any additional information or help is appreciated! 

Not yet...

JessicaMeixner-NOAA:
# Description
Updates the logic to be by run instead of for DO_JEDIOCNVAR to determine how many layers and sets the exception value for MOM6 to be 1e-34 for all scenarios.   Note,  we will no longer have zeros in the ocean grib output and the DA will also run without issues. 

Fixes https://github.com/NOAA-EMC/global-workflow/issues/2615 

# Type of change

# Change characteristics

# How has this been tested?

(this was an earlier version):
Ran a few cycles of WCDA C384mx025 /scratch1/NCEPDEV/climate/Jessica.Meixner/testmarineoutput/test01 
@jiandewang @SulagnaRay-NOAA and @guillaumevernieres all have checked the output is okay. 

@guillaumevernieres did tests with 1e-34 for all and DA worked  as expected with this change. 

# Checklist


parm/config/gefs/config.ufs
```diff
@@ -325,8 +325,8 @@ if [[ "${skip_mom6}" == "false" ]]; then
       MOM6_RIVER_RUNOFF='False'
       eps_imesh="2.5e-1"
       TOPOEDITS="ufs.topo_edits_011818.nc"
-      if [[ "${DO_JEDIOCNVAR:-NO}" = "YES" ]]; then
-        MOM6_DIAG_COORD_DEF_Z_FILE="oceanda_zgrid_75L.nc"
+      if [[ ${RUN} == "enkfgfs" || "${RUN}" == "gdas" ]]; then
```
guillaumevernieres: I guess that works, but wouldn't that cover it:
```bash
if [[ "${RUN}" != "gfs" ]]; then
```
?

guillaumevernieres (approved):
Looks good. Thanks @JessicaMeixner-NOAA .

parm/config/gefs/config.ufs
aerorahul: @NeilBarton-NOAA Would you be able to gauge the the impact for `RUN=gefs`?

JessicaMeixner-NOAA: @guillaumevernieres I think we want this for both GFS and GEFS so that's why I didn't use your suggestion.

JessicaMeixner-NOAA: Although this is the "gfs" config, I don't think we want gfs and gefs to be that out of sync.

aerorahul: @JessicaMeixner-NOAA The comment is for this file `parm/config/gefs/config.ufs`
I agree, ideally, we would want to keep them in sync.
Am I misunderstanding?

JessicaMeixner-NOAA: Maybe I'm missing something? So this is GEFS, so to my knowledge you come into this routine and you would be RUN="GEFS" which means you want the "else" in this loop.  If the if statement was if ~=gfs, then you would get the 75 layers not the 40 layers which is what I assume GEFS wants to my knowledge.

NeilBarton-NOAA: I'm adding @jiandewang I don't know the impact as I don't analyze the ocean output. I would default to keeping it to interpolate_zgrid_40L.nc for gefs

aerorahul: You are right @JessicaMeixner-NOAA

WalterKolczynski-NOAA: I'm late to the party here, but want to confirm we want to run this for `gdas` and `enkfgfs` but not `enkfgdas`.

JessicaMeixner-NOAA: The "else" should be gfs/gefs...   so I think this is my mistake.  Maybe I should switch the logic.  @guillaumevernieres can help confirm too.

JessicaMeixner-NOAA:
FYI @guillaumevernieres @CatherineThomas-NOAA @NeilBarton-NOAA @aerorahul @WalterKolczynski-NOAA 

This has been updated after offline discussions and additional tests by @guillaumevernieres.  The exception value for MOM6 is now 1e-34 reguardless of resolution or RUN or DOJEDI_OCN.   Whether we have 40 or 75 levels in the output is now determined based on if it's GFS/GEFS - 40 and otherwise is 75 (ie gdas, etc).

aerorahul (approved):
Looks good.
Thanks for testing with `-1e34` as the missing value.

WalterKolczynski-NOAA:
Outside the scope of this PR, but why isn't the missing value being read from the metadata?

aerorahul:
> Outside the scope of this PR, but why isn't the missing value being read from the metadata?

Should probably update the `MOM_input_template` in the ufs-weather-model to fix it to `-1e34` instead of making this configurable.

WalterKolczynski-NOAA:
> Outside the scope of this PR, but why isn't the missing value being read from the metadata?

Nevermind, these are for writing, not reading.

TerrenceMcGuinness-NOAA:
CI Tests Failed on Hera.  Then GitHub failed to authenticate due to connection errors when creating the gist below for **C48_ATM** fail in `gfsfcst`.

Looks like it was a disk quota issue.
We all cleared up our files on Hera yesterday upon Kate's request.
Should be fine to re-run.

```
/bin/cp: error writing '/scratch1/NCEPDEV/stmp2/Terry.McGuinness/RUNDIRS/C48_ATM_38470299/gfsfcst.2021032312/fcst.3048188/INPUT/oro_data_ss.tile4.nc': Disk quota exceeded
```

https://gist.github.com/emcbot/52b58c977ed8c647a6b3e5064fd54736

PR #2664, opened 2024-06-06
Gaea c5 added to Jenkinsfile

AnilKumar-NOAA:
# Description
Added Gaea c5 to Jenkinsfile

# Type of change

- Bug fix (fixes something broken)
- New feature (adds functionality)
- Maintenance (code refactor, clean-up, new CI test, etc.)

# Change characteristics
- Is this a breaking change (a change in existing functionality)? YES/NO
- Does this change require a documentation update? YES/NO

# How has this been tested?

# Checklist
- [ ] Any dependent changes have been merged and published
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] I have made corresponding changes to the documentation if necessary

Issue #2572, opened 2024-05-06
Setup CI testing workflow on Gaea-C5


DavidBurrows-NCO:
### What new functionality do you need?

CI testing on Gaea-C5

### What are the requirements for the new functionality?

Update HOMEgfs/ci scripts for Gaea

### Acceptance Criteria

CI workflow to run successfully on Gaea

aerorahul:
@AnilKumar-NOAA Please assign yourself

DavidBurrows-NCO:
@aerorahul Sorry I unassigned myself Rahul...Could you please reassign me?

DavidBurrows-NCO:
Thanks @DavidHuber-NOAA

PR #2701, opened 2024-06-20
Fix xml file setup and complete C48 ATM and S2SW runs for CI on Gaea

DavidBurrows-NCO:
# Description
What?
This PR sets up the ability on Gaea for auto generation of a clean xml file, i.e., an xml file that does not need any alterations before running rocoto.  

Why?
This PR continues the effort to run CI testing on Gaea.

Refs #2572 
Refs #2664 

# Type of change

# Change characteristics

# How has this been tested?
Run forecast-only C48 ATM and S2SW from the CI experiments on Gaea

# Checklist


parm/config/gfs/config.resources
```diff
@@ -113,7 +113,11 @@ case ${step} in
     export nth_waveinit=1
     export npe_node_waveinit=$(( npe_node_max / nth_waveinit ))
     export NTASKS=${npe_waveinit}
-    export memory_waveinit="2GB"
+    if [[ "${machine}" == "GAEA" ]]; then
+      export memory_waveinit=""
```
aerorahul: Can you explain why specifying memory on Gaea is inappropriate (here and elsewhere)?

aerorahul (commented):
Specification of clusters looks ok.
The unsetting of memory is not.  Doing so, will basically need every job to have an `if-else-` for `GAEA`.
Would like to better understand why specifying the maximum memory required for a job on Gaea is problematic.

parm/config/gfs/config.resources
DavidBurrows-NCO: @aerorahul Thanks for looking over the PR.

I tried a few different options for setting the memory on Gaea before contacting Gaea help desk.
1) add <memory>2G</memory>
sbatch: error: Memory specification can not be satisfied
sbatch: error: Batch job submission failed: Requested node configuration is not available

2) <native>--mem=2G</native>
sbatch: error: Memory specification can not be satisfied

3) <memory>0</memory> works

The response from Gaea help desk and ORNL:
"Due to the configuration of slurm on Gaea, users are not expected to set the memory for batch jobs. In cases of node sharing (on a specific partition, on a given set of nodes) among users, you would then be required to explicitly request a certain amount of memory for a job.
I talked to the admins at ORNL to see if it was intentional and with the way slurm is configured memory is a consumable resource which is not shared among jobs meaning exclusivity is assumed in this case. Users should not have to manually set the real memory on the batch partition."

Here's a simple test script that demonstrates the error:
#!/bin/bash
#SBATCH -A ufs-ard
#SBATCH -M c5
#SBATCH --mem=... 0 succeeds; 1G fails
#SBATCH --time=1:00:00

aerorahul: Instead of doing this for every task, can you try this at the end of `config.resources` while we get feedback from the Gaea sysadmins:
```bash
if [[ "${machine}" == "GAEA" ]]; then
  for mem_var in $(env | grep '^memory_' | cut -d= -f1); do
    unset "${mem_var}"
  done
fi
```
It should unset all `memory_` variables for Gaea without having to do so for each task.

DavidBurrows-NCO: Morning @aerorahul. I'm responding to the ticket right now and will test this out today. Thanks.

parm/config/gfs/config.resources
```diff
@@ -1307,4 +1307,10 @@
 
 esac
 
+if [[ "${machine}" == "GAEA" ]]; then
+  for mem_var in $(env | grep '^memory_' | cut -d= -f1); do
```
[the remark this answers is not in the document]
DavidBurrows-NCO: Hi @aerorahul I updated config.resources as you suggested. I tested setup expt and xml, and memory statements no longer go into the xml file on Gaea. Let me know if you want to follow this code suggestion. I'm not 100% sure where it is suggesting to add `|| true`

env/GAEA.env
```diff
@@ -3,6 +3,11 @@
 if [[ $# -ne 1 ]]; then
 
     echo "Must specify an input argument to set runtime environment variables!"
+    echo "argument can be any one of the following:"
+    echo "atmanlrun atmensanlrun aeroanlrun snowanl"
+    echo "anal sfcanl fcst post metp"
+    echo "eobs eupd ecen efcs epos"
+    echo "postsnd awips gempak"
```
aerorahul:
```suggestion
```
I think we are removing this from all.  Its not checked and so serves no point.

parm/config/gfs/config.resources.GAEA
```diff
@@ -20,3 +20,7 @@ case ${step} in
     ;;
 
 esac
+
+for mem_var in $(env | grep '^memory_' | cut -d= -f1); do
```
aerorahul, line 25:
```suggestion
# shellcheck disable=SC2312
for mem_var in $(env | grep '^memory_' | cut -d= -f1); do
```

aerorahul (approved):
looks good.

PR #2715, opened 2024-06-24
Reset to develop

guillaumevernieres:
@RussTreadon-NOAA , I'm not sure if this will work ... Do the differences make sense? I just did ```git reset --hard develop``` to the most recent develop.


guillaumevernieres:
never mind, looks like I started from the wrong commit

PR #2762, opened 2024-07-12
Update Java Agent launching script for Jenkins connections

TerrenceMcGuinness-NOAA:
# Description

Made updates to the Jenkins Launching Script for robustness and less ambiguous documentation:
- Clearer distinction between required user token for the remote api and the systems token for launching
- Added pre-checks: `gh` is authenticating, named compliant token and secret file exists 
- More robust Jason based parser of the remote api response for checking the state of the Node connection
- For `cron` use a 5 minute pause and recheck was added before re-launching of the java agent
- Added concise header documentation of requirements and purpose 

# Type of change
- Maintenance (code refactor, clean-up, and improved readability)

# Change characteristics

# How has this been tested?
Tested on the command line

# Checklist


ci/scripts/utils/launch_java_agent.sh
```diff
@@ -1,8 +1,37 @@
-#!/bin/env bash
+#!/bin/bash
```
DavidHuber-NOAA: To suppress SC2311:

```suggestion
#!/bin/env bash

#shellcheck enable=check-set-e-suppressed
```

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -56,22 +92,49 @@ cd "${JENKINS_AGENT_LANUCH_DIR}"
...
+   echo "User Jenkins authetication TOKEN to the controller for using the Remote API does not exist"
+   exit 1
+fi
+
+JENKINS_TOKEN=$(cat "${controller_user_auth_token}")
+echo -e "#!/usr/bin/env python
+import json,sys
+with open(sys.argv[1], 'r') as file:
+    data = json.load(file)
+print(data[\"offline\"])
```
DavidHuber-NOAA: In case `offline` isn't defined.
```suggestion
print(data.get('offline','True'))
```

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -1,8 +1,37 @@
...
+#        The optional 'now' argument forces the script to launch the Jenkins
+#        agent immediately without waiting, even if the node is detected as offline.
+#
+# ==============================================================================
+
 set -e
+shopt -s inherit_errexit
 
 controller_url="https://jenkins.epic.oarcloud.noaa.gov"
 controller_user="terry.mcguinness"
```
DavidHuber-NOAA: Does `controller_user` need to be updated by the user running the script?  If so, can this be an input to the script with, say, `-u`?  Also, it should be noted in the documentation that this is the username for Jenkins, not the machine.

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -1,8 +1,37 @@
...
+#              the Jenkins agent launch directory, and proper authentication
+#              with GitHub.
+#
+#              It then proceeds to check if the Jenkins node is online and
+#              decides whether to launch the Jenkins agent based on the node's
+#              status. The agent is launched in the background,
+#              and its PID is logged for reference.
+#
+# Prerequisites: JAVA_HOME must be set to a valid JDK installation.
+#                GitHub CLI (gh) must be installed and authenticated.
```
DavidHuber-NOAA: Perhaps a note (or link to a gist) on how to authenticate would be useful.

Also, please mention what version(s) of `gh` are required.

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -1,8 +1,37 @@
...
+#                Jenkins agent launch directory must exist and be specified.
+#                Jenkins controller URL and authentication token must be provided.
```
DavidHuber-NOAA: Doesn't the authentication token need to be in the launch directory?  If so, that should be noted here.  Similarly, a note on how to obtain the token (i.e. contact you, me, or Kris Booker).

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -56,22 +92,49 @@ cd "${JENKINS_AGENT_LANUCH_DIR}"
...
-if [[ "${offline}" == "true" ]]; then
-  echo "Jenkins Agent is offline. Lanuching Jenkins Agent on ${host}"
-  command="nohup ${JAVA} -jar agent.jar -jnlpUrl ${controller_url}/computer/${MACHINE_ID^}-EMC/jenkins-agent.jnlp  -secret @jenkins-secret-file -workDir ${JENKINS_WORK_DIR}"
-  echo -e "Lanuching Jenkins Agent on ${host} with the command:\n${command}" >& "${LOG}"
-  ${command} >> "${LOG}" 2>&1 &
-  nohup_PID=$!
-  echo "Java agent running on PID: ${nohup_PID}" >> "${LOG}" 2>&1
-  echo "Java agent running on PID: ${nohup_PID}"
+if [[ "${offline}" != "False" ]]; then
+  if [[ "${1}" != "now" ]]; then
```
DavidHuber-NOAA:
```suggestion
  if [[ "${1:-}" != "now" ]]; then
```

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -56,22 +92,49 @@ cd "${JENKINS_AGENT_LANUCH_DIR}"
...
+      echo "Jenkins Agent is offline. Waiting 5 more minutes to check again in the event it is a temp network issue"
+      sleep 300
+  fi
+  offline=$(check_node_online)
+  if [[ "${offline}" != "False" ]]; then
+      echo "Jenkins Agent is offline. Lanuching Jenkins Agent on ${host}"
```
DavidHuber-NOAA:
```suggestion
      echo "Jenkins Agent is offline. Launching Jenkins Agent on ${host}"
```

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -56,22 +92,49 @@ cd "${JENKINS_AGENT_LANUCH_DIR}"
...
+      command="nohup ${JAVA} -jar agent.jar -jnlpUrl ${controller_url}/computer/${MACHINE_ID^}-EMC/jenkins-agent.jnlp  -secret @jenkins-secret-file -workDir ${JENKINS_WORK_DIR}"
+      echo -e "Lanuching Jenkins Agent on ${host} with the command:\n${command}" >& "${LOG}"
```
DavidHuber-NOAA:
```suggestion
      echo -e "Launching Jenkins Agent on ${host} with the command:\n${command}" >& "${LOG}"
```

DavidHuber-NOAA (changes requested):

ci/scripts/utils/launch_java_agent.sh
TerrenceMcGuinness-NOAA: oops - usually do do that.

TerrenceMcGuinness-NOAA (commented):
Committing code suggested updates

ci/scripts/utils/launch_java_agent.sh
TerrenceMcGuinness-NOAA: That require it to be in the crontab too and should be clear to the user who is doing that.

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -1,8 +1,55 @@
...
+#       status. The agent is launched in the background,
+#       and its PID is logged for reference.
+#
+# Prerequisites:
+#         JAVA_HOME must be set to a valid JDK installation.
+#         Jenkins agent launch directory must exist and be specified.
+#         GitHub CLI (gh) must be installed and authenticated for messeging
+#           from the Jenkins controller to GitHub PR via shell commands.
+#           Jenkins agent launch directory must exist and be specified.
+#           TODO: Must use GitHub CLI v2.2.5 (newer versoins have issues)
```
DavidHuber-NOAA: We were using 2.25.1.  Version 2.2.5 is not available for download from GitHub (or spack), but 2.25.1 is.  Also, noting that the version requested in https://github.com/JCSDA/spack-stack/issues/1105 is 2.28.0.

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -1,8 +1,55 @@
...
+#         Jenkins controller URL and authentication token must be provided.
+#         jenkins-secret-file:
+#           Must be present in the Jenkins agent launch directory.
+#           This file contains the secret key for the Jenkins agent
+#           established by the Jenkins administrator for each Node.
```
DavidHuber-NOAA, line 32: What do you mean here by Node?  Is that each machine?

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -56,22 +92,121 @@
...
+chmod u+x parse.py
+
+check_node_online() {
+    rm -p curl_response
+    curl_response=$(curl --silent -u "${controller_user}:${JENKINS_TOKEN}" "${controller_url}/computer/${MACHINE_ID^}-EMC/api/json?pretty=true") || true
+    echo -n "${curl_response}" > curl_response
+    ./parse.py curl_response
+}
+
+offline=$(check_node_online)
```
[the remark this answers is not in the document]
DavidHuber-NOAA: Not sure why this is still being flagged.  Perhaps changing this line to
```bash
offline=$(set -x; check_node_online)
```
would satisfy shellcheck.

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -1,8 +1,55 @@
...
+#           Must be present in the Jenkins agent launch directory.
+#           This file contains the user authentication token for the Jenkins controller
+#           to use the Remote API. This token can be generated by the user
+#           on the Jenkins controller.
+#         controller_user:
+#           Must be set to the Jenkins controller username corresponing to the jenkins_token.
+#
+# Usage: ./launch_java_agent.sh [now]
+#        The optional 'now' argument forces the script to launch the Jenkins
+#        agent immediately without waiting, even if the node is detected as offline.
```
DavidHuber-NOAA: I don't think that the `now` option is fully implemented.  If it is passed, then the launcher doesn't wait.  However, it does not force a launch.  Maybe the force option should be a flag (`-f`).

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -1,8 +1,55 @@
 #!/bin/env bash
+
+#shellcheck enable=check-set-e-suppressed
+
```
DavidHuber-NOAA: Looking more closely, this isn't needed.  This shellcheck is performed as part of CI.
```suggestion
```

DavidHuber-NOAA (changes requested):

ci/scripts/utils/launch_java_agent.sh
TerrenceMcGuinness-NOAA: This is not the place to "explain" that.

TerrenceMcGuinness-NOAA (commented):
Ready to merge

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -56,22 +108,49 @@ cd "${JENKINS_AGENT_LANUCH_DIR}"
...
-  echo -e "Lanuching Jenkins Agent on ${host} with the command:\n${command}" >& "${LOG}"
-  ${command} >> "${LOG}" 2>&1 &
-  nohup_PID=$!
-  echo "Java agent running on PID: ${nohup_PID}" >> "${LOG}" 2>&1
-  echo "Java agent running on PID: ${nohup_PID}"
+if [[ "${offline}" != "False" ]]; then
+  if [[ "${1:-}" != "-f" ]]; then
+      echo "Jenkins Agent is offline. Waiting 5 more minutes to check again in the event it is a temp network issue"
+      sleep 300
+  fi
```
DavidHuber-NOAA: Sorry I wasn't clear.  What I meant here was that `now` (which has been changed presently to `-f`), does not force the Jenkins agent to start.  It only prevents sleeping for 300 seconds.  To force the Jenkins agent to start, I think you will need another if block.

What I was trying to suggest was to keep `now` as `now` and **add** a `-f` flag to force the Jenkins agent to start.

TerrenceMcGuinness-NOAA: It should check first, that is the intention of the switch "do not double check" just check once.  Please do not change that intention.  Thanks.

TerrenceMcGuinness-NOAA:
Added both `-f` and `now` option switches

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -56,22 +109,49 @@ cd "${JENKINS_AGENT_LANUCH_DIR}"
...
-  nohup_PID=$!
-  echo "Java agent running on PID: ${nohup_PID}" >> "${LOG}" 2>&1
-  echo "Java agent running on PID: ${nohup_PID}"
+if [[ "${offline}" != "False" ]]; then
+  if [[ "${1:-}" != "now" ]]; then
+      echo "Jenkins Agent is offline. Waiting 5 more minutes to check again in the event it is a temp network issue"
+      sleep 300
+  fi
+  offline=$(check_node_online)
+  if [[ "${offline}" != "False" ]] || [[ "${1:-}" == "-f" ]]; then
```
DavidHuber-NOAA: This assumes that `-f` will be the first argument.  I think this needs to be handled with `getopts` to set a variable, e.g.

near the top:
```bash
force_launch='False'
while getopts 'f' flag; do
  case "${flag}" in
    f) force_launch='True' ;;
    *) echo "Unknown flag: ${flag}"
       exit 1 ;;
  esac
done
```

Then here:
```bash
  if [[ "${offline}" != "False" ]] || [[ "force_launch" == "True" ]]; then
```

DavidHuber-NOAA: @TerrenceMcGuinness-NOAA This issue still remains.

TerrenceMcGuinness-NOAA: oh rats of course, I will need to re-code that now to be a mutually exclusive operation.

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -56,22 +126,61 @@ cd "${JENKINS_AGENT_LANUCH_DIR}"
 
 if ! [[ -f agent.jar ]]; then
   curl -sO "${controller_url}/jnlpJars/agent.jar"
+  echo "Updated agent.jar downloaded"
+fi
```
WalterKolczynski-NOAA, line 127: This will download `agent.jar` if it is not present, but what about updates? Should there be a CLI switch to force a download of the jar?

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -56,22 +126,61 @@ cd "${JENKINS_AGENT_LANUCH_DIR}"
...
+   exit 1
 fi
+JENKINS_TOKEN=$(cat "${controller_user_auth_token}")
+
+echo -e "#!/usr/bin/env python
+import json,sys
+with open(sys.argv[1], 'r') as file:
+    data = json.load(file)
+print(data.get('offline','True'))
+" > parse.py
```
WalterKolczynski-NOAA: Using a heredoc would be a more standard way to do all these multiline strings.
```suggestion
cat << EOF > parse.py
#!/usr/bin/env python3
import json,sys
with open(sys.argv[1], 'r') as file:
    data = json.load(file)
print(data.get('offline','True'))
EOF
```

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -56,22 +126,61 @@ cd "${JENKINS_AGENT_LANUCH_DIR}"
...
-  echo "Java agent running on PID: ${nohup_PID}" >> "${LOG}" 2>&1
-  echo "Java agent running on PID: ${nohup_PID}"
+if [[ "${offline}" != "False" ]]; then
+   if [[ "${skip_wait}" != "True" ]]; then
+      echo "Jenkins Agent is offline. Waiting 5 more minutes to check again in the event it is a temp network issue"
+      sleep 300
+      offline=$(set -e; check_node_online)
+   fi
+   if [[ "${offline}" != "False" ]]; then
+      lauch_agent
```
WalterKolczynski-NOAA:
```suggestion
     lauch_agent
```

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -56,22 +126,61 @@ cd "${JENKINS_AGENT_LANUCH_DIR}"
...
+    fi
+    echo -n "${curl_response}" > curl_response
+    ./parse.py curl_response
+}
+
+lauch_agent () {
+    echo "Launching Jenkins Agent on ${host}"
+    command="nohup ${JAVA} -jar agent.jar -jnlpUrl ${controller_url}/computer/${MACHINE_ID^}-EMC/jenkins-agent.jnlp  -secret @jenkins-secret-file -workDir ${JENKINS_WORK_DIR}"
+    echo -e "Launching Jenkins Agent on ${host} with the command:\n${command}" >& "${LOG}"
+    ${command} >> "${LOG}" 2>&1 &
```
WalterKolczynski-NOAA, line 159: What happens if it fails?

WalterKolczynski-NOAA (changes requested):

ci/scripts/utils/launch_java_agent.sh
TerrenceMcGuinness-NOAA: No

ci/scripts/utils/launch_java_agent.sh
TerrenceMcGuinness-NOAA: Then it didn't work

ci/scripts/utils/launch_java_agent.sh
TerrenceMcGuinness-NOAA: Yes, lets do it that way.

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -1,8 +1,69 @@
...
+    -n (now) causes the script to launch the Jenkins agent without waiting before trying again.
+    -f (force) forces the script to launch the Jenkins regarless of its connection status."
+       exit 0 ;;
+    *) echo "Unknown flag: ${flag}"
+       exit 1;;
+  esac
+done
+
 controller_url="https://jenkins.epic.oarcloud.noaa.gov"
 controller_user="terry.mcguinness"
```
DavidHuber-NOAA:
```suggestion
controller_user=${controller_user:-"terry.mcguinness"}
```

ci/scripts/utils/launch_java_agent.sh
WalterKolczynski-NOAA: Well, duh. But I mean there is no error capture or notification on a failure.

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -1,8 +1,69 @@
...
+#       status. The agent is launched in the background,
+#       and its PID is logged for reference.
+#
+# Prerequisites:
+#         JAVA_HOME must be set to a valid JDK installation.
+#         Jenkins agent launch directory must exist and be specified.
+#         GitHub CLI (gh) must be installed and authenticated for messeging
+#           from the Jenkins controller to GitHub PR via shell commands.
+#           Jenkins agent launch directory must exist and be specified.
+#           TODO: Must use GitHub CLI v2.25.1 (newer versoins have issues)
```
DavidHuber-NOAA, line 26: Can you add the URL for gh?

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -46,6 +107,13 @@ export GH="${HOME}/bin/gh"
 command -v "${GH}"
```
DavidHuber-NOAA: On hera, this causes a silent exit if `gh` is not in `${HOME}/bin`.  Suggest checking that it exists:
```suggestion
[[ -f "${GH}" ]] || echo "gh is not installed in ${HOME}/bin"
```

WalterKolczynski-NOAA (approved):

PR #2755, opened 2024-07-10
Add fixes to products for when REPLAY IC's are used 

EricSinsky-NOAA:
# Description

This PR fixes a couple issues that arise when replay initial conditions are used. These issues only occur when `REPLAY_ICS` is set to `YES` and `OFFSET_START_HOUR` is greater than `0`. The following items are addressed in this PR. 

1. Fix issue that causes ocean_prod tasks not to be triggered (issue [#2725](https://github.com/NOAA-EMC/global-workflow/issues/2725)). A new diag_table was added (called `diag_table_replay`) that is used only when `REPLAY_ICS` is set to `YES`. This diag_table accounts for the offset that occurs when using replay IC's. 
2. Fix issue that causes atmos_prod tasks not to be triggered for the first lead time (e.g. f003) (issue [#2754](https://github.com/NOAA-EMC/global-workflow/issues/2754)). When `OFFSET_START_HOUR` is greater than `0`, the first `fhr` is `${OFFSET_START_HOUR}+(${DELTIM}/3600)`, which is defined in `forecast_predet.sh` and will allow data for the first lead time to be generated. The filename with this lead time will still be labelled with `OFFSET_START_HOUR`. 
3. Minor modifications were made to the extractvars task so that atmos data from replay cases can be processed. 

This PR was split from PR #2680. 

Refs #2725, #2754 

# Type of change

- Bug fix (fixes something broken)

# Change characteristics
- Is this a breaking change (a change in existing functionality)? NO
- Does this change require a documentation update? NO

# How has this been tested?

These changes were cloned, built and ran on WCOSS2. These changes were tested by running GEFS. 

# Checklist
- [ ] Any dependent changes have been merged and published
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] I have made corresponding changes to the documentation if necessary

PR #2778, opened 2024-07-18
Remove f000 from atmos rocoto tasks for replay cases

### LINKED ISSUE #2771: Remove atmos_prod f000 rocoto task when OFFSET_START_HOUR is greater than 0
### What is wrong?

When OFFSET_START_HOUR is greater than 0, the atmos_prod f000 task is still added as a rocoto task. This is an issue because if the OFFSET_START_HOUR is greater than 0, then the atmos_prod f000 task will never be triggered, which results in a stalled atmos_prod task. This may cause downstream tasks to never be triggered, particularly tasks that depend on all lead times to complete for the atmosphere component. 

The atmos_prod f000 task should be removed when OFFSET_START_HOUR is greater than 0. 

All or N/A

### Steps to reproduce

1. Set `REPLAY_ICS` to `YES`.
2. Set `OFFSET_START_HOUR` to a value greater than `0`. 
3. Make sure `atmos` is included as a component. 

### Additional information

Below is a screenshot of an example of the atmos_prod f000 task not being triggered. In this example, `OFFSET_START_HOUR` is 3. 

Screenshot from 2024-07-17 15-22-16

One possible solution is to have an `FHMIN` specifically for the atmosphere, such as `FHMIN_ATMOS`. It should be noted that the wave component has its own `FHMIN`, which is called `FHMIN_WAV`. When `OFFSET_START_HOUR` is greater than 0, `FHMIN_ATMOS` should be set to `OFFSET_START_HOUR`, while the FHMIN for the ocean and ice should be 0.

EricSinsky-NOAA:
# Description

The main purpose of this PR is to remove the f000 from atmosphere-related rocoto tasks when `REPLAY_ICS` is set to `YES`. In cases where `REPLAY_ICS` is `YES` and `OFFSET_START_HOUR `is greater than `0`, it becomes necessary to have the first lead hour set to `OFFSET_START_HOUR ` for the atmosphere-related rocoto tasks. For example, when `OFFSET_START_HOUR ` is set to `3`, then the minimum lead time for the atmos_prod and atmos_ensstat rocoto tasks needs to be 3 and the minimum lead time for the ocean_prod rocoto task needs to be 6 (assuming `FHOUT_OCN` is 6). This PR makes this rocoto workflow set up possible by removing 0 from fhrs for atmosphere-related tasks in `gefs_tasks.py` when replaying. 

This PR also moves where f000 is being removed for the ocean_prod and ice_prod tasks. The if-block that performs this f000 removal has been moved from `tasks.py` to `gefs_tasks.py` and `gfs_tasks.py`.

This PR is a prerequisite for PR #2755. 

Refs #2771 
Resolves #2771 

# Type of change

# Change characteristics

# How has this been tested?

This PR has been tested on WCOSS2 using a C384 GEFS case with `REPLAY_ICS` set to `YES` and `OFFSET_START_HOUR `set to `3`. 

# Checklist


EricSinsky-NOAA:
A C48_S2SWA_gefs CI test case has successfully completed. Also successfully ran a special case where `FHMIN_GFS=3`,  `FHMIN=0` and `REPLAY_ICS=YES`. Marking as ready for review.

aerorahul:
@EricSinsky-NOAA I sent an email discussing the options.  Let's chat offline.

aerorahul (dismissed):
lgtm

EricSinsky-NOAA:
Thank you, @aerorahul. I ran the C48_S2SWA_gefs CI test case, which has successfully completed.

workflow/rocoto/gefs_tasks.py
```diff
@@ -270,6 +270,17 @@ def _atmosoceaniceprod(self, component: str):
                      'maxtries': '&MAXTRIES;'}
 
         fhrs = self._get_forecast_hours('gefs', self._configs[config], component)
+
+        # when replaying, atmos component does not have fhr 0, therefore remove 0 from fhrs
+        local_config = self._configs[config].deepcopy()
+        is_replay = local_config['REPLAY_ICS']
```
WalterKolczynski-NOAA: I don't think we need to make a deep copy here. We're not changing the configs, just reading them. We can use the original for that.
```suggestion
        is_replay = self._configs[config]['REPLAY_ICS']
```

workflow/rocoto/gefs_tasks.py
```diff
@@ -314,6 +325,13 @@ def atmos_ensstat(self):
                      'maxtries': '&MAXTRIES;'}
 
         fhrs = self._get_forecast_hours('gefs', self._configs['atmos_ensstat'])
+
+        # when replaying, atmos component does not have fhr 0, therefore remove 0 from fhrs
+        local_config = self._configs['atmos_ensstat'].deepcopy()
+        is_replay = local_config['REPLAY_ICS']
```
WalterKolczynski-NOAA: Same here
```suggestion
        is_replay = self._configs['atmos_ensstat']['REPLAY_ICS']
```

WalterKolczynski-NOAA (changes requested):

workflow/rocoto/gefs_tasks.py
EricSinsky-NOAA: Sure. Deep copy has been removed here and for atmos_ensstat.

aerorahul (approved):

WalterKolczynski-NOAA (approved):
Conditionally approved

PR #2789, opened 2024-07-24
Replace Analysis class with JEDI class

DavidNew-NOAA:
# Description

This PR replaces the PyGFS ```Analysis``` class with a more general ```JEDI class```. It is a child class of the ```wxflow``` ```Task``` class, like the ```Analysis```, but it's intended for any Global Workflow job that runs one or more JEDI applications. At this point, it has only been applied to the atmospheric variational and ensemble analysis jobs, so the Analysis class still remains part of the code base, but can be fully replaced after this PR using the atmospheric analysis jobs as a template for the other analysis jobs (snow, aerosol, etc).

In order to make things more generic, the input YAML will now always be named after the JEDI application being run. So, for example, if running ```gdas.x```, the name of the input YAML is ```gdas.yaml```, rather than, say, ```gdas.t12z.atmvar.yaml```. In the case of the atmospheric variational analysis, the ```finalize()``` method of the ```AtmAnalysis``` class will ensure that ```gdas.yaml``` is saved as ```gdas.t12z.atmvar.yaml``` in the COMROT directory anyway, so no need to give it such a descriptive name in the run directory.

The ```AtmAnalysis``` and ```AtmEnsAnalysis``` classes, now subclasses of ```JEDI```, now take care of staging observations and bias corrections in their own ```initialize()``` methods, rather than in the ```initialize()``` method of ```JEDI```, but the ```JEDI``` ```initalize()``` method continues to render the input Jinja2-YAMLS/JCB templates and link the JEDI executable file. The ```JEDI``` class ```initalize()``` method now also saves the final YAML in the run directory, rather than having the ```initalize()``` method of the ```AtmAnalysis``` and ```AtmEnsAnalysis``` classes do that (since all JEDI applications require an input YAML file saved to disk.

A generalized ```execute()``` method now exists in ```JEDI``` which takes the APRUN command as input, and optionally a list of additional arguments that might be passed to ```gdas.x``` (eg. ['fv3jedi', 'variational']).

# Type of change

# Change characteristics

# How has this been tested?
- Build on Hera
- Run cycling experiment

# Checklist


ush/python/pygfs/task/jedi.py
```diff
@@ -0,0 +1,300 @@
...
+                    parse_j2yaml, save_as_yaml,
+                    logit,
+                    Task,
+                    Executable,
+                    WorkflowException)
+
+logger = getLogger(__name__.split('.')[-1])
+
+
+class JEDI(Task):
```
guillaumevernieres, line 19: This looks like a class hard-wired for the analysis step, which I think is fine but the name should probably reflect this. `JEDIAnalysis`? or just keep `Analysis`?
I think @CoryMartin-NOAA (or maybe that was you @DavidNew-NOAA ?) started a b-matrix base class as well. We probably want to keep these separate.

ush/python/pygfs/task/jedi.py
```diff
@@ -0,0 +1,300 @@
...
+        super().initialize()
+
+        # Generate and write JEDI input YAML file
+        self.get_jedi_config()
+
+        # Link JEDI executable to run directory
+        self.link_jedi_exe()
+
+    @logit(logger)
+    def execute(self, aprun_cmd: str, jedi_args: Optional[List] = None) -> None:
```
guillaumevernieres, line 54: Running only one executable wouldn't work for the "B-matrix" or "prep obs" jobs for example, thus the suggestion above to rename to something descriptive of the analysis job.

ush/python/pygfs/task/jedi.py
```diff
@@ -0,0 +1,300 @@
...
+            raise KeyError(f"Task config must contain JCB_ALGO, JCB_BASE_YAML, or JEDIYAML")
+
+        logger.debug(f"JEDI config:\n{pformat(self.task_config.jedi_config)}")
+
+        # Save JEDI config to YAML file
+        logger.debug(f"Generate YAML file: {self.task_config.jedi_yaml}")
+        save_as_yaml(self.task_config.jedi_config, self.task_config.jedi_yaml)
+        logger.info(f"Wrote YAML to: {self.task_config.jedi_yaml}")
+
+    def link_jedi_exe(self) -> None:
```
guillaumevernieres, line 132: we probably should have this in a common utility module

ush/python/pygfs/task/jedi.py
```diff
@@ -0,0 +1,300 @@
...
+
+        bias_dict = {
+            'mkdir': [os.path.join(self.task_config.DATA, 'bc')],
+            'copy': copylist
+        }
+        return bias_dict
+
+
+@logit(logger)
+def find_value_in_nested_dict(nested_dict: Dict, target_key: str) -> Any:
```
guillaumevernieres, line 233: same as above, probably belongs to wxflow or some other common utility module

guillaumevernieres (commented):
Thanks for doing this @DavidNew-NOAA . Just a few thoughts/comments.

DavidNew-NOAA:
I'm closing the pull request until I make some modifications. Rahul and I discussed it and agree, based on previous conversations with Dan and Cory, that the JEDI class should not be inherited from the Task class, but rather any child class of Task (eg AtmAnalysis, AtmEnsAnalysis, etc) should have the JEDI class as a member.

ush/python/pygfs/task/jedi.py
DavidNew-NOAA: I agree

ush/python/pygfs/task/jedi.py
DavidNew-NOAA: The intention is for this class to be used anytime a JEDI application is run in general, not just for running an analysis

PR #2791, opened 2024-07-25
Add fix for metplus jobs when gfs_cyc=2 or 4

DavidHuber-NOAA:
# Description
Changes how METplus jobs run so that they run on the last GFS cycle for a given `PDY`.

This is a departure from operations where the METplus jobs run on the 00Z cycle for the previous 3 cycles and 00Z (i.e. `${PDYm1}06` through `${PDY}00`).  With this PR, for gfs_cyc=4, METplus jobs will run on `${PDY}18` for cycles 00-18.  See https://github.com/NOAA-EMC/EMC_verif-global/pull/131 for more details.

`ecf` scripts may need to be modified.

Depends on https://github.com/NOAA-EMC/EMC_verif-global/pull/131
Resolves #2790 

# Type of change
- Bug fix (fixes something broken)
- New feature (adds functionality)

# Change characteristics
- Is this a breaking change (a change in existing functionality)? YES (departure from operations)
- Does this change require a documentation update? NO

# How has this been tested?
Will test a full suite on WCOSS2

# Checklist
- [x] Any dependent changes have been merged and published
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing tests pass with my changes
- [x] I have made corresponding changes to the documentation if necessary

PR #2799, opened 2024-07-30
Check that a PR driver is still running before trying to kill it

### LINKED ISSUE #2798: Bash CI attempts to kill non-existent driver instances
### What is wrong?

Occasionally, a PR fails for one reason or another and the Bash CI attempts to kill the driver via SSH on a different head node.  However, in some circumstances, the driver may have crashed.  In those cases, the script (ci/scripts/driver.sh) still attempts to look for and kill all child processes, but is unable to find the parent process, resulting in a failure.

Also, if a PR directory is manually deleted, the Bash CI will fail to write to a log file, causing a failure

The driver.sh script should first check if the PID is still running before attempting to kill it.  Also, the same script should check for the existence of the PR directory and create it if it does not exist.

WCOSS2

### Steps to reproduce

Rather difficult to reproduce.

### Additional information

Found while testing https://github.com/NOAA-EMC/global-workflow/pull/2791

Add a check to the ssh command that the driver.sh PID is still running.

DavidHuber-NOAA:
# Description
Adds a check to the SSH command used to kill child PIDs of a defunct driver instance on a different head node to prevent invalid kill commands, preventing CI failures.

Resolves #2798

# Type of change

# Change characteristics

# How has this been tested?
Successfully detected the driver was no longer running on WCOSS2 for PR https://github.com/NOAA-EMC/global-workflow/pull/2791.

# Checklist


WalterKolczynski-NOAA (dismissed):
Looks good to me, but hold for @TerrenceMcGuinness-NOAA to get back and review it.

ci/scripts/check_ci.sh
TerrenceMcGuinness-NOAA, line 1: It is my understanding that the "+x" is generally the best practice for checking if a variable is set in BASH.  Please let me know why we should remove these here.

ci/scripts/driver.sh
```diff
@@ -107,7 +108,9 @@ for pr in ${pr_list}; do
            echo -e "${pstree_out}" | grep -Pow "(?<=\()[0-9]+(?=\))" | xargs kill
         fi
       else
-        ssh "${driver_HOST}" 'pstree -A -p "${driver_PID}" | grep -Eow "[0-9]+" | xargs kill'
+        # Check if the driver is still running on the head node; if so, kill it and all child processes
+        #shellcheck disable=SC2029
+        ssh "${driver_HOST}" "if ps -p ${driver_PID} 2>&1; then pstree -A -p \"${driver_PID}\" | grep -Eow \"[0-9]+\" | xargs kill; fi"
```
TerrenceMcGuinness-NOAA:
```suggestion
        ssh "${driver_HOST}" 'pstree -A -p "${driver_PID}" | grep -Eow "[0-9]+" | xargs kill || echo "Failed to kill process with PID: ${driver_PID}, it may not be valid."'
```

TerrenceMcGuinness-NOAA: This might be better.

WalterKolczynski-NOAA:
@TerrenceMcGuinness-NOAA Please review this so it can move forward.

TerrenceMcGuinness-NOAA (changes requested):

ci/scripts/driver.sh
DavidHuber-NOAA: Single quotes prevent variables from being expanded, so this needs to be double quoted.  I do like the 'Failed to kill' message, though.
```suggestion
        ssh "${driver_HOST}" "pstree -A -p \"${driver_PID}\" | grep -Eow \"[0-9]+\" | xargs kill || echo \"Failed to kill process with PID: ${driver_PID}, it may not be valid.\""
```

ci/scripts/check_ci.sh
DavidHuber-NOAA: When checking if a variable (say `${bogus_var}`) is defined, using the `-z` check with `+x` parameter expansion is the way to go, e.g.
```bash
unset bogus_var
[[ -z ${bogus_var+x} ]] && echo "Variable 'bogus_var' is undefined!" || echo "Variable 'bogus_var' is defined as ${bogus_var}"
#> Variable 'bogus_var' is undefined!
```

But if you want to check if the variable is defined but set to an empty string, then you should use `-z` without `+x`:
```bash
unset bogus_var
bogus_var=""
[[ -z ${bogus_var+x} ]] && echo "Variable 'bogus_var' is undefined!" || echo "Variable 'bogus_var' is defined as ${bogus_var}"
#> Variable 'bogus_var' is defined as 
[[ -z ${bogus_var} ]] && echo "Variable 'bogus_var' is an empty string!"
#> Variable 'bogus_var' is an empty string!
```

And lastly, if you want to check if a variable is defined to a non-empty string, use the `-n` test:
```bash
unset bogus_var
bogus_var="foo"
[[ -z ${bogus_var+x} ]] && echo "Variable 'bogus_var' is undefined!" || echo "Variable 'bogus_var' is defined as ${bogus_var}"
#> Variable 'bogus_var' is defined as foo
[[ -n ${bogus_var} ]] && echo "Variable 'bogus_var' is defined as ${bogus_var}!"
#> Variable 'bogus_var' is defined as foo
```

For the tests that are performed in driver.sh, the variables are all defined.  For instance, in
https://github.com/NOAA-EMC/global-workflow/blob/4ef3a00a447e3a78bb1db879f2c39be47b47a380/ci/scripts/check_ci.sh#L52-L53
`rocotostat` is either defined as the path to the `rocotostat` command if it is found or an empty string if it is not.  On line 53, we want to know if `rocotostat` was not found, so we should check for an empty string.

TerrenceMcGuinness-NOAA: Yes yes I reviewed the update in driver on line 52 I saw that `-n` was defiantly the way to go on the first pass.  And yes thank you for taking the time to explicitly pointing out why we don't need `+x` here.

TerrenceMcGuinness-NOAA (approved):
Nice updates.

PR #2795, opened 2024-07-25
Add capability to run forecast in segments

WalterKolczynski-NOAA:
# Description
Adds the ability to run a forecast in segments instead of all at once. To accomplish this, a new local `checkpnts` variable is introduced to `config.base` to contain a comma-separated list of intermediate stopping points for the forecast. This is combined with `FHMIN_GFS` and `FHMAX_GFS` to create a comma-separated string `FCST_SEGMENTS` with all the start/end points that is used by `config.fcst` and rocoto workflow. Capability to parse these into python lists was added to wxflow in an accompanying PR. If `checkpnts` is an empty string, this will result in a single-segment forecast.

To accommodate the new segment metatasks that must be run serially, the capability of `create_task()` was expanded to allow a dictionary key of `is_serial`, which controls whether a metatask is parallel or serial using pre-existing capability in rocoto. The default when not given is parallel (i.e. most metatasks).

Resolves #2274
Refs NOAA-EMC/wxflow#39
Refs NOAA-EMC/wxflow#40

# Type of change
- New feature (adds functionality)

# Change characteristics
- Is this a breaking change (a change in existing functionality)? NO
- Does this change require a documentation update? NO

# How has this been tested?
- Forecast-only on Hercules
- GEFS on Hercules

# Checklist
- [x] Any dependent changes have been merged and published
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing tests pass with my changes
- [x] I have made corresponding changes to the documentation if necessary

PR #2753, opened 2024-07-08
Clean-up temporary rundirs - take 2.

aerorahul:
# Description
This PR:
- is a follow-up to a previous PR that aggressively pruned run directories. 
- removes run directories for the current cycle in the clean-up if the cycle is successful.  If the cycle is not successful, cleanup is not called and all run directories for the cycle are safe from being purged.
- also updates the PR template to list/query for any updates to submodules.

# Type of change
- Bug fix (fixes something broken)
- New feature (adds functionality)
- Maintenance (code refactor, clean-up, new CI test, etc.)

# Change characteristics
- Is this a breaking change (a change in existing functionality)? Hopefully NO
- Does this change require a documentation update? NO

# How has this been tested?
Ran C48_ATM and C96_atm3dvar 

# Checklist
- [ ] Any dependent changes have been merged and published
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] I have made corresponding changes to the documentation if necessary

PR #2804, opened 2024-08-02
Run METplus serially and correct the name of prod tasks

DavidHuber-NOAA:
# Description
Adds 3 hot fixes:

- METplus v9.1.3 has a bug in it that sometimes attempts to create multiple copies of the same directory when running in parallel, causing a Python error and downstream problems.  This PR makes METplus run in serial mode, preventing such issues.
- Corrects the name of the atmos_prod, ocean_prod, and ice_prod tasks in workflow/rocoto/tasks.py (was accidentally changed to e.g. atmosprod)

# Type of change

# Change characteristics
- Is this a breaking change (a change in existing functionality)? YES (For experiments running more than ~20 cycles, the gfsmetpg2o job may time out.)

# How has this been tested?
CI unit tests for the atmos_prod fix, will run CI.

# Checklist

### DIFF
--- a/parm/config/gfs/config.resources
+++ b/parm/config/gfs/config.resources
@@ -889,8 +889,8 @@ case ${step} in
     threads_per_task=1
     walltime_gdas="03:00:00"
     walltime_gfs="06:00:00"
-    ntasks=4
-    tasks_per_node=4
+    ntasks=1
+    tasks_per_node=1
     export memory="80G"
     ;;
 
--- a/workflow/rocoto/gfs_tasks.py
+++ b/workflow/rocoto/gfs_tasks.py
@@ -1104,7 +1104,7 @@ class GFSTasks(Tasks):
         data = f'{history_path}/{history_file_tmpl}'
         dep_dict = {'type': 'data', 'data': data, 'age': 120}
         deps.append(rocoto.add_dependency(dep_dict))
-        dep_dict = {'type': 'task', 'name': f'{self.cdump}fcst'}
+        dep_dict = {'type': 'task', 'name': f'{self.run}fcst'}
         deps.append(rocoto.add_dependency(dep_dict))
         dependencies = rocoto.create_dependency(dep=deps, dep_condition='or')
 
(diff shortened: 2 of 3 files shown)


DavidHuber-NOAA:
@malloryprow Could you add the gdas/nam prepbufr files for 2021032300 through 2021033000 to Hera (under `/scratch1/NCEPDEV/global/Mallory.Row/prepbufr/`) ~and Orion/Hercules (under `/work/noaa/rstprod/verif/prepbufr`)~?

EDIT:
Orion/Hercules actually already has this data.

DavidHuber-NOAA:
@aerorahul @WalterKolczynski-NOAA The failures on WCOSS2 are due to a couple of JJob scripts that do not source `config.resources`.  These will be fixed in #2753.  The failures in gfsmetpg2o1 on Hera are due to missing prepbufr input files and will be uploaded soon.

I believe this PR is ready to merge.

DavidHuber-NOAA:
The prepbufr files have been pushed to Hera.  Thanks @malloryprow!!

parm/config/gfs/config.resources
```diff
@@ -889,8 +889,8 @@ case ${step} in
     threads_per_task=1
     walltime_gdas="03:00:00"
     walltime_gfs="06:00:00"
-    ntasks=4
-    tasks_per_node=4
+    ntasks=1
+    tasks_per_node=1
```
WalterKolczynski-NOAA, line 893: This wasn't what I had in mind when I said "serial", and I don't think fixes anything. Based on my understanding, the tasks need to be serial, not the job execution. Marking a metatask as serial is a capability I'm trying to add in #2795. Recommend taking this out and only addressing the name issue.

WalterKolczynski-NOAA (changes requested):

parm/config/gfs/config.resources
DavidHuber-NOAA: @WalterKolczynski-NOAA I understand now by what you mean by serial.  That said, there is an intermittent bug in METplus when running it with more than 1 task, and this does fix that issue.  See [this comment](https://github.com/NOAA-EMC/global-workflow/pull/2791#issuecomment-2260424914) for more details.

WalterKolczynski-NOAA: That's the one I was referring to as well, I just misunderstood the problem I guess.

WalterKolczynski-NOAA (approved):

Issue #2674, opened 2024-06-11
Missing CICE hist files when cycle is not t00z and FHOUT_GFS_ICE is 24


EricSinsky-NOAA:
### What is wrong?

For cases where `FHOUT_GFS_ICE` is 24 and initialized at a time that is not 00Z, CICE files are not properly linked to COMROOT. This results in missing CICE hist files in COMROOT. After investigating this issue on WCOSS2, this happens because the raw CICE output is not averaged for a full 24 hours for the first lead time. For example, if the initialization time is 12Z and FHOUT_GFS_ICE is 24, the first lead time will only be averaged from 12Z to 00Z. Subsequent lead times are averaged from 00Z to 00Z. The global-workflow predetermines that the CICE output will be averaged from 12Z to 12z, which results in the global workflow creating links that do not match the raw CICE filenames, resulting in the CICE output not to save in COMROOT. 

### What should have happened?

It may be ideal that CICE be averaged for a full 24 hours for the first lead time. For example, if the initialization time is 12Z and `FHOUT_GFS_ICE` is 24, the first lead time should be averaged from 12Z to 12Z. Subsequent lead times should be averaged from 12Z to 12Z. An alternative positive outcome would be for the global-workflow to correctly predetermine the name of the CICE output in these particular cases. 

### What machines are impacted?

WCOSS2

### Steps to reproduce

1. Set `FHOUT_GFS_ICE` to 24 in `config.base`.
2. Run a case that is not initialized at t00Z (e.g. t12z). 

### Additional information

This issue has been tested on WCOSS2, but it most likely impacts all platforms. The list of CICE output below illustrates how the global-workflow is not picking up the raw CICE output. 
![link_mismatch2](https://github.com/NOAA-EMC/global-workflow/assets/48259628/463cd454-6dd9-4f37-be60-c255fb5956f6)

### Do you have a proposed solution?

This issue has been initially addressed in PR #2561. In this PR, CICE_predet and CICE_postdet were modified so that the global-workflow can correctly predetermine the name of the raw CICE output (as illustrated in the image below). Also, the oceanice prod task was modified accordingly to account for the CICE filename changes in these cases. However, further fixes will need to be made to consider cases where `FHMIN` is not 0. 

![CICE_final_output_fix](https://github.com/NOAA-EMC/global-workflow/assets/48259628/31679dac-95c4-420b-ae2c-417a5ca60837)

aerorahul:
[...]

PR #2809, opened 2024-08-06
Remove need for a ice_prod dependency check script 

### LINKED ISSUE #2721: Update dependency check for ice netcdf file for GEFS
### What is wrong?

#2561 introduced a dependency check for ice netcdf file for GEFS via the use of [this](https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/check_ice_netcdf.sh) script.  Rocoto provides an option for checking if the file exists together with the age of the file and should be utilized instead of this script.

Use Rocoto `<datadep age=120>` to check if the file exists and is older than 2 mins, instead of calling a script via `<sh>` command.

All or N/A

### Steps to reproduce

Create the xml and look at the gefs ice product dependency section.

### Additional information

None

Use Rocoto's `datadep` tag to check for dependencies.

EricSinsky-NOAA:
# Description

The main purpose of this PR is to remove the need for an ice_prod dependency check script `ush/check_ice_netcdf.sh`. The original purpose of the ice_prod dependency check script is to check for special case dependencies where `( cyc + FHMIN ) % FHOUT_ICE )) =! 0` (more details on this issue can be found in issue #2674 ). A bugfix for these special cases is expected to come from a PR in the ufs-weather-model. 

Resolves #2721 
Refs #2721, #2674  

# Type of change

# Change characteristics

# How has this been tested?

- Clone, built and tested on WCOSS2. 
- In the WCOSS2 test, `FHOUT_OCN` and `FHOUT_ICE` was set to `24` and the GEFS CI case was successfully tested (202103232 12Z). 
- TODO: Perform test cases with `FHOUT_OCN` and `FHOUT_ICE` set to `6`.

# Checklist


ush/forecast_postdet.sh
```diff
@@ -628,6 +628,36 @@ CICE_out() {
 
 }
 
+CICE_avg() {
+  if (( CICE_cdoavg == 0 )); then
+    echo "SUB ${FUNCNAME[0]}: No further averaging needed for CICE"
+  else
+    echo "SUB ${FUNCNAME[0]}: Averaging output data for CICE"
+    CDO=/apps/prod/hpc-stack/intel-19.1.3.304/cdo/1.9.8/bin/cdo
```
EricSinsky-NOAA: If we choose to use the cdo utility for this, it may need to be added somewhere in `jobs/JGLOBAL_FORECAST`. This line is just a temporary workaround for the cdo utility being undefined in the forecast task environment.

WalterKolczynski-NOAA: The cdo binary directory is already added to the path by the cdo module. You should be able to just call it with `cdo` without this.

WalterKolczynski-NOAA: NB: We have it in the base modulefile used for most jobs, but the model modulefile isn't loading cdo, so it would need to be added to [load_ufswm_modules.sh](https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/load_ufswm_modules.sh).

EricSinsky-NOAA: Thanks, @WalterKolczynski-NOAA. Will module load cdo in [load_ufswm_modules.sh](https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/load_ufswm_modules.sh).

EricSinsky-NOAA:
Bugfix has been removed for special cases where  `( cyc + FHMIN ) % FHOUT_ICE )) =! 0` in commit fcb9003f44892656a05ab7b82f58c1e0107f0753. This bug is expected to be fixed in ufs_weather_model [PR #2437](https://github.com/ufs-community/ufs-weather-model/pull/2437). 
The primary purpose of this PR is to remove ice_prod dependency check script and any temporary bugfixes associated with special cases where  `( cyc + FHMIN ) % FHOUT_ICE )) =! 0`. 
Testing GEFS replay and non-replay cases. This PR will be marked ready for review when these tests pass.

EricSinsky-NOAA:
Marking this PR as ready for review. GEFS replay (`FHOUT_GFS_ICE=24`, `cyc=00`) and non-replay test cases (`FHOUT_GFS_ICE=6`, `cyc=12`) have passed.

workflow/rocoto/gefs_tasks.py
```diff
@@ -219,11 +219,6 @@ def _atmosoceaniceprod(self, component: str):
             dep_dict = {'type': 'metatask', 'name': 'fcst_mem#member#'}
             deps.append(rocoto.add_dependency(dep_dict))
             dependencies = rocoto.create_dependency(dep=deps, dep_condition='or')
-        elif component in ['ice']:
-            command = f"{self.HOMEgfs}/ush/check_ice_netcdf.sh @Y @m @d @H #fhr# &ROTDIR; #member# {fhout_ice_gfs}"
-            dep_dict = {'type': 'sh', 'command': command}
-            deps.append(rocoto.add_dependency(dep_dict))
-            dependencies = rocoto.create_dependency(dep=deps)
```
WalterKolczynski-NOAA, line 226: Should ice be added to the `if` rather than the `else`so it doesn't begin pre-maturely?

WalterKolczynski-NOAA (changes requested):

workflow/rocoto/gefs_tasks.py
EricSinsky-NOAA: You are right. After looking at this further, it looks like there is no need for an if-statement, since the atmos_prod, ocean_prod and ice_prod dependencies are structured the same way (all checking for the existence of a history file or the completion of a fcst metatask). I set this up similar to how the ocn/ice/atmos prod dependencies are set up in `workflow/rocoto/gfs_tasks.py`. These changes have been pushed.

WalterKolczynski-NOAA (approved):

aerorahul (approved):
looks good to me.
thanks for taking out the shell script dependency checker.
It was causing significant delays executing rocoto-stuff for long forecasts (SFS).

WalterKolczynski-NOAA:
Ignore the Orion failure. We already know what that is.

PR #2815, opened 2024-08-07
Jenkins Pipeline Updates

TerrenceMcGuinness-NOAA:
# Description
Pipeline updates:
- get `gh` location on remote machine and define global bash env variable `$GH` for GitHub CLI
-  Failed cases are now displayed accordingly in the Jenkins dashboard (see NOTE below)
-  Added the Build # in messaging for clarity when running from re-runs.
- Replaced Matrix construct for concurrency with the parallel method that can use dynamic case lists
  - With removing of the hard coded list of cases we now get list of cases dynamically from the PR case directory
  - See new look of dashboard below (has more annotations and displays only used cases)

NOTE: **failFast** (quitting all cases on failing of one) still does not work because it isn't quitting the running remote shells.  We can make this a configurable capability in feature request with some custom code.  The current behavior has the remaining cases continuing to run after a FAIL label has been issued and it is incumbent of the code manager to kill the CI job in the controller before resetting another Ready label.

# Type of change

# How has this been tested?
Ran in the authoritative repo with full self-tests

updateddashboard


ci/Jenkinsfile
```diff
@@ -223,6 +227,7 @@ pipeline {
                             expression { return caseList.contains(Case) }
                         }
                         steps {
+                            catchError(buildResult: 'FAILURE', stageResult: 'FAILURE') {
                             script {
```
WalterKolczynski-NOAA: This whole block (to L278) should be indented one more level now.

WalterKolczynski-NOAA (changes requested):

ci/Jenkinsfile
TerrenceMcGuinness-NOAA: Yeah thanks [sic]

ci/Jenkinsfile
```diff
@@ -17,6 +17,10 @@ pipeline {
         parallelsAlwaysFailFast()
     }
 
+    environment {
+        GH = '~/bin/gh'
```
DavidHuber-NOAA: Wonder if this would be better so `gh` doesn't necessarily need to be in the home directory (e.g. if it ends up getting installed into spack-stack or a user has it in ~/.local/bin):
```suggestion
        GH = sh(script: "which gh || echo '~/bin/gh'", returnStdout: true).trim()
```

WalterKolczynski-NOAA (approved):

PR #2727, opened 2024-06-27
Simplify resource-related variables, remove CDUMP where unneeded

DavidHuber-NOAA:
# Description
This overhauls resource-related variables to use a common set of variables for each job.  In the process, this also removed the use of CDUMP in most cases.

Resolves #1299 #2693

Depends on https://github.com/NOAA-EMC/wxflow/pull/35

# Type of change
- New feature (adds functionality)
- Maintenance (code refactor, clean-up, new CI test, etc.)

# Change characteristics
- Is this a breaking change (a change in existing functionality)? NO
- Does this change require a documentation update? NO

# How has this been tested?
Created .xml and compared against develop.  Other tests to follow on Hera and WCOSS2:
- [x] C48_ATM
- [x] C48_S2SW
- [x] C48_S2SWA_gefs
- [x] C48mx500_3DVarAOWCDA (hera only)
- [x] C96C48_hybatmDA
- [x] C96C48_ufs_hybatmDA
- [x] C96_atm3DVar_extended (wcoss2 only)
- [x] C96_atmaerosnowDA

And basic tests on Hercules:
- [x] C48_ATM
- [x] C48_S2SW
- [x] C96C48_hybatmDA

# Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing tests pass with my changes

PR #2839, opened 2024-08-16
Consolidate python configuration dictionaries

### LINKED ISSUE #2783: Consolidate Python resource structure

PR https://github.com/NOAA-EMC/global-workflow/pull/2727 introduced a simplified method to calculate resources for each job and machine.  In doing so, it reworks the way resources are handled when running `setup_xml.py` such that the configuration dictionary is redefined for each required `RUN`.  Rather than redefining the same dictionary each time, a single dictionary with subdictionaries for each `RUN` should be implemented.  This will allow all configuration entries to be retained during execution.

- Methods that were hidden before #2727 should be made hidden again
- Resource calculations do not change

### Acceptance Criteria

No changes to resource calculations and CI tests run successfully on all platforms.

_No response_

DavidHuber-NOAA:
# Description
This modifies the way the `config` dictionary is constructed and referenced.  Rather than updating a single configuration dictionary with each `RUN`, a `RUN`-based dictionary of `config` dictionaries is created and referenced by the appropriate `RUN` when calculating resources.

This also makes the methods that were hidden before #2727 hidden again.

Resolves #2783 

# Type of change

# Change characteristics

# How has this been tested?
XMLs for all CI tests were identical to develop on
- [x] Orion
- [x] Hercules
- [x] Hera
- [x] WCOSS2

# Checklist


DavidHuber-NOAA:
I still need to repeat the XML tests on WCOSS2, Hera, and Hercules, but this should be ready for review.  When ready, I don't think this needs to go through CI or, at most, just through the setup stage.

DavidHuber-NOAA:
I was able to repeat the XML tests on all platforms.  All were identical.  This PR is now ready to be reviewed.

DavidHuber-NOAA:
@aerorahul I just recalled the conversation yesterday about the `_` dictionary entry.  I will update that presently and retest on Hera.

DavidHuber-NOAA:
@aerorahul The no-RUN configs key has been renamed now.

workflow/applications/applications.py
```diff
@@ -32,57 +31,50 @@ def __init__(self, conf: Configuration) -> None:
 
         self.scheduler = Host().scheduler
 
-        # Save the configuration so we can source the config files when
-        # determining task resources
-        self.conf = conf
+        self._base = conf.parse_config('config.base')
```
WalterKolczynski-NOAA: Don't like making this an object field. This version should not be needed beyond initialization. All later uses should be using `self.configs[run]['base']`.
```suggestion
        base = conf.parse_config('config.base')
```

workflow/applications/applications.py
```diff
@@ -91,45 +83,49 @@ def __init__(self, conf: Configuration) -> None:
...
+    def _init_finalize(self, conf: Configuration):
         print("Finalizing initialize")
 
         # Get a list of all possible config_files that would be part of the application
         self.configs_names = self._get_app_configs()
 
-        # Source the config_files for the jobs in the application
-        self.configs = self.source_configs()
+        # Source the config files for the jobs in the application without specifying a RUN
+        self.configs = {'_no_run': self._source_configs(conf)}
```
WalterKolczynski-NOAA, line 104: Not sure why a RUN-less version is necessary. All variables needed outside of a `run` (like mode) should already be an object field set in `init()`. Otherwise, there should be a `run`.

workflow/applications/applications.py
```diff
@@ -156,7 +152,7 @@ def source_configs(self, run: str = "gfs", log: bool = True) -> Dict[str, Any]:
         configs = dict()
 
         # Return config.base as well
-        configs['base'] = self.conf.parse_config('config.base')
+        configs['base'] = conf.parse_config('config.base')
```
WalterKolczynski-NOAA: This should also be based on run
```suggestion
        configs['base'] = conf.parse_config('config.base', RUN=run)
```

workflow/rocoto/tasks.py
```diff
@@ -39,15 +38,16 @@ class Tasks:
...
-        print(f"Source configs with RUN={run}")
-        self._configs = self.app_config.source_configs(run=run, log=False)
+
+        # Get the configs for the specified RUN
+        self._configs = self.app_config.configs[run]
 
         # Update the base config for the application
-        self._configs['base'] = self.app_config.update_base(self._configs['base'])
-        # Save dict_configs and base in the internal state (never know where it may be needed)
+        self._configs['base'] = self.app_config._update_base(self._configs['base'])
```
WalterKolczynski-NOAA, line 48: Is there a reason the limited function of `_update_base` can't be handled at init time instead of deferring it to this point?

WalterKolczynski-NOAA (changes requested):

workflow/applications/applications.py
DavidHuber-NOAA: This is because of the subclass `update_base` call in `init_finalize`.  Currently, this method is only called for the `_no_run` configuration dictionary, so only that `config['base']` has, e.g., `SDATE_GFS` in it.  I could rework this so that `update_base` is called for all configuration dictionaries.  Would you prefer that?

If we go this route, and we want to specify a `run` for all downstream references to the configuration dictionaries, then we will need to know which `run` to call at the appropriate time.  For instance, a cycled experiment may or may not have a `gfs` `run` while a forecast-only experiment will only have a `gfs` or `gefs` run.  We could solve this by retrieving an arbitrary `run`, i.e.
```python
arb_run=next(iter(configs.values()))
```

But if we go that route, then a `_no_run` configuration dictionary looks more attractive as you will always end up with the same `config.base`.  But if we plan on adding `RUN`-specific variables to `config.base` that are needed at setup time, then both of these approaches may not suffice long term.

The last option I can think of is to abandon `_no_run` and add a function with logic to check for available `run`s and pick the most logical choice.  E.g.
```python
def select_run(configs, mode)
  cycled = mode == "cycled"
  if cycled:
    if "gfs" in configs:
      return "gfs"
    elif "gefs" in configs: # Assuming gefs will one day cycle
      return "gefs"
    elif "gdas" in configs:
      return "gdas"
    else:
      raise Exception("Invalid configuration setup!")
  else:
    if "gfs" in configs:
      return "gfs"
    elif "gefs" in configs":
      return "gefs"
    else:
      raise Exception("Invalid configuration setup!")
```

Is this last option what you would like to see?  Or is there more down the road that we need to consider before selecting another option?

WalterKolczynski-NOAA: The last option is closer to what I was thinking, but I think we need to do more here. We're already building a list of component runs in init(), but even that I think should be changed so that `do_*` is run-based. This lets us cleanly just change the APP based on RUN in `config.base` and have everything that follows from that just work. This might require moving them to properties of Task instead of AppConfig.

workflow/applications/applications.py
DavidHuber-NOAA: Alright, I removed the `_base` attribute from the `AppConfig` class and subclasses.  `_no_run` is no longer referenced anywhere outside of `_init_finalize`.  This is still required to determine which `run`s are valid for the experiment.

workflow/applications/applications.py
DavidHuber-NOAA: Fixed.

workflow/rocoto/tasks.py
DavidHuber-NOAA: I'm going to test this out, but I think that since `_update_base` is defined in the subclass, this needs to be in `_init_finalize` or copied into each of the subclass `__init__` methods.

DavidHuber-NOAA: The `_update_base` isn't an issue.  However, the subclass `__init__` method (at least the `GFSCycledAppConfig` class) sets some flags needed by the subclass's `_get_app_configs` method.  I could resolve that by moving these flag declarations to the `AppConfig` `__init__` method (e.g. `do_jediatmvar`, `do_jediocnvar`, `do_hybvar`), but that seems like a kludge.  I prefer the `_init_finalize` method as it is.

workflow/applications/applications.py
DavidHuber-NOAA: I wasn't able to implement the last option completely.  Instead, I isolated the `_no_run` to just the `_init_finalize` method where the valid `run`s are determined.  After that, `run` is used when referencing any config.

I started to look into moving the `do_*` flags to `Tasks`.  Since `_get_app_configs` needs to know these flags when `_init_finalize` is called in order to determine which configs to include, I don't think it is feasible to move them.  However, I do think they can be set based on `RUN` and determined within `AppConfig` subclasses.  I'll look into this more next week.

WalterKolczynski-NOAA:
Deferring any final elimination of the no_run to a future PR.

WalterKolczynski-NOAA (dismissed):

DavidHuber-NOAA:
@WalterKolczynski-NOAA The `gdasverfozn` job failed on Hercules due to walltime exceedance for both the C96_atm3DVar and C96C48_hybatmDA cases.  All other tests were successful.  I have increased the walltime request to 10 minutes from 5.  Do the tests need to be rerun?

WalterKolczynski-NOAA (approved):

PR #2842, opened 2024-08-16
Fixed test on UNAVAILBLE in python Rocoto check

TerrenceMcGuinness-NOAA:
# Description

Update the test for UNAVAILABLE from inclusion to count: `if ['UNAVAILABLE'] > 0`
in `rocotocheck.py `script.

# Type of change

The test for UNAVAILABLE was for inclusion so the tool was always return the state UNAVAILABLE until DONE

# Change characteristics

# How has this been tested?
Ran tool on EXPDIR that had zero UNAVAILABLE and UNKNOWN states


DavidHuber-NOAA (approved):
Looks good, thanks for fixing this @TerrenceMcGuinness-NOAA!

WalterKolczynski-NOAA (approved):

PR #2868, opened 2024-08-27
Correct ocean `MOM.res_#` stage copy

### LINKED ISSUE #2864: recent updates break ocean cycling at 0.25 degree resolution
### What is wrong?

if `resdetocean=0.25` gdasstage_ic fails because of the logic in ocean.yaml.  This patch allows it to complete

```
diff --git a/parm/stage/ocean.yaml.j2 b/parm/stage/ocean.yaml.j2
index 75a472df..6dd9afcc 100644
--- a/parm/stage/ocean.yaml.j2
+++ b/parm/stage/ocean.yaml.j2
@@ -10,8 +10,8 @@ ocean:
         {% set imem = mem - first_mem %}
         {% set COMOUT_OCEAN_RESTART_PREV_MEM = COMOUT_OCEAN_RESTART_PREV_MEM_list[imem] %}
         - ["{{ ICSDIR }}/{{ COMOUT_OCEAN_RESTART_PREV_MEM | relpath(ROTDIR) }}/{{ m_prefix }}.[MOM.res.nc](http://mom.res.nc/)", "{{ COMOUT_OCEAN_RESTART_PREV_MEM }}"]
-        {% if OCNRES == "025" %}
-        {% for nn in range(1, 3) %}
+        {% if OCNRES == 25 or OCNRES == "025" %}
+        {% for nn in range(1, 4) %}
         - ["{{ ICSDIR }}/{{ COMOUT_OCEAN_RESTART_PREV_MEM | relpath(ROTDIR) }}/{{ m_prefix }}.MOM.res_{{ nn }}.nc", "{{ COMOUT_OCEAN_RESTART_PREV_MEM }}"]
         {% endfor %}
         {% endif %}
```

However, it seems like OCNRES should not be an integer (it should be converted to a string with a 0 prepended).  I believe there may be other places in the workflow where OCNRES is an integer but is treated as a string (and either used to build a filename, or is compared to a string '025')

the ocean restart files should have been staged to the COMROOT directory

All or N/A

### Steps to reproduce

run a cycling experiment with ocean DA enabled and resdetocean set to 0.25

### Additional information

N/A

no

KateFriedman-NOAA:
# Description

This PR corrects a bug in the staging job for ocean `MOM.res_#` IC files. The `OCNRES` value was coming in as an integer (e.g. `25`) but the `ocean.yaml.j2` file was checking for `"025"`. Correct to now set OCNRES to be three digits in staging script and also correct the for loop range to include third file.

Resolves #2864

# Type of change

# Change characteristics

# How has this been tested?

Repeated the tests that the users who reported this ran.

### DIFF
--- a/parm/stage/ocean.yaml.j2
+++ b/parm/stage/ocean.yaml.j2
@@ -11,7 +11,7 @@ ocean:
         {% set COMOUT_OCEAN_RESTART_PREV_MEM = COMOUT_OCEAN_RESTART_PREV_MEM_list[imem] %}
         - ["{{ ICSDIR }}/{{ COMOUT_OCEAN_RESTART_PREV_MEM | relpath(ROTDIR) }}/{{ m_prefix }}.MOM.res.nc", "{{ COMOUT_OCEAN_RESTART_PREV_MEM }}"]
         {% if OCNRES == "025" %}
-        {% for nn in range(1, 3) %}
+        {% for nn in range(1, 3 + 1) %}
         - ["{{ ICSDIR }}/{{ COMOUT_OCEAN_RESTART_PREV_MEM | relpath(ROTDIR) }}/{{ m_prefix }}.MOM.res_{{ nn }}.nc", "{{ COMOUT_OCEAN_RESTART_PREV_MEM }}"]
         {% endfor %}
         {% endif %}
--- a/parm/stage/ocean.yaml.j2
+++ b/parm/stage/ocean.yaml.j2
@@ -11,7 +11,7 @@ ocean:
         {% set COMOUT_OCEAN_RESTART_PREV_MEM = COMOUT_OCEAN_RESTART_PREV_MEM_list[imem] %}
         - ["{{ ICSDIR }}/{{ COMOUT_OCEAN_RESTART_PREV_MEM | relpath(ROTDIR) }}/{{ m_prefix }}.MOM.res.nc", "{{ COMOUT_OCEAN_RESTART_PREV_MEM }}"]
         {% if OCNRES == "025" %}
-        {% for nn in range(1, 3 + 1) %}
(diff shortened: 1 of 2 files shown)


jswhit2 (approved):

RuiyuSun (approved):

parm/stage/ocean.yaml.j2
```diff
@@ -11,7 +11,7 @@ ocean:
         {% set COMOUT_OCEAN_RESTART_PREV_MEM = COMOUT_OCEAN_RESTART_PREV_MEM_list[imem] %}
         - ["{{ ICSDIR }}/{{ COMOUT_OCEAN_RESTART_PREV_MEM | relpath(ROTDIR) }}/{{ m_prefix }}.MOM.res.nc", "{{ COMOUT_OCEAN_RESTART_PREV_MEM }}"]
         {% if OCNRES == "025" %}
-        {% for nn in range(1, 3) %}
+        {% for nn in range(1, 3 + 1) %}
```
WalterKolczynski-NOAA:
```suggestion
        {% for nn in range(1, 4) %}
```

scripts/exglobal_stage_ic.py
```diff
@@ -26,6 +26,9 @@ def main():
 
     stage_dict = AttrDict()
     for key in keys:
+        # Make sure OCNRES is three digits
+        if key == "OCNRES":
+            stage.task_config.OCNRES = f"{stage.task_config.OCNRES :03d}"
```
WalterKolczynski-NOAA, line 31: There's something unsettling about this. But fine for now.

WalterKolczynski-NOAA (changes requested):

DavidHuber-NOAA (approved):

WalterKolczynski-NOAA (approved):

PR #2953, opened 2024-09-24
Increased the No. of default levels to 40

AminIlia-NOAA:
# Description

 GFSv17 and GEFSv13 require 40 levels for 3D variables for [operational forecast](https://docs.google.com/document/d/1AwYbHArLJ6KvD0HW5bUbNOiFJaVAWgWEmnMQ2g076TA/edit ) . The default levels in case user doesn't define number of levels was 12 in the shell. I suggest to increase the levels to 40 as modified in this shell code (line238).   

  Resolves #1234

# Type of change

# Change characteristics

  - [ ] EMC verif-global 
  - [ ] GDAS 
  - [ ] GFS-utils 
  - [ ] GSI 
  - [ ] GSI-monitor 
  - [ ] GSI-utils 
  - [ ] UFS-utils 
  - [ ] UFS-weather-model 
  - [ ] wxflow 

# How has this been tested?

# Checklist

### DIFF
--- a/ush/oceanice_nc2grib2.sh
+++ b/ush/oceanice_nc2grib2.sh
@@ -235,7 +235,9 @@ component=${1:?"Need a valid component; options: ice|ocean"}
 grid=${2:-"0p25"} # Default to 0.25-degree grid
 current_cycle=${3:-"2013100100"} # Default to 2013100100
 avg_period=${4:-"0-6"} # Default to 6-hourly average
-ocean_levels=${5:-"5:15:25:35:45:55:65:75:85:95:105:115:125"} # Default to 12-levels
+ocean_levels=${5:-"5:15:25:35:45:55:65:75:85:95:105:115\
+:125:135:145:155:165:175:185:195:205:215:226:241:267:309\
+:374:467:594:757:960:1204:1490:1817:2184:2587:3024:3489:3977:4481"} # Default to 40-levels
 
 case "${grid}" in
   "0p25")


ush/oceanice_nc2grib2.sh
```diff
@@ -235,7 +235,9 @@ component=${1:?"Need a valid component; options: ice|ocean"}
 grid=${2:-"0p25"} # Default to 0.25-degree grid
 current_cycle=${3:-"2013100100"} # Default to 2013100100
 avg_period=${4:-"0-6"} # Default to 6-hourly average
-ocean_levels=${5:-"5:15:25:35:45:55:65:75:85:95:105:115:125"} # Default to 12-levels
+ocean_levels=${5:-"5:15:25:35:45:55:65:75:85:95:105:115\
```
aerorahul, line 238: We don't use this script.  The global-workflow calls https://github.com/NOAA-EMC/global-workflow/blob/develop/scripts/exglobal_oceanice_products.py and https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/python/pygfs/task/oceanice_products.py.  The levels are specified in the yaml https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/post/oceanice_products.yam

aerorahul: there is a copy of this in global-workflow https://github.com/NOAA-EMC/global-workflow/blob/ec634926f3b1ba1984b0735e17b76869659ceabe/ush/oceanice_nc2grib2.sh#L238

AminIlia-NOAA, line 238: Ok, Thanks for clarification!

aerorahul:
@AminIlia-NOAA 
Apologies.  The script is used. I thought I was looking at the gfs-utils.
One note, on the levels, for `mx500`, the number of levels are 25.  Changing to 40 will cause a failure when generating ocean products at `mx500` resolution.  I believe that is why the default number of levels were capped at 12.  
The operational system will use 40 levels, as prescribed by the yaml configuration [here](https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/post/oceanice_products.yaml#L34)

ush/oceanice_nc2grib2.sh
aerorahul: I think we should keep the default to under 25 (under `267m`) so that the tests will pass for `mx500`.
In ops, the levels are provided via the yaml, so we should make sure they are the correct ones.

AminIlia-NOAA: Ok, That's right.

PR #2960, opened 2024-09-25
SFS Runs at C96mx100 

NeilBarton-NOAA:
# Description
Adds the ability to run C96mx100 with SFS options.  Included in this PR is the ability to use the interpolated/coarse grained 1 degree MOM6 ICs.  Staging the ATM and OCN perturbation files without using the +03 replay ICs. 

# Type of change
- [X] New feature (adds functionality)

# Change characteristics

  - [ ] EMC verif-global 
  - [ ] GDAS 
  - [ ] GFS-utils 
  - [ ] GSI 
  - [ ] GSI-monitor 
  - [ ] GSI-utils 
  - [ ] UFS-utils 
  - [ ] UFS-weather-model 
  - [ ] wxflow 

# How has this been tested?
This has been tested in the gefs forecast-only configurations on hera and hercules. 

# Checklist
- [X] I have documented my code, including function, input, and output descriptions
- [X] This change is covered by an existing CI test or a new one has been added


parm/config/gefs/config.ufs
```diff
@@ -98,13 +98,13 @@ case "${fv3_res}" in
...
-        export layout_x_gfs=2
-        export layout_y_gfs=2
+        export layout_x_gfs=4
+        export layout_y_gfs=4
         export nthreads_fv3=1
         export nthreads_fv3_gfs=1
         export nthreads_ufs=1
         export nthreads_ufs_gfs=1
-        export xr_cnvcld=".false."  # Do not pass conv. clouds to Xu-Randall cloud fraction
+        export xr_cnvcld=".true."  # Pass conv. clouds to Xu-Randall cloud fraction
```
lisa-bengtsson, line 107: @NeilBarton-NOAA we could perhaps just remove this xr_cnvcld variable from the workflow to avoid any confusion. It is default .true. in the code, and when we wanted it to be false for C96 only it made sense to keep it here. But now when we moved it back to true, and instead changed the scaling in the code - it is probably cleaner to just remove.

lisa-bengtsson (approved):

aerorahul:
@pjpegion has proposed a `system=sfs` in addition to `gfs|gefs` variants.  
@WalterKolczynski-NOAA we should ensure that it is brought in sooner rather than later, before conditionals for `SFS` get if-ed into the `RUN=gefs` code-blocks.
@JacobCarley-NOAA FYI

ci/cases/sfs/C96mx100_S2S.yaml
```diff
@@ -0,0 +1,20 @@
+experiment:
+  system: gefs
```
aerorahul, line 2: This is now a SFS test using `gefs` system.

sorc/build_all.sh
```diff
@@ -37,6 +37,8 @@ Usage: ${BASH_SOURCE[0]} [-a UFS_app][-c build_config][-d][-f][-h][-j n][-v][-w]
     Execute all build scripts with -v option to turn on verbose where supported
   -w:
     Use structured wave grid
+  -y:
+    Use hystrostatic version of FV3
```
aerorahul:
```suggestion
    Use hydrostatic version of FV3
```

sorc/build_all.sh
```diff
@@ -129,7 +133,7 @@ declare -A build_opts
 big_jobs=0
 build_jobs["ufs"]=8
 big_jobs=$((big_jobs+1))
-build_opts["ufs"]="${_wave_opt} ${_verbose_opt} ${_build_ufs_opt} ${_build_debug}"
+build_opts["ufs"]="${_wave_opt} ${_hydro_opt} ${_verbose_opt} ${_build_ufs_opt} ${_build_debug}"
```
aerorahul, line 136: This requires `build_ufs.sh` to be updated as well to make use of the `-y`.

aerorahul (commented):
How is this actually tested?  It appears that the model has to be built with a hydrostatic option `-y` but that is missing from `build_ufs.sh`.
Also, this requires the CI system to build the ufs-weather-model in 3 configurations; unstructured waves for GFS, structured waves for GEFS (`-w`) and hydrostatic dynamical code for SFS (`-y`).
I think we need to re-consider the CI system where we are cloning 2 copies of global-workflow one each for GFS and GEFS, and now this third one for SFS.
Instead, we could refactor the current CI system to build 3 executables with a suffix of `.gfs.x`, `.gefs.x`, `.sfs.x`  or `gfs_model.x`, `gefs_model.x`, `sfs_model.x` for the variants of the model options and still use the same `HOMEgfs`.

This will require some prep-work in the global-workflow.

aerorahul:
@WalterKolczynski-NOAA @DavidHuber-NOAA  Please see [this](https://github.com/NOAA-EMC/global-workflow/pull/2960#pullrequestreview-2373552502) comment and provide your thoughts.

WalterKolczynski-NOAA:
> How is this actually tested? It appears that the model has to be built with a hydrostatic option `-y` but that is missing from `build_ufs.sh`. Also, this requires the CI system to build the ufs-weather-model in 3 configurations; unstructured waves for GFS, structured waves for GEFS (`-w`) and hydrostatic dynamical code for SFS (`-y`). I think we need to re-consider the CI system where we are cloning 2 copies of global-workflow one each for GFS and GEFS, and now this third one for SFS. Instead, we could refactor the current CI system to build 3 executables with a suffix of `.gfs.x`, `.gefs.x`, `.sfs.x` or `gfs_model.x`, `gefs_model.x`, `sfs_model.x` for the variants of the model options and still use the same `HOMEgfs`.
> 
> This will require some prep-work in the global-workflow.

I agree. We've already talked about doing this for the two existing variations, though it hasn't been a priority as yet. I'd like to name the executables by options rather than system, since there is nothing inherent preventing GEFS from running with an unstructured wave grid, for example. But that also may get cumbersome if we accumulate too many different build options.

I can roll this into my existing compute-build effort (indeed was already considering doing so).

sorc/build_all.sh
NeilBarton-NOAA: @aerorahul I didn't catch an error with the merge/commit. The updated build_all.sh and other files are now uploaded

ci/cases/sfs/C96mx100_S2S.yaml
NeilBarton-NOAA: The SFS_POST=YES is currently a proxy to system=sfs and will need to be edited.

NeilBarton-NOAA:
The GEFS and GFS coupled forecasts CIs are successful with the latest commits. The pynorm error is unclear to me. @WalterKolczynski-NOAA any suggestions on the pynorm error?

WalterKolczynski-NOAA:
> The GEFS and GFS coupled forecasts CIs are successful with the latest commits. The pynorm error is unclear to me. @WalterKolczynski-NOAA any suggestions on the pynorm error?

If you click on 'details' it will take you to the pynorms log. In this case, it is just one thing:
```
checking ./scripts/exglobal_stage_ic.py
./scripts/exglobal_stage_ic.py:26:40: E231 missing whitespace after ','
```

ci/cases/sfs/C96mx100_S2S.yaml
WalterKolczynski-NOAA, line 1: This test should be disabled on wcoss2 like the other gefs tests until we change the build system to build different executables.

parm/config/gefs/config.efcs
```diff
@@ -39,7 +39,7 @@ if [[ "${RUN}" == "enkfgfs" ]]; then
 fi
 
 # Stochastic physics parameters (only for ensemble forecasts)
-export DO_SKEB="YES"
+export DO_SKEB=${DO_SKEB:-"YES"} # SKEB turned off for C96
```
WalterKolczynski-NOAA: I don't think this is going to work, because `DO_SKEB` is set in `config.fcst` first, so the default here will never be used.

parm/config/gefs/config.fcst
```diff
@@ -43,6 +43,7 @@ export FHOUT=${FHOUT_GFS}
 export FHOUT_HF=${FHOUT_HF_GFS}
 export FHOUT_OCN=${FHOUT_OCN_GFS}
 export FHOUT_ICE=${FHOUT_ICE_GFS}
+export FHZER=@FHZER@
```
WalterKolczynski-NOAA, line 47: This variable is not used.

parm/config/gefs/yaml/defaults.yaml
```diff
@@ -15,7 +15,20 @@ base:
...
   REPLAY_ICS: "NO"
   USE_OCN_PERTURB_FILES: "false"
+  FHOUT_GFS: 6
   FHOUT_OCN_GFS: 6
   FHOUT_ICE_GFS: 6
   HPSSARCH: "NO"
   LOCALARCH: "NO"
+  SFS_POST: "NO"
+fcst:
+  reforecast: "YES"
```
WalterKolczynski-NOAA: The default should be "NO" and turned on for the tests that need it.

parm/config/gefs/yaml/defaults.yaml
```diff
@@ -15,7 +15,20 @@ base:
...
+  FHZER: 6
+  TYPE: "nh"
+  MONO: "non-mono"
+stage_ic:
+  USE_OCN_ENS_PERTURB_FILES: "YES"
+  USE_ATM_ENS_PERTURB_FILES: "YES"
```
WalterKolczynski-NOAA: Do we really want to default to using these?

parm/stage/master_gefs.yaml.j2
```diff
@@ -114,7 +114,7 @@
 {% endfilter %}
 {% endif %}
 
-{% if REPLAY_ICS %}
+{% if REPLAY_ICS or USE_ATM_ENS_PERTURB_FILES %}
```
WalterKolczynski-NOAA: Should replay just not turn on this setting?
```suggestion
{% if USE_ATM_ENS_PERTURB_FILES %}
```

parm/stage/master_gefs.yaml.j2
```diff
@@ -135,7 +135,7 @@
 {% include "ocean_rerun.yaml.j2" %}
 {% endfilter %}
 {% endif %}
-{% if REPLAY_ICS %}
+{% if REPLAY_ICS or USE_OCN_ENS_PERTURB_FILES %}
```
WalterKolczynski-NOAA:
```suggestion
{% if USE_OCN_ENS_PERTURB_FILES %}
```

sorc/build_ufs.sh
```diff
@@ -7,15 +7,17 @@ cwd=$(pwd)
...
-while getopts ":da:fj:vw" option; do
+while getopts ":da:fj:vwy" option; do
   case "${option}" in
     d) BUILD_TYPE="Debug";;
     a) APP="${OPTARG}";;
     f) FASTER="ON";;
     j) BUILD_JOBS="${OPTARG}";;
     v) export BUILD_VERBOSE="YES";;
     w) PDLIB="OFF";;
+    y) HYDRO="ON";;
```
WalterKolczynski-NOAA, line 20: Is there a need to create a separate executable, or can we run hydrostatic mode with the non-hydrostatic executable?

parm/config/gfs/config.ocn
```diff
@@ -24,6 +24,11 @@ else
...
 
-
-
+MOM6_INTERP_ICS=@MOM6_INTERP_ICS@
+if [[ "${MOM6_INTERP_ICS}" == "YES" ]]; then
+    export MOM6_RESTART_SETTING='n'
+    export MOM6_WARMSTART_FILE="MOM.res.nc"
+    export MOM6_INIT_FROM_Z='False'
+    export MOM6_INIT_UV='file'
+fi
```
WalterKolczynski-NOAA, line 33: Don't we need a similar block in `gefs/config.ocn`?

WalterKolczynski-NOAA: Or maybe these should be set elsewhere in the run scripts based on `$MOM6_INTERP_ICS`

ush/forecast_predet.sh
```diff
@@ -332,19 +332,27 @@ FV3_predet(){
     if [[ "${TYPE}" == "nh" ]]; then  # monotonic and non-hydrostatic
       hord_mt=${hord_mt_nh_mono:-"10"}
       hord_xx=${hord_xx_nh_mono:-"10"}
+      hord_dp=-${hord_xx_nh_nonmono:-"5"}
```
WalterKolczynski-NOAA: Changes you have here would change the behavior of `hord_dp` in NH cases from what is now.
```suggestion
      hord_dp=-${hord_xx_nh_nonmono:-"-10"}
```

ush/forecast_predet.sh
```diff
@@ -332,19 +332,27 @@ FV3_predet(){
...
     else  # monotonic and hydrostatic
       hord_mt=${hord_mt_hydro_mono:-"10"}
       hord_xx=${hord_xx_hydro_mono:-"10"}
+      hord_dp=-${hord_xx_nh_nonmono:-"10"}
```
WalterKolczynski-NOAA:
```suggestion
      hord_dp=-${hord_xx_nh_nonmono:-"-10"}
```

ush/forecast_predet.sh
```diff
@@ -332,19 +332,27 @@ FV3_predet(){
...
     fi
   else  # non-monotonic options
     d_con=${d_con_nonmono:-"1."}
     do_vort_damp=".true."
     if [[ "${TYPE}" == "nh" ]]; then  # non-monotonic and non-hydrostatic
       hord_mt=${hord_mt_nh_nonmono:-"5"}
       hord_xx=${hord_xx_nh_nonmono:-"5"}
+      hord_dp=${hord_xx_hydro_mono:-"10"}
```
WalterKolczynski-NOAA:
```suggestion
      hord_dp=${hord_xx_hydro_mono:-"-5"}
```

WalterKolczynski-NOAA (changes requested):

sorc/build_ufs.sh
aerorahul: @junwang-noaa might know.

junwang-noaa: Yes, you do need to compile with model with "HYDRO" turned on for the SFS case.

WalterKolczynski-NOAA: Okay, thanks @junwang-noaa

parm/config/gefs/config.efcs
NeilBarton-NOAA: DO_SKEB="NO" for C96 in config.ufs which is being propagated to the namelist. See log on hercules at /home/nbarton/GW/EXPDIR/SFS_C96mx100_S2S/LOGS_COMROOT/1994050100/fcst_mem001_seg0.log

parm/config/gfs/config.ocn
NeilBarton-NOAA: gefs/config.ocn is currently a soft link from gfs/config.ocn. i.e., each is set. If preferred to be set someplace else, I'm happy to edit

parm/config/gefs/yaml/defaults.yaml
NeilBarton-NOAA: I changed the default to NO

parm/config/gefs/yaml/defaults.yaml
NeilBarton-NOAA: I changed the default to NO

parm/config/gefs/config.fcst
NeilBarton-NOAA: ? The variable is being propagated to the input.nml file in my test runs. see hercules at /home/nbarton/GW/EXPDIR/SFS_C96mx100_S2S/LOGS_COMROOT/1994050100/fcst_mem001_seg0.log

parm/config/gfs/config.ocn
WalterKolczynski-NOAA: Okay, I don't always remember which gefs files are separate and which are just links.

parm/config/gefs/config.fcst
WalterKolczynski-NOAA: Okay, I was looking it in the changes since it was new here, but it is already set up to be used in input.nml.

parm/config/gefs/config.efcs
WalterKolczynski-NOAA: The point was this "YES" will never be used because it is _always_ set upstream.

WalterKolczynski-NOAA:
@NeilBarton-NOAA Still a couple minor suggestions from me & Rahul that haven't been addressed yet.

parm/config/gefs/config.efcs
NeilBarton-NOAA: I see, I removed DO_SKEB in config.ufs and added an if statement in config.efcs

NeilBarton-NOAA:
@WalterKolczynski-NOAA I believe I addressed everything. Let me know if not.

WalterKolczynski-NOAA:
> @WalterKolczynski-NOAA I believe I addressed everything. Let me know if not.

The new test hasn't been disabled on wcoss. Actually, I think it needs to be disabled everywhere to start with, because it needs a would need a pipeline to build the hydrostatic model executable separate from gfs and gefs.

Edit: Unless it is okay to run with the nh GEFS executable for now, in which case just disable on wcoss.

sorc/build_all.sh
WalterKolczynski-NOAA: @NeilBarton-NOAA This typo hasn't been fixed.

NeilBarton-NOAA:
I fixed the typo. Let's discuss the testing after the GEFS g-w meeting today.

parm/config/gefs/config.efcs
```diff
@@ -39,7 +39,11 @@ if [[ "${RUN}" == "enkfgfs" ]]; then
 fi
 
 # Stochastic physics parameters (only for ensemble forecasts)
-export DO_SKEB="YES"
+if [[ "${CASE}" == "C96" ]] ; then
+    export DO_SKEB="NO" # SKEB turned off for C96
+else
+    export DO_SKEB="YES" # SKEB turned off for C96
+fi
```
KateFriedman-NOAA: Only turned off for C96? What about the other resolutions, SKEB is turned on? Also, update comment on line 45 to say SKEB is turned **_on_** for C96.

parm/config/gefs/config.resources
```diff
@@ -188,7 +188,7 @@ case ${step} in
 
     case "${CASE}" in
       "C48" | "C96" | "C192")
-        declare -x "walltime"="03:00:00"
+        declare -x "walltime"="08:00:00"
```
KateFriedman-NOAA: Is the longer walltime due to forecast length? Would GEFS also need 8hrs?

KateFriedman-NOAA (changes requested):

WalterKolczynski-NOAA (dismissed):
Conditionally approved pending CI testing. Know the "SFS" case won't be tested.

parm/config/gefs/config.efcs
NeilBarton-NOAA: updated

parm/config/gefs/config.resources
NeilBarton-NOAA: GEFS does not run these resolutions. Forecast time is for long SFS forecast, but I rolled it back to 03 if needed for CI testing. An easy method to change this would be nice.

WalterKolczynski-NOAA:
@NeilBarton-NOAA The gdas conflict needs to be resolved. Use whichever hash of the two is more recent.

parm/config/gefs/config.resources
KateFriedman-NOAA: Gotcha, thanks for explaining!

KateFriedman-NOAA (dismissed):
Approve pending successful completion of CI tests. Thanks @NeilBarton-NOAA !

NeilBarton-NOAA:
@WalterKolczynski-NOAA gdas updated and develop merged into the branch. I believe it's ready for CI testing.

NeilBarton-NOAA:
The fix for the gefs CI test is uploaded.

WalterKolczynski-NOAA:
Why are there so many submodule updates in this PR? Were those needed for this?

NeilBarton-NOAA:
> Why are there so many submodule updates in this PR? Were those needed for this?

@WalterKolczynski-NOAA There shouldn't be. I assume I missed something in a merge. I'll put each submodule to the develop branch and upload

parm/config/gefs/config.ufs
WalterKolczynski-NOAA: If this is a general change and not just to shoehorn C96 SFS in until the separate run is available, it should be made for C48 as well.

NeilBarton-NOAA: added for C48

NeilBarton-NOAA:
@WalterKolczynski-NOAA modules are matching to develop and performed a quick DA test. This is ready for CI testing in my opinion

NeilBarton-NOAA:
I just built this branch on hera.

WalterKolczynski-NOAA:
The GEFS replay test is failing because it is looking for a non-existent increment file.

NeilBarton-NOAA:
@WalterKolczynski-NOAA fix for CI test added

WalterKolczynski-NOAA (approved):

Issue #2951, opened 2024-09-24
Accelerated RFC to place PlanetiQ RO data in monitor mode 


RussTreadon-NOAA:
### Description

PlanetiQ data has added noise.  This issue issue is opened in case we decide to place this data in monitor mode until we more fully examine it.

### Target version

v16.3.19

### Expected workflow changes

The only file changed is `global_convinfo.txt`.  This requires a new GSI tag and a new GFS tag.

### Tasks

- [x] Create release branch
- [x] Make workflow changes for upgrade in release branch (add additional checklist items as needed)
- [x] Create release notes
- [x] Cut hand-off tag for CDF - [EMC-v16.3.19](https://github.com/NOAA-EMC/global-workflow/releases/tag/EMC-v16.3.19)
- [x] Submit CDF to NCO
- [x] Implementation into operations complete
- [x] Merge release branch into operational branch
- [x] Cut version tag from operational branch
- [x] Release new version tag
- [x] Announce to users

KateFriedman-NOAA:
Shuffled other current release branches around and recut [`release/gfs.v16.3.19`](https://github.com/NOAA-EMC/global-workflow/tree/release/gfs.v16.3.19) from `dev/gfs.v16` for this ARFC.

KateFriedman-NOAA:
@RussTreadon-NOAA sounds like this may or may not happen? Let me know the status when you do.

Also, let me know the new GSI tag name when/if created. I will prep release notes and be ready to update the GSI tag in the release branch when/if available. I'll ask you to review a PR into the release branch from my fork once ready.

FYI, we will be handing off a new GFS version that is a companion to the `obsproc/v1.3` update that is being handed off in early October. Hopefully this issue can be decided on before we need to hand-off for that.

KateFriedman-NOAA:
Additional update from @RussTreadon-NOAA in #2591 (https://github.com/NOAA-EMC/global-workflow/issues/2591#issuecomment-2371619577):

```
DA team members are meeting this afternoon to make a final decision as to what course of action to take. 
Submitting an ARFC to place PlanetiQ data in monitor mode is one option. Another option is to let the 
PlanetiQ data flow into and be assimilated by the GFS. This later option requires no EIB action.

This issue is opened as a "heads up" to EIB in case the DA team goes with the first option.
```

Thanks for the update @RussTreadon-NOAA ! Will see what the DA team decides and go from there.

RussTreadon-NOAA:
[...]

PR #2966, opened 2024-09-30
Increase enkfgdas_update walltime to 35mins

KateFriedman-NOAA:
# Description

NCO increased the walltime for the `enkfgdas_update` job by 5mins (from 30mins to 35mins) ahead of implementation today.

This PR increases the walltime similarly in the relevant ecf script and also updates the two `config.resource` configs walltime settings for the eupd job.

Refs #2951

# Type of change

Production update


RussTreadon-NOAA (approved):
The increase in the enkfgdas_update wallclock from 30 to 35 minutes is correct.  This change was implemented as part of gfs.v16.3.19.

Approve.

PR #3018, opened 2024-10-18
GitHub Actions Pipeline Updates for Self-Hosted Runners on PW

TerrenceMcGuinness-NOAA:
# Description

Updates to the GItHub Self-hosted Runner Pipelines for Parallel Works:
- User now inputs PR number when dispatching the CI workflow
- User can also select which operating system to test on during the transition to Rocky 8

**NOTE:** When this pipeline is dispatched it should be run with [PR 3023](https://github.com/NOAA-EMC/global-workflow/pull/3023) to demonstrate that it can pass an end-to-end case.

# Type of change
- [ ] Bug fix (fixes something broken)
- [x] New feature (adds functionality)
- [ ] Maintenance (code refactor, clean-up, new CI test, etc.)

# How has this been tested?

Full CI tests using this pipeline were built and ran using the  [PR 3023](https://github.com/NOAA-EMC/global-workflow/pull/3023) branch and the results can be found here for [Centos 7 ](https://github.com/TerrenceMcGuinness-NOAA/global-workflow/actions/runs/11447462304) and [Rocky 8](https://github.com/TerrenceMcGuinness-NOAA/global-workflow/actions/runs/11446734722/job/31852433846)

PR #3023, opened 2024-10-21
Build aws updates prototype

TerrenceMcGuinness-NOAA:
# Description

This "_Prototype_" PR is used to demonstrate changes that may be needed in order to build and run on AWS for Centos 7 and Rocky 8.

It was used to demonstrate and end-to-end build and run of the C48_ATM free forecasts case for using GitHub Self-hosted Runners on Parallel works (see [PR #3018](https://github.com/NOAA-EMC/global-workflow/pull/3018))

**Disclamer**

This was done basically by finding working version numbers agains the current installs of Spack Stack pre installed in `/contrib/spack-stack `using [`$HOMEgfs/version/spack.ver`](https://github.com/TerrenceMcGuinness-NOAA/global-workflow/blob/build_aws_updates_prototype/versions/spack.ver) with Spack Stack v1.6.0

No claims are made about the validity of these choice and where used essential to create a process in order to build and run global-workflow on the Parllel Works platform and to test the CI development.  It should only be used as guid and proof of concept.

# How has this been tested?

The build and runs using the brach can be found here for [Centos 7 ](https://github.com/TerrenceMcGuinness-NOAA/global-workflow/actions/runs/11447462304) and [Rocky 8](https://github.com/TerrenceMcGuinness-NOAA/global-workflow/actions/runs/11446734722/job/31852433846)

# Change characteristics


sorc/ufs_noaacloud.intel.diff
weihuang-jedi, line 1: This is a diff out put, do we really need it?

PR #3036, opened 2024-10-25
Add ability to add tag to pslots with generate_workflows

WalterKolczynski-NOAA:
# Description
`generate_workflows.sh` would only use the case name as the pslot, leading to conflicts if you try to run two different sets at the same time. Now there is a new option that allows a 'tag' to the case name when determining the pslots:

```
generate_workflows.sh -t <tag>

tag: string to be added to end of case name
```

For example, this:

```
generate_workflows.sh -t test
```

Will result in pslots of `C48_ATM_test`, `C48_S2SW_test`, etc.

This is similar to how the CI system appends the hash to the pslots, but allows flexibility since this is a command-line tool.

# Type of change

# Change characteristics

# How has this been tested?
- Ran `generate_workflows.sh` both with and without the new `-t` option.

# Checklist

### DIFF
--- a/workflow/generate_workflows.sh
+++ b/workflow/generate_workflows.sh
@@ -71,6 +71,9 @@ function _usage() {
        If this option is not specified, then the existing email address in
        the crontab will be preserved.
 
+    -t Add a 'tag' to the end of the case names in the pslots to distinguish
+       pslots between multiple sets of tests.
+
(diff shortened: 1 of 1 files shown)


DavidHuber-NOAA (approved):
Looks good, thanks for this feature Walter!

Issue #2591, opened 2024-05-10
GFSv16.3.21 - Remove WAFS from GFS


KateFriedman-NOAA:
### Description

Remove the WAFS (formerly EMC_gfs_wafs) from the GFS, it will now be a standalone system in operations.

WAFS issue: https://github.com/NOAA-EMC/WAFS/issues/23

### Target version

v16.3.21 (previously v16.3.22 (previously v16.3.23))

### Expected workflow changes

Remove WAFS pieces:

- jobs in ecf def
- ecf scripts
- wafs job, ex-scripts, and ush scripts
- WAFS jobs from rocoto mesh
- configs
- set `WAFSF="NO"`
- discontinue GTG checkout and `-o` flag for checkout (?)

KateFriedman-NOAA:
Timeline update from @aerorahul :
```
WAFS has a code delivery in Q1FY25 with an implementation in Q2FY25.
The GFS will be delivered and implemented at the same time removing WAFS from the GFS. 
```

KateFriedman-NOAA:
~~This will be v16.4.0~~

KateFriedman-NOAA:
Created branch: https://github.com/NOAA-EMC/global-workflow/tree/release/gfs.v16.3.18

KateFriedman-NOAA:
Just renamed the release branch to `release/gfs.v16.3.21` to match the new version in ops. The ARFC for `GFSv16.3.19` took the version number and the `obsproc/v1.3` update will take `GFSv16.3.20`. Will update release notes and internal version shortly, need to prep release branch for ARFC `GFSv16.3.19` first.

RussTreadon-NOAA:
Thank you @KateFriedman-NOAA for your quick action.  

DA team members are meeting this afternoon to make a final decision as to what course of action to take.  Submitting an ARFC to place PlanetiQ data in monitor mode is one option.  Another option is to let the PlanetiQ data flow into and be assimilated by the GFS.  This later option requires no EIB action.  

This issue is opened as a "heads up" to EIB in case the DA team goes with the first option.

KateFriedman-NOAA:
Synced the `release/gfs.v16.3.21` branch with `dev/gfs.v16` after the GFSv16.3.19 branch was merged into `dev/gfs.v16`. Also updated the release branch for the new version `v16.3.21`.

KateFriedman-NOAA:
[...]

PR #3047, opened 2024-10-30
Update enkfgdas_update job resource/env settings

KateFriedman-NOAA:
# Description

Updates for the `enkfgdas_update` job come from testing by NCO/GDIT and EMC. They produce a more stable runtime and bitwise identical outputs to ops.

Add `export OMP_PROC_BIND=true` and update `APRUN_ENKF` to use `--cpu-bind core` instead of `--cpu-bind depth` in WCOSS2.env. Also, adjust npe (280) and thread (16) values.

NCO requests that we include these updates in the GFSv16.3.21 package.

Refs #2591

# Type of change

Resource and job environment improvements

### DIFF
--- a/docs/Release_Notes.md
+++ b/docs/Release_Notes.md
@@ -155,7 +155,9 @@ and are being retired from the WAFS package:
 ENVIRONMENT AND RESOURCE CHANGES
 --------------------------------
 
-* No changes from GFS v16.3.20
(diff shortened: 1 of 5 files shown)


RussTreadon-NOAA (approved):
Looks good to me.

Approve.

DavidHuber-NOAA (approved):
Looks good!

PR #3105, opened 2024-11-15
Archive the experiment directory along with git status/diff output

### LINKED ISSUE #2994: Ability to save experiment directory (EXPDIR)

A switch or similar setting to allow users to save their `EXPDIR` (e.g. to HPSS).

That the contents of the `EXPDIR` are saved/archived.

### Acceptance Criteria

- [ ] `EXPDIR` saved/archived to HPSS

_No response_

DavidHuber-NOAA:
# Description
This adds the capability to archive the experiment directory.  Additionally, this adds options to run `git status` and `git diff` on the `HOMEgfs` global workflow (but not the submodules) and store that information within the experiment directory's archive.  These options are specified in `config.base` with the following defaults:

```bash
export ARCH_EXPDIR='YES'     # Archive the EXPDIR configs, XML, and database
export ARCH_EXPDIR_FREQ=0    # How often to archive the EXPDIR in hours or 0 for first and last cycle only
export ARCH_HASHES='YES'     # Archive the hashes of the GW and submodules and 'git status' for each; requires ARCH_EXPDIR
export ARCH_DIFFS='NO'       # Archive the output of 'git diff' for the GW; requires ARCH_EXPDIR
```

Resolves #2994
# Type of change

# Change characteristics

  - [x] wxflow https://github.com/NOAA-EMC/wxflow/pull/45

# How has this been tested?
- [x] Local archiving on Hercules for a C48_ATM case
- [x] Cycled testing on Hercules with `ARCH_DIFFS=YES` and `ARCH_EXPDIR_FREQ=6,12`
- [x] Testing with `ARCH_EXPDIR=NO` or `ARCH_HASHES=NO`

# Checklist


DavidHuber-NOAA:
FYI @AndrewEichmann-NOAA

DavidHuber-NOAA:
Tests were performed on Hercules and the experiment directory tarballs for both cycled and forecast-only modes were found to have been created as expected.  Opening for review.

parm/archive/expdir.yaml.j2
```diff
@@ -0,0 +1,26 @@
...
+    # Copy the experiment files from the EXPDIR into the ROTDIR for archiving
+    {% set copy_expdir = "expdir." ~ cycle_YMDH %}
+    FileHandler:
+        mkdir:
+            - "{{ ROTDIR }}/{{ copy_expdir }}"
+        copy:
+            {% for config in glob(EXPDIR ~ "/config.*") %}
+            - [ "{{ config }}", "{{ ROTDIR }}/{{ copy_expdir }}/." ]
+            {% endfor %}
+            - [ "{{ EXPDIR }}/{{ PSLOT }}.db", "{{ ROTDIR }}/{{ copy_expdir }}/." ]
```
WalterKolczynski-NOAA: Would recommend *not* saving the database.
```suggestion
```

parm/archive/expdir.yaml.j2
```diff
@@ -0,0 +1,26 @@
...
+            - [ "{{ EXPDIR }}/{{ PSLOT }}.xml", "{{ ROTDIR }}/{{ copy_expdir }}/." ]
+            {% if ARCH_HASHES or ARCH_DIFFS %}
+            - [ "{{ EXPDIR }}/git_info.log", "{{ ROTDIR }}/{{ copy_expdir }}/." ]
+            {% endif %}
+    target: "{{ ATARDIR }}/{{ cycle_YMDH }}/expdir.tar"
+    required:
+        - "{{ copy_expdir }}/config.*"
+        - "{{ copy_expdir }}/{{ PSLOT }}.db"
```
WalterKolczynski-NOAA:
```suggestion
```

parm/archive/master_gdas.yaml.j2
```diff
@@ -97,3 +97,10 @@ datasets:
 
     # End of restart checking
 {% endif %}
+
+# Archive the EXPDIR if requested
+{% if archive_expdir %}
+{% filter indent(width=4) %}
+{% include "expdir.yaml.j2" %}
+{% endfilter %}
+{% endif %}
```
WalterKolczynski-NOAA, line 106: Should only need this in one of gdas or gfs, but there is a bit of a coordination problem since either can be run without the other.

ush/python/pygfs/task/archive.py
```diff
@@ -417,3 +438,151 @@ def replace_string_from_to_file(filename_in, filename_out, search_str, replace_s
         replace_string_from_to_file(in_track_p_file, out_track_p_file, "AVNO", pslot4)
 
         return
+
+    @staticmethod
```
WalterKolczynski-NOAA: Why is this static? Especially as a "hidden" method.

ush/python/pygfs/task/archive.py
```diff
@@ -417,3 +438,151 @@ def replace_string_from_to_file(filename_in, filename_out, search_str, replace_s
...
+        # Otherwise, the frequency is in hours
+        elif mode == "forecast-only" and (sdate - current_cycle).total_seconds() % freq != 0:
+            return False
+        elif mode == "cycled" and (sdate + assim_freq - current_cycle).total_seconds() % freq != 0:
+            return False
+
+        # Looks like we are archiving the EXPDIR
+        return True
+
+    @staticmethod
```
WalterKolczynski-NOAA: Same. Why static?

ush/python/pygfs/task/archive.py
```diff
@@ -417,3 +438,151 @@ def replace_string_from_to_file(filename_in, filename_out, search_str, replace_s
...
+            elif mode == "cycled" and (current_cycle != sdate + assim_freq or current_cycle != edate):
+                return False
+        # Otherwise, the frequency is in hours
+        elif mode == "forecast-only" and (sdate - current_cycle).total_seconds() % freq != 0:
+            return False
+        elif mode == "cycled" and (sdate + assim_freq - current_cycle).total_seconds() % freq != 0:
+            return False
+
+        # Looks like we are archiving the EXPDIR
+        return True
```
WalterKolczynski-NOAA: Reversing the polarity of the ifs, because I think we always want to save the first and last, even if the end time is not a multiple of the save frequency.
```suggestion
        first_full = sdate
        if mode in ["cycled"]:
            first_full += assim_freq
        if current_cycle in [first_full, edate]:
            # Always save the first and last
            return True
        elif (current_cycle - first_full).total_seconds() % freq == 0:
            # Otherwise, the frequency is in hours
            return True
        else
            return False
```

ush/python/pygfs/task/archive.py
```diff
@@ -417,3 +438,151 @@ def replace_string_from_to_file(filename_in, filename_out, search_str, replace_s
...
+            # Are we running git to get hashes?
+            if arch_hashes:
+                output += "Global workflow hash:\n"
+
+                try:
+                    output += git("rev-parse", "HEAD", output=str)
+                    output += "\nSubmodule hashes:\n"
+                    output += git("submodule", "status", output=str)
+                except ProcessError:
+                    raise OSError("FATAL ERROR Failed to run git")
```
WalterKolczynski-NOAA: If we are going to catch and then raise new exceptions, we should use `from` to keep the info from the original (repeat for subsequent instances):
```suggestion
                except ProcessError as pe:
                    raise OSError("FATAL ERROR Failed to run git") from pe
```

WalterKolczynski-NOAA (changes requested):

parm/archive/master_gdas.yaml.j2
DavidHuber-NOAA: Agreed.  I added a method to archive.py to determine which `RUN` to archive this in and set the `archive_expdir` boolean accordingly.

ush/python/pygfs/task/archive.py
DavidHuber-NOAA: Force of habit...
```suggestion
```

ush/python/pygfs/task/archive.py
DavidHuber-NOAA:
```suggestion
```

ush/python/pygfs/task/archive.py
```diff
@@ -417,3 +438,151 @@ def replace_string_from_to_file(filename_in, filename_out, search_str, replace_s
...
+                    output += git("diff", "--submodule=diff", output=str)
+                except ProcessError:
+                    # The version of git may be too old.  See if we can run just a surface diff.
+                    try:
+                        output += git("diff", output=str)
+                        print("WARNING git was unable to do a recursive diff.\n"
+                              "Only a top level diff was performed.\n"
+                              "Note that the git version must be >= 2.14 for this feature.")
+                    except ProcessError:
+                        raise OSError("FATAL ERROR Failed to run 'git diff'")
```
DavidHuber-NOAA:
```suggestion
                    except ProcessError as pe:
                        raise OSError("FATAL ERROR Failed to run 'git diff'") from pe
```

ush/python/pygfs/task/archive.py
```diff
@@ -109,6 +113,16 @@ def configure(self, arch_dict: Dict[str, Any]) -> (Dict[str, Any], List[Dict[str
...
 
+        # Determine if we are archiving the EXPDIR this cycle
+        if arch_dict.ARCH_EXPDIR:
+            self.archive_expdir = Archive._archive_expdir(arch_dict)
+            arch_dict.archive_expdir = self.archive_expdir
+
+            if self.archive_expdir:
+                # If requested, get workflow hashes/statuses/diffs for EXPDIR archiving
+                if arch_dict.ARCH_HASHES or arch_dict.ARCH_DIFFS:
+                    Archive._pop_git_info(arch_dict)
```
DavidHuber-NOAA:
```suggestion
            self.archive_expdir = self._archive_expdir(arch_dict)
            arch_dict.archive_expdir = self.archive_expdir

            if self.archive_expdir:
                # If requested, get workflow hashes/statuses/diffs for EXPDIR archiving
                if arch_dict.ARCH_HASHES or arch_dict.ARCH_DIFFS:
                    self._pop_git_info(arch_dict)
```

ush/python/pygfs/task/archive.py
DavidHuber-NOAA: Good catch.  Thanks.

ush/python/pygfs/task/archive.py
```diff
@@ -417,3 +438,151 @@ def replace_string_from_to_file(filename_in, filename_out, search_str, replace_s
...
+                    except ProcessError:
+                        raise OSError("FATAL ERROR Failed to run 'git diff'")
+
+        # Write out to the log file
+        try:
+            with open(os.path.join(expdir, git_filename), 'w') as output_file:
+                output_file.write(output)
+        except OSError:
+            fname = os.path.join(expdir, git_filename)
+            raise OSError(f"FATAL ERROR Unable to write git output to '{fname}'")
```
DavidHuber-NOAA:
```suggestion
        except OSError as ose:
            fname = os.path.join(expdir, git_filename)
            raise OSError(f"FATAL ERROR Unable to write git output to '{fname}'") from ose
```

ush/python/pygfs/task/archive.py
```diff
@@ -417,3 +438,147 @@ def replace_string_from_to_file(filename_in, filename_out, search_str, replace_s
...
+        first_full = sdate
+        if mode in ["cycled"]:
+            first_full += assim_freq
+        if current_cycle in [first_full, edate]:
+            # Always save the first and last
+            return True
+        elif (current_cycle - first_full).total_seconds() % freq == 0:
+            # Otherwise, the frequency is in hours
+            return True
+        else
```
DavidHuber-NOAA:
```suggestion
        else:
```

ush/python/pygfs/task/archive.py
```diff
@@ -439,7 +439,6 @@ def replace_string_from_to_file(filename_in, filename_out, search_str, replace_s
 
         return
 
-    @staticmethod
     @logit(logger)
     def _archive_expdir(arch_dict: Dict[str, Any]) -> bool:
```
WalterKolczynski-NOAA: You removed the decorator, but did not actually make this an instance method by adding `self`. Also, once this is a instance method, you should no longer need to pass in a dict, since the object should already have the information contained within the instance.
```suggestion
    def _archive_expdir(self) -> bool:
```

WalterKolczynski-NOAA (changes requested):

ush/python/pygfs/task/archive.py
DavidHuber-NOAA: Thanks for catching my slop.  I added `self`, but opted to keep the `arch_dict` as the `task_config` contains all environmental variables and the configuration/execution methods should use the more limitedly scoped dictionaries (i.e. `arch_dict`).  Tested on WCOSS2 with `HPSSARCH=YES`.

WalterKolczynski-NOAA: Fine for now, we can revisit later.

WalterKolczynski-NOAA (approved):

Issue #2639, opened 2024-05-30
Re-enable tracker jobs on Hercules and Orion


WalterKolczynski-NOAA:
### What new functionality do you need?

The tracker external package is not yet available for Rocky 9 on Orion and Hercules.
~A hack was added to the tracker jobs in #2608 to immediately exit those jobs on Hercules.~
These jobs have thus been disabled on Orion and Hercules.  Once the external tracker package is available, those jobs should be reenabled.

### What are the requirements for the new functionality?

Tracker jobs run in full instead of exiting immediately.

### Acceptance Criteria

- [ ] Tracker jobs run in full without error on MSU machines

DavidHuber-NOAA:
Updated title and description to reflect that this is an issue for both Orion and Hercules.

Note that the upstream dependency issue is https://github.com/NOAA-EMC/TC_tracker/issues/3.  FYI @JiayiPeng-NOAA.

JessicaMeixner-NOAA:
@KateFriedman-NOAA - I thought I'd check to see if there were any updates on this.

KateFriedman-NOAA:
> [@KateFriedman-NOAA](https://github.com/KateFriedman-NOAA) - I thought I'd check to see if there were any updates on this.

I need to install the new tracker tag and test it on MSU. It's on the todo list.

KateFriedman-NOAA:
Related to issue #3371 . The new tag is installed on MSU and will be tested.

PR #3138, opened 2024-12-03
Update HPC Tier Definitions

### LINKED ISSUE #3135: Update Machine Support Documentation

Add/amend documentation to reflect the new 3-tier system for supported HPCs/cloud platforms.

Documentation is clear and all supported platforms are categorized.

### Acceptance Criteria

Documentation builds successfully.

_No response_

DavidHuber-NOAA:
# Description
Updates the tier definitions for supported platforms.  This also updates and reformats the README.
~Note that the README badges will display errors until this PR is merged.~ @TerrenceMcGuinness-NOAA updated the gist, so the badges are now showing up correctly in the README.

Resolves #3135 
# Type of change
- [x] Documentation update

# Change characteristics

# How has this been tested?
Checked the [rendered README.md](https://github.com/DavidHuber-NOAA/global-workflow/blob/doc/hpc_tiers/README.md).

# Checklist


README.md
```diff
@@ -4,27 +4,42 @@
...
 ![Custom badge](https://img.shields.io/endpoint?url=https://gist.githubusercontent.com/emcbot/e35aa2904a54deae6bbb1fdc2d960c71/raw/orion.json)
+![Custom badge](https://img.shields.io/endpoint?url=https://gist.githubusercontent.com/emcbot/e35aa2904a54deae6bbb1fdc2d960c71/raw/hercules.json)
+![Custom badge](https://img.shields.io/endpoint?url=https://gist.githubusercontent.com/emcbot/e35aa2904a54deae6bbb1fdc2d960c71/raw/wcoss2.json)
 
-# global-workflow
-Global Workflow currently supporting the Global Forecast System (GFS) with the [UFS-weather-model](https://github.com/ufs-community/ufs-weather-model) and [GSI](https://github.com/NOAA-EMC/GSI)-based Data Assimilation System.
+The Global Workflow supporting the Global Forecast System (GFS) and the Global Ensemble Forecasting System (GEFS) with the [UFS-weather-model](https://github.com/ufs-community/ufs-weather-model).  Data assimilation, currently only available for the GFS, is provide by both the [GSI](https://github.com/NOAA-EMC/GSI)- and [GDASApp (JEDI)](https://github.com/NOAA-EMC/GDASApp)-based Data Assimilation systems.
```
KateFriedman-NOAA:
```suggestion
The Global Workflow supporting the Global Forecast System (GFS), the Global Ensemble Forecasting System (GEFS), and the Seasonal Forecast System (SFS) with the [UFS-weather-model](https://github.com/ufs-community/ufs-weather-model).  Data assimilation, currently only available for the GFS, is provide by both the [GSI](https://github.com/NOAA-EMC/GSI)- and [GDASApp (JEDI)](https://github.com/NOAA-EMC/GDASApp)-based Data Assimilation systems.
```

KateFriedman-NOAA (changes requested):
Can't comment on the `.github/workflows/*.yaml` files but the `README.md` updates look good. Left one suggested change.

WalterKolczynski-NOAA (dismissed):

docs/source/hpc.rst
```diff
@@ -44,6 +44,39 @@ The following system software requirements are the minimum for any new or existi
 |              |             |   WCOSS2                              |
 +--------------+-------------+---------------------------------------+
 
+===========================
+Feature availability by HPC
+===========================
```
KateFriedman-NOAA, line 49: So this feels like something that shouldn't be in the HPC section and should instead be in either the "GFS Configuration" or perhaps "Global Workflow Components" section or become a new section. We should separate the feature/science stuff from the purely HPC stuff.

Also, I see the new "Wave Component" section...that should move under the "Global Workflow Components" section. Perhaps we can move it in this PR as well or a follow-up one?

Thoughts? @WalterKolczynski-NOAA @aerorahul

DavidHuber-NOAA: I think the Wave Component section should be moved as well.  There's a lot going on in that section and some of it could be expanded to other components.

I would suggest making a component subsection for each of the major forecast/DA components (ATM, Waves, Ocean/Ice, Aerosols) with submodule names (e.g. Ocean DA via the GDASApp and forecasting via the UFS/MOM6), valid grids, and associated fix files.  This could probably be a table.  Then, maybe under the "Contributing to the Global Workflow" section, we could add subsections to cover adding new grids and fix files.

The GFS Configuration section is in need of a complete revamp, but it would be difficult to maintain the graphics as JEDI-based DA jobs are rapidly changing.  I wonder if there is a way to code a flowchart in either RST or a Python-drawn image/set of images.  That would make it much easier to update.

I still think the capabilities matrix fits well in the HPC Settings section, but it could refer back to the Global Workflow Components subsections.

Aside from the GFS Configuration section, I think everything could be tackled in this PR.  I can open an issue for the flowcharts if there is consensus on going that way.

WalterKolczynski-NOAA: I think a lot of the docs need to be rethought for the new multiply coupled world. Fix stuff here if it is easy, but otherwise let's not bog this PR down with correcting a bunch of ancillary things.

DavidHuber-NOAA: Yeah, I'm finding I have opened a can of worms.  I'll make a couple of cleanup changes here and open another issue to rework the documentation.

KateFriedman-NOAA: Works for me!

DavidHuber-NOAA:
I'm not sure how to get consistent spacing inside of the table column headers.  I may try some other options tomorrow.

docs/source/hpc.rst
DavidHuber-NOAA: I think I am as happy as I'm going to be with this table now.  I don't like how long the code had to be, but it was either this, wonky headers, or raw HTML.  I'm OK with returning to wonky headers if this is too much.

README.md
```diff
@@ -3,28 +3,42 @@
...
+## Supported platforms
 
-Documentation (in progress) is available [here](https://global-workflow.readthedocs.io/en/latest/).
+The Global Workflow currently supports the following machines at the indicated tier.
+
+| HPC                                     | Tier | Notes                                                                      |
+| --------------------------------------- |:----:|:--------------------------------------------------------------------------:|
+| WCOSS2<br>NCO                           | 1    | GEFS testing is not regularly performed.                                   |
+| Hera<br>NOAA RDHPCS                     | 1    |                                                                            |
+| Hercules<br>MSU                         | 1    | Currently does not support the TC Tracker.                                 |
```
JessicaMeixner-NOAA, line 30: @DavidHuber-NOAA  This table is great! I learned for example, that TC tracker is not supported.  I tried to look for an issue for this, but couldn't find one.  Any chance you could point me towards additional information.

DavidHuber-NOAA: Sure, here is the GW issue: https://github.com/NOAA-EMC/global-workflow/issues/2639.

And here is the TC Tracker issue: https://github.com/NOAA-EMC/TC_tracker/issues/3.

Thinking about it now, I suspect we could use the Orion executables on Hercules until this is resolved.  That may be worth a test.

DavidHuber-NOAA: Actually, it seems that the tracker is disabled on both Orion and Hercules as it has not been rebuilt for either system since the Rocky 9 update, so using Orion executables is impossible.  I will update the issues and this table accordingly.

KateFriedman-NOAA (approved):
Looks good, thanks @DavidHuber-NOAA !

Issue #2558, opened 2024-05-01
GFSv16.3.20 - Annual CO2 update for 2025 and turn on assimilation of PlanetiQ GPS Radio Occultation (GPS-RO) data


KateFriedman-NOAA:
### Description

The annual update to CO2 fix files, as well as updates to storm names for the 2025, 2026, and 2027 hurricane seasons.

Updated fix files will be provided by @Qingfu-Liu .

This issue is on hold until fall 2024 for the regular annual update time.

Document from @Qingfu-Liu : https://docs.google.com/document/d/1sWaex-8cfETPajB7DKfG-fN_TK5WK_et/edit#heading=h.gjdgxs

### Target version

v16.3.20

### Expected workflow changes

None

KateFriedman-NOAA:
The new CO2 files have been pulled into the para install for `gfs.v16.3.20` by @DianeStokes-NCO . This issue will close when #2913 is completed.

KateFriedman-NOAA:
> as well as updates to storm names for the 2025, 2026, and 2027 hurricane seasons.

@Qingfu-Liu were the storm name updates included in the files you provided recently?

KateFriedman-NOAA:
Files copied into GFSv16 fix set on WCOSS2 for pickup by NCO:
```
WCOSS2 (PRIMARYSYS-D) fix_am> pwd
/lfs/h2/emc/global/save/emc.global/FIX/fix_nco_gfsv16.3.0/fix_am
WCOSS2 (PRIMARYSYS-D) fix_am> ll co2dat_4a/global_co2historicaldata_2023.txt co2dat_4a/global_co2historicaldata_2024.txt_proj_u co2dat_4a/global_co2historicaldata_2025.txt_proj fix_co2_proj/global_co2historicaldata_2025.txt fix_co2_update/global_co2historicaldata_2024.txt
-rw-r--r-- 1 emc.global global 24484 Oct  7 20:55 co2dat_4a/global_co2historicaldata_2023.txt
-rw-r--r-- 1 emc.global global 24484 Oct  7 20:55 co2dat_4a/global_co2historicaldata_2024.txt_proj_u
-rw-r--r-- 1 emc.global global 24484 Oct  7 20:55 co2dat_4a/global_co2historicaldata_2025.txt_proj
-rw-r--r-- 1 emc.global global 24484 Oct  7 20:55 fix_co2_proj/global_co2historicaldata_2025.txt
[...]

PR #3144, opened 2024-12-06
Reduce scope to annual CO2 update and PlanetIQ assimilation

KateFriedman-NOAA:
# Description

The `obsproc/v1.3` upgrade is delayed, therefore, the `GFSv16.3.20` will now update the following:

- install 2025 carbon dioxide (CO2) fix file updates before the start of the new year
- updated GSI global_convinfo.txt to turn on assimilation of PlanetiQ GPS Radio Occultation (GPS-RO) data
- memory increases for a number of jobs (provided by NCO)
- load the cmdaccel module at the start of wave point postprocessor jobs to stabilize runtime (provided by NCO)

FYI @DianeStokes-NCO 

Refs #2558

# Type of change

Production update

# How has this been tested?

NCO para testing

### DIFF
--- a/versions/run.ver
+++ b/versions/run.ver
@@ -1,5 +1,5 @@
-export version=v16.3.19
-export gfs_ver=v16.3.19
+export version=v16.3.20
+export gfs_ver=v16.3.20
 export ukmet_ver=v2.2
 export ecmwf_ver=v2.1
 export nam_ver=v4.2
@@ -47,5 +47,7 @@ export g2_ver=3.4.5
 export zlib_ver=1.2.11
 export sp_ver=2.3.3
 export ip_ver=3.3.3
+
+export cmdaccel_ver=1.1
 
 export NET=gfs
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -8,7 +8,7 @@ protocol = git
 required = True
 
 [GSI]
-tag = gfsda.v16.3.19
+tag = gfsda.v16.3.20
 local_path = sorc/gsi.fd
 repo_url = https://github.com/NOAA-EMC/GSI.git
 protocol = git
--- a/docs/Release_Notes.md
+++ b/docs/Release_Notes.md
@@ -26,7 +26,7 @@ The checkout script extracts the following GFS components:
 | --------- | ----------- | ----------------- |
 | MODEL     | GFS.v16.3.1   | Jun.Wang@noaa.gov |
 | GLDAS     | gldas_gfsv16_release.v.2.1.0 | Helin.Wei@noaa.gov |
-| GSI       | gfsda.v16.3.19 | Andrew.Collard@noaa.gov |
+| GSI       | gfsda.v16.3.20 | Andrew.Collard@noaa.gov |
 | UFS_UTILS | ops-gfsv16.3.0 | George.Gayno@noaa.gov |
 | POST      | upp_v8.3.0 | Wen.Meng@noaa.gov |
 | WAFS      | gfs_wafs.v6.3.3 | Yali.Mao@noaa.gov |
@@ -124,3 +124,4 @@ DOCUMENTATION
 PREPARED BY
 -----------
 Kate.Friedman@noaa.gov
+Andrew.Collard@noaa.gov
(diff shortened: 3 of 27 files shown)


KateFriedman-NOAA:
Awaiting final confirmation on whether an additional GSI code change will be included.

aerorahul (approved):
looks ok to me.
quite an update to the memory requests.  Can they tell us what changed to get the new estimates? Should note it in the release notes for documentation that they came from testing by the SPA team.

KateFriedman-NOAA:
Folded in a new GSI tag and additional memory increase, with mention in release notes. Awaiting potential final resource change to a walltime from NCO but otherwise this PR should be reviewed and approved. Thanks!

KateFriedman-NOAA:
Added increase to enkfgdas_update job walltime (35mins -> 40mins) @ [7552cf4](https://github.com/NOAA-EMC/global-workflow/pull/3144/commits/7552cf44da6893b1b8951beaf2603e0edb256547). Update was merged into ops this morning, so this is a complete changeset now. Please review and approve, thanks!

versions/run.ver
```diff
@@ -48,4 +48,6 @@ export zlib_ver=1.2.11
 export sp_ver=2.3.3
 export ip_ver=3.3.3
 
+export cmdaccel_ver=1.1
```
WalterKolczynski-NOAA, line 51: What is this new module?

WalterKolczynski-NOAA (approved):

DianeStokes-NCO:
Here is the description of the cmdaccel module:

------------------------------------------------------------
--------------------------------------------
   /apps/prod/lmodules/sys_core/cmdaccel/1.1.lua:
------------------------------------------------------------
--------------------------------------------
Description:
Sets up environment-modifying functions for the most frequently used
scripting commands from /usr/bin.
These functions adjust the LD_LIBRARY_PATH environment variable temporarily
while running the command to
avoid unnecessary searches through application paths in /apps, potentially
saving considerable execution
time.

In cmdaccel/1.1, functions are created for the following commands:
  awk  cat  chmod  chown  cp  cut  date  env  grep  ln  mkdir  rm  sed
tar  wc
The function for cut, eg, looks like:
function cut() { LD_PRELOAD= LD_LIBRARY_PATH="/lib64:/usr/lib64"
/usr/bin/cut "$@"; }

The gfs_wave_post*pnt* jobs run many small scripts in parallel and those
scripts run many basic commands.   We see wall clock failures for thes jobs
too often because, for some reason *(system load, slow node, etc),* they
run much more slowly than usual.  Additionally, runtime for these jobs have
increased since the SP4 upgrade.  GDIT has made several suggestions to
improve performance of these jobs.  Some require script or source code
changes.  One simple suggestion that provided obvious improvement was to
use the cmdaccel module.

Following are typical run times in minutes before and after the SP4 upgrade
and after SP4 when we load cmdaccel/1.1:
           *job*             *SP3*   *SP4*   *SP4 w/cmdaccel*
  jgfs_wave_post_bndpnt     39    45    30 minutes
  jgfs_wave_post_bndpntbll  34    36    22 minutes
  jgfs_wave_postpnt         48    57    40 minutes

I also added the load of cmdaccel to jgdas_wave_postpnt: 1) for consistency
with jgfs_wave_postpnt; and 2) because it helped rather than hurt.  Typical
runtimes for that job:
           *job*             *SP3*   *SP4*   *SP4 w/cmdaccel*
  jgdas_wave_postpnt       1.6   2.3   1.8 minutes


I compared output from runs with and without cmdaccel and content of the
output files matched.
Please let me know if you have further questions or concerns.
    Diane

versions/run.ver
KateFriedman-NOAA: See Diane's comment https://github.com/NOAA-EMC/global-workflow/pull/3144#issuecomment-2533190930

KateFriedman-NOAA:
Thanks @DianeStokes-NCO for the description on what the `cmdaccel` module is and how it helps.

aerorahul:
Thanks @DianeStokes-NCO for the explaination.

PR #3171, opened 2024-12-16
Set runtime shell requirements within Jenkins Pipeline

TerrenceMcGuinness-NOAA:
# Description

This PR updates the Jenkins Pipeline script to source `gw_setup.sh` from within the PR itself for when ever a shell command is issued that needs it.

Currently the runtime bash shell environment requirements for running the **global-workflow** (set by sourcing `$HOMEgfs/workflow/gw_setup.sh`) is established by the user's shell that makes the connection from the Java remote agent to the Jenkins Controller.  This update simply sources `gw_setup.sh` directly during run time when ever needed instead.

**NOTE:**  This does not apply for **GitHub CLI** and **pyGitHub** since they are not currently supported by **Spack Stack** and therefore not incorporated in to `gw_setup.sh`.  These remaining dependances are the responsibility of the _role account_ holders that manage the Jenkins Controller to Node Agent connections.

# Type of change


AnilKumar-NOAA (commented):
We have added HOMEGFS to the runtime shell environment for gaea C5/C6 both in our epic role account .bashrc file as follows:  
### Added for Global Workflow
if [ -n "${HOMEGFS+x}" ] && [ -n "${HOMEGFS}" ]; then
   source ${HOMEGFS}/workflow/gw_setup.sh
fi

TerrenceMcGuinness-NOAA:
@AnilKumar-NOAA Oh I see now that you were referring to the `.bashrc` in the roll account.  In a Jenkins Pipeline the shell commands are not login shells but are issued from the shell environment from which the agent had been made to make the connection to the controller.  Having said that, I will have to make some adjustments to the wrapper scripts so the needed modules can be set at runtime.  Let me make some updates to the wrapper scripts in the CI framework and run some tests from a generic shell.  In the meantime this PR has been reverted to a Draft

ci/scripts/utils/launch_java_agent.sh
```diff
@@ -102,9 +102,10 @@ JAVA="${JAVA_HOME}/bin/java"
 echo "JAVA VERSION: "
 ${JAVA} -version
 
-export GH="${HOME}/bin/gh"
-[[ -f "${GH}" ]] || echo "gh is not installed in ${HOME}/bin"
+GH=$(command -v gh || echo "${HOME}/bin/gh")
+[[ -f "${GH}" ]] || echo "ERROR: GitHub CLI (gh) not found. (exiting with error)"; exit 1
```
aerorahul: I think this is what I think you intend.  Otherwise, `exit 1`will be called regardless of the result.
```suggestion
[[ -f "${GH}" ]] || ( echo "ERROR: GitHub CLI (gh) not found. (exiting with error)"; exit 1 )
```

aerorahul (commented):
changes to source `gw_setup.sh` are good.  
A comment at the end of this PR re. `GH` might solve the shellnorm error.  Please check.
Approve.

ci/scripts/utils/launch_java_agent.sh
TerrenceMcGuinness-NOAA: Yes, **shellnorms** spotted this too.

TerrenceMcGuinness-NOAA (commented):
committing update to launch script so exit works correctly when gh is not found

aerorahul (approved):
looks good.

@TerrenceMcGuinness-NOAA 
Just a note here that "launch" is misspelled as "lanuch".  Consequently, variables such as `JENKINS_AGENT_LAUNCH_DIR` is misspelled as `JENKINS_AGENT_LANUCH_DIR`.  Not sure where that variable is defined.

This should probably be corrected in a future PR.

aerorahul:
@TerrenceMcGuinness-NOAA 
I just merged this PR and got a message about a Hercules jenkins-node.env fail.  I hope this PR does not need additional work to be taken on the HPCs

PR #3186, opened 2024-12-20
Create compute build option

### LINKED ISSUE #3131: Build using compute nodes

The capability to build on compute nodes using the job scheduler

Build all components using the system job scheduler, whether it be slurm or pbspro. Most programs should be able to build on compute nodes, but the GDASapp will need to build in the service queue for now as it (currently) contacts the outside world to build.

### Acceptance Criteria

- [ ] All components build without error using slurm
- [ ] All components build without error using pbspro
- [ ] Output unchanged (maybe, building on compute nodes may result in changes; at very least all programs should run successfully)

_No response_

DavidHuber-NOAA:
# Description
This creates scripts to run compute-node builds and also refactors the build_all.sh script to make it easier to build all executables.

In place of various options to control what components are built when using `build_all.sh`, instead it takes in a list of one or more systems to build:

- `gfs` builds everything needed for forecast-only gfs (UFS model with unstructured wave grid, gfs_utils, ufs_utils, upp, ww3 pre/post for unstructured wave grid)
- `gefs` builds everything needed for GEFS (UFS model with structured wave grid, gfs_utils, ufs_utils, upp, ww3 pre/post for structured wave grid)
- `sfs` builds everything needed SFS (UFS model in hydrostatic mode with unstructured wave grid, gfs_utils, ufs_utils, upp, ww3 pre/post for structured wave grid)
- `gsi` builds GSI-based DA components (gsi_enkf, gsi_monitor, gsi_utils)
- `gdas` builds JEDI-based DA components (gdas app, gsi_monitor, gsi_utils)

`all` will build all of the above (mostly for testing)

Examples:
Build for forecast-only GFS:
```./build_all.sh gfs```
Build cycled GFS including coupled DA:
``` ./build_all.sh gfs gsi gdas```
Build GEFS:
```./build_all.sh gefs```
Build everything (for testing purposes):
```./build_all.sh all```
Other options, such as `-d` to build in debug mode, remain unchanged.

The full script signature is now:
```
./build_all.sh [-a UFS_app][-c build_config][-d][-f][-h][-v] [gfs] [gefs] [sfs] [gsi] [gdas] [all]
```

Additionally, there is a new script to build components on the compute nodes using the job scheduler instead of the login node. This method takes the load off of the login nodes and may be faster in some cases. Compute build is invoked using the build_compute.sh script, which behaves similarly to the new `build_all.sh:`
```
./build_compute.sh [-h][-v][-A <hpc-account>] [gfs] [gefs] [sfs] [gsi] [gdas] [all]
```
Compute build will generate a rocoto workflow and then call `rocotorun` itself repeatedly until either a build fails or all builds succeed, at which point the script will exit. Since the script is calling `rocotorun` itself, you don't need to set up your own cron to do it, but advanced users can also use all the regular rocoto tools on `build.xml` and `build.db` if you wish.

Some things to note with the compute build:
- When a build fails, other build jobs are not cancelled and will continue to run.
- Since the script stops running `rocotorun` once one build fails, the rocoto database will no longer update with the status of the remaining jobs after that point.
- Similarly, if the terminal running `build_compute.sh` gets disconnected, the rocoto database will no longer update.
- In either of the above cases, you could run `rocotorun` yourself manually to update the database as long as the job information hasn't aged off the scheduler yet.

Resolves #3131

# Type of change

# Change characteristics
- Is this a breaking change (a change in existing functionality)? YES (build jobs are no longer specifiable)

# How has this been tested?
- [x] Built and linked all executables with both `build_all.sh` and `compute_build.sh`
- [ ] Run a C96C48_hybatmDA case on Hera
- [ ] Run a C48_S2SWA_gefs case on Hera

# Checklist

### CODE REVIEW

File: workflow/build_opts.yaml [hunk 1]
```diff
@@ -0,0 +1,90 @@
...
+    walltime: "00:30:00"
+
+  gefs_ww3_prepost:
+    command: "./build_ww3prepost.sh -w -j 4"
+    log: "build_ww3prepost_gefs.log"
+    cores: 4
+    walltime: "00:10:00"
+
+  sfs_model:
+    command: "./build_ufs.sh -e sfs_model.x -j 12"
```
aerorahul:
```suggestion
    command: "./build_ufs.sh -y -e sfs_model.x -j 12"
```
I think the SFS model is built with a hydrostatic option.

workflow/build_opts.yaml
```diff
@@ -0,0 +1,90 @@
...
+    - "ufs_utils"
+    - "gfs_utils"
+    - "upp"
+  gfs:
+    - "gfs_model"
+    - "gfs_ww3prepost"
+    - "gsi_enkf"
+    - "gdas"
+    - "gsi_utils"
+    - "gsi_monitor"
```
aerorahul: We could probably offer more systems here:
`gfs` - as is here (this would cover the JEDI and GSI cycled options as well as the forecast only option)
`gfs_forecast_only` - just `gfs_model` and `gfs_ww3prepost`
`gfs_gsi` - everything here except `gdas`

ush/compute_build.py
aerorahul, line 1: Where is this script used?

workflow/build_compute.py
```diff
@@ -0,0 +1,176 @@
...
+    if user_inputs.account is not None:
+        host_specs.account = user_inputs.account
+
+    build_specs = AttrDict(parse_yaml(user_inputs.yaml))
+
+    # Determine systems to build
+    builds = set()
+    if user_inputs.system == "all":
+        for system, build_set in build_specs.systems.items():
+            builds.update(build_set)
```
aerorahul: Isn't this just the `build_specs.build` instead of going through systems?

sorc/build_compute.sh
```diff
@@ -0,0 +1,108 @@
...
+   else
+      echo "FATAL ERROR: ${BASH_SOURCE[0]} rocoto failed with state '${state}'"
+      exit 3
+   fi
+done
+
+echo "All builds completed successfully!"
+
+# Remove the database file
+rm -f "${build_db}"
```
aerorahul: Can we keep this file, in case someone wants to go and recompile, they can do rocotorewind and rocotorun.

aerorahul (commented):
looks good.  thanks for taking the suggestions and updating to accommodate all systems.
A few more suggestions based on the code.

workflow/build_compute.py
DavidHuber-NOAA: Ah, yes, we only have to loop through systems if something besides `all` is selected.

sorc/build_compute.sh
DavidHuber-NOAA: Yep.

workflow/build_opts.yaml
DavidHuber-NOAA: Good idea, added.

workflow/build_opts.yaml
DavidHuber-NOAA: Yep, fixed.

ush/compute_build.py
DavidHuber-NOAA: It's not.  Deleted.

DavidHuber-NOAA:
@aerorahul Thanks for the quick review.  I'm rerunning build tests now and will re-request a review when complete.

sorc/build_all.sh
```diff
@@ -48,30 +39,21 @@ readonly HOMEgfs=$(cd "$(dirname "$(readlink -f -n "${BASH_SOURCE[0]}" )" )/.."
...
 _verbose_opt=""
-_wave_opt=""
-_hydro_opt=""
 _build_job_max=20
 _quick_kill="NO"
 _ufs_exec="-e gfs_model.x"
 # Reset option counter in case this script is sourced
 OPTIND=1
-while getopts ":a:dfghj:kuvwy" option; do
+while getopts ":a:dfhj:kv" option; do
```
aerorahul: The `-j` option is gone, I think.

DavidHuber-NOAA: Yep, removed.

ci/Jenkinsfile
```diff
@@ -120,9 +120,7 @@ pipeline {
                             def error_logs_message = ""
                             dir("${HOMEgfs}/sorc") {
                                 try {
-                                    sh(script: './build_all.sh -kgu')  // build the global-workflow executables for GFS variant (UFS-wx-model, WW3 pre/post executables)
-                                    sh(script: './build_ww3prepost.sh -w > ./logs/build_ww3prepost_gefs.log 2>&1')  // build the WW3 pre/post processing executables for GEFS variant
-                                    sh(script: './build_ufs.sh -w -e gefs_model.x > ./logs/build_ufs_gefs.log 2>&1')  // build the UFS-wx-model executable for GEFS variant
+                                    sh(script: './build_all.sh gfs gefs gsi gdas')  // build the global-workflow executables
```
aerorahul: Wouldn't the `gdas` also build `gsi_monitor` and `gsi_utils`

DavidHuber-NOAA: Yes, but `gsi` is still needed to build the `gsi_enkf`.

aerorahul: Perhaps `gdas_builds` should only include `gdas` and not `gsi_utils gsi_monitor`.

aerorahul: L79 of `build_all.sh`.

DavidHuber-NOAA: Maybe, but I think GDASApp-based DA still makes use of some gsi-utils and gsi-monitor executables during cycling.  @RussTreadon-NOAA @CoryMartin-NOAA is that still the case?

sorc/build_all.sh
```diff
@@ -82,19 +64,95 @@ while getopts ":a:dfghj:kuvwy" option; do
...
+# shellcheck disable=SC2034
+gefs_builds="gefs gfs_utils ufs_utils upp ww3_gefs"
+# shellcheck disable=SC2034
+sfs_builds="sfs gfs_utils ufs_utils upp ww3_gefs"
+# shellcheck disable=SC2034
+gsi_builds="gsi_enkf gsi_monitor gsi_utils"
+# shellcheck disable=SC2034
+gdas_builds="gdas gsi_monitor gsi_utils"
+# shellcheck disable=SC2034
+all_builds="gfs gfs_utils ufs_utils upp ww3_gfs ww3_gefs gdas gsi_enkf gsi_monitor gsi_monitor gsi_utils"
```
aerorahul: I think this is missing `gefs` and `gsi_monitor` is duplicated
```suggestion
all_builds="gfs gfs_utils ufs_utils upp ww3_gfs gefs ww3_gefs gdas gsi_enkf gsi_monitor gsi_utils"
```

ci/Jenkinsfile
aerorahul: it might. but the point here is that this invocation is going to build `gsi_monitor` and `gsi_utils` twice.  
Could we run into race conditions?

DavidHuber-NOAA: Ah, I see.  No, it will only build once.  `build_all.sh` uses associative arrays, so `build[gsi_monitor]=yes` will be assigned, so it will only build once.

sorc/build_all.sh
```diff
@@ -82,19 +64,95 @@ while getopts ":a:dfghj:kuvwy" option; do
...
+      (( system_count += 1 ))
+      build_list_name="${system}_builds"
+      for build in ${!build_list_name}; do
+         builds["${build}"]="yes"
+      done
+   else
+      echo "Unsupported build system: ${system}"
+      _usage
+   fi
+done
```
DavidHuber-NOAA: @aerorahul Here is where the magic happens.

aerorahul: excellent.  Thanks for pointing this out.

ci/Jenkinsfile
aerorahul: Ok. thanks I understand now from the other comment, that an intersection is performed and all duplicate entries are removed.

aerorahul (dismissed):
lgtm.

aerorahul (dismissed):
LGTM

aerorahul (dismissed):
👍🏾

DavidHuber-NOAA:
I'm trying out compute-node builds in Jenkins on Hercules.

sorc/build_compute.sh
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env bash
+
+function _usage() {
+  cat << EOF
+Builds all of the global-workflow components on compute nodes.
+
+Usage: ${BASH_SOURCE[0]} [-h][-v][-A <hpc-account>][-s gfs,gfs_forecast_only,gfs_gsi,gefs,sfs,all]
```
WalterKolczynski-NOAA: I would have the script signature match build_all (no `-s`; allow any number of gfs, gefs, sfs, gsi, gdas; support `-d`, etc.).

DavidHuber-NOAA:
Hercules build times follow (HH:MM:SS):

```
gfs_model   00:16:18
gfs_ww3prepost   00:02:21
gefs_model   00:09:43
gefs_ww3_prepost   00:02:13
sfs_model   00:16:51
gsi_enkf   00:08:21
gsi_monitor   00:01:11
gsi_utils   00:00:34
ufs_utils   00:00:58
gfs_utils   00:00:34
gdas   00:31:30
```

sorc/build_all.sh
```diff
@@ -82,20 +64,94 @@ while getopts ":a:dfghj:kuvwy" option; do
...
+# If no build system was specified, build for gfs forecast-only
+if [[ $# -eq 0 ]]; then
+   selected_systems="gfs"
+else
+   selected_systems="$*"
+fi
+
+supported_systems=("gfs" "gefs" "sfs" "gsi" "gdas" "all")
+# shellcheck disable=SC2034
+gfs_builds="gfs gfs_utils ufs_utils upp ww3_gfs"
```
WalterKolczynski-NOAA: Would call the UFS build `ufs_gfs`, `ufs_gefs`, etc. to avoid confusion.

sorc/build_all.sh
```diff
@@ -82,20 +64,94 @@ while getopts ":a:dfghj:kuvwy" option; do
...
+build_jobs=(
+    ["gfs"]=8 ["gefs"]=8 ["sfs"]=8 ["gdas"]=8 ["gsi_enkf"]=2 ["gfs_utils"]=1 ["ufs_utils"]=1
+    ["ww3_gfs"]=1 ["ww3_gefs"]=1 ["gsi_utils"]=1 ["gsi_monitor"]=1 ["gfs_utils"]=1 ["upp"]=1
+)
+
+# Establish build options for each job
+_gfs_exec="gfs_model.x"
+_gefs_exec="gefs_model.x"
+_sfs_exec="sfs_model.x"
+build_opts=(
```
WalterKolczynski-NOAA, line 105: Might be clearer to not separate the build script from the build options and just have one dict.

sorc/build_all.sh
```diff
@@ -82,20 +64,94 @@ while getopts ":a:dfghj:kuvwy" option; do
...
-if [[ ! -d "${HOMEgfs}/exec" ]]; then
-  echo "Creating ${HOMEgfs}/exec folder"
-  mkdir -p "${HOMEgfs}/exec"
-fi
+# Jobs per build ("min max")
+declare -A build_jobs build_opts build_scripts
+build_jobs=(
+    ["gfs"]=8 ["gefs"]=8 ["sfs"]=8 ["gdas"]=8 ["gsi_enkf"]=2 ["gfs_utils"]=1 ["ufs_utils"]=1
+    ["ww3_gfs"]=1 ["ww3_gefs"]=1 ["gsi_utils"]=1 ["gsi_monitor"]=1 ["gfs_utils"]=1 ["upp"]=1
+)
```
WalterKolczynski-NOAA: Would love to pull this out of the yaml at some future point to eliminate duplication (unless we think we'll use fewer CPUs in interactive mode to avoid stressing the login nodes).

sorc/build_all.sh
WalterKolczynski-NOAA: Could also use an associative array to hold the builds. Looks cleaner than a bunch of variable names:
```suggestion
builds['gfs']="ufs_gfs gfs_utils ufs_utils upp ww3_gfs"
```

WalterKolczynski-NOAA (changes requested):
Other than aligning the build signatures, these are all "might be nicer this way" rather than hard suggestions.

sorc/build_compute.sh
DavidHuber-NOAA: Agreed.  I'd like to tackle this in a follow-up, though.  I kept this script as simple as possible for the time being.  I'm also not going to document this script in RTD at this time.

sorc/build_all.sh
DavidHuber-NOAA: Yeah, I think they will need to be separate.  I've overloaded the head nodes in the past (and had a finger-wagging from RDHPCS), while for the compute nodes we can hit them as hard as we want.  This is also why I am getting rid of the `-j` option.

aerorahul: would love for this script to be able to read that yaml (or build_config) file to get that information.

sorc/build_compute.sh
DavidHuber-NOAA: Simpler than I thought.  Done.

WalterKolczynski-NOAA:
UPP compute build fails immediately on WCOSS because it can't find `cmake`:

```
+ cd build
+ cmake -DCMAKE_INSTALL_PREFIX=../install -DBUILD_WITH_NEMSIO=ON -DBUILD_WITH_WRFIO=ON -DBUILD_WITH_GTG=OFF -DBUILD_WITH_IFI=OFF /lfs/h2/emc/ptmp/walter.kolczynski/PR/PR_3186/global-workflow/sorc/ufs_model.fd/FV3/upp
./compile_upp.sh: line 142: cmake: command not found
PBS: Running epilogue on parent mom node: nid001702...
```

WalterKolczynski-NOAA:
gdas also fails, but that was expected since it does not build in develop either.

DavidHuber-NOAA:
The UPP modules do not load cmake, so to make this work, the UPP will need to be built on the dev_transfer queue.  Fortunately, the transfer queue does let you assign at least up to 8 CPUs, so the YAML doesn't have to change.  I will just need a bit of a hack in compute_build.py for WCOSS2.

aerorahul:
@WenMeng-NOAA 
We would like to build upp on the compute nodes on all platforms.  
Can we include cmake in wcoss2.lua in the upp modulefiles?

RussTreadon-NOAA:
Adding cmake to `sorc/ufs_model.fd/FV3/upp/modulefiles/wcoss2.lua` is preferable.  upp builds on Dogwood with compute nodes using the following change to `wcoss2.lua`
```
 load(pathJoin("craype", craype_ver))
 load(pathJoin("cray-mpich", cray_mpich_ver))
 
+cmake_ver=os.getenv("cmake_ver") or "3.20.2"
+load(pathJoin("cmake", cmake_ver))
+
 hdf5_ver=os.getenv("hdf5_ver") or "1.10.6"
 netcdf_ver=os.getenv("netcdf_ver") or "4.7.4"
```

aerorahul:
@DavidHuber-NOAA 
Creating a git diff patch and applying it via build_compute.sh on WCOSS2 might be sufficient until we get the permanent fix in the upp module file. 
Just a thought.

DavidHuber-NOAA:
@aerorahul What do you think of the workaround I came up with to add `cmake` to the default modules.  This will cause `cmake` to load when the UPP `compile_upp.sh` script runs `module reset`.

DavidHuber-NOAA:
I opened issue NOAA-EMC/UPP#1106 to add `cmake` to the WCOSS2 module file.

aerorahul:
> @aerorahul What do you think of the workaround I came up with to add `cmake` to the default modules.  This will cause `cmake` to load when the UPP `compile_upp.sh` script runs `module reset`.

It works.  Though, will this not be applied for all machines? It might have unintended consequences.

DavidHuber-NOAA:
It will be applied to all machines that do not have `cmake` in the search path.  Having tested compute nodes on Hera, Gaea-C5, Gaea-C6, Hercules, and S4, I can verify that these all have `cmake` in their search paths.

I could go one step further and 1) add a `detect_machine.sh` call to `build_upp.sh` and 2) only run this block if we are on WCOSS2.  Thoughts?

DavidHuber-NOAA:
> @DavidHuber-NOAA Creating a git diff patch and applying it via build_compute.sh on WCOSS2 might be sufficient until we get the permanent fix in the upp module file. Just a thought.

This could also work, though the patch can only be applied once.  Thus, if `build_compute.sh` needs to be called again, it will need to be smart enough to know that the patch was already applied or just ignore the resulting error.

aerorahul:
Since all of this is temporary, just do what's necessary and easy.  Hopefully, upp will make a fix soon.

sorc/build_all.sh
DavidHuber-NOAA: This makes most things easier, but logging errors is a little tricky here:

https://github.com/NOAA-EMC/global-workflow/blob/72c8d7ac65a4f260aafad302c4ce50a45a1b1253/sorc/build_all.sh#L286-L290

I think I will save this for a later rendition.

sorc/build_all.sh
DavidHuber-NOAA: Done.

WenMeng-NOAA:
> @WenMeng-NOAA We would like to build upp on the compute nodes on all platforms. Can we include cmake in wcoss2.lua in the upp modulefiles?

@aerorahul Sure, please submit a PR to the UPP repos.

sorc/build_all.sh
DavidHuber-NOAA: Maybe these could be set differently in the YAML depending on whether it's a compute-node or head-node build...

DavidHuber-NOAA:
Test compute build on WCOSS2 was successful.

aerorahul (approved):
Lgtm

RussTreadon-NOAA:
FYI: updating `$HOMEgfs/sorc/gdas.cd/modulefiles/GDAS/wcoss2.intel.lua` to NCO's test installation of `spack-stack/1.6.0` along with adding `cmake` to the UPP wcoss2 modulefile allows `./build_compute.sh -v gdas` to successfully run to completion on Dogwood.

WalterKolczynski-NOAA:
UPP now builds successfully on compute nodes.

WalterKolczynski-NOAA (approved):

PR #3199, opened 2025-01-03
Ensure OCNRES and ICERES have 3 digits in the archive script

### LINKED ISSUE #3198: OCNRES variable not always 3 digits in archive task which may result in archiving issues
### What is wrong?

When `OCNRES` is 050 (mx050) or 025 (mx025), the ocean products are not archiving properly because python is removing the leading zeros. 

The leading zeros in "050" and "025" should remain to ensure proper archiving of the ocean products. In other words, `OCNRES` should be 3 digits in the archive task. 

All or N/A

1c37f907ecaadd835580d5bb73a0eaf2a146f2fc

### Steps to reproduce

Run a case that includes ocean at resolution of mx025 or mx050. Then run global workflow up to the archive task. 

### Additional information

This bug has been fixed in the feature/gefs_reforecast branch and needs to be brought to the develop branch. 

The following should be changed in [scripts/exglobal_archive.py](https://github.com/NOAA-EMC/global-workflow/blob/develop/scripts/exglobal_archive.py):

```
    archive_dict = AttrDict()
    for key in keys:
        archive_dict[key] = archive.task_config.get(key)
```

The above python code should be changed to the following:

```
    archive_dict = AttrDict()
    for key in keys:
        # Make sure OCNRES is three digits
        if key == "OCNRES":
            archive.task_config.OCNRES = f"{archive.task_config.OCNRES :03d}"
        archive_dict[key] = archive.task_config.get(key)
```

aerorahul:
# Description
This PR:
- ensures the `OCNRES` and `ICERES` variables in `task_config` are 3 digits

Resolves #3198 

# Type of change

# Change characteristics

- Is this a breaking change (a change in existing functionality)? YES/NO
- Does this change require a documentation update? YES/NO
- Does this change require an update to any of the following submodules? YES/NO (If YES, please add a link to any PRs that are pending.)
  - [ ] EMC verif-global 
  - [ ] GDAS 
  - [ ] GFS-utils 
  - [ ] GSI 
  - [ ] GSI-monitor 
  - [ ] GSI-utils 
  - [ ] UFS-utils 
  - [ ] UFS-weather-model 
  - [ ] wxflow 

# How has this been tested?

# Checklist

### DIFF
--- a/scripts/exglobal_archive.py
+++ b/scripts/exglobal_archive.py
@@ -37,9 +37,9 @@ def main():
 
     archive_dict = AttrDict()
     for key in keys:
+        if key in ['OCNRES', 'ICERES']:  # update this key to be 3 digits
+            archive.task_config[key] = f"{archive.task_config.get(key):03d}"
         archive_dict[key] = archive.task_config.get(key)
-        if key in ['OCNRES', 'ICERES']:
-            archive_dict[key] = f"{archive_dict.get(key):03d}"
         if archive_dict[key] is None:
             logger.warning(f"WARNING: key ({key}) not found in task_config!")
 
--- a/scripts/exglobal_archive.py
+++ b/scripts/exglobal_archive.py
@@ -38,15 +38,17 @@ def main():
     archive_dict = AttrDict()
     for key in keys:
         archive_dict[key] = archive.task_config.get(key)
+        if key in ['OCNRES', 'ICERES']:
+            archive_dict[key] = f"{archive_dict.get(key):03d}"
         if archive_dict[key] is None:
-            print(f"Warning: key ({key}) not found in task_config!")
+            logger.warning(f"WARNING: key ({key}) not found in task_config!")
 
     # Also import all COMIN* and COMOUT* directory and template variables
     for key in archive.task_config.keys():
         if key.startswith("COM_") or key.startswith("COMIN_") or key.startswith("COMOUT_"):
             archive_dict[key] = archive.task_config.get(key)
             if archive_dict[key] is None:
-                print(f"Warning: key ({key}) not found in task_config!")
+                logger.warning(f"WARNING: key ({key}) not found in task_config!")
 
     with chdir(config.ROTDIR):
 
--- a/scripts/exglobal_archive.py
+++ b/scripts/exglobal_archive.py
@@ -18,12 +18,11 @@ def main():
     archive = Archive(config)
 
     # update these keys to be 3 digits if they are part of archive.task_config.keys
-    if {'OCNRES', 'ICERES'} <= archive.task_config.keys():
-        for key in ['OCNRES', 'ICERES']:
-            try:
-                archive.task_config[key] = f"{archive.task_config[key]:03d}"
-            except KeyError as ee:
-                logger.info(f"key ({key}) not found in archive.task_config!")
+    for key in ['OCNRES', 'ICERES']:
+        try:
+            archive.task_config[key] = f"{archive.task_config[key]:03d}"
+        except KeyError as ee:
+            logger.info(f"key ({key}) not found in archive.task_config!")
 
     # Pull out all the configuration keys needed to run the rest of archive steps
     keys = ['ATARDIR', 'current_cycle', 'FHMIN', 'FHMAX', 'FHOUT', 'RUN', 'PDY',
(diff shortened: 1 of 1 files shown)


EricSinsky-NOAA (dismissed):
@aerorahul Thanks for submitting a PR for this issue so quickly. Agreed that `ICERES` will need to be checked as well (we didn't need `ICERES` in the reforecast, which is why it wasn't considered in the reforecast branch). All looks good.

DavidHuber-NOAA (dismissed):
Looks good, thanks for this fix @aerorahul!

aerorahul:
@DavidHuber-NOAA @EricSinsky-NOAA 
Apologies for pushing an update after your review.
I found that if keys are not present, they would be None, which shouldn't happen.

scripts/exglobal_archive.py
```diff
@@ -17,6 +17,14 @@ def main():
     # Instantiate the Archive object
     archive = Archive(config)
 
+    # update these keys to be 3 digits if they are part of archive.task_config.keys
+    if {'OCNRES', 'ICERES'} <= archive.task_config.keys():
```
aerorahul: This will check if the 2 keys are part of the `task_config` dictionary.

scripts/exglobal_archive.py
```diff
@@ -17,6 +17,14 @@ def main():
...
+        for key in ['OCNRES', 'ICERES']:
+            try:
+                archive.task_config[key] = f"{archive.task_config[key]:03d}"
+            except KeyError as ee:
+                logger.info(f"key ({key}) not found in archive.task_config!")
```
aerorahul: This will handle the case if `OCNRES` is present but `ICERES` is not.  Say someone wants to run w/ Ocean but no SeaIce.

EricSinsky-NOAA: Given the above check (`if {'OCNRES', 'ICERES'} <= archive.task_config.keys():`), won't `OCNRES` and `ICERES` always both be present within the if-statement? If either `OCNRES` or `ICERES` are not present, it appears everything within the if-statement will be skipped, right?

aerorahul: correct. if either is present, the if-block evaluates to True and enters here.  Otherwise, its skipped.
The for loop ensures the present ones are covered.

DavidHuber-NOAA: No, if only one is present, then this will evaluate to false and the for block will not execute:

```python
>>> a={"ocnres": 50, "iceres": 10}
>>> b = {"ocnres", "iceres"} <= a.keys()
>>> b
True
>>> a={"ocnres": 50}
>>> b = {"ocnres", "iceres"} <= a.keys()
>>> b
False
```

aerorahul: Thanks for spotting the error @DavidHuber-NOAA 
```python
>>> a={"ocnres": 50, "iceres": 10}
>>> any(map(lambda v: v in a.keys(), ['ocnres', 'iceres']))
True
>>> a={"ocnres": 50}
>>> any(map(lambda v: v in a.keys(), ['ocnres', 'iceres']))
True
```

EricSinsky-NOAA: Agreed @DavidHuber-NOAA. Does that mean we can just change

```
if {'OCNRES', 'ICERES'} <= archive.task_config.keys():
        for key in ['OCNRES', 'ICERES']:
            try:
                archive.task_config[key] = f"{archive.task_config[key]:03d}"
            except KeyError as ee:
                logger.info(f"key ({key}) not found in archive.task_config!")
```

to this:

```
    for key in ['OCNRES', 'ICERES']:
        try:
            archive.task_config[key] = f"{archive.task_config[key]:03d}"
        except KeyError as ee:
            logger.info(f"key ({key}) not found in archive.task_config!")
```

If either `OCNRES` or `ICERES` are not present, or both are not present, `KeyError` will be raised.

EricSinsky-NOAA (approved):
Thanks for the modifications. The changes look good to me.

DavidHuber-NOAA (approved):
LGTM.  Thanks for cleaning up the logging as well.

DavidHuber-NOAA:
Running CI on Hera to test archiving.

WalterKolczynski-NOAA (approved):

PR #3208, opened 2025-01-06
Reinstate product groups

### LINKED ISSUE #2999: rocotorun is too slow for an 11-member 35-day GEFS case
### What is wrong?

When performing a rocotorun for 11-member 35-day GEFS, rocotorun takes an unreasonably long time to execute. This is mostly due to the large number of atmos_prod tasks in a 11-member 35-day run. 

rocotorun should be executed in a reasonable amount of time. 

All or N/A

a0771f849935a55db54f3a4f217d7c67efe97639

### Steps to reproduce

1.  Set `nens `to 10.
2.  Set `FHMAX_GFS `to 840
3.  Set `FHMAX_HF_GFS `to 240
4.  Set `FHOUT_HF_GFS `to 3
5.  Set `FHOUT_GFS` to 6
6. Set up experiment and run GEFS

### Additional information

_No response_

_No response_

### LINKED ISSUE #3210: wave gridded post is triggered on end of forecast and not run when data is available
### What is wrong?

The wave gridded post jobs should be run when data is available versus only run at the end of the forecast job. 

Wave gridded post should be triggered when data is available versus the end of the forecast: https://github.com/NOAA-EMC/global-workflow/blob/develop/workflow/rocoto/gfs_tasks.py#L1193

All or N/A

a branch on top of develop 

### Steps to reproduce

Run a forecast job with waves, once the forecast is running do a rocotocheck, you can see that it's waiting for the forecast job to finish not if the input data is ready: 
```

  dependencies
    SOME is not satisfied
      gfs_fcst_seg0 of cycle 202002130000 is not SUCCEEDED


```

### Additional information

none. 

Trigger job off of data availability and not end of forecast.

WalterKolczynski-NOAA:
# Description
Testing with full-sized GEFS found that the sheer number of tasks overloads rocoto, resulting in `rocotorun` taking over 10 min to complete or hanging entirely. To reduce the number of tasks, product groups are reimplemented so that multiple forecast hour are processed in a single task. However, the implementation is a little different than previously.

The jobs where groups are enabled (atmos_products, oceanice_products, wavepostsbs, atmos_ensstat, and gempak) have a new variable, `MAX_TASKS`, that controls how many groups to use. This setting is currently *per member*. The forecast hours to be processed are then divided into this many groups as evenly as possible without crossing forecast segment boundaries. The walltime for those jobs is then multiplied by the number of times in the largest group. For the gridded wave post job, the dependencies were also updated to trigger off of either the data being available or the appropriate segment completing (the dependencies had not been updated when the job was initially broken into fhrs).

A number of helper methods are added to Tasks to determine these groups and make a standard metatask variable dict in a centralized location. There is also a function to multiply the walltime, but this may be better off relocated to wxflow with the other time functions.

As part of switching from a single value to a list, hours are no longer passed by rocoto as zero-padded values. The lists are comma-delimited (without spaces) and split apart in the job stub (`jobs/rocoto/*`), so each j-job call is still a single forecast hour.

The offline post (upp) job is not broken into groups, since it really isn't used outside the analysis anymore.
Resolves #2999
Resolves #3210

# Type of change

# Change characteristics
  - [ ] EMC verif-global 
  - [ ] GDAS 
  - [ ] GFS-utils 
  - [ ] GSI 
  - [ ] GSI-monitor 
  - [ ] GSI-utils 
  - [ ] UFS-utils 
  - [ ] UFS-weather-model 
  - [ ] wxflow 

# How has this been tested?
- Full-sized GEFS test on Hercules
- Standard forecast-only test on Hercules
- Standard atm-only cycled test on Hercules

# Checklist


workflow/rocoto/tasks.py
```diff
@@ -176,6 +176,203 @@ def _get_forecast_hours(run, config, component='atmos') -> List[str]:
...
+        Parameters
+        ----------
+        hms_timedelta: str
+                       String represnting a time delta in HH:MM:SS format
+        multiplier: int | float
+                    Value to multiply the time delta by
+
+        Returns
+        -------
+        str: String represnting a time delta in HH:MM:SS format
```
KateFriedman-NOAA: Small typo fix:
```suggestion
                       String representing a time delta in HH:MM:SS format
        multiplier: int | float
                    Value to multiply the time delta by

        Returns
        -------
        str: String representing a time delta in HH:MM:SS format
```

workflow/rocoto/tasks.py
```diff
@@ -176,6 +176,203 @@ def _get_forecast_hours(run, config, component='atmos') -> List[str]:
...
+        fhrs: List[int]
+              List of forecast hours to break into groups
+        ngroups: int
+                 Number of groups to split the forecast hours into
+
+        Returns
+        -------
+        dict: Several variable lists for use in rocoto metatasks:
+              fhr_list: list of comma-separated lists of fhr groups
+              fhr_label: list of labels corrsponding to the fhr range
```
KateFriedman-NOAA: Small typo fix:
```suggestion
              fhr_label: list of labels corresponding to the fhr range
```

workflow/rocoto/tasks.py
```diff
@@ -176,6 +176,203 @@ def _get_forecast_hours(run, config, component='atmos') -> List[str]:
...
+        assert Tasks.get_job_groups(fhrs=test_array, ngroups=4) == test_groups
+
+        # Test with a break point that aligns with normal split point
+        test_groups = [{'fhrs': [0, 1, 2, 3, 4, 5], 'seg': 0},
+                       {'fhrs': [6, 7, 8, 9, 10, 11], 'seg': 0},
+                       {'fhrs': [12, 13, 14, 15, 16, 17], 'seg': 1},
+                       {'fhrs': [18, 19, 20, 21, 22, 23], 'seg': 1}]
+        assert Tasks.get_job_groups(fhrs=test_array, ngroups=4, breakpoints=[11]) == test_groups
+
+        # Test with a break point not at a normal splilt point
```
KateFriedman-NOAA: Small typo fix:
```suggestion
        # Test with a break point not at a normal split point
```

workflow/rocoto/gefs_tasks.py
```diff
@@ -308,22 +310,35 @@ def atmos_ensstat(self):
...
         deps = []
-        dep_dict = {'type': 'metatask', 'name': f'gefs_fcst_mem#member#'}
+        dep_dict = {'type': 'task', 'name': f'gefs_fcst_mem#member#_#seg_dep#'}
         deps.append(rocoto.add_dependency(dep_dict))
         dependencies = rocoto.create_dependency(dep=deps)
 
+        fhrs = self._get_forecast_hours('gefs', self._configs['wavepostsbs'], 'wave')
+        is_replay = self._configs['wavepostsbs']['REPLAY_ICS']
+        if is_replay:
+            fhrs = [fhr for fhr in fhrs if fhr not in [0, 1, 2]]
```
aerorahul, line 335: Can you place a comment on what this is doing and why?

workflow/rocoto/tasks.py
```diff
@@ -176,6 +176,203 @@ def _get_forecast_hours(run, config, component='atmos') -> List[str]:
...
+        # Now that we know how many groups each forecast segment should be split into,
+        # Split them and flatten to a single list.
+        groups = []
+        for seg_num, (fhrs_seg, ngroups_seg) in enumerate(zip(fhrs_segs, ngroups_segs)):
+            [groups.append({'fhrs': grp.tolist(), 'seg': seg_num}) for grp in np.array_split(fhrs_seg, ngroups_seg)]
+
+        return groups
+
+    @staticmethod
+    def test_job_groups():
```
aerorahul: Should we create and place the tests in `test_tasks.py`?  How are these tests executed?

aerorahul (commented):
I have no major concerns with this PR.  It looks good and clean to me.  I have not created the XML and looked at it or run any tests.

A few comments inline with the review.

KateFriedman-NOAA (changes requested):
Changes look good, just leaving a few suggested changes to resolve some typos I noticed.

workflow/rocoto/gfs_tasks.py
```diff
@@ -1194,13 +1197,21 @@ def wavepostsbs(self):
         deps.append(rocoto.add_dependency(dep_dict))
         dependencies = rocoto.create_dependency(dep=deps)
```
aerorahul: Lets update the dependency here to launch the forecast hour (group) post-processing sbs job when data is available.  This will resolve #3210

jobs/rocoto/atmos_products.sh
```diff
@@ -15,13 +15,17 @@ if (( status != 0 )); then exit "${status}"; fi
...
-# Execute the JJOB
-###############################################################
-"${HOMEgfs}/jobs/JGLOBAL_ATMOS_PRODUCTS"
+export FORECAST_HOUR
+for FORECAST_HOUR in "${fhr_list[@]}"; do
+	###############################################################
+	# Execute the JJOB
+	###############################################################
+	"${HOMEgfs}/jobs/JGLOBAL_ATMOS_PRODUCTS"
+	status=$?
```
EricSinsky-NOAA, line 28: I recommend moving `jobid` here and naming it to something like the suggestion below. Please see my comment above. 
```suggestion
  jobid="${job}_${FORECAST_HOUR}.$$"
	"${HOMEgfs}/jobs/JGLOBAL_ATMOS_PRODUCTS"
	status=$?
```

jobs/rocoto/atmos_products.sh
```diff
@@ -15,13 +15,17 @@ if (( status != 0 )); then exit "${status}"; fi
 export job="atmos_products"
 export jobid="${job}.$$"
```
EricSinsky-NOAA: If `KEEPDATA` is `YES`, the `DATA` (`DATA=${DATA:-"${DATAROOT}/${jobid}"}`) of the current `FORECAST_HOUR` may overwrite the `DATA` of the previous `FORECAST_HOUR`, which may potentially create issues. To avoid these potential issues, particularly when `KEEPDATA` is needed for debugging, can `jobid` be moved inside the for `FORECAST_HOUR` loop? I think this suggestion also can be applied to `oceanice_products.sh` and `wavepostsbs.sh`. 
```suggestion
```

EricSinsky-NOAA (changes requested):

jobs/rocoto/atmos_products.sh
aerorahul: excellent suggestion @EricSinsky-NOAA

jobs/rocoto/oceanice_products.sh
```diff
@@ -15,11 +15,17 @@ if (( status != 0 )); then exit "${status}"; fi
...
-###############################################################
-# Execute the JJOB
-###############################################################
-"${HOMEgfs}/jobs/JGLOBAL_OCEANICE_PRODUCTS"
+export FORECAST_HOUR
+for FORECAST_HOUR in "${fhr_list[@]}"; do
+	###############################################################
+	# Execute the JJOB
+	###############################################################
+	"${HOMEgfs}/jobs/JGLOBAL_OCEANICE_PRODUCTS"
```
aerorahul, line 27: @EricSinsky-NOAA's suggestion for jobid should be done here as well.  In addition, since rocoto sets `COMPONENT`, may be we can use that as well in `jobid` such that `${job}_${COMPONENT}_${FORECAST_HOUR}.$$`
```suggestion
  jobid="${job}_${COMPONENT}_${FORECAST_HOUR}.$$"
	"${HOMEgfs}/jobs/JGLOBAL_OCEANICE_PRODUCTS"
```

jobs/rocoto/wavepostsbs.sh
```diff
@@ -5,17 +5,24 @@ source "${HOMEgfs}/ush/preamble.sh"
...
-status=$?
-[[ ${status} -ne 0 ]] && exit ${status}
+# shellcheck disable=SC2153
+IFS=', ' read -r -a fhr_list <<< "${FHR_LIST}"
+
+export FHR3
+for FORECAST_HOUR in "${fhr_list[@]}"; do
+	FHR3=$(printf '%03d' "${FORECAST_HOUR}")
+	# Execute the JJOB
+	"${HOMEgfs}/jobs/JGLOBAL_WAVE_POST_SBS"
```
aerorahul, line 23:
```suggestion
  jobid="${job}_${FORECAST_HOUR}.$$"
	"${HOMEgfs}/jobs/JGLOBAL_WAVE_POST_SBS"
```

jobs/rocoto/atmos_products.sh
WalterKolczynski-NOAA: Added these for all the groupified jobs.

workflow/rocoto/gefs_tasks.py
WalterKolczynski-NOAA: Done

workflow/rocoto/tasks.py
WalterKolczynski-NOAA: Right now, they are only executed in an interactive python interpreter by importing the class and then calling the test methods.

Instinctually I like the tests in the same file they are testing, but after doing some cursory research, it seems the preferred way is a separate file, so should probably go in that direction. I can either do it here, or in a future PR.

WalterKolczynski-NOAA:
All reviewer comments other than the location of test methods have been addressed, so I am going to go ahead and run CI for the first machine.

workflow/rocoto/tasks.py
aerorahul: Since tests were added (a good thing), I suggest placing them in the separate file such as `test_tasks.py`.  
It will take a little effort to make them part of a `pytest`, but that can be addressed in another PR.

workflow/rocoto/gefs_tasks.py
```diff
@@ -308,22 +310,43 @@ def atmos_ensstat(self):
         return task
 
     def wavepostsbs(self):
+
```
EricSinsky-NOAA, line 318: In addition to wavepostsbs, should the atmos_ensstat tasks be grouped as well?

WalterKolczynski-NOAA: Yes, also the dependency there is wrong anyway and I need to fix. Not sure how that passed.

WalterKolczynski-NOAA: Changes made and a test is running. It passed the first time because I ran the default case instead of the full-sized one. The default case has fewer output times than the max tasks, so grouping never kicked in.

WalterKolczynski-NOAA: First couple ensstat groups have now run successfully. Will keep test running, but can proceed now.

aerorahul (dismissed):
looks good.  thanks for including the suggestions.

parm/config/gefs/config.resources
```diff
@@ -242,14 +243,15 @@ case ${step} in
     ;;
 
   "atmos_ensstat")
```
EricSinsky-NOAA, line 245:
```suggestion
  "atmos_ensstat")
   # Walltime is per forecast hour; will be multipled by group size
```

EricSinsky-NOAA (dismissed):
Changes overall look good to me. I just included one minor optional suggestion. Thank you for addressing my previous suggestions.

parm/config/gefs/config.resources
WalterKolczynski-NOAA: Added. Thanks for catching that.

WalterKolczynski-NOAA:
Need to look at the gempak jobs a little closer.

KateFriedman-NOAA (approved):
Looks good, thanks @WalterKolczynski-NOAA !

PR #3325, opened 2025-02-13
Fix memory unsetting on Gaea

### LINKED ISSUE #3322: Gaea C6  - wavepostsbs memory variable still in xml
### What is wrong?

For Gaea C6 the [parm/config/gfs/config.resources.GAEAC6 ](https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/config/gfs/config.resources.GAEAC6) removes the memory line because Gaea does not use that.  However, for whatever reason for wavepostsbs Memory is still present in the xml file, which causes errors. 

When generating the xml file for say, C48_S2SW on gaea C6, there should be no memory variable for wavepostsbs

PS - There is not a machine selection for Gaea, so I put all. 

All or N/A

https://github.com/NOAA-EMC/global-workflow/pull/3289 d5813441fa6a4d7e83b682c04972d5e85aa643fd

### Steps to reproduce

git clone https://github.com/NOAA-EMC/global-workflow
cd global-workfow 
git submodule update --init --recursive 
cd sorc 
./build_all.sh gfs 
./link_workflow.sh 
cd ../../
mkdir t01
cd t01 
source ../global_workflow/workflow/gw_setup.sh
 ../global-workflow/workflow/create_experiment.py --yaml ../global-workflow/ci/cases/pr/C48_S2SW.yaml

### Additional information

I was able to change the xml file to remove this line and get the job to go through. However, then jobs that succeeded before were also being resubmitted which was strange. 

Not sure how to get the xml file to generate without the memory line being created. 

Currently I only know how to manually fix this issue.

DavidHuber-NOAA:
# Description
Memory requests are not valid on Gaea, so the variable must be unset when running the setup scripts there.  This fixes memory unsetting when variables like `memory_gdas` are set in config.resources.

Resolves #3322 

# Type of change

# Change characteristics

# How has this been tested?
C48_S2SW on Gaea C6

# Checklist

### DIFF
--- a/parm/config/gfs/config.resources.GAEAC5
+++ b/parm/config/gfs/config.resources.GAEAC5
@@ -5,7 +5,4 @@
 export FI_VERBS_PREFER_XRC=0
 
 unset memory
-# shellcheck disable=SC2312
-for mem_var in $(env | grep '^memory_' | cut -d= -f1); do
-  unset "${mem_var}"
-done
+unset "memory_${RUN}"
--- a/parm/config/gfs/config.resources.GAEAC6
+++ b/parm/config/gfs/config.resources.GAEAC6
@@ -3,7 +3,4 @@
 # GaeaC6-specific job resources
 
 unset memory
-# shellcheck disable=SC2312
-for mem_var in $(env | grep '^memory_' | cut -d= -f1); do
-  unset "${mem_var}"
-done
+unset "memory_${RUN}"


KateFriedman-NOAA (approved):
Looks good and fixes the issue. Thanks @DavidHuber-NOAA !

aerorahul (approved):
thanks for the quick fix @DavidHuber-NOAA

Issue #3318, opened 2025-02-12
Check if the HOMEDIR, STMP, and PTMP are writable at setup time


DavidHuber-NOAA:
### What new functionality do you need?

When setting up an experiment, in particular at `setup_xml.py` time, a check should be added to verify that the user running the experiment is able to write to those directories.  This will save debugging time if a user forgets/doesn't know to change these directories.

### What are the requirements for the new functionality?

`setup_xml.py` will fail if it is not possible to create/write to `HOMEDIR`, `STMP`, or `PTMP`.  Bonus points if other read directories are also checked (such as `rstprod` directories).

### Acceptance Criteria

- [ ] `setup_xml.py` runs without error when `HOMEDIR`, `STMP`, and `PTMP` are writable
- [ ] `setup_xml.py` raises an error if these directories are not writeable
- [ ] A test is written and incorporated into CI testing to verify this feature works

DavidHuber-NOAA:
FYI @wx20jjung

mingshichen-noaa:
@WalterKolczynski-NOAA @DavidHuber-NOAA 
Hi both,
This issue is interesting.  I would like to resolve the issue.

DavidHuber-NOAA:
Resolved via https://github.com/NOAA-EMC/global-workflow/pull/3430.

PR #3426, opened 2025-03-07
Chang permission modes of HOMEDIR STMP PTMP as writable at setup time

mingshichen-noaa:
# Description
When setting up an experiment, in particular at setup_xml.py/setup_expt.py time, a check should be added to verify that the user running the experiment is able to write to those directories. This will save debugging time if a user forgets/doesn't know to change these directories.
Refs https://github.com/NOAA-EMC/global-workflow/issues/3318

# Type of change

# Change characteristics

# How has this been tested?
 
- Clone and build on RDHPCS
- Cycled tests on Hercules
- Forecast-only tests on Hercules
 
# Checklist

### DIFF
--- a/workflow/setup_expt.py
+++ b/workflow/setup_expt.py
@@ -29,6 +29,8 @@ def makedirs_if_missing(dirname):
     if not os.path.exists(dirname):
         os.makedirs(dirname)
 
+    return
+
 def chmod_dir_permission(dirname):
     """
     change the permission mode of a directory as writable
@@ -44,6 +46,7 @@ def chmod_dir_permission(dirname):
     except Exception as e:
         print(f"An unexpected error occurred: {e}")
 
+    return
 
 def fill_expdir(inputs):
     """
--- a/workflow/setup_expt.py
+++ b/workflow/setup_expt.py
@@ -5,6 +5,7 @@ Entry point for setting up an experiment in the global-workflow
 """
 
 import os
+import stat
 import glob
 import shutil
 from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter, SUPPRESS, ArgumentTypeError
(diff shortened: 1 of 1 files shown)


DavidHuber-NOAA:
@mingshichen-noaa A few things about this PR/issue #3318:

1. `setup_expt.py` sets the defaults for `STMP`, `PTMP`, and `HOMEDIR`, but these can be overridden by the user before running `setup_xml.py`.  So we don't want to check if they can be written to during the `setup_expt.py` step, but instead the `setup_xml.py` step.
2. All three of these directories may or may not exist at the time `setup_xml.py` is called.  So determining if they can be written to is a little tricky.  What I would suggest is something like (I am guessing at function names here)
```
def check_dir_writable(dir)
    if os.path.is_dir(dir):
        if os.access(directory_path, os.W_OK):
            return True
        else:
            return False
    elif os.path.is_file(dir):
         # raise exception
    else:  # Find the nearest parent directory that already exists
        test_parent = os.path.dirname(dir)
        while test_parent:
            if os.path.exists(test_parent):
                # Call check_dir_writable on the parent
                return check_dir_writable(test_parent)
            test_parent = os.path.dirname(test_parent)
       # If we exited the loop, then something went wrong with my logic
       # raise exception 'Could not find a parent directory of {dir}'
```
If this returns `True`, then the directory is writable, otherwise, it's not and you can raise an exception notifying the user.
3. Lastly, `STMP`, `PTMP`, and `HOMEDIR` should be read from `config.base`, not the host file.

I'd suggest closing this PR and reworking the issue on the setup_xml side of things.

mingshichen-noaa:
@DavidHuber-NOAA 
Thank you for your comments. I will close the PR.

PR #3462, opened 2025-03-17
Make marine LETKF optional

shlyaeva:
# Description

Make marine LETKF optional (default v17 behavior) to save computational resources when it doesn't need to be run.
GFSv17 yaml is changed to not run LETKF. GDASApp low-res yaml is set up to run LETKF as it does on develop now (I think we should keep testing this).

Resolves part of https://github.com/NOAA-EMC/GDASApp/issues/1544

# Type of change

# Change characteristics

# How has this been tested?
Tested with low-res GDASApp C48mx500_hybAOWCDA tests with LETKF on (like in this PR) and LETKF off (manually changed and run experiments). The behavior is as expected (LETKF run in the first case and not run in the second case); 1.5 cycle successfully run and the output is as expected. All run on hera with intel.

# Checklist


scripts/exglobal_archive_tars.py
```diff
@@ -37,7 +37,7 @@ def main():
             'FHOUT_HF_WAV', 'FHMAX_WAV', 'FHMAX_HF_WAV', 'FHMAX_WAV_GFS',
             'restart_interval_gdas', 'restart_interval_gfs', 'DO_ARCHCOM',
             'DO_AERO_ANL', 'DO_AERO_FCST', 'DO_CA', 'DOIBP_WAV', 'DO_JEDIOCNVAR', 'DOHYBVAR_OCN',
-            'NMEM_ENS', 'DO_JEDIATMVAR', 'FHMAX_FITS', 'waveGRD',
+            'DOLETKF_OCN', 'NMEM_ENS', 'DO_JEDIATMVAR', 'FHMAX_FITS', 'waveGRD',
```
DavidHuber-NOAA, line 40: I do not see where `DOLETKF_OCN` is used by the archiving YAMLs.  Are there files that need to be archived?

aerorahul (approved):
looks good to me.
Q. Should `DOLETKF_OCN` be conditioned on `DOHYBVAR_OCN`?  i.e. Can `DOLETKF_OCN=YES` when `DOHYBVAR_OCN=NO`?

shlyaeva:
> Q. Should DOLETKF_OCN be conditioned on DOHYBVAR_OCN? i.e. Can DOLETKF_OCN=YES when DOHYBVAR_OCN=NO?

It's a little complicated. DOHYBVAR_OCN may be a misnomer: from what I find in the code it is used to decide whether to do ensemble recentering, and what the bmatrix task dependencies are (it can process ensembles among other things). In gfs v17 we plan to do ensemble recentering for marine without using LETKF, so in this respect the options are not conditioned on each other.

guillaumevernieres (approved):
:+1:

DavidHuber-NOAA (changes requested):

scripts/exglobal_archive_tars.py
shlyaeva: There's currently no archiving for marine LETKF, but we'll definitely be adding it in the future. I can remove this change though if you prefer, let me know.

DavidHuber-NOAA: OK, that's fine to leave it as a placeholder.  A note for the future, if there will be items that need to be archived from the ensemble members, then this flag will also need to be added to `scripts/exgdas_enkf_earc_tars.py` as well.  No need to change presently.  Thanks!

DavidHuber-NOAA (approved):

scripts/exglobal_archive_tars.py
shlyaeva: Got it, thank you!

PR #28, opened 2020-03-05
Issue #8 - add switch for MET+ jobs

KateFriedman-NOAA:

PR #3472, opened 2025-03-18
add capability to dynamically generate GSI *info files

### LINKED ISSUE #3293: add the capabilty to dynamically generate GSI satinfo, convinfo ozinfo files for reanalysis

Currently config.anal has an if/then/else loop that sets env vars SATINFO, OZINFO, CONVINFO that contain paths to different gsi info files  for different periods over the last several years.  For reanalysis, we need a solution that works back to 1979.  Rather than add to the if/then/else block in config.anal, I have created a set of scripts that generate *info files dynamically given a date (https://github.com/NOAA-PSL/build_gsinfo).  What is needed is a way to use those scripts in global-workflow.   Note this is only needed for GSI - for JEDI the needed functionality is included in observation_chronicle (https://github.com/NOAA-EMC/jcb-gdas/tree/develop/observation_chronicle/atmosphere).

the global-workflow can dynamically generate and use GSI satinfo, ozinfo and radinfo files based on the analysis date.

### Acceptance Criteria

- the workflow dynamically generates the correct satinfo/ozinfo/convinfo files when using GSI and uses them

Create a new RUN_ENVIR ('reanl') and have the [build_gsinfo](https::/github.com/NOAA-PSL/build_gsinfo) scripts generate satinfo, ozinfo and convinfo files based on ${PDY}${cyc}

```
diff --git a/parm/config/gfs/config.anal b/parm/config/gfs/config.anal
index 123bd6de..4b56e515 100644
--- a/parm/config/gfs/config.anal
+++ b/parm/config/gfs/config.anal
@@ -146,6 +146,11 @@ if [[ ${RUN_ENVIR} == "emc" ]]; then
   #   if [[ "${PDY}${cyc}" -ge "2021110312" && "${PDY}${cyc}" -lt "YYYYMMDDHH" ]]; then
   #     export SATINFO=${FIXgfs}/gsi/gfsv16_historical/global_satinfo.txt.2021110312
   #   fi
+elif [[ ${RUN_ENVIR} == "reanl" ]]; then
+   # these will be set based on date in scripts/exglobal_atmos_analysis.sh
+   export SATINFO="generate"
+   export CONVINFO="generate"
+   export OZINFO="generate"
 fi

 # Flag to turn on (.true.) or off (.false.) the infrared cloud and aerosol detection software

diff --git a/scripts/exglobal_atmos_analysis.sh b/scripts/exglobal_atmos_analysis.sh
index bd11b81c..b549e273 100755
--- a/scripts/exglobal_atmos_analysis.sh
+++ b/scripts/exglobal_atmos_analysis.sh
@@ -363,14 +363,26 @@ cd ${DATA} || exit 99
 # Fixed files
 ${NLN} ${BERROR}       berror_stats
 ${NLN} ${SATANGL}      satbias_angle
-${NLN} ${SATINFO}      satinfo
+if [ $SATINFO == "generate" ]; then
+   sh ${HOMEgfs}/sorc/build_gsinfo/create_satinfo.sh  ${PDY}${cyc} > satinfo
+else
+   ${NLN} ${SATINFO}      satinfo
+fi
 ${NLN} ${RADCLOUDINFO} cloudy_radiance_info.txt
 ${NLN} ${ATMSFILTER}   atms_beamwidth.txt
 ${NLN} ${ANAVINFO}     anavinfo
-${NLN} ${CONVINFO}     convinfo
+if [ $CONVINFO == "generate" ]; then
+   sh ${HOMEgfs}/sorc/build_gsinfo/create_convinfo.sh  ${PDY}${cyc} > convinfo
+else
+   ${NLN} ${CONVINFO}     convinfo
+fi
 ${NLN} ${vqcdat}       vqctp001.dat
 ${NLN} ${INSITUINFO}   insituinfo
-${NLN} ${OZINFO}       ozinfo
+if [ $OZINFO == "generate" ]; then
+   sh ${HOMEgfs}/sorc/build_gsinfo/create_ozinfo.sh  ${PDY}${cyc} > ozinfo
+else
+   ${NLN} ${OZINFO}       ozinfo
+fi
 ${NLN} ${PCPINFO}      pcpinfo
 ${NLN} ${AEROINFO}     aeroinfo
 ${NLN} ${SCANINFO}     scaninfo
```

jswhit2:
# Description

  Currently config.anal has an if/then/else loop that sets env vars SATINFO, OZINFO, CONVINFO that contain paths to different gsi info files for different periods over the last several years. For reanalysis, we need a solution that works back to 1979. Rather than add to the if/then/else block in config.anal, I have created a set of scripts that generate *info files dynamically given a date (https://github.com/NOAA-PSL/build_gsinfo). What is needed is a way to use those scripts in global-workflow. Note this is only needed for GSI - for JEDI the needed functionality is included in observation_chronicle (https://github.com/NOAA-EMC/jcb-gdas/tree/develop/observation_chronicle/atmosphere).

Resolves https://github.com/NOAA-EMC/global-workflow/issues/3293

Requires https://github.com/NOAA-EMC/GSI-fix/pull/28

Enabled via USE_BUILD_GSINFO env var that can be set to YES in config.base (default is NO).
3 new scripts added (create_satinfo.sh, create_ozinfo.sh, create_convinfo.sh).  These scripts generate the GSI *info for a given analysis date using data from build_gsinfo (which will live inside GSI-fix once https://github.com/NOAA-EMC/GSI-fix/pull/28 is merged). 

The OBS_INPUT table in the GSI namelist is removed from exglobal_atmos_analysis.sh to allow for separate options for NCEP ops and reanalysis.  Both versions are included as text files in [build_gsinfo/obs_input](https://github.com/NOAA-PSL/build_gsinfo/tree/main/obs_input).  The OBS_INPUT env var can be used to choose which version do use.  The NCEP ops version is the default in config.anal.   

A workaround for https://github.com/NOAA-EMC/GSI/issues/752 is included in exglobal_atmos_analysis.sh (pointing to a separate directory for HIRS coefficient files).  This hack can be removed once https://github.com/NOAA-EMC/GSI/issues/783 is merged.

# Type of change

# Change characteristics
- Is this a breaking change (a change in existing functionality)?NO
- Does this change require an update to any of the following submodules? YES
 
  - [x] GSI-fix https://github.com/NOAA-EMC/GSI-fix/pull/28

# How has this been tested?

- Clone and build on gaeac6 and hera
- Cycled test on gaeac6 and hera

# Checklist
- [ ] This change is covered by an existing CI test or a new one has been added (to be added later)


jswhit2:
@jack-woollen @jswhit2  Jeff I didn't figure out how to update your branch with the convinfo file containing complete satwnd definitions. Instead you can copy from /work2/noaa/da/jwoollen/RAEXPS/scripts/2021ozn1/build_gsinfo/convinfo/merged_convinfo.txt.

parm/config/gfs/config.anal
```diff
@@ -148,6 +148,17 @@ if [[ ${RUN_ENVIR} == "emc" ]]; then
...
+if [[ ${USE_BUILD_GSINFO} == "YES" ]]; then
+   # these will be set based on date in scripts/exglobal_atmos_analysis.sh
+   export SATINFO="generate"
+   export CONVINFO="generate"
+   export OZINFO="generate"
+   export OBS_INPUT=${FIXgfs}/gsi/build_gsinfo/obs_input/obs_input_reanl_nasa_ozone.txt
+else
+   export OBS_INPUT=${FIXgfs}/gsi/build_gsinfo/obs_input/obs_input_ops.txt
+fi
+export HIRS_FIX=/gpfs/f6/drsa-precip4/proj-shared/Jeffrey.S.Whitaker/hirs_fix
```
CoryMartin-NOAA: noting a hard coded Gaea path here

scripts/create_convinfo.sh
```diff
@@ -0,0 +1,18 @@
+# create global_ozinfo file for a given date
```
CoryMartin-NOAA: Need a shebang defined here?

scripts/create_ozinfo.sh
```diff
@@ -0,0 +1,27 @@
+# create global_ozinfo file for a given date
```
CoryMartin-NOAA: Another shebang needed

CoryMartin-NOAA: Generally: These new scripts likely need to go in `ush/` and need the style issues addressed

parm/config/gfs/config.anal
jswhit2: Thanks @CoryMartin-NOAA - I think this needs to default to the CRTM_FIX directory and be configurable in config.anal.  We won't need any of this when CRTM is updated though (hopefully).

ClaraDraper-NOAA (changes requested):
@jswhit2 To include the soil analysis in the reanalysis / scout runs, we'll need to make changes to the convinfo and anavinfo. For the NRT system, those changes are here. 

Can you please add these changes into your [conv/anav]info files? I'm not sure if it's better to include them always, or as an option.

jswhit:
@ClaraDraper-NOAA I think you forgot to include the link.   Would you be able to tell us specifically which lines to add to the. reanalysis version of convinfo (https://github.com/NOAA-PSL/build_gsinfo/blob/main/convinfo/merged_convinfo.txt)?  The anavinfo changes won't be needed for the scout runs until we start running ensemble DA.

ClaraDraper-NOAA:
> @jswhit2 To include the soil analysis in the reanalysis / scout runs, we'll need to make changes to the convinfo and anavinfo. For the NRT system, those changes are here.
> 
> Can you please add these changes into your [conv/anav]info files? I'm not sure if it's better to include them always, or as an option.

I did forget the link! It's [here](https://github.com/NOAA-EMC/global-workflow/blob/6956eebebc68d1270e87284fc86decbe4fd3b008/parm/config/gfs/config.anal#L51)

ClaraDraper-NOAA:
If we want to assimilate 2m obs into the scout run (maybe not a great idea???), to see the changes you need, in fix/gsi/ 

>diff  global_convinfo_2mObs.txt global_convinfo.txt

If we just want to monitor them, you can leave convinfo as is. For monitoring or assim, we need to add t2m and q2m to the met_guess and state_derivative namelists in anavinfo. There's an example here on hera: 

/scratch2/BMC/gsienkf/Clara.Draper/gerrit-hera/global-workflow_CNTRL/fix/global_anavinfo_2mDiag.l127.txt

jswhit:
OK got it - thanks @ClaraDraper-NOAA

DavidHuber-NOAA:
Moving to draft while upstream issues are addressed.  @jswhit feel free to re-mark this as ready for review when it is so.

scripts/exglobal_atmos_analysis.sh
```diff
@@ -359,14 +367,26 @@ cd ${DATA} || exit 99
 # Fixed files
 ${NLN} ${BERROR}       berror_stats
 ${NLN} ${SATANGL}      satbias_angle
-${NLN} ${SATINFO}      satinfo
+if [[ ${SATINFO} == "generate" ]]; then
+   sh ${USHgfs}/create_satinfo.sh  ${PDY}${cyc} > satinfo
```
DavidHuber-NOAA: Prefer that the `create_satinfo.sh` file be called directly.  If it isn't already, that script should be made executable.

```suggestion
   ${USHgfs}/create_satinfo.sh  ${PDY}${cyc} > satinfo
```

scripts/exglobal_atmos_analysis.sh
```diff
@@ -359,14 +367,26 @@ cd ${DATA} || exit 99
...
+else
+   ${NLN} ${SATINFO}      satinfo
+fi
 ${NLN} ${RADCLOUDINFO} cloudy_radiance_info.txt
 ${NLN} ${ATMSFILTER}   atms_beamwidth.txt
 ${NLN} ${ANAVINFO}     anavinfo
-${NLN} ${CONVINFO}     convinfo
+if [[ ${CONVINFO} == "generate" ]]; then
+   sh ${USHgfs}/create_convinfo.sh  ${PDY}${cyc} > convinfo
```
DavidHuber-NOAA:
```suggestion
   ${USHgfs}/create_convinfo.sh  ${PDY}${cyc} > convinfo
```

scripts/exglobal_atmos_analysis.sh
```diff
@@ -359,14 +367,26 @@ cd ${DATA} || exit 99
...
+else
+   ${NLN} ${CONVINFO}     convinfo
+fi
 ${NLN} ${vqcdat}       vqctp001.dat
 ${NLN} ${INSITUINFO}   insituinfo
-${NLN} ${OZINFO}       ozinfo
+if [[ ${OZINFO} == "generate" ]]; then
+   sh ${USHgfs}/create_ozinfo.sh  ${PDY}${cyc} > ozinfo
```
DavidHuber-NOAA:
```suggestion
   ${USHgfs}/create_ozinfo.sh  ${PDY}${cyc} > ozinfo

File: ush/create_satinfo.sh [hunk 7]
```diff
@@ -0,0 +1,27 @@
...
+    # find matching date
+    usedate=""
+    for f in "${sat}"/*; do
+        if [[ ${f} != "${sat}/readme" ]]; then # skip readme file
+           datex=$(basename "${f}")
+	   if [[ ${date} -ge ${datex} ]]; then
+              usedate=${datex}
+           fi
+        fi
+    done
```
DavidHuber-NOAA: Please use consistent white spacing and avoid `tab` characters.

ush/create_satinfo.sh
```diff
@@ -0,0 +1,27 @@
...
+    # cat matching date file, or quit if date not found
+    if [[ ${usedate} != "" ]]; then
+        cat "${sat}/${usedate}" || exit 1
+    else
+        echo "date not found for ${sat}"
+	exit 1
```
DavidHuber-NOAA: Same

DavidHuber-NOAA (changes requested):
A few minor nit-picks.  There are also a number of shell-check issues noted here: https://github.com/NOAA-EMC/global-workflow/actions/runs/14601534590/job/40960623756?pr=3472.  Could you please also address these?

scripts/exglobal_atmos_analysis.sh
jswhit2: done

scripts/exglobal_atmos_analysis.sh
jswhit2: done

scripts/exglobal_atmos_analysis.sh
jswhit2: done

ush/create_satinfo.sh
jswhit2: done

ush/create_satinfo.sh
jswhit2: done

jswhit2:
> A few minor nit-picks. There are also a number of shell-check issues noted here: https://github.com/NOAA-EMC/global-workflow/actions/runs/14601534590/job/40960623756?pr=3472. Could you please also address these?

@DavidHuber-NOAA those shell-check errors do not appear to coming from lines I modified so I hesitate to change them.

RussTreadon-NOAA:
@jswhit2 :  GSI-fix PR [#28](https://github.com/NOAA-EMC/GSI-fix/pull/28) has been merged into GSI-fix `develop`.  Done at [35c06c0](https://github.com/NOAA-EMC/GSI-fix/commit/35c06c041d144bee42f7b744ef3e3718b17b3a5d)

ush/create_convinfo.sh
```diff
@@ -0,0 +1,19 @@
...
+#      usedate=$datex
+#   fi
+#done
+## cat matching date file, or quit if date not found
+#if [[ ${usedate} != "" ]]; then
+#    cat ${usedate}
+#else
+#    echo "date not found"
+#    exit 1
+#fi
```
aerorahul: Just checking here -- is this intended as is or should it be generating a convinfo file based on `date`

ush/create_ozinfo.sh
```diff
@@ -0,0 +1,29 @@
+#! /usr/bin/env bash
+# create global_ozinfo file for a given date
+date=$1
```
aerorahul:
```suggestion
date_in=$1
```
Please use a different name for this variable so that it does not conflict with the linux utility `date`

ush/create_ozinfo.sh
```diff
@@ -0,0 +1,29 @@
...
+# loop over satellites
+cd "${BUILD_GSINFO_DIR}/ozinfo" || exit 1
+grep -v '^ *#' satellites | while IFS= read -r sat 
+do
+    # find matching date
+    usedate=""
+    for f in "${sat}"/*; do
+        if [[ ${f} != "${sat}/readme" ]]; then # skip readme file
+           datex=$(basename "${f}")
+           if [[ ${date} -ge ${datex} ]]; then
```
aerorahul:
```suggestion
           if [[ ${date_in} -ge ${datex} ]]; then
```

ush/create_satinfo.sh
```diff
@@ -0,0 +1,27 @@
+#! /usr/bin/env bash
+# create global_satinfo file for a given date
+date=$1
```
aerorahul:
```suggestion
date_in=$1
```

ush/create_satinfo.sh
```diff
@@ -0,0 +1,27 @@
...
+# loop over satellites
+cd "${BUILD_GSINFO_DIR}/satinfo" || exit 1
+grep -v '^ *#' satellites | while IFS= read -r sat 
+do
+    # find matching date
+    usedate=""
+    for f in "${sat}"/*; do
+        if [[ ${f} != "${sat}/readme" ]]; then # skip readme file
+           datex=$(basename "${f}")
+           if [[ ${date} -ge ${datex} ]]; then
```
aerorahul:
```suggestion
           if [[ ${date_in} -ge ${datex} ]]; then
```

scripts/exglobal_diag.sh
```diff
@@ -83,8 +83,13 @@ if [[ "${GENDIAG}" == "YES" ]] ; then
 
    diagtype[0]="conv conv_gps conv_ps conv_pw conv_q conv_sst conv_t conv_tcp conv_uv conv_spd"
    diagtype[1]="pcp_ssmi_dmsp pcp_tmi_trmm"
+   if [[ ${USE_BUILD_GSINFO} == "YES" ]]; then
+   diagtype[2]=$(cat ${BUILD_GSINFO_DIR}/ozinfo/satellites)
+   diagtype[3]=$(cat ${BUILD_GSINFO_DIR}/satinfo/satellites)
```
aerorahul:
```suggestion
     diagtype[2]=$(cat ${BUILD_GSINFO_DIR}/ozinfo/satellites)
     diagtype[3]=$(cat ${BUILD_GSINFO_DIR}/satinfo/satellites)
```

dev/parm/config/gfs/config.anal
```diff
@@ -148,6 +148,18 @@ if [[ ${RUN_ENVIR} == "emc" ]]; then
   #   fi
 fi
 
+if [[ ${USE_BUILD_GSINFO} == "YES" ]]; then
+   # these will be set based on date in scripts/exglobal_atmos_analysis.sh
+   export SATINFO="generate"
+   export CONVINFO="generate"
+   export OZINFO="generate"
+   export OBS_INPUT="${BUILD_GSINFO_DIR}/obs_input/obs_input_reanl_nasa_ozone.txt"
```
aerorahul: Is there a reason why this cannot be used for the various date-specific if statements above?  I was under the impression, this feature would eliminate them.

aerorahul (changes requested):
Some suggestions and a couple of questions.

ush/create_convinfo.sh
jswhit: currently we're pointing to a single convinfo file, that's why the date matching code is commented out

dev/parm/config/gfs/config.anal
jswhit: it could, but we would have to create a build_gsinfo branch that exactly matches the choices in those case statements.  For reanalysis we made different choices (mostly to use platforms earlier than was done in operations).

DavidHuber-NOAA:
> > A few minor nit-picks. There are also a number of shell-check issues noted here: https://github.com/NOAA-EMC/global-workflow/actions/runs/14601534590/job/40960623756?pr=3472. Could you please also address these?
> 
> @DavidHuber-NOAA those shell-check errors do not appear to coming from lines I modified so I hesitate to change them.

@jswhit2 I have addressed the shellcheck issues in `exglobal_atmos_analysis.sh` and `exglobal_diag.sh`.

ush/create_ozinfo.sh
```diff
@@ -0,0 +1,29 @@
...
+        fi
+    done
+    # cat matching date file, or quit if date not found
+    if [[ ${usedate} != "" ]]; then
+        cat "${sat}/${usedate}" || exit 1
+    else
+        echo "date not found for ${sat}"
+        exit 1
+    fi
+done
```
DavidHuber-NOAA: The output of the `grep` command should be passed into the loop so that error codes are not masked:

```suggestion
while IFS= read -r sat 
do
    # find matching date
    usedate=""
    for f in "${sat}"/*; do
        if [[ ${f} != "${sat}/readme" ]]; then # skip readme file
           datex=$(basename "${f}")
           if [[ ${date_in} -ge ${datex} ]]; then
              usedate=${datex}
           fi
        fi
    done
    # cat matching date file, or quit if date not found
    if [[ ${usedate} != "" ]]; then
        cat "${sat}/${usedate}" || exit 1
    else
        echo "date not found for ${sat}"
        exit 1
    fi
done < <(grep -v '^ *#' satellites)
```

ush/create_satinfo.sh
```diff
@@ -0,0 +1,27 @@
...
+        fi
+    done
+    # cat matching date file, or quit if date not found
+    if [[ ${usedate} != "" ]]; then
+        cat "${sat}/${usedate}" || exit 1
+    else
+        echo "date not found for ${sat}"
+        exit 1
+    fi
+done
```
DavidHuber-NOAA:
```suggestion
while IFS= read -r sat 
do
    # find matching date
    usedate=""
    for f in "${sat}"/*; do
        if [[ ${f} != "${sat}/readme" ]]; then # skip readme file
           datex=$(basename "${f}")
           if [[ ${date_in} -ge ${datex} ]]; then
              usedate=${datex}
           fi
        fi
    done
    # cat matching date file, or quit if date not found
    if [[ ${usedate} != "" ]]; then
        cat "${sat}/${usedate}" || exit 1
    else
        echo "date not found for ${sat}"
        exit 1
    fi
done < <(grep -v '^ *#' satellites)

File: ush/create_convinfo.sh [hunk 18]
```diff
@@ -0,0 +1,33 @@
+#! /usr/bin/env bash
+# create global_convinfo file for a given date
+date=$1
+# 2m ob use file (DO_GSISOILDA)
+if [ -z ${2} ]; then
```
DavidHuber-NOAA: Prefer testing the argument count first:
```suggestion
if [[ $# -ge 2 && -z "${2}" ]]; then
```

ush/create_convinfo.sh
```diff
@@ -0,0 +1,33 @@
...
+   use2mobs="NO" # NO if not present
+else
+   use2mobs=$2
+fi
+cd "${BUILD_GSINFO_DIR}/convinfo" || exit
+usedate=""
+for datex in [1-2]*; do
+   if [[ ${date} -ge ${datex} ]]; then
+      usedate=$datex
```
DavidHuber-NOAA:
```suggestion
      usedate=${datex}

File: ush/create_convinfo.sh [hunk 20]
```diff
@@ -0,0 +1,33 @@
...
+# cat matching date file, or quit if date not found
+if [[ ${usedate} != "" ]]; then
+    if [[ ${use2mobs} == "YES" ]]; then
+        # turn on 2m t,q obs over land
+        sed -e "s/t        181    0   -1/t        181    0    1/g" \
+            -e "s/t        183    0   -1/t        183    0    1/g" \
+            -e "s/t        187    0   -1/t        187    0    1/g" \
+            -e "s/q        181    0   -1/q        181    0    1/g" \
+            -e "s/q        183    0   -1/q        183    0    1/g" \
+            -e "s/q        187    0   -1/q        187    0    1/g" ${usedate}
```
DavidHuber-NOAA:
```suggestion
            -e "s/q        187    0   -1/q        187    0    1/g" "${usedate}"

File: ush/create_convinfo.sh [hunk 21]
```diff
@@ -0,0 +1,33 @@
...
+    else
+        cat ${usedate}
```
DavidHuber-NOAA:
```suggestion
        cat "${usedate}"

File: ush/create_ozinfo.sh [hunk 22]
```diff
@@ -0,0 +1,29 @@
...
+        fi
+    done
+    # cat matching date file, or quit if date not found
+    if [[ ${usedate} != "" ]]; then
+        cat "${sat}/${usedate}" || exit 1
+    else
+        echo "date not found for ${sat}"
+        exit 1
+    fi
+done < <(grep -v '^ *#' satellites)
```
DavidHuber-NOAA:
```suggestion
done < <(grep -v '^ *#' satellites || true)

File: ush/create_satinfo.sh [hunk 23]
```diff
@@ -0,0 +1,27 @@
...
+        fi
+    done
+    # cat matching date file, or quit if date not found
+    if [[ ${usedate} != "" ]]; then
+        cat "${sat}/${usedate}" || exit 1
+    else
+        echo "date not found for ${sat}"
+        exit 1
+    fi
+done < <(grep -v '^ *#' satellites)
```
DavidHuber-NOAA:
```suggestion
done < <(grep -v '^ *#' satellites || true)
```

DavidHuber-NOAA (commented):
Just some suggestions to move through the shellcheck issues.

scripts/exglobal_atmos_analysis.sh
```diff
@@ -687,22 +725,22 @@ EOFunzip
       else
          fname=$(echo ${diag_file} | cut -d'.' -f1)
```
DavidHuber-NOAA:
```suggestion
         fname=$(echo "${diag_file}" | cut -d'.' -f1)
```

scripts/exglobal_atmos_analysis.sh
```diff
@@ -687,22 +725,22 @@ EOFunzip
...
          date=$(echo ${diag_file} | cut -d'.' -f2)
```
DavidHuber-NOAA:
```suggestion
         date=$(echo "${diag_file}" | cut -d'.' -f2)
```

scripts/exglobal_atmos_analysis.sh
```diff
@@ -92,12 +93,13 @@ OSCATBF=${OSCATBF:-${COMIN_OBS}/${OPREFIX}oscatw.tm00.bufr_d${OSUFFIX}}
 RAPIDSCATBF=${RAPIDSCATBF:-${COMIN_OBS}/${OPREFIX}rapidscatw.tm00.bufr_d${OSUFFIX}}
 GSNDBF=${GSNDBF:-${COMIN_OBS}/${OPREFIX}goesnd.tm00.bufr_d${OSUFFIX}}
 GSNDBF1=${GSNDBF1:-${COMIN_OBS}/${OPREFIX}goesfv.tm00.bufr_d${OSUFFIX}}
-#B1HRS2=${B1HRS2:-${COMIN_OBS}/${OPREFIX}1bhrs2.tm00.bufr_d${OSUFFIX}} # HIRS temporarily disabled due to CRTM versioning issues
+B1HRS2=${B1HRS2:-${COMIN_OBS}/${OPREFIX}1bhrs2.tm00.bufr_d${OSUFFIX}} # HIRS temporarily disabled due to CRTM versioning issues
 B1MSU=${B1MSU:-${COMIN_OBS}/${OPREFIX}1bmsu.tm00.bufr_d${OSUFFIX}}
-#B1HRS3=${B1HRS3:-${COMIN_OBS}/${OPREFIX}1bhrs3.tm00.bufr_d${OSUFFIX}} # HIRS temporarily disabled due to CRTM versioning issues
-#B1HRS4=${B1HRS4:-${COMIN_OBS}/${OPREFIX}1bhrs4.tm00.bufr_d${OSUFFIX}} # HIRS temporarily disabled due to CRTM versioning issues
+B1HRS3=${B1HRS3:-${COMIN_OBS}/${OPREFIX}1bhrs3.tm00.bufr_d${OSUFFIX}} # HIRS temporarily disabled due to CRTM versioning issues
+B1HRS4=${B1HRS4:-${COMIN_OBS}/${OPREFIX}1bhrs4.tm00.bufr_d${OSUFFIX}} # HIRS temporarily disabled due to CRTM versioning issues
```
CoryMartin-NOAA: Since these are uncommented out, can we remove the inline comments?

ush/create_gsi_info.sh
```diff
@@ -0,0 +1,129 @@
...
+    if [[ ${date_in} -ge ${datex} ]]; then
+      usedate=${datex}
+    fi
+  done
+
+  echo "${usedate}"
+}
+
+# Get the starting directory
+starting_dir="${DATA:-${PWD}}"
```
CoryMartin-NOAA: Something to consider, I think it would be cleaner/more explicit if the script took in as an argument where it was writing out files to rather than it being implicitly $DATA. Just my $0.02

CoryMartin-NOAA (commented):
Few minor comments, but things look good overall (have not tested)

scripts/exglobal_atmos_analysis.sh
jswhit: done

jswhit:
To reproduce the gfsv17_historical GSI *info files, the build_gsinfo dir in GSI-fix will need to be updated to match https://github.com/NOAA-PSL/build_gsinfo-fix/pull/3.  It will not reproduce them bit for bit, since I've turned off some instruments that no longer exist, but are still turned on in satinfo (like amsua_metop-a and amsua_aqua).

ush/create_gsi_info.sh
```diff
@@ -0,0 +1,129 @@
...
+date_in=${2:-}
+use2mobs=${3:-NO}
+
+if [[ -z "${type_in}" || -z "${date_in}" ]]; then
+  echo "Usage: ${0} <type> <date> [use2mobs]"
+  echo "  <type>: conv, oz, or sat"
+  echo "  <date>: date string to match"
+  echo "  [use2mobs]: (optional, only for conv, whether to use 2m observations) YES or NO"
+  exit 1
+fi
```
DavidHuber-NOAA: Per Cory's suggestion:
```suggestion
#   ./create_gsi_info.sh <type> <date> <directory> [use2mobs]
#   <type>: conv, oz, or sat
#   <date>: date string to match
#   <directory>: directory to write the info file into
#   [use2mobs]: (optional, only for conv) YES or NO

set -eu
# Inherit set -e inside of the get_usedate function below
shopt -s inherit_errexit

# Input arguments
type_in=${1:-}
date_in=${2:-}
dir_in=${3:-}
use2mobs=${4:-NO}

if [[ -z "${type_in}" || -z "${date_in}" || -z "${dir_in}" ]]; then
  echo "Usage: ${0} <type> <date> <directory> [use2mobs]"
  echo "  <type>: conv, oz, or sat"
  echo "  <date>: date string to match"
  echo "  <directory>: where to write the new GSI info files into"
  echo "  [use2mobs]: (optional, only for conv, whether to use 2m observations) YES or NO"
  exit 1
fi

File: ush/create_gsi_info.sh (hunk 27 above)
Comment (DavidHuber-NOAA) [inline ush/create_gsi_info.sh]:
```suggestion
starting_dir="${PWD}"
```

ush/create_gsi_info.sh
```diff
@@ -0,0 +1,129 @@
...
+  done
+
+  echo "${usedate}"
+}
+
+# Get the starting directory
+starting_dir="${DATA:-${PWD}}"
+
+# Filename to write the info to
+info_file="${starting_dir}/${type_in}info"
```
DavidHuber-NOAA:
```suggestion
info_file="${dir_in}/${type_in}info"
```

DavidHuber-NOAA (commented):
Some suggestions for Cory's suggestion.

dev/parm/config/gfs/config.anal
```diff
@@ -150,6 +150,30 @@ if [[ ${RUN_ENVIR} == "emc" ]]; then
...
+   else # after 2022052712 use NCEP ozone bufr with hoices from gfsv17_historical
+      export OBS_INPUT="${BUILD_GSINFO_DIR}/obs_input/obs_input_reanl_ncep_ozone.txt"
+   fi
+
+   # Use 2m observations when constructing conventional obs info file
+   if [[ ${DO_GSISOILDA} == "YES" ]]; then
+      export BUILD_2M_OBS="YES"
+   else
+      export BUILD_2M_OBS="NO"
+   fi
```
aerorahul:
```suggestion
   # Use 2m observations when constructing conventional obs info file when DO_GSISOILDA is YES
   export BUILD_2M_OBS="${DO_GSISOILDA}"
```

aerorahul:
```suggestion
    export BUILD_2M_OBS="${DO_GSISOILDA}"
```

dev/parm/config/gfs/config.anal
```diff
@@ -150,6 +150,30 @@ if [[ ${RUN_ENVIR} == "emc" ]]; then
...
+   # these will be set based on date in scripts/exglobal_atmos_analysis.sh
+   export SATINFO="generate"
+   export CONVINFO="generate"
+   export OZINFO="generate"
+   if [[ "${PDY}${cyc}" -lt "2022052712" ]]; then
+      # use NASA ozone files for reanalysis before 2022052712
+      export OBS_INPUT="${BUILD_GSINFO_DIR}/obs_input/obs_input_reanl_nasa_ozone.txt"
+   else # after 2022052712 use NCEP ozone bufr with hoices from gfsv17_historical
+      export OBS_INPUT="${BUILD_GSINFO_DIR}/obs_input/obs_input_reanl_ncep_ozone.txt"
+   fi
```
aerorahul: Shouldn't the program be able to `generate` the correct list?

ush/create_gsi_info.sh
```diff
@@ -0,0 +1,129 @@
...
+      if [[ ${usedate} != "" ]]; then
+        cat "${sat}/${usedate}" >> "${info_file}"
+      else
+        echo "ERROR: No valid satellite info was found for satellite target '${sat}'!"
+        exit 1
+      fi
+    done <<< "${satellite_list}"
+    ;;
+  *)
+    echo "ERROR: Unknown info file type: '${type_in}'. Must be one of: conv, oz, sat"
```
aerorahul:
```suggestion
    echo "FATAL ERROR: Unknown info file type: '${type_in}'. Must be one of: conv, oz, sat"
```

ush/create_gsi_info.sh
```diff
@@ -0,0 +1,129 @@
...
+      cd "${sat}" || exit 1
+
+      usedate=$(get_usedate)
+
+      cd "${BUILD_GSINFO_DIR}/satinfo" || exit 1
+
+      if [[ ${usedate} != "" ]]; then
+        cat "${sat}/${usedate}" >> "${info_file}"
+      else
+        echo "ERROR: No valid satellite info was found for satellite target '${sat}'!"
```
aerorahul:
```suggestion
        echo "FATAL ERROR: No valid satellite info was found for satellite target '${sat}'!"
```

ush/create_gsi_info.sh
```diff
@@ -0,0 +1,129 @@
...
+    fi
+
+    # Header line
+    echo '!sensor/instr/sat      chan iuse  error  error_cld  ermax   var_b    var_pg  icld_det icloud iaerosol' >> "${info_file}"
+    while IFS= read -r sat
+    do
+      usedate=""
+      # Check that the satellite directory exists
+      if [[ ! -d "${sat}" ]]; then
+        echo "ERROR: Directory ${sat} does not exist!"
```
aerorahul:
```suggestion
        echo "FATAL ERROR: Directory ${sat} does not exist!"
```

ush/create_gsi_info.sh
```diff
@@ -0,0 +1,129 @@
...
+      echo "ERROR: No valid ozone info was found!"
+      exit 1
+    fi
+    ;;
+  sat)
+    cd "${BUILD_GSINFO_DIR}/satinfo" || exit 1
+    # Read in the satellite list, ignoring comment lines and the readme
+    satellite_list=$(grep -Ev '^ *#|readme' satellites)
+    if [[ -z "${satellite_list}" ]]; then
+      echo "ERROR: No satellites found in the satellite file list!"
```
aerorahul:
```suggestion
      echo "FATAL ERROR: No satellites found in the satellite file list!"
```

ush/create_gsi_info.sh
```diff
@@ -0,0 +1,129 @@
...
+    if [[ ${usedate} != "" ]]; then
+      # Header lines
+      {
+        echo '! For mls data, pressure and obs errors are pulled from bufr, so not listed here'
+        echo '! sens/instr/sat lev  use pressure gross   obs    b_oz  pg_oz'
+        echo '!                                  error  error variational qc'
+      } >> "${info_file}"
+      cat "${usedate}" >> "${info_file}"
+    else
+      echo "ERROR: No valid ozone info was found!"
```
aerorahul:
```suggestion
      echo "FATAL ERROR: No valid ozone info was found!"
```

ush/create_gsi_info.sh
```diff
@@ -0,0 +1,129 @@
...
+        # Turn on 2m t,q obs over land
+        sed -e "s/t        181    0   -1/t        181    0    1/g" \
+            -e "s/t        187    0   -1/t        187    0    1/g" \
+            -e "s/q        181    0   -1/q        181    0    1/g" \
+            -e "s/q        187    0   -1/q        187    0    1/g" "${usedate}" >> "${info_file}"
+      else
+        cat "${usedate}" >> "${info_file}"
+      fi
+    else
+      echo "ERROR: No valid conventional info was found!"
```
aerorahul:
```suggestion
      echo "FATAL ERROR: No valid conventional info was found!"
```

ush/create_gsi_info.sh
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env bash
+# Generic script to create convinfo, ozinfo, or satinfo for a given date.
+# Usage:
+#   ./create_gsi_info.sh <type> <date> [use2mobs]
+#   <type>: conv, oz, or sat
+#   <date>: date string to match
+#   [use2mobs]: (optional, only for conv) YES or NO
+
+set -eu
```
aerorahul: Isn't this already set via `set_strict` in [preamble.sh](https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/preamble.sh) @DavidHuber-NOAA?

aerorahul (changes requested):
A few comments and a question regarding the ozinfo generation.

dev/parm/config/gfs/config.anal
DavidHuber-NOAA: @jswhit2 ?

jswhit: we're planning to use NASA ozone obs (in netcdf format) for reanalysis, these require different settings in the OBS_INPUT namelist block in GSI.   I've also added a bunch of old instrurements (like msu_tirosn) to the namelist block so we can run in 1979.  I don't see how to generate this stuff automatically in GSI, since they depend on certain user choices. For now the user will have to modify config.anal to point to the right obs_input namelist file.

dev/parm/config/gfs/config.anal
jswhit: It's possible that we may want to use BUILD_2M_OBS=YES for reanalysis, with setting DO_GSISOILDA=YES.  This will use t2m obs to update the atmosphere, but not do a soil analysis (which requires an ensemble).

DavidHuber-NOAA: @jswhit2 OK, I can make this overridable with

```bash
export BUILD_2M_OBS="${BUILD_2M_OBS:-DO_GSISOILDA}"
```

Then you have the option of adding `export BUILD_2M_OBS=YES` to `config.base` alongside `export DO_GSISOILDA=NO`.  Would that satisfy that case?

dev/parm/config/gfs/config.anal
aerorahul: Thanks for that information. I am good with this choice.

dev/parm/config/gfs/config.anal
jswhit: Yes, that should work fine.

dev/parm/config/gfs/config.anal
jswhit: I think I will remove that if/then block anyway and just go with using 
`export OBS_INPUT="${BUILD_GSINFO_DIR}/obs_input/obs_input_reanl_ncep_ozone.txt"` for simplicity's sake, and we can just override that to use NASA's files.

ush/create_gsi_info.sh
jswhit: done

ush/create_gsi_info.sh
jswhit: done

ush/create_gsi_info.sh
jswhit: done

ush/create_gsi_info.sh
jswhit: done

ush/create_gsi_info.sh
jswhit: done

ush/create_gsi_info.sh
jswhit: done

dev/parm/config/gfs/config.anal
jswhit: done

ush/create_gsi_info.sh
jswhit: done

dev/parm/config/gfs/config.anal
jswhit: done

jswhit:
I think all the requested changes have been made and this PR is ready to go once @RussTreadon-NOAA and @ClaraDraper-NOAA finish their reviews.

scripts/exglobal_atmos_analysis.sh
```diff
@@ -80,6 +80,7 @@ export NTHREADS_CHGRES=${NTHREADS_CHGRES:-24}
 CALCINCPY=${CALCINCPY:-${USHgfs}/calcinc_gfs.py}
 
 export hofx_2m_sfcfile=${hofx_2m_sfcfile:-".false."}
+export ignore_2mQM=${ignore_2mQM:-".false."}
```
ClaraDraper-NOAA, line 83: @jswhit Can you remember the details of this? For a PREPBUFR file with the obs_error set to missing (which results in the obs being discarded in read_prepbufr), it will override that and set the QM to 2, and provide a hard-coded obs_error. 

Older PREPBUFR had missing obs_errors for the 2m obs (thus the above was necessary), but for recent files the obs_error is set. I can't remember if ignore_2mQM is out-dated (and should be removed), or if we need to keep it for the reanalysis (which would be the case if the latter is assimilating old & unchanged PREPBUFR files).

jswhit2: hmmm - sorry, but I don't remember either.  The reanalysis prepbufr files are old (they are from CFSR), so I guess we still need it.

ClaraDraper-NOAA: That's my memory ( I think). In that case, for the reanalysis we'll need to set ignore_2mQM = .true. whenever hofx_2m_sfcfile is .true. But the default  ignore_2mQM for GFS should still be false / undefined. 

Once we have a workflow set-up specific to the reanalysis, I think this would go in config.eupd

ClaraDraper-NOAA (dismissed):
I've only looked at the parts relevant to the 2m obs, but these look good to me. Thanks Jeff.

scripts/exglobal_atmos_analysis.sh
jack-woollen: I checked the prepbufr files and the gsi code and verified ignore_2mQM is required for reanalysis using t2m and q2m. By 2020 the operational files did fill in the qm=2 and obs error for t and q, but I think leaving the ignore logical set true does no harm since convinfo is the final arbiter of usage.  The new scout run is assimilating both t2m and q2m. Maybe hofx_2m_sfcfile is a jedi thing?

ClaraDraper-NOAA: hofx_2m_sfcfile is a flag to calculate the 2m h(x) from the model's diag files (without it, the lowest model level is used). 

If we set ignore_2mQM to true, it will over-ride all quality mark = 9 for the T2m and q2m over land. My memory from when I coded this (which is very foggy), is that even with the new obs-processing there can be a handful of individual obs that were set with quality flag 9 (for other quality reasons), even though most now have 2. So I recommend that we leave that flag with a default of false.

jack-woollen: That's a very good point. I wish I thought of it. But this can be solved by also checking two other codes, which specify why the qm=9. If the reason code is 3, and the program code is 4, then qm=9 comes only from a missing oberror. If the reason code is not 3, then it's a different problem. So, without checking the other codes, using ignore_2mQM as is can promote bad data. I checked several prepbufr files and did find qm=9, pc=4,  with reason codes other than 3. I'll check a few months to see the extent of this, and code up a check to take this into account.

jswhit2:
@DavidHuber-NOAA curious about why you made [this change](https://github.com/jswhit2/global-workflow/commit/ee61b30f0d6bed31001dbe6d88af636550c7b654).  Seems like the correct place to look is ```${FIXgsi}/build_gsinfo```.

scripts/exglobal_atmos_analysis.sh
ClaraDraper-NOAA: Sounds great - thanks Jack!

DavidHuber-NOAA:
@jswhit that change goes along with this one: https://github.com/jswhit2/global-workflow/pull/2/commits/8167654653d5f97a8bb625840208e87badae4f66.  The latter creates links from the `build_gsinfo` contents (`satinfo`, `ozinfo`, etc) to `parm/gsinfo` (I made some additional commits after this to correct some bugs in the linking; you can see the full, correct list [here](https://github.com/jswhit2/global-workflow/blob/399af6bc127183379483619bcdb4f5146e98b884/sorc/link_workflow.sh#L311-L325)).

The reasons for this change set is that we can link directly to the `gsinfo` files in the `build_gsinfo` submodule and thus not have to stage new GSI fix file sets on all of the platforms.  Since the `build_gsinfo` submodule is version controlled, I wanted to avoid staging it.

DavidHuber-NOAA:
@jswhit2 @ClaraDraper-NOAA I ran a test case that generates the 2m observation info files along with the other GSI *info files for a C96 analysis and for cycle 2025090100.  The run directory for the analysis can be found here: `/scratch4/NCEPDEV/stmp/David.Huber/RUNDIRS/gsinfo/gdas.2025090100/anal.2297506`

Can you take a look at this directory and verify that everything looks as it should (and/or ping others to take a look)?  I would like to start the process of merging upstream submodule PRs so we can move this PR forward.

The `expdir` for the experiment can be found here: `/scratch3/NCEPDEV/stmp/David.Huber/para_gsinfo/expdir/gsinfo`

It is worth noting that the `gdas_analdiag` job failed for this case.  I have not investigated the cause of the failure yet, but it suggests that the `diag` jobs will need to be reworked some to support dynamic GSI *info files.  I'm not sure if this needs to happen for this PR or if this can happen in a follow-up.  What do you think?

DavidHuber-NOAA (dismissed):
Clearing my requested changes.

CoryMartin-NOAA:
> It is worth noting that the `gdas_analdiag` job failed for this case. I have not investigated the cause of the failure yet, but it suggests that the `diag` jobs will need to be reworked some to support dynamic GSI *info files. I'm not sure if this needs to happen for this PR or if this can happen in a follow-up. What do you think?

I don't think we can let this get merged in if it breaks `analdiag`

DavidHuber-NOAA:
@CoryMartin-NOAA just clarifying that this only breaks ‘analdiag’ if the ‘build_gsinfo’ is enabled, which it is not by default.  I’m fine with continuing to work on this issue if that’s still a blocking issue.

CoryMartin-NOAA:
@DavidHuber-NOAA got it, I misunderstood. As long as it passes with the default/old configuration, that's fine with me

jswhit:
> @jswhit that change goes along with this one: [jswhit2@8167654](https://github.com/jswhit2/global-workflow/commit/8167654653d5f97a8bb625840208e87badae4f66). The latter creates links from the `build_gsinfo` contents (`satinfo`, `ozinfo`, etc) to `parm/gsinfo` (I made some additional commits after this to correct some bugs in the linking; you can see the full, correct list [here](https://github.com/jswhit2/global-workflow/blob/399af6bc127183379483619bcdb4f5146e98b884/sorc/link_workflow.sh#L311-L325)).
> 
> The reasons for this change set is that we can link directly to the `gsinfo` files in the `build_gsinfo` submodule and thus not have to stage new GSI fix file sets on all of the platforms. Since the `build_gsinfo` submodule is version controlled, I wanted to avoid staging it.

OK that makes sense. Thanks @DavidHuber-NOAA

jswhit:
> @jswhit2 @ClaraDraper-NOAA I ran a test case that generates the 2m observation info files along with the other GSI *info files for a C96 analysis and for cycle 2025090100. The run directory for the analysis can be found here: `/scratch4/NCEPDEV/stmp/David.Huber/RUNDIRS/gsinfo/gdas.2025090100/anal.2297506`
> 
> Can you take a look at this directory and verify that everything looks as it should (and/or ping others to take a look)? I would like to start the process of merging upstream submodule PRs so we can move this PR forward.
> 
> The `expdir` for the experiment can be found here: `/scratch3/NCEPDEV/stmp/David.Huber/para_gsinfo/expdir/gsinfo`
> 
> It is worth noting that the `gdas_analdiag` job failed for this case. I have not investigated the cause of the failure yet, but it suggests that the `diag` jobs will need to be reworked some to support dynamic GSI *info files. I'm not sure if this needs to happen for this PR or if this can happen in a follow-up. What do you think?

Unfortunately, I don't have RDHPCS access yet (after my recent re-hire) so I can't see those files. I do still have access to orion, so if you can copy the files there (including the log from the failed analdiag step)  I can take a look.

DavidHuber-NOAA:
@jswhit2 sure, I've copied over the expdir, comroot, and anal.2297506 directories to Orion here: `/work2/noaa/global/dhuber/gsinfo`.

ush/create_gsi_info.sh
DavidHuber-NOAA: Yes, fixed.

parm/config/gfs/config.anal
DavidHuber-NOAA: The `hirs_fix` data has been added to the `build_gsinfo` repository, so there are no longer any C6 paths in config.anal.

ush/create_gsi_info.sh
DavidHuber-NOAA, line 1: @jswhit2 Could you add this file to `.github/CODEOWNERS`?  Here is the patch:

```patch
diff --git a/.github/CODEOWNERS b/.github/CODEOWNERS
index 51046f05..0c534496 100644
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -149,6 +149,7 @@ ush/atmos_extractvars.sh @EricSinsky-NOAA
 ush/bash_utils.sh
 ush/calcanl_gfs.py @CoryMartin-NOAA
 ush/calcinc_gfs.py @CoryMartin-NOAA
+ush/create_gs_info.sh @jswhit2 @DavidHuber-NOAA
 ush/detect_machine.sh
 ush/extractvars_tools.sh @EricSinsky-NOAA
 ush/file_utils.sh
```

aerorahul (dismissed):
looks good.

jswhit2:
the analdiag step fails because a diag_tcp file is not created.  This should be fixed by https://github.com/jswhit2/global-workflow/commit/c3518bb392a4ca73c24c156fa54d821dc35decfd

jswhit2:
Checked the satinfo, ozinfo and convinfo files in @DavidHuber-NOAA's test and compared to the files in [gfsv17_historical](https://github.com/ADCollard/GSI-fix/tree/develop/gfsv17_historical).  Aside from one error in avhrr3_metop-c (which is now fixed in [build_gsinfo-fix](https://github.com/NOAA-PSL/build_gsinfo-fix/tree/gfsv17_historical) the files are functionally equivalent (but not identical).  The only differences are for instruments that no longer exist, or are turned off.

DavidHuber-NOAA:
@jswhit thanks for the fixes!  That did get the gdas_analdiag job to run and the rest of the cycle ran successfully as well.

CatherineThomas-NOAA:
@jswhit2 @DavidHuber-NOAA - Is the GSI hash in this branch correct?  I'm not able to clone/build:
```
fatal: Fetched in submodule path 'gsi_enkf.fd', but it did not contain 1460f419c700dadb706c3c85f2cbe71cc280aab6. Direct fetching of that commit failed.
```
Github is also [a bit confused](https://github.com/NOAA-EMC/GSI/compare/b806a99515d02de3fbfa80bb6336a454eabf8d8b...1460f419c700dadb706c3c85f2cbe71cc280aab6) about the hash.

DavidHuber-NOAA:
@CatherineThomas-NOAA the branch lives in my fork.  I will update the `.gitmodules` file to tell Git/GitHub where to find it.

DavidHuber-NOAA:
@CatherineThomas-NOAA It should be clonable now.  I had a couple of local commits on Ursa that weren't pushed.  I updated the `.gitmodules` file as well for good measure.

CatherineThomas-NOAA:
@DavidHuber-NOAA - I was able to clone and run successfully on GaeaC6 after the change.  Thanks!

CatherineThomas-NOAA:
@DavidHuber-NOAA @jswhit 
GSI-fix PR https://github.com/NOAA-EMC/GSI-fix/pull/50 has been successfully merged to its repo.  Do you need someone else to update the hash in the GSI as well?  Are there other changes in the GSI hash associated with this PR or is it just the GSI-fix hash update?

jswhit:
I think it's just the GSI-fix hash in GSI that needs updating. @DavidHuber-NOAA could you update the hash in your [fork](https://github.com/DavidHuber-NOAA/GSI/tree/feature/update_build_gsinfo) and issue a PR?  Once that is merged I can update the GSI hash here and this PR should be ready to merge.

dev/parm/config/gfs/config.base.j2
```diff
@@ -384,6 +385,8 @@ export SMOOTH_ENKF="NO"
 export l4densvar=".true."
 export lwrite4danl=".true."
 export DO_CALC_INCREMENT="NO"
+export USE_BUILD_GSINFO="NO"
```
CatherineThomas-NOAA: Since we plan on using this for the retros (imminently), is there any reason to not make this "YES"?

CatherineThomas-NOAA:
The associated GSI PR https://github.com/NOAA-EMC/GSI/pull/945 has been merged. Looks like from here we need to:

- Update the GSI hash
- Update the glopara staged GSI-fix directory (with new date)
- Update gsi fix version date
- Turn on default behavior to build_gsinfo in gfs/config.base

Did I miss anything?

dev/parm/config/gfs/config.base.j2
jswhit: No, I think it should be set to "YES"

DavidHuber-NOAA:
@CatherineThomas-NOAA would you mind making the GSI fix directory?  I'm not sure how it should be constructed given that the head of the GSI is still pointing to the `20250529` dataset while the global workflow points to `20250827` (should this be updated in the GSI? e.g. [gsi_ursa.intel.lua#L21](https://github.com/NOAA-EMC/GSI/blob/840ad1d01af82b5d0fe6a6369399f9ee509ab7db/modulefiles/gsi_ursa.intel.lua#L21)).  Once constructed, please point me to it and I will stage it.  Also, should there be another GSI hash update to point to the new dataset?

CatherineThomas-NOAA:
@RussTreadon-NOAA @CoryMartin-NOAA @jswhit

Can someone spot check the GSI-fix I created for glopara staging?  The new directory is on Ursa/Hera here:
`/scratch3/NCEPDEV/da/Catherine.Thomas/git/GSI-fix/glopara.update`

I copied the existing glopara directory (`/scratch3/NCEPDEV/global/role.glopara/fix/gsi/20250827/`) and replaced these files based on the differences I found with GSI-fix develop and recent GSI-fix commits:
* rejectlist_global.txt
* global_scaninfo.txt
* global_satinfo.txt
* global_convinfo.txt
* cloudy_radiance_info.txt 

And then I also added build_gsinfo.

RussTreadon-NOAA:
@CatherineThomas-NOAA :  I cloned GSI `develop` at [840ad1d](https://github.com/NOAA-EMC/GSI/commit/840ad1d01af82b5d0fe6a6369399f9ee509ab7db) on Hera.  As @DavidHuber-NOAA  noted, I updated `GSI_BINARY_SOURCE_DIR` in the local copy of `modulefiles/gsi_hera.intel.lua`
```
@@ -24,7 +24,7 @@ load("impi/2022.1.2")
 pushenv("CFLAGS", "-xHOST")
 pushenv("FFLAGS", "-xHOST")

-pushenv("GSI_BINARY_SOURCE_DIR", "/scratch3/NCEPDEV/global/role.glopara/fix/gsi/20250529")
+pushenv("GSI_BINARY_SOURCE_DIR", "/scratch3/NCEPDEV/global/role.glopara/fix/gsi/20250827")
 setenv("CRTM_FIX", pathJoin("/scratch3/NCEPDEV/global/role.glopara/fix/crtm", "v" .. crtm_fix_ver))

 whatis("Description: GSI environment on Hera with Intel Compilers")
```

`ush/build.sh` was then executed.  After the build finished the local `fix` directory was compared with `/scratch3/NCEPDEV/da/Catherine.Thomas/git/GSI-fix/glopara.update`
```
Hera(hfe04):/scratch3/NCEPDEV/da/Russ.Treadon/git/gsi/develop$ diff -r fix /scratch3/NCEPDEV/da/Catherine.Thomas/git/GSI-fix/glopara.update
Only in fix: .git
Only in fix: .gitignore
Only in fix: .gitmodules
diff -r fix/build_gsinfo/.git /scratch3/NCEPDEV/da/Catherine.Thomas/git/GSI-fix/glopara.update/build_gsinfo/.git
1c1
< gitdir: ../../.git/modules/fix/modules/build_gsinfo
---
> gitdir: ../.git/modules/build_gsinfo
Only in /scratch3/NCEPDEV/da/Catherine.Thomas/git/GSI-fix/glopara.update: tmp.txt
```

I don't think the _Only_ or _diff_ in `.git*` matters.  

This check confirms consistency between `/scratch3/NCEPDEV/da/Catherine.Thomas/git/GSI-fix/glopara.update` and GSI `develop` at [840ad1d](https://github.com/NOAA-EMC/GSI/commit/840ad1d01af82b5d0fe6a6369399f9ee509ab7db).  

It does not confirm that the contents of the fix files are correct.  Since we will use `fix/build_gsinfo` in GFS v17 retrospectives we also need confirmation that all the files in the directories inside `build_gsinfo` are correct.

DavidHuber-NOAA:
@CatherineThomas-NOAA may I begin staging the GSI-fix data in the glopara spaces from `/scratch3/NCEPDEV/da/Catherine.Thomas/git/GSI-fix/glopara.update`?

CatherineThomas-NOAA:
@RussTreadon-NOAA - Thank you for checking the directory I staged.

> It does not confirm that the contents of the fix files are correct. Since we will use fix/build_gsinfo in GFS v17 retrospectives we also need confirmation that all the files in the directories inside build_gsinfo are correct.

This type of confirmation was done as part of https://github.com/NOAA-EMC/GSI-fix/pull/50.

@DavidHuber-NOAA - You can start staging the GSI-fix directory that I provided.  Thanks!

jswhit2:
I've updated .gitmodules to point to NOAA-EMC/GSI develop, updated the GSI hash, and set USE_BUILD_GSINFO=YES in config.base.j2

DavidHuber-NOAA:
I have staged the new GSI fix data on Ursa under the version date 20251105.  I will now start testing on that platform.  I will also go ahead and stage on the other platforms, but wait for Ursa to pass before launching other tests.

.gitmodules
```diff
@@ -14,6 +14,7 @@
 [submodule "sorc/verif-global.fd"]
   path = sorc/verif-global.fd
   url = https://github.com/NOAA-EMC/EMC_verif-global.git
+# TODO: Point back to NOAA-EMC
```
CoryMartin-NOAA:
```suggestion
```

CoryMartin-NOAA (dismissed):
Looks good, just a leftover comment

DavidHuber-NOAA:
The new GSI fix dataset has been staged on all platforms (Hera/Ursa, Gaea C6, Cactus, Dogwood, and Orion/Hercules) as well as to the S3 bucket and HPSS.

DavidHuber-NOAA:
FYI @TaylorRoper-NOAA there is a new GSI fix dataset available on the AWS cloud (20251105).

DavidHuber-NOAA:
FYI @weihuang-jedi There is a new GSI fix dataset that will soon be required if you wish to run the GSI on Derecho.  It can be synched from Ursa: `/scratch3/NCEPDEV/global/role.glopara/fix/gsi/20251105`

DavidHuber-NOAA:
All cycled test cases have completed the first-cycle analyses on Ursa and all first-cycle forecasts are either running or completed successfully.  Launching CI tests on WCOSS2, C6, and Hercules.

CatherineThomas-NOAA:
GSI PR https://github.com/NOAA-EMC/GSI/pull/950 has been opened to update the GSI modulefiles and is undergoing ctesting.

DavidHuber-NOAA:
@CatherineThomas-NOAA Since the GSI hash update will be a non-value changing update, I suggest just running a single `C96C48_hybatmDA` test through the first full cycle and comparing forecasts with the current run on Ursa.  This assumes that all currently-running tests pass.

DavidHuber-NOAA:
All tests passed on WCOSS2.

CatherineThomas-NOAA:
@DavidHuber-NOAA - Where can I find the glopara CI tests on Ursa?

DavidHuber-NOAA:
@CatherineThomas-NOAA That would be here: `/scratch3/NCEPDEV/global/role.glopara/GFS_CI_CD/URSA/BUILDS/GITLAB/pr_cases_3472_6e3ec86b_5544/RUNTESTS`.

CatherineThomas-NOAA:
I ran the `C96C48_hybatmDA` CI test on Ursa with `jswhit2:build_gsinfo` + updating the GSI hash to [7da30e7](https://github.com/NOAA-EMC/GSI/commit/7da30e79df0c986712c92b30de8136d7caa00adc).  The only change is the updating of the GSI modulefiles to the new glopara staged fix directory date.  

The second full cycle of the new CI test reproduces the test in `/scratch3/NCEPDEV/global/role.glopara/GFS_CI_CD/URSA/BUILDS/GITLAB/pr_cases_3472_6e3ec86b_5544/RUNTESTS` as checked by identical gsistats and restarts.

DavidHuber-NOAA:
@CatherineThomas-NOAA Great!  I will go ahead and update the GSI hash and then seek final approvals for merging.

CatherineThomas-NOAA (dismissed):

aerorahul:
@jswhit 
I am checking for EE2 dead links and replacing `ln` with `cpreq` where required, when I ran into a strange behaviour with the `C96C48_hybatmDA` test.
https://github.com/NOAA-EMC/global-workflow/blob/22926afdcf1c8edff47f611fee39f25ed885eb3d/scripts/exgdas_enkf_update.sh#L109-L111

When `BUILD_GSINFO=YES`, these values of `SATINFO`, `CONVINFO`, and `OZINFO` are `generate`.

When we reach https://github.com/NOAA-EMC/global-workflow/blob/22926afdcf1c8edff47f611fee39f25ed885eb3d/scripts/exgdas_enkf_update.sh#L124-L127, the links are created to a non-existent file `generate`.
Here is a snippet of the log: 
```
+ exgdas_enkf_update.sh[124]/bin/ln -sf generate satinfo
+ exgdas_enkf_update.sh[125]/bin/ln -sf /gpfs/f6/drsa-precip3/world-shared/global/CI/GITLAB/nightly_0_6ef102de_5821/global-workflow/fix/gsi/global_scaninfo.txt scaninfo
+ exgdas_enkf_update.sh[126]/bin/ln -sf generate convinfo
+ exgdas_enkf_update.sh[127]/bin/ln -sf generate ozinfo
```

For EE2, we need to replace the `ln -sf` with `cpreq`.  There is no file named `generate`.  The script fails here.

I wonder if the EnKF actually does not need these files anymore, or worse assumes something in its place when a file is not found.

FYI @CatherineThomas-NOAA

CatherineThomas-NOAA:
Unfortunately, it looks like the EnKF is making some use of the satinfo file.  From our new retrospectives enkfstat file (`/lfs/h2/emc/gfstemp/emc.global/comroot/retrov17_01_stream2/enkfgdas.20240515/12/ensstat/analysis/atmos/enkfgdas.t12z.enkfstat.txt`): 
```
nobs_conv, nobs_oz, nobs_sat =      2006906       35170           0
```
It looks like we need to add these blocks for each info file into the exgdas_enkf_update.sh script – 
https://github.com/NOAA-EMC/global-workflow/blob/22926afdcf1c8edff47f611fee39f25ed885eb3d/scripts/exglobal_atmos_analysis.sh#L369-L373

I tested this in my branch https://github.com/CatherineThomas-NOAA/global-workflow/tree/gsinfo_eupd.  While testing, I also noticed a typo in the config.anal file:
```
- export USE_2M_OBS="${USE_2M_OBS:-DO_GSISOILDA}
+ export USE_2M_OBS="${USE_2M_OBS:-$DO_GSISOILDA}
```
The resulting C96C48_hybatmDA test resulted in radiances being assimilated:
```
 nobs_conv, nobs_oz, nobs_sat =      2261461       34187      243633
```
I also tested it with the C96C48mx500_S2SW_cyc_gfs test and copied some output to Hercules here:
`/work2/noaa/stmp/cthomas/gsinfo`

@jswhit @DavidHuber-NOAA @aerorahul  - Does this look correct?

aerorahul:
Looks like the satobs are now assimilated with a valid satinfo file.
Why are the number of conv and oz obs changing?
What does the gsi output say (when run in eobs) regarding the number of these obs?

CatherineThomas-NOAA:
> Why are the number of conv and oz obs changing?

The two lines of `nobs` that I showed are not directly comparable.  I pulled the first line from our full res and the second was from my low res test with the fix.  I'll run a clean comparison against develop.

> What does the gsi output say (when run in eobs) regarding the number of these obs?

The summary stats aren't easily comparable (separated vs summary, include qc or not, etc), but ozone is easy enough to compare since its just a few instruments and the counts match.  Both the gsistat.ensmean and enkfstat have radiances by channel and comparing atms_n20 between the two files are identical for all channels.

jswhit:
> Unfortunately, it looks like the EnKF is making some use of the satinfo file. From our new retrospectives enkfstat file (`/lfs/h2/emc/gfstemp/emc.global/comroot/retrov17_01_stream2/enkfgdas.20240515/12/ensstat/analysis/atmos/enkfgdas.t12z.enkfstat.txt`):
> 
> ```
> nobs_conv, nobs_oz, nobs_sat =      2006906       35170           0
> ```
> 
> It looks like we need to add these blocks for each info file into the exgdas_enkf_update.sh script –
> 
> https://github.com/NOAA-EMC/global-workflow/blob/22926afdcf1c8edff47f611fee39f25ed885eb3d/scripts/exglobal_atmos_analysis.sh#L369-L373
> 
> I tested this in my branch https://github.com/CatherineThomas-NOAA/global-workflow/tree/gsinfo_eupd. While testing, I also noticed a typo in the config.anal file:
> 
> ```
> - export USE_2M_OBS="${USE_2M_OBS:-DO_GSISOILDA}
> + export USE_2M_OBS="${USE_2M_OBS:-$DO_GSISOILDA}
> ```
> 
> The resulting C96C48_hybatmDA test resulted in radiances being assimilated:
> 
> ```
>  nobs_conv, nobs_oz, nobs_sat =      2261461       34187      243633
> ```
> 
> I also tested it with the C96C48mx500_S2SW_cyc_gfs test and copied some output to Hercules here: `/work2/noaa/stmp/cthomas/gsinfo`
> 
> @jswhit @DavidHuber-NOAA @aerorahul - Does this look correct?

Yes, this looks correct.  I should have caught this, sorry @CatherineThomas-NOAA.  The &satobs_enkf and &ozobs_enkf namelist blocks will also need to be updated for reanalysis to include historical instruments. I will create a PR.

NWM: I see @CatherineThomas-NOAA already created a PR and it's been merged.

PR #3489, opened 2025-03-24
Save marine LETKF yaml

shlyaeva:
# Description

Saves marine LETKF yaml for posterity and for EVA verifications plots.

Note: this is not urgent as it's not required for v17.

# Type of change

# Change characteristics

# How has this been tested?
Tested with the hybrid C384m025 experiment in ci/cases/gfsv17, with hera/intel. The yaml is saved in COMROOT as expected and successfully used for diagnostics plots (see https://github.com/NOAA-EMC/gdas-marine-viz/pull/511 for an example).

# Checklist


AndrewEichmann-NOAA (dismissed):
Looks good

ush/python/pygfs/task/marine_letkf.py
```diff
@@ -205,5 +205,9 @@ def finalize(self):
         # stage the desired diag files
         FileHandler({'mkdir': [os.path.join(letkfsaveconf.COMOUT_OCEAN_LETKF, 'diags')]}).sync()
         FileHandler({'copy': obs_files_to_copy}).sync()
+        # yaml configurations
+        yamls_to_copy = []
+        yamls_to_copy.append([letkfsaveconf.letkf_yaml_file, os.path.join(letkfsaveconf.COMOUT_OCEAN_LETKF, 'letkf.yaml')])
```
aerorahul: This action is fine.   
However, we are saving all the configurations to directory named `COM_CONF` instead of `COMOUT_OCEAN_LETKF`.  `COM_CONF` is a directory in `COM` that has all the configurations of all the executables that are run in a cycle.

shlyaeva: Nice, I didn't know this! I'll take a look later this week to change this and maybe also some other marine yamls that get copied to a different directory.

guillaumevernieres (dismissed):
:+1:

ush/python/pygfs/task/marine_letkf.py
shlyaeva: Finally getting back to this. Is there a place where things from COM_CONF get archived? I haven't found it in the code yet. I'm wondering if I need a custom addition for archiving this yaml, or if it's already handled some place.

aerorahul: E.g. https://github.com/NOAA-EMC/global-workflow/blob/b1c60d4506a973d1a0684018e2adc3225bc2e7a4/parm/archive/gfsa.yaml.j2#L14-L15
shows that the `gfsa.tar` will contain the `ufs.input.nml` from `COM_CONF`.

aerorahul: A potential place to add this file to the archive list would be in [this](https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/archive/enkf_marine_grp.yaml.j2) yaml.

jobs/JGLOBAL_MARINE_ANALYSIS_LETKF
```diff
@@ -30,6 +30,10 @@ YMD=${PDY} HH=${cyc} declare_from_tmpl -rx \
   COMOUT_OCEAN_LETKF:COM_OCEAN_LETKF_TMPL \
   COMOUT_ICE_LETKF:COM_ICE_LETKF_TMPL
 
+RUN=${GDUMP_ENS} YMD=${PDY} HH=${cyc} MEMDIR="ensstat" declare_from_tmpl -rx \
```
aerorahul, line 34: I think this should be `RUN=${RUN}` because the output is going to land in the `enkfgdas` or `enkfgfs` based on if this job is run in the early or late cycle.
```suggestion
YMD=${PDY} HH=${cyc} MEMDIR="ensstat" declare_from_tmpl -rx \
```
Incidentally, in the current case for this line, `GDUMP_ENS=RUN=enkfgdas`, but that will not be true when `RUN=enkfgfs`.

aerorahul (changes requested):
Rest of the changes looks ok, but I have one comment that needs your attention.

jobs/JGLOBAL_MARINE_ANALYSIS_LETKF
shlyaeva: Cool, this makes sense!
Unfortunately right now this won't work: the LETKF job is `gdas_marineanlletkf`. I ran a test, and it's putting the output in the `gdas` directory. Looks like this is incorrect, but I'd rather not fix it as part of this PR if that's OK.
In general, where is it defined which `RUN` goes with which task?

aerorahul: Understood @shlyaeva 
This job should be run in the `RUN=enkfgdas` but is currently running in `RUN=gdas`.  I had not noticed that.  Thanks for catching.
If you wouldn't mind opening an issue and placing a `TODO:` or `FIXME:` in this job pointing to the issue number, we can address it via the issue.  Moving this job to the `RUN=enkfgdas` cycle will impact dependencies and triggers.

shlyaeva: Done!

aerorahul (approved):
thanks @shlyaeva

parm/archive/master_enkf.yaml.j2
```diff
@@ -74,6 +74,12 @@ datasets:
 
     {% endfor %}
 
+    {% set tmpl_dict = ({ '${ROTDIR}':ROTDIR,
+                          '${RUN}':RUN,
+                          '${YMD}':cycle_YMD,
+                          '${HH}':cycle_HH,
+                          '${MEMDIR}': "ensstat"}) %}
+    {% set COMIN_CONF = COM_CONF_TMPL | replace_tmpl(tmpl_dict) %}
```
DavidHuber-NOAA, line 82: I think this should be moved to JGLOBAL_ENS_ARCHIVE_TARS by inserting a declaration here:

https://github.com/NOAA-EMC/global-workflow/blob/e131fb155c782159ac676b383d93fe0410e7f456/jobs/JGLOBAL_ENS_ARCHIVE_TARS#L9-L15

But I'm OK with this being handled as part of https://github.com/NOAA-EMC/global-workflow/issues/3767.

DavidHuber-NOAA (approved):
Approve with a note on a future change request for the `COMIN_CONF` declaration in `master_enkf.yaml.j2` to be moved to `JGLOBAL_ENS_ARCHIVE_TARS`.

PR #3483, opened 2025-03-20
CSPs sync to GW develop

weihuang-jedi:
AWS uses two partitions for GW, "compute" partition for fcst models, and few wave components which need more than one node; "process" partition for products and others which need single node. Recent code broke this, and only compute was used. 
The first cut of Jenkinsfile4AWS compile "gfs" and "gefs" in two commands, we want to combine "gfs", "gefs", and "sfs" together.
Another issue is on AWS, crontab files need to include shell options, but the order was wrong.

  Resolves #3482

# Type of change
- [x] Bug fix (fixes something broken)
- [ ] New feature (adds functionality)
- [x] Maintenance (code refactor, clean-up, new CI test, etc.)

# Change characteristics
- Is this a breaking change (a change in existing functionality)? NO
- Does this change require a documentation update? NO
- Does this change require an update to any of the following submodules? NO (If YES, please add a link to any PRs that are pending.)
  - [ ] EMC verif-global 
  - [ ] GDAS 
  - [ ] GFS-utils 
  - [ ] GSI 
  - [ ] GSI-monitor 
  - [ ] GSI-utils 
  - [ ] UFS-utils 
  - [ ] UFS-weather-model 
  - [ ] wxflow 

# How has this been tested?
- Clone and build on AWS
-  Clone and test on Hera, Hercules

# Checklist
- [x] Any dependent changes have been merged and published
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have documented my code, including function, input, and output descriptions
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] This change is covered by an existing CI test or a new one has been added
- [ ] Any new scripts have been added to the .github/CODEOWNERS file with owners
- [ ] I have made corresponding changes to the system documentation if necessary

PR #3543, opened 2025-04-03
Check that partition should be set

DavidHuber-NOAA:
# Description
This adds a check to see if partition should be set at all before attempting to override the default partition with the task partition.  This resolves a bug discovered on WCOSS2, which does not have a partition, introduced by #3483.

# Type of change

# Change characteristics

# How has this been tested?
Generated an XML on WCOSS2 and C6.

# Checklist


DavidHuber-NOAA:
FYI @weihuang-jedi

aerorahul (approved):
looks good.

PR #3411, opened 2025-02-28
Use jinja templates instead of `@VARNAME@` in config files

aerorahul:
# Description
This PR:
- replaces the template pattern used in configs e.g. `@MACHINE@` in `config.base` with jinja template pattern `{{ MACHINE }}`
- `setup_expt.py` script is updated to remove custom code that was written to search and replace `@VARNAME@` with values from the experiment yamls.  Jinja provides a much cleaner way to interface those two.
- updated configs that are jinja templates now have an extension of `.j2` in `parm/config/<system>` directory.  e.g. `config.base` is now named `config.base.j2` in `parm/config/gfs`.  It gets rendered as `config.base` in the experiment directory after `setup_expt.py` completes the execution.
- makes use of `--account` argument in `setup_expt.py` as opposed to replacing `ACCOUNT` from `hosts/host.yaml`. 
- organizes `hosts/host.yaml` into sections for `Paths`, `BQS properties`, `HPSS properties` and `Features` for that host.  

In addition, this PR also:
- improves logging of `create_experiment.py`, `setup_expt.py` and `setup_xml.py`.
- allows `host.py` to honor `MACHINE_ID` set by sourcing `detect_machine.sh`.

Resolves #3439 

# Type of change
- [X] Maintenance (code refactor, clean-up, new CI test, etc.)

# Change characteristics
- Is this a breaking change (a change in existing functionality)? NO
- Does this change require a documentation update? NO
- Does this change require an update to any of the following submodules? NO 

# How has this been tested?
Experiment directories were created before and after this change and compared.  Changes were in the expected places where template patterns were updated.

# Checklist
- [X] Any dependent changes have been merged and published
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have documented my code, including function, input, and output descriptions
- [X] My changes generate no new warnings
- [X] New and existing tests pass with my changes
- [X] This change is covered by an existing CI test or a new one has been added
- [X] Any new scripts have been added to the .github/CODEOWNERS file with owners
- [X] I have made corresponding changes to the system documentation if necessary

PR #3573, opened 2025-04-10
Remove duplicate SFS config.fcst to resolve failing SFS fcst jobs

KateFriedman-NOAA:
# Description

This PR fixes an issue that is causing the SFS fcst jobs in the C96mx100_S2S CI test to fail. Several templated variables in `input.nml` are not getting set due to incorrect `config.fcst` being used. The old sfs config.fcst was not removed in work done for issue #3411 and was being chosen instead of the gefs config.fcst.j2 (via symlink from `parm/config/sfs` folder. This PR:

- deletes `parm/config/sfs/config.fcst`; should be using symlink to the gefs `config.fcst.j2` instead (symlink in place)
- adds o3forc_params setting into gefs `config.fcst.j2` from the now-deleted sfs `config.fcst`

Refs #3411

# Type of change

# How has this been tested?

Tested within on-going CI tests for another PR on WCOSS2 and a dev CI test on WCOSS2.


aerorahul (commented):
Ooh.  
I don't think we want to change this in GEFS.
We may need to update the SFS config.fcst into a j2

KateFriedman-NOAA:
> Ooh. I don't think we want to change this in GEFS. We may need to update the SFS config.fcst into a j2

Aye aye, I'll close this and reopen with that instead. Thanks!

PR #3791, opened 2025-06-11
If merged, this PR will Remove a blank string that is causing  exglobal_cleanup.sh to crash

### LINKED ISSUE #3786: GCAFS failing in gcdas_cleanup after 5 cycles.
### What is wrong?

GCAFS is failing in gcdas_cleanup starting in the 5th cycle.  The error occurs on line 42 of gcdas_cleanup.sh:

    find "${directory}" -type f -not \( ${find_exclude_string} \) -ignore_readdir_race -delete

and is caused by the fact that find_exclude_string is empty.  I've created a temporary workaround that keeps the code from crashing by checking to see if the string is empty, and ignores it if it is empty. This may not be the correct solution (eg the files might not be deleted properly), but for now its not crashing.  And perhaps the string shouldn't be empty in the first place.  I haven't tested this yet on Hera to see if it is machine dependent. 

This needs a more robust solution that 1. Finds out why find_exclude_string is empty starting on the 4th cycle and 2. either corrects this if it shouldn't be empty, or simply avoids crashing it is supposed to be empty. 

Gaea-C6

https://github.com/CoryMartin-NOAA/global-workflow,  feature/stage-gen-anl

### Steps to reproduce

1. Run the CI test o a cycling experiment beyond 4 cycles. 

### Additional information

_No response_

I have a fix which keeps the cleanup step from failing, but not sure if it is quite right. It simply checks to see if the string find_exclude_string is empty, and leaves it out of the script if it is.

andytangborn:
# Description

This draft PR is a potential fix to the crashing of the cleanup step for GCAFS. The problem is caused by an empty string and this fix checks for the empty string and changes the script to leave it out. 

 

  

  Resolves #3786 

# Type of change
- Bug fix

# Change characteristics

# How has this been tested?

Cycling many days on gaea.

# Checklist

### DIFF
--- a/scripts/exglobal_cleanup.sh
+++ b/scripts/exglobal_cleanup.sh
@@ -24,6 +24,7 @@ fi
 last_date=$(date --utc +%Y%m%d%H -d "${PDY} ${cyc} -${RMOLDEND:-24} hours")
 first_date=$(date --utc +%Y%m%d%H -d "${PDY} ${cyc} -${RMOLDSTD:-120} hours")
 last_rtofs=$(date --utc +%Y%m%d%H -d "${PDY} ${cyc} -${RMOLDRTOFS:-48} hours")
+
 function remove_files() {
     local directory=$1
     shift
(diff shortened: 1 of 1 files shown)


scripts/exglobal_cleanup.sh
```diff
@@ -27,26 +27,44 @@ last_rtofs=$(date --utc +%Y%m%d%H -d "${PDY} ${cyc} -${RMOLDRTOFS:-48} hours")
...
+    # Remove trailing -or if it exists
+    find_exclude_string="${find_exclude_string% -or}"
+
+    # Print for debugging
+    echo "DEBUG: directory = '${directory}'"
+    echo "DEBUG: find_exclude_string = ${find_exclude_string}"
+
+    if [[ -n "$find_exclude_string" ]]; then
+        # shellcheck disable=SC2086
+        eval find "${directory}" -type f -not \( ${find_exclude_string} \) -ignore_readdir_race -delete
```
CoryMartin-NOAA: why is `eval` added here?

CoryMartin-NOAA (commented):
I think this is the right general idea, but there are changes here that shouldn't be made like extra spaces, prints, and we need to keep all the shellcheck directives. Also, we will need to clean up the PR title/description.

scripts/exglobal_cleanup.sh
andytangborn: Leftover  from trying to debug.  Probably not needed.

CoryMartin-NOAA:
Superseded by a newer PR (that might have not yet fixed the issue completely). Closing this.

PR #3892, opened 2025-07-23
Updates for code hand-off, land orog sfc_data consistency check, etc.

### LINKED ISSUE #3664: Wave gempak needs different output frequencies
### What is wrong?

In operations, wave gempak is only for every 3 hours out to 180, see for example the output here: 
/lfs/h1/ops/prod/com/gfs/v16.3/gfs.20250508/00/wave/gempak

Currently in dev/workflow/rocoto/gfs_tasks.py https://github.com/NOAA-EMC/global-workflow/blob/develop/dev/workflow/rocoto/gfs_tasks.py#L1324-L1363 uses the  wave gridded post output for frequency which means we get more gempak jobs than we need for the wave model. 

We should create a new list of outputs based on wavegempak specific output times and only run wave gempak jobs out to 180 hours (or however long the forecast is) and every 3 hours. 

WCOSS2

https://github.com/JessicaMeixner-NOAA/global-workflow/tree/wavefixanddownstream (but this is in develop too)

### Steps to reproduce

I'm using https://github.com/JessicaMeixner-NOAA/global-workflow/tree/wavefixanddownstream branch to get the S2SW_extended test running on wcoss2 again and noticed this issue.  This issue is in develop as well. 

### Additional information

FYI: @AminIlia-NOAA 

Update https://github.com/NOAA-EMC/global-workflow/blob/develop/dev/workflow/rocoto/gfs_tasks.py#L1324-L1363  and add extra time related variables to https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/config/gfs/config.wavegempak similar to how https://github.com/NOAA-EMC/global-workflow/blob/develop/parm/config/gfs/config.waveawipsgridded has it's own variables for output frequency.

### LINKED ISSUE #3825: CI test ICS inconsistent with default orography files
### What is wrong?

For at least the C96C48_hybatmaerosnowDA.yaml and C96_atm3DVar.yaml tests the C96 surface ICS files on hera are inconsistent with the default orography files used by global_workflow, in that they have different land masks.

ICS files here: 
/scratch1/NCEPDEV/global/glopara/data/ICSDIR/C96C48/20250327/gdas.20211220/ 

Default orog here: 
/scratch1/NCEPDEV/global/glopara/fix/orog/20240917/C96/C96.mx500_oro_data.tile*

The ICS files have about 2000 fewer land grid points than the orog files. This is causing confusion when people use these files as examples to look into land-related issues, and could lead to crashes in the future.

@yuanxue2870 has checked all the ocn res in that directory, and none of them match. I suspect that the ICS were generated with older orog versions.

The land mask used to create the ICS should be the same as the orog file being used in the tests. If the number of land grid cells match, this is sufficient to be confident that the land masks are the same.

Specifically, the number of grid cells that are classified as land in the restart (using either vtype or stype > 0) should match the number of grid cells classified as land in the orography files (using land_frac>0).

The number of land points can be checked using:  /scratch2/BMC/gsienkf/Clara.Draper/count_land_restart_orog.py 

All or N/A

b1c60d4506a973d1a0684018e2adc3225bc2e7a4 (fix/orog or ICS dir have not changed in newer versions)

### Steps to reproduce

Update restart file directory, orog directory, and ocn and atm res in this script: 

/scratch2/BMC/gsienkf/Clara.Draper/count_land_restart_orog.py 

then check output has same number of land grid points in both files.

### Additional information

_No response_

This can be fixed by re-generating restart files for the test. We should probably check the other tests too.

### LINKED ISSUE #3838: Check all ICS for global_workflow tests
### What is wrong?

This [issue](https://github.com/NOAA-EMC/global-workflow/issues/3825) points out inconsistencies in the land mask between the C96 CI test ICS and the orog_data files being used (specifically different land masks), which can cause model crashes. 

Fixing these inconsistencies fixed the crash @CatherineThomas-NOAA was experiencing in this [issue](https://github.com/NOAA-EMC/global-workflow/issues/3522).

[This](https://github.com/ufs-community/ufs-weather-model/issues/2738) mystery crash also looks like it has inconsistencies between the ICS and orog_files - we're yet to confirm if this is causing the crash (but this explanation is consistent with the crash). 

All ICS in the global_workflow tests should have consistent land mask as the orog_data files being used. It is sufficient to check the total number of land points, for example using:

/scratch2/BMC/gsienkf/Clara.Draper/count_land_restart_orog.py

Hera, All or N/A

b1c60d4 (fix/orog or ICS dir have not changed in newer versions)

### Steps to reproduce

Count the land grid cells in the ICS and orog_files for the examples above.

### Additional information

Everything I've written above is for hera - I haven't looked at other machines.

Run the above check on the hera ICS to see which need updating (all?), then update them! 

Will also need to check other machines / synch the updates.

### LINKED ISSUE #3893: gdas_fcst should depend on gdas_waveinit
### What is wrong?

If we have waves, the gdas_fcst job should depend on the gdas_waveinit job.   I've been running some tests that update the model to the new spack-stack version (w/out running aerosols), and it seems like waveinit needs more memory.  However, I see both the waveinit job and gdas_fcst running at the same time.  

My XML file looks like this: for that task 

```
                <dependency>
                        <or>
                                <and>
                                        <taskdep task="gdas_sfcanl"/>
                                        <taskdep task="gdas_waveinit"/>
                                        <taskdep task="gdas_marineanlfinal"/>
                                        <taskdep task="gdas_snowanl"/>
                                </and>
                                <taskdep task="gdas_stage_ic"/>
                        </or>
                </dependency>
```

I'm a little confused by this, but I'm thinking for the first half cycle we just have to have the gdas_stage_ic and not the waveinit?   It's likely very rare when waveinit doesn't finish about the same time stage_ic does so that's why we probably haven't run into this before.  

gdas_fcst should wait for gdas_waveinit to finish before starting 

All or N/A

I have a lot of updates beyond develop, but none that change this part of the code. 

### Steps to reproduce

Probably could force this with a C48 case, but I'm using a C1152 case and just force the waveinit to fail.  

### Additional information

Maybe it should be something like: 

```
                <dependency>
                        <or>
                                <and>
                                        <taskdep task="gdas_sfcanl"/>
                                        <taskdep task="gdas_waveinit"/>
                                        <taskdep task="gdas_marineanlfinal"/>
                                        <taskdep task="gdas_snowanl"/>
                                </and>
                                </and>
                                      <taskdep task="gdas_waveinit"/>
                                      <taskdep task="gdas_stage_ic"/>
                                </and>
                        </or>
                </dependency>
```
assuming my understanding is correct... 

_No response_

aerorahul:
# Description
This PR:
- moves `load_fv3gfs_modules.sh`, `load_ufswm_modules.sh`, and `load_ufsda_modules.sh` from `ush/` to `dev/ush/` as these scripts are not intended for use in ops and hence should be moved to `dev/` space.
- Fixes #3664 - adds control for wave gempak output generation and caps it at 180h.
- Fixes #3838 - adds a checker to the forecast job.  Presently shows a warning, but a fatal error facility is added waiting to be enabled.  The checker can be run as a standalone script. 
- Fixes #3825
- Fixes #3893 - checks for first half cycle if it exists and updates the dependency tree for cycling.  Tested via `C96C48mx500_S2SW_cyc_gfs` on WCOSS2.
- Removes a section of TODO from `build_upp.sh` that depended on UFSWM.  The UFSWM dependency was satisfied.

# Type of change
- [X] New feature (adds functionality)

# Change characteristics

# How has this been tested?
Tests on Gaea and WCOSS are run at:
`/gpfs/f6/drsa-precip4/world-shared/Rahul.Mahajan/RUNTESTS/testing/gwpr3892`
`/lfs/h2/emc/eib/noscrub/rahul.mahajan/RUNTESTS/testing/gwpr3892`

# Checklist


aerorahul:
@ClaraDraper-NOAA 
When an inconsistency is encountered between the `oro_data` and `sfc_data`, the checker script will show it like this as an example:
```
2025-07-23 13:13:20,823 - INFO     - __main__    : BEGIN: __main__.count_points_from_input
2025-07-23 13:13:20,871 - INFO     - __main__    :   END: __main__.count_points_from_input
2025-07-23 13:13:20,871 - INFO     - __main__    : BEGIN: __main__.count_points_from_orog
2025-07-23 13:13:20,886 - INFO     - __main__    :   END: __main__.count_points_from_orog
2025-07-23 13:13:20,886 - INFO     - __main__    : BEGIN: __main__.compare_counts
2025-07-23 13:13:20,886 - WARNING  - root        : Count mismatch for tile1: input=1056, orography=1055
2025-07-23 13:13:20,886 - WARNING  - root        : Count mismatch for tile2: input=949, orography=954
2025-07-23 13:13:20,886 - WARNING  - root        : Count mismatch for tile3: input=1460, orography=1462
2025-07-23 13:13:20,886 - INFO     - __main__    :   END: __main__.compare_counts
```
This is an example output from the `C48_S2SW` test that uses cold-start initial conditions for `sfc_data`.  
The script will do the same for warm start restart provided files or `sfcanl_data` files that are used to initialize the model.

CoryMartin-NOAA:
@aerorahul @ClaraDraper-NOAA can/should we make it a failure rather than a warning?

aerorahul:
> @aerorahul @ClaraDraper-NOAA can/should we make it a failure rather than a warning?

Yes.  We absolutely should make it a failure.  If we do that now, all our tests will fail because we need an updated consistent ICs.  When they are provided, and the ICs are replaced for the tests, we can turn this into a fatal error.

It is a warning for now to allow testing to proceed. All we need to do is pass `--fatal` to the checker call.

JessicaMeixner-NOAA (dismissed):
@aerorahul  - I have looked through all of the wave related changes and they look good to me.  Did I need to run a wcoss2 test for the wave gempak to confirm or have you?

sorc/build_upp.sh
```diff
@@ -21,36 +21,8 @@ while getopts ":dj:v" option; do
...
-      set -x
-   fi
-fi
-
 cd ufs_model.fd/FV3/upp/tests
 # shellcheck disable=SC2086
-BUILD_JOBS=${BUILD_JOBS:-8} ./compile_upp.sh ${_opts}
+BUILD_JOBS=${BUILD_JOBS:-8} "./compile_upp.sh" ${_opts}
+
+exit 0
```
CoryMartin-NOAA: Missing EOL

dev/jobs/arch_tars.sh
```diff
@@ -4,7 +4,7 @@ set -x
 
 ###############################################################
 # Source FV3GFS workflow modules
-. "${HOMEgfs}"/ush/load_fv3gfs_modules.sh
+source "${HOMEgfs}"/ush/load_fv3gfs_modules.sh
```
CatherineThomas-NOAA: Missed /dev/ush here

dev/jobs/arch_vrfy.sh
```diff
@@ -4,7 +4,7 @@ set -x
 
 ###############################################################
 # Source FV3GFS workflow modules
-. "${HOMEgfs}"/ush/load_fv3gfs_modules.sh
+source "${HOMEgfs}"/ush/load_fv3gfs_modules.sh
```
CatherineThomas-NOAA: Missed /dev/ush here

dev/jobs/cleanup.sh
```diff
@@ -4,7 +4,7 @@ set -x
 
 ###############################################################
 # Source FV3GFS workflow modules
-. "${HOMEgfs}"/ush/load_fv3gfs_modules.sh
+source "${HOMEgfs}"/ush/load_fv3gfs_modules.sh
```
CatherineThomas-NOAA: One more `dev`

dev/jobs/globus_arch.sh
```diff
@@ -4,7 +4,7 @@ set -x
 
 ###############################################################
 # Source FV3GFS workflow modules
-. "${HOMEgfs}"/ush/load_fv3gfs_modules.sh
+source "${HOMEgfs}"/ush/load_fv3gfs_modules.sh
```
CatherineThomas-NOAA: Another `dev`

sorc/build_upp.sh
aerorahul: fixed. thx.

dev/jobs/arch_tars.sh
aerorahul: My `sed` magic loop didn't apply to `"${HOMEgfs}"/ush`.   Should be fixed now.  Thx.

aerorahul:
> @aerorahul - I have looked through all of the wave related changes and they look good to me. Did I need to run a wcoss2 test for the wave gempak to confirm or have you?

@JessicaMeixner-NOAA 
Thanks for the quick review @JessicaMeixner-NOAA 
I have not run the extended test (because it only runs on WCOSS2), and gempak jobs don't run on GaeaC6.
I did however, generate the XML for the gempak jobs and confirmed that the wave gempak jobs are run every 3h upto 72h and then 6h upto 180h (or 120h if FHMAX_GFS=120).

CoryMartin-NOAA (dismissed):

JessicaMeixner-NOAA:
Thanks @aerorahul  - Are you planning to run CI on WCOSS2 or should I run the extended test their to double check things?

CatherineThomas-NOAA (dismissed):
Looks good.  Thanks @aerorahul!

aerorahul:
> Thanks @aerorahul - Are you planning to run CI on WCOSS2 or should I run the extended test their to double check things?

I will start one this afternoon on WCOSS2.

JessicaMeixner-NOAA:
Awesome! Thanks @aerorahul !

ush/check_land_input_orography.py
```diff
@@ -0,0 +1,201 @@
...
+    Parameters
+    ----------
+    input_dir : str
+        Path to the directory containing the input NetCDF files.
+
+    Returns
+    -------
+    counts: Dict
+        Dictionary with:
+        - vtype: Number of points with 'vtype' > 0 for each tile.
```
ClaraDraper-NOAA, line 64: Note: you only need one of these - I suggest vtype.

ush/check_land_input_orography.py
```diff
@@ -0,0 +1,201 @@
...
+NTILES = 6  # Number of tiles expected in the input and orography data
+
+
+@logit(logger)
+def count_points_from_input(input_dir: str) -> Dict:
+    """
+    Counts the number of valid points in the 'vtype' and 'stype' variables from surface data NetCDF files
+    in the specified input directory.
+    For each tile (from 1 to NTILES), attempts to open the corresponding 'sfc_data.tile{tt}.nc' file and
+    counts the number of elements greater than zero in the 'vtype' and 'stype' arrays.
```
ClaraDraper-NOAA: @GeorgeGayno-NOAA This is OK, right? We wouldn't expect any vtype=17 (water) in the restarts would we?

ClaraDraper-NOAA (commented):
Thanks for much for doing this @aerorahul . It looks good to me. 

For simplicity, I'd probably remove the counting of the stype, as it's not needed.

aerorahul:
> Awesome! Thanks @aerorahul !

@JessicaMeixner-NOAA 
Thanks for making me run a test for `gfs_wavegempak` on WCOSS.
The `gfs_wavegempak` tasks depend on `gfs_wavepostsbs` tasks.  The `gfs_wavegempak` tasks run at a different forecast frequency than the `gfs_wavepostsbs`, and also run for a maximum of 180h, while the `gfs_wavepostsbs` runs till the end of the forecast (384h).  This creates a mismatch in the rocoto dependency with metatasks etc.

E.g. `gfs_wavepostsbs` metatask has 25 tasks for the forecast hour groups of `000-008`, `009-017`, etc. while, `gfs_wavegempak` metatask has 25 tasks for forecast hour groups of `000-003`, `006-009`, etc.  
`gfs_wavegempak` groups have a dependency of `gfs_wavepostsbs` groups. 
A simple fix would be to make `gfs_wavegempak` depend on `gfs_wavepostsbs` to complete.  That means `gfs_wavegempak` wait till all the `gfs_wavepostsbs` to finish.  I think I can make `gfs_wavegempak` depend on the `gfs_wavepostsbs` task containing the 180h,  That still means a wait for `gfs_wavegempak` till 180h `gfs_wavepostsbs` to finish.  Would this be acceptable for rocoto.  For ecflow, none of this matters as we don't generate the operational ecflow suite with this logic.
Please let me know.

If you wish to look at the test I ran on WCOSS, the relevant directories are at:
```
/lfs/h2/emc/eib/noscrub/rahul.mahajan/RUNTESTS/testing/gwpr3892/EXPDIR/C48_S2SW_extended_3892
/lfs/h2/emc/eib/noscrub/rahul.mahajan/RUNTESTS/testing/gwpr3892/COMROOT/C48_S2SW_extended_3892
```

JessicaMeixner-NOAA:
Could the gempak jobs be triggered off the availability of the grib.idx files instead of the sbs job? 

The only concern of the rocoto not starting the gempak job would be for a realtime run.

aerorahul:
> Could the gempak jobs be triggered off the availability of the grib.idx files instead of the sbs job?
> 
> The only concern of the rocoto not starting the gempak job would be for a realtime run.

Yes.  That can be done.  What would the name of the `grib.idx` file be for the gempak job?  E.g. for the `000-003` `gfs_wavegempak` group, what `003` `grib.idx` file should it depend on?

JessicaMeixner-NOAA:
Hmm... OKay this could get a little tricky, because it depends on what export GEMPAK_GRIDS='aoc_9km gnh_10m gsh_15m' is and then you'd need to convert that to the proper grid name, and then it'd be something like: 

/lfs/h2/emc/eib/noscrub/rahul.mahajan/RUNTESTS/testing/gwpr3892/COMROOT/C48_S2SW_extended_3892/gfs.20210323/12/products/wave/gridded/global.0p16/gfs.wave.t12z.global.0p16.f003.grib2.idx 

Maybe we just launch them at the end?  Hopefully these jobs run fast enough that it wouldn't slow progress down.

aerorahul:
> Hmm... OKay this could get a little tricky, because it depends on what export GEMPAK_GRIDS='aoc_9km gnh_10m gsh_15m' is and then you'd need to convert that to the proper grid name, and then it'd be something like:
> 
> /lfs/h2/emc/eib/noscrub/rahul.mahajan/RUNTESTS/testing/gwpr3892/COMROOT/C48_S2SW_extended_3892/gfs.20210323/12/products/wave/gridded/global.0p16/gfs.wave.t12z.global.0p16.f003.grib2.idx
> 
> Maybe we just launch them at the end? Hopefully these jobs run fast enough that it wouldn't slow progress down.

@JessicaMeixner-NOAA 
I was able to find a way to get the appropriate wavesbs labels for wavegempak triggers.
I tested it on WCOSS2, and they can be seen here;
`/lfs/h2/emc/eib/noscrub/rahul.mahajan/RUNTESTS/testing/gwpr3892/EXPDIR/C48_S2SW_extended_3892`

Let me know your thoughts.

JessicaMeixner-NOAA:
Thanks @aerorahul that is great!

ush/check_land_input_orography.py
aerorahul: `stype` has been removed.  `vtype` is retained and checked against `land_frac`.

aerorahul:
I ran the tests with `generate_workflows.sh` on gaeaC6 and wcoss, and all passed.
I will run the CI now on Hercules for final testing unless there are more reviewer comments that will require modifications to this PR.

CoryMartin-NOAA (approved):

JessicaMeixner-NOAA (approved):
Thanks again @aerorahul for the updates.

CatherineThomas-NOAA (approved):

ClaraDraper-NOAA (approved):

DavidHuber-NOAA (approved):
Looks great!

PR #3942, opened 2025-08-13
Re-enable the `MAKE_PREPBUFR` capability

### LINKED ISSUE #3941: MAKE_PREPBUFR inadvertently removed?
### What is wrong?

The env var MAKE_PREPBUFR was removed in PR #3683.  This env var was previously used to skip the prepobs step when pre-generated prepbufr files were available. I believe this was inadvertent and this option should be added back in.    

See the if else block that was removed in prep.sh via

https://github.com/NOAA-EMC/global-workflow/pull/3683/files#diff-8c2f2e3d35212976715385d4e373bd0258b8651f9eaca7beaf515adc4bc0e43aR106-L104

and the env var definition in config.base.j2

https://github.com/NOAA-EMC/global-workflow/pull/3683/files#diff-8c2f2e3d35212976715385d4e373bd0258b8651f9eaca7beaf515adc4bc0e43aL40

when MAKE_PREPBUFR is set to NO in config.base the obsprep step is skipped and an existing prepbufr file is used.

All or N/A

develop

### Steps to reproduce

1. set MAKE_PREPBUFR=NO in config.base
2. run workflow
3. obsprep step is run

### Additional information

_No response_

_No response_

aerorahul:
# Description
This PR:
- Resolves #3941
- removes the options `MAKE_NSSTBUFR` and `MAKE_ACFTBUFR`. They are options in name-only.  
- reinstates the option to create or copy `prepbufr`, `prepbufr.acft_profiles` and `nsstbufr` files by running obsproc programs, or copying them from `DMPDIR` or a user-prescribed path `PREPBUFR_DIR`.  `PREPBUFR_DIR` can be passed via `config.prep` by the user.  This was requested for the reanalysis project.

# Type of change

# Change characteristics

# How has this been tested?
Atm-only cycling test on Gaea in progress

# Checklist


CoryMartin-NOAA (dismissed):
Didn't test, looks reasonable though

aerorahul:
`C96_atm3Dvar` test has passed.

dev/parm/config/sfs/config.base.j2
```diff
@@ -318,12 +318,6 @@ if [[ ${DONST} = "YES" ]]; then export FNTSFA="        "; fi
 # The switch to apply SST elevation correction or not
 export nst_anl=.true.
 
-# Make the nsstbufr file on the fly or use the GDA version
-export MAKE_NSSTBUFR="{{ MAKE_NSSTBUFR }}"
-
-# Make the aircraft prepbufr file on the fly or use the GDA version
-export MAKE_ACFTBUFR="{{ MAKE_ACFTBUFR }}"
-
```
jswhit, line 326: Shouldn't MAKE_PREPBUFR be set in here (to "NO" by default)? Maybe also USE_PREPBUFR_FROM_OPS?

aerorahul: @jswhit thanks for your review.

`MAKE_PREPBUFR` can be set to `NO` in `config.prep` 
It is only used in the `prep.sh` job, and by default the behaviour is `YES` such that the `prepbufr`, `prepbufr.acft_profiles` are created with each execution.
If the user wishes to use operational files, they can set `USE_PREPBUFR_FROM_OPS=YES`.  This is also the default when `MAKE_PREPBUFR=NO`.
Finally, if the user wishes to use their own `prepbufr` files, place the following in `config.prep`:
```bash
export MAKE_PREPBUFR="NO"
export USE_PREPBUFR_FROM_OPS="NO"
export PREPBUFR_DIR="/path/to/user/staged/prepbufr_files/"
```
Note however, `prepbufr` and `prepbufr.acft_profiles` should have the prefix `OPREFIX`.

jswhit2: I guess I was suggesting adding MAKE_PREPBUFR=YES (the default) explicitly in config.prep, with what you just described as a comment (by way of documentation). Otherwise, it will be pretty hard for a new user to figure out.

aerorahul: Ok. Yes. I can make that happen.

Here is what I think will work:

In `config.prep`, I can place this:
```bash
export MAKE_PREPBUFR={{ "MAKE_PREPBUFR" | default("YES") }}
if [[ "${MAKE_PREPBUFR} != "YES" ]]; then
  if [[ ${USE_PREPBUFR_FROM_OPS} != "YES" ]]; then
    export PREPBUFR_DIR="/path/to/user/staged/prepbufr_files"
  fi
fi
```

I hope this gives the user enough to figure out what to edit and where.
I'll beef this section up with some inline documentation as well.

Thoughts?

aerorahul: @jswhit Let me know if https://github.com/NOAA-EMC/global-workflow/pull/3942/commits/16c19e7f539d07457f03f7d5fabea33d2287bd6a meets your needs.

jswhit2 (approved):

DavidHuber-NOAA (approved):
Looks good to me.

dev/parm/config/gfs/config.prep
```diff
@@ -15,6 +15,17 @@ export PROCESS_TROPCY=${PROCESS_TROPCY:-NO}
 export TROPCYQCRELOSH="${SCRgfs}/exglobal_atmos_tropcy_qc_reloc.sh"
 export COMINsyn=${COMINsyn:-$(compath.py "${envir}/com/gfs/${gfs_ver}")/syndat}
 
+# Allow users to control the generation or use of either operational or
+# their processed prepbufr, prepbufr.acft_profiles, nsstbufr files
+export MAKE_PREPBUFR="YES"  # Generate prepbufr, etc. files by executing obsproc
```
CoryMartin-NOAA, line 20: Did what you wrote before not work?
```
export MAKE_PREPBUFR={{ "MAKE_PREPBUFR" | default("YES") }}
```

CoryMartin-NOAA: I guess because this isn't a .j2

aerorahul: yes. that. and, we should for completeness, add a `prep` section in the `defaults.yaml` as well.
All of which will take a little more work than this.

CoryMartin-NOAA (approved):

DavidHuber-NOAA:
@TaylorRoper-NOAA noted that he makes use of the MAKE_NSSTBUFR and MAKE_ACFTBUFR options on the cloud to create those BUFR files on the fly.  @aerorahul is the default option to do the same?

aerorahul:
> @TaylorRoper-NOAA noted that he makes use of the MAKE_NSSTBUFR and MAKE_ACFTBUFR options on the cloud to create those BUFR files on the fly. @aerorahul is the default option to do the same?

@TaylorRoper-NOAA 
`MAKE_NSSTBUFR` and `MAKE_ACFTBUFR` are rolled into `MAKE_PREPBUFR`  to create those on the fly.  The default (`YES`) should allow you to create those on the cloud as before.

PR #3953, opened 2025-08-18
sorc: use absolute HOMEgfs_ paths in build scripts

TerrenceMcGuinness-NOAA:
# Description

When running build scripts in `HOMEgfs_/sorc` from within pipeline scripts would sometimes fail because they mostly contain non-full paths to subsequent bash calls relative to $PWD.  This update uses the pattern where `HOMEgfs_` is applied to use full paths within the build scripts so they are not depended on being launched from $PWD.

# Type of change

# Change characteristics

# How has this been tested?

This PR will be tested directly on the pipeline running on Gaea C6

### DIFF
--- a/sorc/build_gdas.sh
+++ b/sorc/build_gdas.sh
@@ -28,6 +28,6 @@ shift $((OPTIND-1))
 BUILD_JOBS="${BUILD_JOBS:-8}" \
 WORKFLOW_BUILD="${WORKFLOW_BUILD:-"ON"}" \
 WORKFLOW_TESTS="${WORKFLOW_TESTS:-"OFF"}" \
-./gdas.cd/build.sh ${_opts} -f -w ${HOMEgfs_}
+"${HOMEgfs_}/gdas.cd/build.sh" ${_opts} -f -w ${HOMEgfs_}
 
 exit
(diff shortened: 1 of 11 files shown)


sorc/build_gdas.sh
```diff
@@ -28,6 +28,6 @@ shift $((OPTIND-1))
 BUILD_JOBS="${BUILD_JOBS:-8}" \
 WORKFLOW_BUILD="${WORKFLOW_BUILD:-"ON"}" \
 WORKFLOW_TESTS="${WORKFLOW_TESTS:-"OFF"}" \
-./gdas.cd/build.sh ${_opts} -f -w ${HOMEgfs_}
+"${HOMEgfs_}/gdas.cd/build.sh" ${_opts} -f -w ${HOMEgfs_}
```
aerorahul:
```suggestion
"${HOMEgfs_}/sorc/gdas.cd/build.sh" ${_opts} -f -w ${HOMEgfs_}
```

aerorahul (commented):
Just 1 comment, otherwise looks ok to me.

DavidHuber-NOAA (approved):
Looks good!  Thanks for the comprehensive cleanup!

PR #3956, opened 2025-08-19
GitLab pipeline support for the Hercules and Orion 

### LINKED ISSUE #3936: Add Orion and Hercules to GitLab PR Case Pipeline

As the Control Manager we need to enable Hercules and Orion RDHPCS systems to the CI/CD GitLab Pipelines so developers can be insured that their updates will not break the build and still run on all tier 1 supported platforms.

Hercules and Orion function in all manner of respects as does Hera and Gaea C6 within the GitLab CI/CD Pipelines.

### Acceptance Criteria

- [ ] All PR Cases for each supported platform run end-to-end and report results to their respective PR comment feeds.

_No response_

### LINKED ISSUE #3955: Robust fail checks in CI pipeline for build step

As a CM we need a consistent single point of entry for building that returns a fail exit code with a list of associated logs corresponding to the fail (a feature that has been supported) that now also needs to be reflected in the GitLab PR CI/CD pipeline.

Updated build scripts with exit codes.
Update pipeline scripts that captures build fails

### Acceptance Criteria

- [ ] On the build side (formally supported) with exit 1 and logs
- [ ] On the pipeline side that captures fail and messages logs to GitHub PR feed

_No response_

TerrenceMcGuinness-NOAA:
# Description

This PR adds GitLab CI/CD pipeline support for the Hercules and Orion HPC platforms to the Global Workflow.

The implementation adds platform-specific configuration files for both Hercules and Orion, updates the GitLab CI pipeline configurations to include test matrices for these platforms, and enhances the CI infrastructure to support GitLab runner deployment on these systems. This follows the existing patterns established for Hera and other supported platforms.

Key changes include:
- Added `dev/ci/platforms/config.hercules` and `dev/ci/platforms/config.orion` with platform-specific paths and settings
- Extended `dev/ci/gitlab-ci-hosts.yml` to include test case matrices for both platforms
- Updated GitLab CI workflow files to support the new platforms
- Added and tested the necessary files and support to create the badges for Orion and Hercules 
- Added capture of artifacts from build fails and pushed them to PR comment feeds with links to appropriate error logs.

image

This expansion allows developers to test workflow changes on a broader range of NOAA HPC systems, improving confidence in cross-platform compatibility and reducing the risk of platform-specific issues in production deployments.

Resolves #3936
Resolves #3955

# Type of change

# Change characteristics
 
# How has this been tested?

CI Infrastructure Testing:
- Validated CI script functionality on target platforms (ran branch directly on GitLab)
- Confirmed proper job routing and runner assignment and CI label updates on GitHub PR
- Ran pipelines from user account and created build fails and fine tuned the[ user output formatting:](https://github.com/TerrenceMcGuinness-NOAA/global-workflow/pull/348#issuecomment-3234797290)
image
Platform Validation:

- Hercules: Verified CI environment setup and basic workflow functionality
- Orion: Confirmed platform detection and CI job execution capabilities
- Cross-platform: Ensured existing CI functionality on Hera remains unchanged

# Checklist


dev/ci/scripts/run_check_gitlab_ci.sh
```diff
@@ -153,9 +153,13 @@ while true; do
   # Get job statistics
   echo "Gather Rocoto statistics"
   # shellcheck disable=SC2312 # We want to use the exit code of the command
-  full_state=$("${HOMEgfs}/dev/ci/scripts/utils/rocotostat.py" -w "${xml}" -d "${db}" -v)
+
+  caseName="${caseName:-${pslot%_*-*}}"
```
aerorahul: Can you explain what the `${pslot%_*_*}` is supposed to capture?

dev/workflow/hosts.py
```diff
@@ -39,6 +39,44 @@ def __str__(self) -> str:
...
+            'actions'
+        ]
+
+        if any(pattern in hostname for pattern in github_hostname_patterns):
+            return True
+
+        # Check for GitHub runner user patterns
+        user = os.getenv('USER', '').lower()
+        if user in ['runner', 'github']:
+            return True
```
aerorahul: What are the chances that these would be exercised?  The [documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/variables) suggests that when in the running `GITHUB-*` variables will be declared.

dev/workflow/hosts.py
```diff
@@ -51,6 +89,14 @@ def detect(self) -> None:
...
             if pw_csp != 'UNKNOWN':
                 self.machine = f"{pw_csp.upper()}PW"
+            else:
+                self.machine = machine_id.upper()
+            return
+
+        # Check if this is a GitHub Actions runner first
+        if self._is_github_runner():
+            print('Detected GitHub Actions runner; defaulting to HERA configuration.')
+            self.machine = 'HERA'
```
aerorahul:
```suggestion
        else:  # 'UNKNOWN'
            # Check if this is a GitHub Actions runner first
            if self._is_github_runner():
                print('Detected GitHub Actions runner; defaulting to HERA configuration.')
            self.machine = 'HERA'
```

dev/workflow/hosts.py
```diff
@@ -94,6 +138,11 @@ def detect(self) -> None:
                 raise ValueError(
                     f'cloud service provider "{pw_csp}" is not supported.')
             self.machine = f"{pw_csp.upper()}PW"
+        else:
+            # Unknown environment - could be development or testing
+            machine = socket.gethostname().upper()
+            print(f'Unknown host environment detected on {machine}; defaulting to HERA configuration.')
+            self.machine = 'HERA'
```
aerorahul: not so clear about this `else`.

aerorahul (commented):
A few comments.
My main comment is that there are a few places where a defaulting to Hera is being performed.  Not sure if that is the best approach.
For e.g. one could just create a path `/scratch1` on the runner (or anywhere else for that matter) that tricks host.py to return Hera.  This would eliminate all the checks in this script. Something to think about.

dev/ci/scripts/run_check_gitlab_ci.sh
TerrenceMcGuinness-NOAA: It backs out the TAG to get the proper Case Name [fome here
](https://github.com/TerrenceMcGuinness-NOAA/global-workflow/blob/594685c0a0e50eb40d909005f1acb4f6846428b7/dev/ci/scripts/utils/ci_utils.sh#L131)

dev/workflow/hosts.py
TerrenceMcGuinness-NOAA: Yeah those are [here](https://github.com/TerrenceMcGuinness-NOAA/global-workflow/blob/594685c0a0e50eb40d909005f1acb4f6846428b7/dev/workflow/hosts.py#L50).

Oh I see what you mean, yes chances are slim to none that we make it this far without a true to exit already.

dev/workflow/hosts.py
TerrenceMcGuinness-NOAA: @DavidHuber-NOAA please review these changes.

dev/workflow/hosts.py
TerrenceMcGuinness-NOAA: @DavidHuber-NOAA Rahul got ahead of your review, looks like we added to much here to our last resort logic.

dev/ci/scripts/run_check_gitlab_ci.sh
DavidHuber-NOAA: @TerrenceMcGuinness-NOAA I think this removes everything after the first `_` and/or `-`.  I'm not sure this is what you want, is it?

```console
> pslot="C96_atm3dvar-123456789abc"
> caseName=${pslot%_*-*}
> echo $caseName
C96
```

dev/workflow/hosts.py
DavidHuber-NOAA: @aerorahul Should this be:

```python
        else:  # 'UNKNOWN'
            # Check if this is a GitHub Actions runner first
            if self._is_github_runner():
                print('Detected GitHub Actions runner; defaulting to HERA configuration.')
                self.machine = 'HERA'
```

Otherwise, `HERA` will be the default rather than `UNKNOWN`, right?

dev/workflow/hosts.py
DavidHuber-NOAA: I'm going to open a hotfix today with corrected checking for Hera/Ursa.  Maybe we can return to checking for the GitLab runner after that hotfix goes in.

dev/ci/scripts/run_check_gitlab_ci.sh
TerrenceMcGuinness-NOAA: That is exactly what it does and it work as expected, what is the issue here?

dev/workflow/hosts.py
TerrenceMcGuinness-NOAA: This is OCB.

dev/workflow/hosts.py
TerrenceMcGuinness-NOAA: @DavidHuber-NOAA Agreed, this was not intended to be here at all.  We were in the middle of an SM session and when you had to cut out to a meeting Rahul picked it up, remember I said I was going make it draft until you worked out host.py

.github/workflows/trigger-gitlab-pipelines.yml
```diff
@@ -131,7 +139,7 @@ jobs:
 
           curl --verbose --request POST \
             --form "token=${{ secrets.GITLAB_TRIGGER_TOKEN }}" \
-            --form "ref=develop" \
+            --form "ref=add_hercules_orion" \
```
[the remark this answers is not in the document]
TerrenceMcGuinness-NOAA: almost forgot - thanks AI

dev/ci/scripts/run_check_gitlab_ci.sh
DavidHuber-NOAA: Aren't you going to have identical case names?  For instance for the tests `C48_S2SW` and `C48_S2SW_gefs`, the `caseName` will be `C48` for both.  You're constructing and exporting `ROCOTOSTAT_LOG_FILE` on the next line.  So doesn't that mean you'll overwrite the log file?

TerrenceMcGuinness-NOAA: But your example is malformed the system intentionally adds the tags with an underscore ahead of time in the from [from here](https://github.com/TerrenceMcGuinness-NOAA/global-workflow/blob/594685c0a0e50eb40d909005f1acb4f6846428b7/dev/ci/scripts/utils/ci_utils.sh#L131) above it goes from left to right and picks up the first "_".

aerorahul: @TerrenceMcGuinness-NOAA 
The result is `C48`.  I think you mean to capture `C48_S2SW` from `C48_S2SW-123456789abc` for e.g.

The issue at the core here, is that there can be multiple `_` in the `PSLOT` name itself, I think.

aerorahul: L131 recaptured here 
```bash
export pslot=${case}_${TAG}
```
`${case}` can and does contain sometimes multiple `_`, so, when going from left to right, as the above comment https://github.com/NOAA-EMC/global-workflow/pull/3956/files#r2289025699 says, is going to stop at `C48` for a `case=C48_S2SW`.  And that is not the `caseName`

TerrenceMcGuinness-NOAA: I thought it goes from left to right from the first instance of underscore. Both of your examples are malformed it will always be:  `C48_S2SW_bla-bla_foo_123456789abcfghij`.  Am I still wrong about the reg ex?

TerrenceMcGuinness-NOAA: Ok, you guys are usually right of course, so this is the correct reg ex then?
`caseName="${pslot%_*}"`

TerrenceMcGuinness-NOAA: Ok updated.  Thanks for your diligence and my apologies.

DavidHuber-NOAA: After looking at a recent run, I see that the construction of `TAG` includes both the hash and a workflow identifier (e.g. `90f10fc1-3517`.  So the `pslot` will be, e.g. `C48_S2SW_90f10fc1-3517`.  And for this case, this does what it is intended to:

```console
> pslot="C48_S2SW_90f10fc1-3517"
> caseName=${pslot%_*-*}
> echo $caseName
C48_S2SW
```

dev/ci/scripts/utils/ci_utils.sh
```diff
@@ -177,6 +177,45 @@ function publish_logs() {
...
+    # First, upload to repo (retain original behavior) if desired
+    # shellcheck disable=SC2027,SC2086
+    ${HOMEgfs_}/dev/ci/scripts/utils/publish_logs.py --file "${files[@]}" --repo ${PR_header} > /dev/null || true
+
+    # For gist, if more than one file use --multiple --format github
+    if (( ${#files[@]} > 1 )); then
+      URL="$("${HOMEgfs_}/dev/ci/scripts/utils/publish_logs.py" --file "${files[@]}" --multiple --format github --gist "${PR_header}")"
+    else
+      URL="$("${HOMEgfs_}/dev/ci/scripts/utils/publish_logs.py" --file "${files[0]}" --gist "${PR_header}")"
+    fi
```
aerorahul:
```suggestion
    if (( ${#files[@]} > 1 )); then
      cmd_args="--multiple --format github"
    fi
    URL="$("${HOMEgfs_}/dev/ci/scripts/utils/publish_logs.py" --file "${files[0]}" ${cmd_args:-} --gist "${PR_header}")"
```

dev/workflow/hosts.py
DavidHuber-NOAA: The hotfix is in.

.gitlab-ci.yml
```diff
@@ -87,6 +87,10 @@ include:
       echo "PIPELINE_TYPE: ${PIPELINE_TYPE}"
       echo "RUN_ON_MACHINES: ${RUN_ON_MACHINES}"
 
+      # This is needed to support PyGitHub not being in Spack-Stack 1.9
```
DavidHuber-NOAA: Could you add a `TODO` here for spack-stack 2.x?

.gitlab-ci.yml
```diff
@@ -155,7 +175,32 @@ include:
...
 
+        error_log_file="sorc/logs/error.logs"
         if [[ "${GFS_CI_RUN_TYPE}" == "pr_cases" && "${PR_NUMBER}" != 0 ]]; then
+          if [[ -f "${error_log_file}" ]]; then
+            pr_error_log_links=$(dev/ci/scripts/utils/ci_utils.sh publish_logs_from_file "PR_${PR_NUMBER}" "${error_log_file}")
+          else
+            pr_error_log_links="(no error.logs file found)"
+          fi
+
+          if [[ -f "${error_log_file}" && -s "${error_log_file}" ]]; then
```
DavidHuber-NOAA: Nitpick: The `-f` test isn't necessary here.  `-s` also tests for existence.

```suggestion
          if [[ -s "${error_log_file}" ]]; then

File: .gitlab-ci.yml [hunk 9]
```diff
@@ -155,7 +175,32 @@ include:
...
+            while IFS= read -r full_log_path; do
+              [[ -n "${full_log_path}" ]] || continue
+
+              if [[ -f "${full_log_path}" && -s "${full_log_path}" ]]; then
+                error_logs_for_gist="${error_logs_for_gist} ${full_log_path}"
+                error_logs_markdown=$(echo -e "${error_logs_markdown}\n${full_log_path}")
+              fi
+            done < "${error_log_file}"
+
+            comment_body=$(printf "**BUILD FAILED** on %s (pipeline ID: %s)\n\nIn directory: \`%s/sorc\`\n\n**Error log files:**\n\`\`\`\n%s\n\`\`\`\n\n🔗 [View Error Logs](%s)\n\n_This failure was detected automatically by global-workflow's CI/CD Pipeline_" "${Machine}" "${CI_PIPELINE_ID}" "${GW_HOMEgfs}" "${error_logs_markdown}" "${pr_error_log_links}")
```
DavidHuber-NOAA: Should we be using non-ascii characters?  i.e. 🔗

.gitlab-ci.yml
```diff
@@ -155,7 +175,32 @@ include:
...
+        
+            ${GH} pr comment ${PR_NUMBER} --repo ${GW_REPO_URL} --body "${comment_body}" || true
```
DavidHuber-NOAA: Why the `|| true`?  I don't think it's necessary since `set -e` isn't being invoked.  It also seems like `PR_NUMBER` and `GW_REPO_URL` should be in double quotes.

dev/ci/gitlab-ci-hosts.yml
```diff
@@ -22,11 +22,17 @@
...
+  - caseName: ["C48_ATM", "C48_S2SW", "C48_S2SWA_gefs", "C48mx500_3DVarAOWCDA", "C48mx500_hybAOWCDA", "C96C48_hybatmDA", "C96C48_hybatmsnowDA", "C96C48_ufs_hybatmDA", "C96C48mx500_S2SW_cyc_gfs", "C96_atm3DVar", "C96_gcafs_cycled", "C96_gcafs_cycled_noDA", "C96mx100_S2S"]
 
 .gaeac6_cases_matrix: &gaeac6_cases
   - caseName: ["C48_ATM", "C48_S2SW", "C48_S2SWA_gefs", "C48mx500_3DVarAOWCDA", "C48mx500_hybAOWCDA", "C96C48_hybatmDA", "C96C48_hybatmsnowDA", "C96C48mx500_S2SW_cyc_gfs", "C96_atm3DVar", "C96_gcafs_cycled", "C96_gcafs_cycled_noDA", "C96mx100_S2S"]
 
+.orion_cases_matrix: &orion_cases
+  - caseName: ["C48_ATM", "C48_S2SW", "C48_S2SWA_gefs", "C96C48_hybatmDA", "C96C48mx500_S2SW_cyc_gfs", "C96_atm3DVar", "C96mx100_S2S"]
+
+.hercules_cases_matrix: &hercules_cases
+  - caseName: ["C48_ATM", "C48_S2SW", "C48_S2SWA_gefs", "C48mx500_3DVarAOWCDA", "C48mx500_hybAOWCDA", "C96C48_hybatmDA", "C96C48mx500_S2SW_cyc_gfs", "C96_atm3DVar", "C96mx100_S2S"]
```
DavidHuber-NOAA, line 34: Should these be determined dynamically by checking the `skip_ci_on_hosts` list in the `dev/ci/cases/*.yaml` files?  Maybe a feature for later.

dev/ci/gitlab-ci-hosts.yml
```diff
@@ -81,6 +87,60 @@ run_experiments-gaeac6:
...
+  variables:
+    machine: orion
+  tags:
+    - orion
+  parallel:
+    matrix: *orion_cases
+  needs:
+    - build-orion
+  rules:
+    - if: $PIPELINE_TYPE == "pr_cases" && ($CI_PIPELINE_SOURCE == "trigger" || $CI_PIPELINE_SOURCE == "schedule") && ($RUN_ON_MACHINES =~ /\borion\b|all/)
```
DavidHuber-NOAA, line 102: Can these be templated?  e.g. `setup_experiments_on_host`?  Probably also a feature for later.

dev/ci/scripts/run_check_gitlab_ci.sh
```diff
@@ -144,42 +164,35 @@ rc=99
...
+  export ROCOTOSTAT_LOG_FILE="${RUNTESTS}/EXPDIR/${pslot}/logs/${caseName}_rocotostat.log"
+  source <("${HOMEgfs}/dev/ci/scripts/utils/rocotostat.py" -w "${xml}" -d "${db}" --declare --thread-logging) || true
 
-  echo -e "(${pslot} on ${MACHINE_ID^})\n\tTotal Cycles: ${CYCLES_TOTAL}\n\tNumber Cycles done: ${CYCLES_DONE}\n\tState: ${ROCOTO_STATE}"
+  echo -e "\tCompleted Cycles: ${CYCLES_DONE}/${CYCLES_TOTAL}
+  \tCompleted Jobs  : ${JOBS_DONE}/${JOBS_TOTAL}
+  \tState           : ${ROCOTO_STATE}"
 
-  if [[ ${error_stat} -ne 0 ]]; then
+  if [[ ${ROCOTO_STATE} == "FAIL" ]]; then
```
DavidHuber-NOAA: Should this also test for `STALE`, `UNKNOWN`, etc?

dev/ci/scripts/utils/ci_utils.sh
```diff
@@ -177,6 +177,44 @@ function publish_logs() {
...
+  # output formatting.
+  # Usage: publish_logs_from_file <ID> <list_file>
+  local PR_header="$1"
+  local list_file="$2"
+  local files=()
+
+  # Read the list file and build an array of existing full paths
+  while IFS= read -r line || [[ -n "${line}" ]]; do
+    # skip empty lines
+    [[ -z "${line// /}" ]] && continue
```
DavidHuber-NOAA: Prefer this as an if block:

```suggestion
    if [[ -z "${line// /}" ]]; then
        continue
    fi
```

dev/ci/scripts/utils/ci_utils.sh
```diff
@@ -177,6 +177,44 @@ function publish_logs() {
...
+  # and publishes them using publish_logs.py. When more than one valid file is
+  # found, call the python utility with --multiple --format github for gist
+  # output formatting.
+  # Usage: publish_logs_from_file <ID> <list_file>
+  local PR_header="$1"
+  local list_file="$2"
+  local files=()
+
+  # Read the list file and build an array of existing full paths
+  while IFS= read -r line || [[ -n "${line}" ]]; do
```
DavidHuber-NOAA, line 192: I don't think the `|| [[ -n "${line}" ]]` is necessary here, or am I missing something?  It seems like if `read -r line` fails, then we shouldn't keep trying to work with the previous `${line}`.

dev/ci/scripts/utils/ci_utils.sh
```diff
@@ -177,6 +177,44 @@ function publish_logs() {
...
+    [[ -z "${line// /}" ]] && continue
+    if [[ -f "${line}" ]]; then
+      files+=("${line}")
+    else
+      echo "File ${line} does not exist"
+    fi
+  done < "${list_file}"
+
+  local URL=""
+  if (( ${#files[@]} )); then
```
DavidHuber-NOAA: Just noting, we prefer `[[ ]]` logic for operational code because it offers better logging.  I'm fine with this here.

dev/ci/scripts/utils/ci_utils.sh
```diff
@@ -177,6 +177,44 @@ function publish_logs() {
...
+    # First, upload to repo (retain original behavior) if desired
+    # shellcheck disable=SC2027,SC2086
+    ${HOMEgfs_}/dev/ci/scripts/utils/publish_logs.py --file "${files[@]}" --repo ${PR_header} > /dev/null || true
```
DavidHuber-NOAA: I don't think you want to disable either 2027 or 2086 here.  `PR_header` and `HOMEgfs_` are simple strings:

```suggestion
    "${HOMEgfs_}/dev/ci/scripts/utils/publish_logs.py" --file "${files[@]}" --repo "${PR_header}" > /dev/null || true
```

dev/ci/scripts/utils/rocotostat.py
```diff
@@ -425,13 +428,12 @@ def is_stalled(rocoto_status):
             else:
                 print(f'Number of {status} : {rocoto_status[status]}')
 
-    if args.export:
+    if args.declare:
         for status in rocoto_status:
-            print(f'export {status}={rocoto_status[status]}')
+            print(f'declare {status}={rocoto_status[status]}')
```
DavidHuber-NOAA, line 433: Is `declare` necessary here?  Would `{status}={rocoto_status[status]}` be adequate since `declare` doesn't export by default, nor doe assignment?  I'm fine with keeping it, just wondering why `declare` is used.

DavidHuber-NOAA (changes requested):
A few question and suggestions.

dev/ci/gitlab-ci-hosts.yml
TerrenceMcGuinness-NOAA: They _**are**_ templated [here](https://github.com/NOAA-EMC/global-workflow/blob/d7e815218b4e117bda44519736808c717170eb19/dev/ci/gitlab-ci-cases.yml#L17)

dev/ci/gitlab-ci-hosts.yml
TerrenceMcGuinness-NOAA: Yes I was going to suggest to make it a Unit Test in the event that some one adds a yaml file that is not in this list.

.gitlab-ci.yml
TerrenceMcGuinness-NOAA: `set -e` is set as the default behavior of all shell jobs.

.gitlab-ci.yml
TerrenceMcGuinness-NOAA: Ok I can take those out.

.gitlab-ci.yml
TerrenceMcGuinness-NOAA: Ok, added the quotes.

.gitlab-ci.yml
TerrenceMcGuinness-NOAA: Done.

dev/ci/scripts/utils/ci_utils.sh
TerrenceMcGuinness-NOAA: The (( )) version is actually more idiomatic for numeric checks,

dev/ci/scripts/utils/ci_utils.sh
TerrenceMcGuinness-NOAA: It's a defensive programming practice to handle files that might not be perfectly formatted.

dev/ci/scripts/utils/ci_utils.sh
TerrenceMcGuinness-NOAA: Oh I used to have to not use the quotes on the files argument.  Let me try to remove the SC's them again and add the quotes on the first part.

dev/ci/scripts/utils/rocotostat.py
TerrenceMcGuinness-NOAA: Yeah there is no specific need to add or remove the "declare" BASH quantifier.

aerorahul (dismissed):
looks good.  Approve pending other reviewer's approval.

dev/ci/scripts/run_check_gitlab_ci.sh
TerrenceMcGuinness-NOAA: Yes, indeed, we need to check for STALLED!  And for UNKNOWN, and UNAVAILABLE when it could not recover from those (not using exit code in this update).

TerrenceMcGuinness-NOAA (commented):
Need some code changes still after catching another review comment regarding the main run check's responses to `rocotostaty.py.`

dev/ci/scripts/utils/ci_utils.sh
DavidHuber-NOAA: I think I would prefer a logical operator here for clarity:

```suggestion
  if (( ${#files[@]} > 0 )); then
```

dev/ci/scripts/utils/ci_utils.sh
DavidHuber-NOAA: OK, thanks for the clarification.

TerrenceMcGuinness-NOAA (commented):
Looks good, should be ready for merge (there are no links to the log files because the pipeline code that updates PYTHONPATH needs to be in the develop branch)

DavidHuber-NOAA:
@TerrenceMcGuinness-NOAA Are the Gaea C6 failures expected?

dev/ci/scripts/utils/ci_utils.sh
DavidHuber-NOAA: @TerrenceMcGuinness-NOAA Could you address this suggestion, please.

dev/ci/gitlab-ci-hosts.yml
DavidHuber-NOAA: OK, why isn't that template used?

TerrenceMcGuinness-NOAA:
Of course they are, the test is for the pipeline not the workflow.  Looks like archive jobs failed to access HPSS or something, but the pipeline flagged the error and messaged the GitHub feed.  The lack of actual links to the the error logs on GitHub are not present is because [PYTHONPATH updates ](https://github.com/TerrenceMcGuinness-NOAA/global-workflow/blob/fb5605f35ec7ef2adb264d4dd02e7b58a0699c3c/.gitlab-ci.yml#L92)are in the pipeline scripts here in this PR and thus do not run on the GitLab server yet but all the bash scripts do ([gfs_arch_tars.log](https://gist.github.com/emcbot/361fa6d1ee76111c8109fd12db26bc59))

dev/ci/gitlab-ci-hosts.yml
TerrenceMcGuinness-NOAA: David, `set_up_eperiments-orion` **_is_** an instantiation of the template `.setup_experiment_template`

DavidHuber-NOAA (approved):

PR #3957, opened 2025-08-20
HOTFIX: enable compute builds on Hera; always build the GDASApp locally

DavidHuber-NOAA:
# Description
These hotfixes are being incorporated:
1. Fix the detection of Hera/Ursa and the GitHub runner
2. The GDASApp will now always build on the head node
3. Related to 1, `detect_machine.sh` will be copied from the global workflow to all submodules when running compute builds
4. The call to `build_gdas.sh` has been removed from ci_utils.sh as it is now handled in `build_compute.sh`

2 and 3 are temporary pending a more permanent solution.

# Type of change

# Change characteristics
- Is this a breaking change (a change in existing functionality)? YES (compute builds are disabled for the GDASApp)

# How has this been tested?
Build test on Hera

# Checklist


dev/workflow/build_compute.py
```diff
@@ -184,6 +184,13 @@ def main(*argv):
     user_yaml_dict = AttrDict(parse_yaml(user_inputs.yaml))
     build_specs = get_build_specs(user_yaml_dict, host_specs)
 
+    # Temporarily prevent the GDASApp from building on the compute node
+    # TODO restore the GDASApp when it can be built on compute nodes again and/or 'compute' builds are enabled on head nodes.
+    #      See issue 3933
+    if "gdas" in build_specs.build:
+        print("Skipping the compute-node build of GDASApp.  This is being built locally.")
+        build_specs.build.pop("gdas")
```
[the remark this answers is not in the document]
DavidHuber-NOAA: The key check is already being performed.

DavidHuber-NOAA:
Starting manual CI on Hera.

TerrenceMcGuinness-NOAA (dismissed):
Looks great. Practically like removing TODO from CI/CD utils and the rapid response in solving the issue.

sorc/build_ufs_utils.sh
```diff
@@ -22,6 +22,8 @@ while getopts ":j:dv" option; do
 done
 shift $((OPTIND-1))
 
+source "${HOMEgfs_}/sorc/ufs_model.fd/tests/detect_machine.sh"
```
aerorahul:
```suggestion
source "${HOMEgfs_}/ush/detect_machine.sh"
```

sorc/build_upp.sh
```diff
@@ -24,6 +24,8 @@ while getopts ":dj:v" option; do
 done
 shift $((OPTIND-1))
 
+source "${HOMEgfs_}/sorc/ufs_model.fd/tests/detect_machine.sh"
```
aerorahul:
```suggestion
source "${HOMEgfs_}/ush/detect_machine.sh"
```

DavidHuber-NOAA:
FYI @RussTreadon-NOAA @CoryMartin-NOAA this PR will change the way GDASApp is built so build_compute.sh can be used again.

CoryMartin-NOAA:
Thanks @DavidHuber-NOAA as an FYI I'm hopeful we can remove GDASApp's dependence on the repo that has been the cause of our new compute build issues "soon"

sorc/build_ufs_utils.sh
DavidHuber-NOAA: Oops.

sorc/build_upp.sh
DavidHuber-NOAA: Fixed.

ush/detect_machine.sh
```diff
@@ -107,3 +108,5 @@ elif [[ -d /data/prod ]]; then
 else
   echo WARNING: UNKNOWN PLATFORM 1>&2
 fi
+
+export MACHINE_ID
```
aerorahul: I don't think this is necessary.
Also, this will not be executed if the condition on L68 is satisfied.

DavidHuber-NOAA: It is necessary.  If it is not exported then the log for `build_gfs_utils.sh` shows the following when executing e.g. [gfs_utils.fd/ush/detect_machine.sh#12](https://github.com/NOAA-EMC/gfs-utils/blob/4f2806c15dc49359e0eb02b497edb00cb2e0c2f4/ush/detect_machine.sh#L12):

```
 [[ -n '' ]]
```

But when exported the log shows
```
[[ -n hera ]]
```

I will export on line 68 as well.

aerorahul: Should we use `declare -rx MACHINE_ID` to make sure this does not get modified?

DavidHuber-NOAA: Sure, that seems like a good idea.

DavidHuber-NOAA: Done.

DavidHuber-NOAA:
Running manual CI on Hera.

DavidHuber-NOAA:
Also launching on C6 for good measure.

DavidHuber-NOAA:
The upp failed to compile on C6 with the following error when trying to load modules:

```console
> /gpfs/f6/drsa-precip3/world-shared/global/CI/GITLAB/pr_cases_3957_90f10fc1_3477/global-workflow/sorc/ufs_model.fd/FV3/upp/tests/compile_upp.sh
Running "module reset". Resetting modules to system default. The following $MODULEPATH directories have been removed: None
0x00007fa79cb64be6 in wait4 () from /lib64/libc.so.6
$1 = 0
$2 = -1
[Inferior 1 (process 2073441) detached]
Modulefile does not exist for machine gaeac6, compiler intel
```

I'm not sure what exactly happened here, but will debug.  The GitLab scripts failed to capture the error, which resulted in the jobs to crash later on due to missing/bad links.  I think @TerrenceMcGuinness-NOAA is already aware of this and working on a solution.

DavidHuber-NOAA:
The UPP identifies Gaea C6 as `gaea`, not `gaeac6`, which is why sourcing the global-workflow's `detect_machine.sh` does not work for that system.  This should be fixed at the UPP level since C5 also exists, but I will put in a stop-gap fix in `build_upp.sh`.

DavidHuber-NOAA:
Opened issue https://github.com/NOAA-EMC/UPP/issues/1308 to address this on the UPP side.

DavidHuber-NOAA:
I'm running compute builds on C6 now.  When complete and verified, I will relaunch GitLab CI on C6.

DavidHuber-NOAA:
Manual CI passed overnight on Hera.

Relaunching CI on C6.

PR #4018, opened 2025-09-05
Hotfix/pipeline pythonpath

TerrenceMcGuinness-NOAA:
The fix for updating the PYTHONPATH in the pipeline bash section of the build job had to also be translated into the run check bash script (due to **PyGitHub** not making it to the latest **Spack Stack** release.


dev/ci/scripts/run_check_gitlab_ci.sh
```diff
@@ -135,6 +135,8 @@ EOF
 # Source modules and setup logging
 echo "Source modules."
 source "${HOMEgfs}/dev/ush/gw_setup.sh"
+PYTHONPATH="${PYTHONPATH}:${HOME}/.local/lib/python3.11/site-packages"
+echo "Updated PYTHONPATH: ${PYTHONPATH}"
```
aerorahul: Not a problem, but just checking if `3.11` is the version on all platforms running this.
Instead, should this be delegated to` config.<platform>` and included from there?
Just a thought.

aerorahul (approved):

Issue #4001, opened 2025-09-03
metplus jobs are unsupported on C6 and MSU and fail silently


DavidHuber-NOAA:
### What is wrong?

`gfs_metp[g2g1, g2o1, pcp1]` jobs fail on Orion, Hercules, and C6 because the paths `HOMEMET` and `HOMEMETplus` are not defined.  This is because met/metplus v9.1.3/3.1.1 are not available.

### What should have happened?

Modules are available to run the metp jobs on supported platforms and the jobs run correctly.

### What machines are impacted?

Gaea-C6, Hercules, Orion

### What global-workflow hash are you using?

https://github.com/NOAA-EMC/global-workflow/commit/4764ea1d0f6b3d55439516b8d9a00f46ad667b8b

### Steps to reproduce

Launch a metg2o1 job on MSU or C6 and observe that `HOMEmet` and `HOMEmetplus` are undefined resulting in 0-sized statistics files.

### Do you have a proposed solution?

Either load spack-stack 1.6.0 at metp job runtime or migrate EMC_verif-global to met/metplus 12.0.x/6.0.x (the versions available in spack-stack 1.9.3+).

DavidHuber-NOAA:
Related to #3992.

DavidHuber-NOAA:
These jobs have been disabled on C6 and MSU.  The capability will be reenabled following the upgrade to MET/METplus in #4127.  Closing as a duplicate of #4127.

PR #4058, opened 2025-09-12
HOTFIX: Disable metp jobs on MSU and C6

DavidHuber-NOAA:
C6, Hercules, and Orion currently do not support EMC_verif-global due to the unavailability of MET/METplus v9.1.3/v3.1.1, which are required by the package.  This is a short-term stopgap PR to disable those jobs from running on MSU and C6 while a fix is put in place.

This also removes `DO_*` options from host yaml files that are ignored/overridden.

Refs #4001


dev/workflow/hosts/ursa.yaml
```diff
@@ -27,6 +27,5 @@ CHGRP_CMD: 'chgrp rstprod'
 # Features
 SUPPORTED_RESOLUTIONS: ['C1152', 'C768', 'C384', 'C192', 'C96', 'C48']
 DO_ARCHCOM: 'YES'
-DO_AWIPS: 'NO'
 MAKE_NSSTBUFR: 'NO'
 MAKE_ACFTBUFR: 'NO'
```
aerorahul:
```suggestion
```
Pretty sure these were removed from Hera with #3942

aerorahul (dismissed):
looks good.  just one suggestion for `ursa.yaml`.

DavidHuber-NOAA:
Thanks for the suggestion @aerorahul.

PR #4091, opened 2025-09-23
Fix for exglobal_cleanup.sh so that it won't choke on empty strings.

### LINKED ISSUE #4090: Cleanup step in GCAFS fails due to empty string
### What is wrong?

The script, exglobal_cleanup.sh, contains this line

find "${directory}" -type f -not \( ${find_exclude_string} \) -ignore_readdir_race -delete

which fails when find_exclude_string is empty. The proposed change 

find "${directory}" -type f -ignore_readdir_race -delete

has already been tested. 

The cleanup script should finish without failure. It fails due to the empty string. 

Gaea-C6

commit f294f5e39437b8bddbe08f4b0238b43d2431e046

### Steps to reproduce

1. Clone and build the global workflow with GCAFS and GDAS.
2. Run a cycling experiment 4 cycles (failure is in the 4th cycle).

### Additional information

_No response_

The proposed solution is shown at the top. A PR will be issued with this change shortly.

andytangborn:
If merged, this PR will fix a bug in exglobal_cleanup.sh that causes it to fail when it encounters an empty string. 

# Description

GCAFS fails in the cleanup step, specifically the script exglobal_cleanup.sh crashes due to an empty string. The fix changes one line to remove the string from the delete command, and this happens on the 4th cycle. This is why the CI test does not fail.  Cycling experiments are now running without failure. 

  Describe your changes. Focus on the *what* and *why*. The *how* will be evident from the changes. In particular, be sure to note any interface changes, such as command line syntax, that will need to be communicated to users.

  Resolves #4090  
 
-->

# Type of change
- [x ] Bug fix (fixes something broken)

# Change characteristics

  - [ ] EMC verif-global 
  - [ ] GDAS 
  - [ ] GFS-utils 
  - [ ] GSI 
  - [ ] GSI-monitor 
  - [ ] GSI-utils 
  - [ ] UFS-utils 
  - [ ] UFS-weather-model 
  - [ ] wxflow 

# How has this been tested?
- Clone and build on  gaea c6. It has not yet been run on other machines. 

# Checklist - Note: most of these don't apply. CI tests don't encounter this bug.
- [x ] Any dependent changes have been merged and published
- [x ] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [x ] I have documented my code, including function, input, and output descriptions
- [x ] My changes generate no new warnings
- [x ] New and existing tests pass with my changes
- [x ] This change is covered by an existing CI test or a new one has been added
- [x ] Any new scripts have been added to the .github/CODEOWNERS file with owners
- [ x] I have made corresponding changes to the system documentation if necessary


scripts/exglobal_cleanup.sh
```diff
@@ -39,10 +39,10 @@ function remove_files() {
     find_exclude_string="${find_exclude_string[*]/%-or}"
     # Remove all regular files that do not match
     # shellcheck disable=SC2086
-    find "${directory}" -type f -not \( ${find_exclude_string} \) -ignore_readdir_race -delete
```
CoryMartin-NOAA, line 42: @andytangborn I'm not sure @DavidHuber-NOAA or @aerorahul are going to accept removing this outright. We need a different solution to solve this issue, I think it came up before, where we check if that variable is empty or not and then have an if-block, is that right?

andytangborn: I did have something earlier that was a bit more complicated.  It might have done something like that. But let me see if I can come up with a solution as you describe.

CoryMartin-NOAA (approved):
looks good

scripts/exglobal_cleanup.sh
DavidHuber-NOAA: This is a good solution.  Thanks @andytangborn!

DavidHuber-NOAA (approved):

DavidHuber-NOAA:
@andytangborn have you been able to run a GCAFS test with the latest version?  If not, I can launch automated CI on Hera.

CoryMartin-NOAA:
@DavidHuber-NOAA FYI, I don't think this issue shows up in the CI tests, it requires several days of cycles before it runs into this condition.

DavidHuber-NOAA:
@CoryMartin-NOAA Yes, of course you're right.  OK, I won't launch tests.

PR #4179, opened 2025-10-24
Update checks for MOM6 restarts when performing a re-run on failure

### LINKED ISSUE #3822: Restart runs are not ensuring MOM6 output has completed writing.
### What is wrong?

We have had multiple failures of ocean post caused when the forecast job fails (for various reasons, reasons unrelated to this issue here).  The underlying issue is that we check on restarts if the MOM6 restart file is written, however that does not mean that the last time-averaged output has been written.  Therefore the model successfully restarts from the restart and then we start getting ocean post failures that we cannot fix because the ocean output is not fully written. 

We should restart the model where we have both output and restarts up to that point in time. 

All or N/A

develop  54dc7d2931317be11cbaa6b28e036d7e6cd68a11

### Steps to reproduce

Potentially hard to reproduce because you essentially have to have the model crash when the MOM6 restart file is written and the MOM6 output has not finished writing.  (Note, if restart write times are weird intervals of averaged output, this could be an issue). 

### Additional information

This is a rewording of the issue originally reported here: https://github.com/NOAA-EMC/global-workflow/issues/3773 

After discussions with @jiandewang and @sanAkel  - we have determined that the writing of restarts and output are asynchronous so the restart being written out is not an indicator for the output having being written out. 

If needed or it would help, @sanAkel has offered to modify mom code to echo when:
a. output is written out - as in https://github.com/NOAA-EMC/MOM6/pull/148
b. restart has (also) been written out - to another ascii file (different from a) that can be queried, say be the workflow. 

Several options: 
* Go back one restart (this however requires more restarts to be written to make sure we can restart within a window and potentially restarting earlier than you have to in other cases. 
* Add a check to make sure the ocean model also has written out it's output.  This will be a several step process: 

1. Update MOM6 to write something to the log file to indicate that the model output has been written (https://github.com/NOAA-EMC/MOM6/issues/151) 
2. Ensure the MOM6 update makes its way to global-workflow after it's in the ufs-weather-model 
3. In g-w when we check to see if a restart exists for determining if we can use that restart, we should also do that same log-check for the MOM6 output as well.  This code is here: https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/forecast_det.sh#L83-L86

aerorahul:
# Description
This PR:
- Resolves #3822 
- employs additional checks on MOM6 state when restarting upon re-run for robustness.  This check is skipped when `FHOUT_OCN > 6` such as in the `SFS` runs where `FHOUT_OCN=24`.  This causes problems noted in the issue comment [here](https://github.com/NOAA-EMC/global-workflow/issues/3822#issuecomment-3461900670)

This PR also:
- has a minor fix in `ush/preamble.sh` that prints the `Begin` date time in human readable format similar to `End` date time message.

# Type of change

# Change characteristics

# How has this been tested?
Tests 1, 2, 3 below have FHMAX_GFS=384 and restart_interval_gfs=48

1. Ran C48_S2SW case forecast out to FHMAX_GFS without stopping
2. Start running C48_S2SW forecast
  a. kill forecast job at fhr=126.  Restarts are available at hrs 48, 96
  b. re-submit forecast job.  The job picks the restarts at hr 96 and completes
3. Start running C48_S2SW forecast
  a. kill forecast job at fhr = 102.  Restarts are available at hrs 48, 96
  b. re-submit forecast job.  The job picks restarts at hr 48 (ocean forecast history file at 105 is not available, rendering restart at hr 96 invalid) and completes.
4. Ran C48_S2SW test start to end.
5. Ran C48_S2SWA_gefs test to test the segment functionality.

# Checklist


ush/forecast_det.sh
```diff
@@ -80,9 +80,22 @@ UFS_det(){
...
         cmeps_rst_ok="NO"
       fi
-      if [[ ! -f "${DATArestart}/MOM6_RESTART/${rdate:0:8}.${rdate:8:2}0000.MOM.res.nc" ]]; then
       # TODO: add checks for other MOM6 restarts as well
+      if [[ ! -f "${DATArestart}/MOM6_RESTART/${rdate:0:8}.${rdate:8:2}0000.MOM.res.nc" ]]; then
         mom6_rst_ok="NO"
+      else
+        # Also check for MOM6 history file availability
+        # TODO: Need to adapt for SFS where averaging period may be different, need to generalize
+        hdate=$(date -u -d "${rdate:0:8} ${rdate:8:2}:00:00 + 3 hours" +"%Y_%m_%d_%H")  # MOM6 history is averaged 6 hrs
```
JessicaMeixner-NOAA: Should this be generalized now with FHOUT_OCN ? 

Also - This should have some sort of if not gdas/enkfgdas because the output is different for those runs.  Should those runs have different checks here?

aerorahul: Generalized to use `FHOUT_OCN`.  I had committed what I had on Friday before leaving work, and a TODO to remind me about it when I pick it up on Monday.
2f1650f addresses the first part of this comment.

`RERUN` is not available for `RUN = gdas|enkfgdas` as their forecast lengths are less than 12 hours (15 if you count IAU).
https://github.com/aerorahul/global-workflow/blob/2f1650f2eb7237ab95373e50cb081097545781d0/ush/forecast_det.sh#L38-L42
`enkfgfs` does not have a forecast segment.
I'll test this for a `RUN=gefs` with segments to ensure this change does not break that.

aerorahul:
GEFS test passed. 
Merged `develop` after renaming PR.
Opening up for review.

JessicaMeixner-NOAA (dismissed):
This resolves the open issue.

jiandewang (dismissed):

DavidHuber-NOAA (dismissed):

aerorahul:
I did not test the SFS case in my testing.  I will do that and report back.    SFS has a `restart_interval_gfs=12` and does 24hr averaging of the ocean output.  The checks for restarts are probably either not enough or too long.

aerorahul:
See [comment](https://github.com/NOAA-EMC/global-workflow/issues/3822#issuecomment-3461900670) in the issue #3822 and advise if the approach there is acceptable.

ush/forecast_det.sh
```diff
@@ -80,9 +84,26 @@ UFS_det(){
       if [[ ! -f "${DATArestart}/CMEPS_RESTART/ufs.cpld.cpl.r.${rdate:0:4}-${rdate:4:2}-${rdate:6:2}-${seconds}.nc" ]]; then
         cmeps_rst_ok="NO"
       fi
-      if [[ ! -f "${DATArestart}/MOM6_RESTART/${rdate:0:8}.${rdate:8:2}0000.MOM.res.nc" ]]; then
       # TODO: add checks for other MOM6 restarts as well
+      if [[ ! -f "${DATArestart}/MOM6_RESTART/${rdate:0:8}.${rdate:8:2}0000.MOM.res.nc" ]]; then
         mom6_rst_ok="NO"
+      else
+        # Also check for MOM6 history file availability
+        # TODO: SFS runs with 24-hr averaging of ocean output, which causes issues with restart checks,
```
DavidHuber-NOAA, line 92: @aerorahul Can you open an issue for this?

ush/forecast_det.sh
```diff
@@ -80,9 +84,26 @@ UFS_det(){
...
+        # TODO: so we will skip them for now, and revisit this logic later
+        if [[ ${FHOUT_OCN} -le 6 ]]; then
+          fhout_ocn_by_2=$((FHOUT_OCN / 2))
+          hdate=$(date -u -d "${rdate:0:8} ${rdate:8:2} + ${fhout_ocn_by_2} hours" +"%Y%m%d%H")
```
DavidHuber-NOAA, line 96: nitpick:
```suggestion
          fhout_ocn_half=$((FHOUT_OCN / 2))
          hdate=$(date -u -d "${rdate:0:8} ${rdate:8:2} + ${fhout_ocn_half} hours" +"%Y%m%d%H")
```

DavidHuber-NOAA (approved):
Looks good.  Just a nitpick on variable name.

DavidHuber-NOAA:
Launching tests on C6.

ush/forecast_det.sh
aerorahul: https://github.com/NOAA-EMC/global-workflow/issues/4199 has been opened.

DavidHuber-NOAA: Thanks!

ush/forecast_det.sh
aerorahul: I'll leave it as is for now since this will get revamped when https://github.com/NOAA-EMC/global-workflow/issues/4199 gets fixed. I am expecting soon-ish, with some work in the UFS-weather-model.

DavidHuber-NOAA:
All tests passed on C6.  Merging.

PR #4183, opened 2025-10-27
GFSv16.3.28 - annual CO2 update

DavidHuber-NOAA:
# Description
This update is for the already-in-production GFS v16.3.28, which includes updates CO2 fix files.  There are no code changes.

# Type of change

# Change characteristics

# How has this been tested?
Tested by SPA Xiaoxue Wang.

# Checklist

### DIFF
--- a/ush/ozn_xtrct.sh
+++ b/ush/ozn_xtrct.sh
@@ -4,26 +4,27 @@
 #
 #  This script performs the data extraction from the oznstat
 #  diagnostic files.  The resulting data (*.ieee_d) files, GrADS
-#  control files and stdout files will be moved to the
-#  $TANKverf_ozn.
+#  control files and stdout files will be moved to the 
+#  $TANKverf_ozn.  
 #
-#  Calling scripts must define:
+#  Calling scripts must define: 
 #	$TANKverf_ozn
 #	$HOMEoznmon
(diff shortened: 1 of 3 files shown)


ush/ozn_xtrct.sh
aerorahul, line 1: Why did this file change? Looks like whitespace changes.

DavidHuber-NOAA: This is a copy from the production directory.  I believe this was done by a SPA, but I'm not sure.

aerorahul: ok. Can we not commit this? NCO does not follow version control on upgrades -- they just copy the previous operational version and make local mods and call it a new version.  When they eventually get a release, they will get a correct script.

DavidHuber-NOAA: Sure, I'll revert.

aerorahul: That directory also has `chmod` changes -- NCO turned every file in the GFS directory to have executable permissions.  We are not bringing that in either.

DavidHuber-NOAA: Done.

aerorahul (approved):
lgtm

Issue #4231, opened 2025-11-06
Reduce surface increment files that are regridded


tsga:
### What new functionality do you need?

The Gaussian to FV3 regridding of soil increments currently copies and calls the regridding code for files that are not necessarily needed for a given cycle/run (e.g., when land IAU is on, the only non-IAU increment needed is for the gaussian surface analysis).

Changes to GW scripts to to turn on/off which files to be regirded could save significant run time. 

### Acceptance Criteria

Relevant CI tests with Soil DA pass.

### Suggest a solution (optional)

Make revisions to relevant scripts (Start with the latest regridding code from https://github.com/NOAA-EMC/global-workflow/pull/4194)  to 
- turn off copying and regridding of increments not relevant for a cycle/run
- rename the i006 file for the IAU case to avoid overwriting of the non-IAU file
- regird the correct non-IAU increment for Gaussian sfc analysis
- modify the gaussian sfc_analysis with "add increment"

CoryMartin-NOAA:
Also noting here that currently the workflow overwrites a i006 file with one time step with a i006 file with 3 time steps, the file with 3 time steps should be renamed to 1) not overwrite and 2) more clearly explain the file

CoryMartin-NOAA:
We also need to apply the i006 file with one time step to the sfc restart files as part of the gaussian_sfcanl generation process. @tsga if you would like me to work on any of these (perhaps I can take the lead on the increment add part?) let me know

CatherineThomas-NOAA:
Thanks @tsga and @CoryMartin-NOAA.  Note that some of the Land IAU IF statements are adjusted in https://github.com/NOAA-EMC/global-workflow/pull/4194.

tsga:
Thank you @CoryMartin-NOAA and @CatherineThomas-NOAA . I have updated the todo list. @CoryMartin-NOAA I will let you know once I looked at it.

CatherineThomas-NOAA:
Since https://github.com/NOAA-EMC/global-workflow/pull/4238 was merged, is this issue now considered complete?

tsga:
@CatherineThomas-NOAA yes it is, thank you!

PR #4238, opened 2025-11-12
Add soil increments to Gaussian SFC Analysis

tsga:
# Description

The gaussian sfc analysis do not contain soil DA increments for the current cycle if land IAU is on. Previously, the increments were added through global cycle with no land IAU.

This PR adds the regirded soil DA increments on FV3 grid to the sfc analysis variables before they are converted to Gaussian fields. Related script changes ensure only needed surface increments are being re-gridded, resolve file name conflict for soil increments, and update the soil increment archiving.

Resolves NOAA-EMC/global-workflow/issue#https://github.com/NOAA-EMC/global-workflow/issues/4231
Refs https://github.com/NOAA-EMC/gfs-utils/issues/160

# Type of change

# Change characteristics

  - [ ] EMC verif-global 
  - [ ] GDAS 
  - [x] GFS-utils NOAA-EMC/gfs-utils#161
  - [ ] GSI 
  - [ ] GSI-monitor 
  - [ ] GSI-utils 
  - [ ] UFS-utils 
  - [ ] UFS-weather-model 
  - [ ] wxflow 

# How has this been tested?

On Ursa, ran the CI test C96C48_hybatmsoilDA for this branch and the develop branch with gfs_utils updated to commit [02a16f65](https://github.com/NOAA-EMC/gfs-utils/commit/02a16f655ab36a55980305ee5bfe7e2e960763af). Made sure that
- the develop and this branch forecast files are identical
- the difference in gaussian_sfc_analysis file between develop and this branch are explained by the increments added

All CI/Pr tests pass on GAEAC6

# Checklist


parm/archive/enkf.yaml.j2
```diff
@@ -106,7 +106,9 @@ enkf:
         # Ensemble mean analyses/increments
         # soil DA increments
         {% if DO_GSISOILDA %}
-        - "{{ COMIN_ATMOS_ANALYSIS_ENSSTAT | relpath(ROTDIR) }}/{{ head }}increment.sfc.i006.nc"
+            {% for fhr in IAUFHRS %}
```
CoryMartin-NOAA, line 109: do we need to archive this for every hour? what does IAUFHRS resolve to in the default configuration for v17?

scripts/exglobal_atmos_sfcanl.sh
```diff
@@ -123,7 +123,7 @@ fi
 gcycle_dates=("${PDY}${cyc}")  # Always update surface restarts at middle of window
 soilinc_fhrs=("${assim_freq}") # increment file at middle of window
 LFHR=${assim_freq}
-if [[ "${DOIAU:-}" == "YES" ]]; then  # Update surface restarts at beginning of window
+if [[ "${DOIAU:-}" == "YES" ]]; then #Update surface restarts at beginning of window
```
CoryMartin-NOAA: This space should be returned?

ush/gaussian_sfcanl.sh
```diff
@@ -171,6 +171,26 @@ ${NLN} "${COMOUT_ATMOS_ANALYSIS}/${APREFIX}analysis.sfc.a006.nc" "./sfc.gaussian
 # Namelist uses booleans now
 if [[ ${DONST} == "YES" ]]; then do_nst='.true.'; else do_nst='.false.'; fi
 
+#Add soil increments to gdas gaussian sfcanal if they are not added by gcycle (i.e., when landiau=true)
+GSANAL_DO_SOILINCR=".false."
+if [[ "${RUN}" == "gdas" && "${DO_GSISOILDA:-NO}" == "YES" && "${DO_LAND_IAU:-.false.}" == ".true." ]]; then
```
CoryMartin-NOAA: @CatherineThomas-NOAA do we need this for the GFS cycle too?

ush/gaussian_sfcanl.sh
```diff
@@ -171,6 +171,26 @@ ${NLN} "${COMOUT_ATMOS_ANALYSIS}/${APREFIX}analysis.sfc.a006.nc" "./sfc.gaussian
...
+
+# sfc increment files
+if [[ "${GSANAL_DO_SOILINCR}" == ".true." ]]; then
+    for i in $(seq 1 6); do
+	sfc_inc="${COMOUT_ATMOS_ANALYSIS}/increment.sfc.i006.tile${i}.nc"
+        if [[ ! -f "${sfc_inc}" ]]; then
+            echo "Error! gaussian sfc analysis missing increment file ${sfc_inc}"
+            exit 1
+        else
+            ${NLN} "${sfc_inc}" "./sfc_inc.tile${i}.nc"
```
CoryMartin-NOAA, line 183: I don't quite understand this, if the file is missing, do a symbolic link?

parm/archive/enkf.yaml.j2
tsga: They resolve to hours 3, 6, and 9. Given the file sizes (the gaussian ones here are small compared to the re-gridded fv3) and given the corresponding atm and snow increments are being saved, I thought we should save all the three (And will help during the evaluation similar to that we did these past few weeks). At worst we can get rid them of later.

ush/gaussian_sfcanl.sh
tsga: The other way round. If increment is missing, error and exit; else link.

CoryMartin-NOAA: oh, duh!

ush/gaussian_sfcanl.sh
tsga: In the current configuration, we may not have the increments yet in gfs_analcalc as the gdas_sfcanal generates the soil increments.

aerorahul: If `DO_GSISOILDA==NO`, would `GSANAL_DO_SOILINCR` be `.true.`?  Also, lets try and keep the `DO_` and `GSI` a little consistent. ` DO_GSISOILINCR`  instead of `GSANAL_DO_SOILINCR`?

CoryMartin-NOAA: Good point @tsga , and I agree with @aerorahul that we should 1) make names consistent and 2) ask if we need these booleans added

tsga: @aerorahul no it (DO_GSANL_SOILINCR) would not be true if DO_GSISOILDA is NO. The name "DO_GSISOILINCR" is 'taken' (for non-IAU soil increments added to restarts, in contrast to Gaussian analysis here) so renamed GSANAL_DO_SOILINCR to DO_GSANAL_SOILINCR.

CatherineThomas-NOAA: @CoryMartin-NOAA - We intentionally did not include the soil increments in the early cycle for sfcanl and esfc due to the timing constraints.  For those jobs, `DO_GSISOILDA=NO` is set at the config level.  So it seems like we can either leave this as is, or allow this line to include the `gfs` case but set `DO_GSISOILDA=NO` in the analcalc config (which would be more correct and possibly save headaches if we change things in the future).

aerorahul: @tsga 
Please shorten to `DO_GSISOILINCR`.

aerorahul: I see I am mistaken.  The `GS` here is likely referring to `Gaussian Surface`.  Hmm ...  Would just `DO_SOILINCR` work for you?

tsga: @aerorahul yes that is true. And DO_SOILINCR is also taken somewhere :) I have now change it to DO_GSSOILINCR.

ush/gaussian_sfcanl.sh
```diff
@@ -183,6 +203,9 @@ cat <<EOF > fort.41
   donst=${do_nst},
   imp_physics=${imp_physics:-8},
   landsfcmdl=${landsfcmdl:-2},
+  add_soil_inc=${DO_GSANAL_SOILINCR},
```
CoryMartin-NOAA: @tsga does this require changes to the source code, or does it already have a namelist argument of `add_soil_inc`?

tsga: yes, it requires changes to the source code.

ush/gaussian_sfcanl.sh
```diff
@@ -171,6 +171,26 @@ ${NLN} "${COMOUT_ATMOS_ANALYSIS}/${APREFIX}analysis.sfc.a006.nc" "./sfc.gaussian
...
 
+#Add soil increments to gdas gaussian sfcanal if they are not added by gcycle (i.e., when landiau=true)
+DO_GSANAL_SOILINCR=".false."
+if [[ "${RUN}" == "gdas" && "${DO_LAND_IAU:-.false.}" == ".true." ]]; then   
+    DO_GSANAL_SOILINCR=".true."
+fi
+LSOIL_INCR=${LSOIL_INCR:-2}
+
+# sfc increment files
+if [[ "${DO_GSANAL_SOILINCR}" == ".true." ]]; then
```
CoryMartin-NOAA:
```suggestion
DO_GAUSS_ANL_SOILINCR=".false."
if [[ "${RUN}" == "gdas" && "${DO_LAND_IAU:-.false.}" == ".true." ]]; then   
    DO_GAUSS_ANL_SOILINCR=".true."
fi
```

ush/gaussian_sfcanl.sh
```diff
@@ -171,6 +171,26 @@ ${NLN} "${COMOUT_ATMOS_ANALYSIS}/${APREFIX}analysis.sfc.a006.nc" "./sfc.gaussian
 # Namelist uses booleans now
 if [[ ${DONST} == "YES" ]]; then do_nst='.true.'; else do_nst='.false.'; fi
 
+#Add soil increments to gdas gaussian sfcanal if they are not added by gcycle (i.e., when landiau=true)
+DO_GSANAL_SOILINCR=".false."
+if [[ "${RUN}" == "gdas" && "${DO_LAND_IAU:-.false.}" == ".true." ]]; then   
+    DO_GSANAL_SOILINCR=".true."
+fi
```
CoryMartin-NOAA:
```suggestion
```

tsga: @CoryMartin-NOAA I had already made these changes per @aerorahul recommendation but weirdly they were not showing

ush/gaussian_sfcanl.sh
tsga: github seems to show the old (reviewed) code.

ush/gaussian_sfcanl.sh
CoryMartin-NOAA: Could we save on a boolean by just having the source code check if the increment filename is not empty and assume that if it is defined in the namelist that it is expected to be used?

tsga: We could, but the Boolean saves multiple steps in the source code by avoiding having to open the file and reading etc, when it is false (by default): https://github.com/NOAA-EMC/gfs-utils/pull/161

CoryMartin-NOAA: ok makes sense

dev/parm/config/gfs/config.analcalc
```diff
@@ -8,4 +8,10 @@
 # Get task specific resources
 source "${EXPDIR}/config.resources" analcalc
 
+if [[ "${RUN}" == "gfs" ]]; then
+    echo "turning off gsi soilda for gfs analcalc"
+    DO_GSISOILDA="NO"
```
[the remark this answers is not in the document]
tsga: fixed

ush/gaussian_sfcanl.sh
```diff
@@ -171,6 +171,26 @@ ${NLN} "${COMOUT_ATMOS_ANALYSIS}/${APREFIX}analysis.sfc.a006.nc" "./sfc.gaussian
 # Namelist uses booleans now
 if [[ ${DONST} == "YES" ]]; then do_nst='.true.'; else do_nst='.false.'; fi
 
+#Add soil increments to gdas gaussian sfcanal if they are not added by gcycle (i.e., when landiau=true)
+DO_GSSOILINCR=".false."
+if [[ "${RUN}" == "gfs" || "${RUN}" == "gdas" ]] && [[ "${DO_LAND_IAU:-.false.}" == ".true." ]]; then
```
CoryMartin-NOAA: last question, why do we need to check what the run is, shouldn't DO_LAND_IAU be sufficient?

tsga: ah, true (now that we are setting it in config).

tsga: on second thought, the reason we are adding gfs now is for "future" as @CatherineThomas-NOAA suggested, unless even later on we only rely on do_land_iau.

CatherineThomas-NOAA: I think it's okay to remove the $RUN check from this line.  Unlike in the regridding script, the gaussian_sfcanl script is only called from `gfs` or `gdas`, not the `enkf`.

JessicaMeixner-NOAA:
@tsga  - I see that the gfs-utils PR was merged on Friday.  Other than resolving the merge conflicts in this branch, is there other work that needs to be done for this PR?

tsga:
@JessicaMeixner-NOAA no it is ready to go (I think the merge conflicts are resolved).

JessicaMeixner-NOAA:
Thanks @tsga ! My guess is this PR is waiting for reviews and then CI testing.  Thanks for making these updates!

CoryMartin-NOAA (dismissed):

dev/parm/config/gfs/config.analcalc
```diff
@@ -8,4 +8,10 @@ echo "BEGIN: config.analcalc"
 # Get task specific resources
 source "${EXPDIR}/config.resources" analcalc
 
+if [[ "${RUN}" == "gfs" ]]; then
+    echo "turning off gsi soilda for gfs run"
+    export DO_GSISOILDA="NO"
+    export DO_LAND_IAU=".false."
+fi
```
DavidHuber-NOAA: Should this be turned off for all `gfs` jobs or just `gfs_analcalc`?  If the former, then this probably belongs in `config.base`.

https://github.com/NOAA-EMC/global-workflow/blob/9901082002d4bf22a9eca734ebe4d853c4cbad33/dev/parm/config/gfs/config.base.j2#L422-L423

could be expanded to

```bash
# Expand this to enkfgdas if needed there as well
if [[ "${RUN}" == "gdas" ]]; then
    export DO_GSISOILDA="{{ DO_GSISOILDA }}"
    export DO_LAND_IAU="{{ DO_LAND_IAU }}"
else
    export DO_GSISOILDA="NO"
    export DO_LAND_IAU="NO"
fi
```

If this is only needed for the `analcalc` job, then there should probably be another variable here rather than overriding the `DO_GSISOILDA` variable.

```bash
if [[ "${RUN}" == "gfs"]]; then
    # I just made these names up, pick what makes sense
    export calc_gsisoilda="NO"
    export calc_land_iau="NO"
else
    export calc_gsisoilda="${DO_GSISOILDA}"
    export calc_land_iau="${DO_LAND_IAU}"
fi
```

The script would then need to be updated to look for the new variable names.

parm/archive/enkf_restarta_grp.yaml.j2
```diff
@@ -57,7 +57,7 @@ enkf_restarta_grp:
                 {% endfor %}  # iaufhr in IAUFHRS
                 {% if DO_LAND_IAU %}   #sfc inc tiles needed for warm start
                     {% for itile in range(1,7) %}
-        - "{{ COMIN_ATMOS_ANALYSIS_MEM | relpath(ROTDIR) }}/increment.sfc.i006.tile{{ itile }}.nc"
+        - "{{ COMIN_ATMOS_ANALYSIS_MEM | relpath(ROTDIR) }}/increment.sfc.tile{{ itile }}.nc"
```
DavidHuber-NOAA, line 60: Do each of the tiles contain multiple time levels or just one?

tsga: They contain multiple time levels (the three IAU times)

DavidHuber-NOAA (approved):
Looks good.

DavidHuber-NOAA:
Launching CI on C6 and Ursa.

DavidNew-NOAA (approved):

CatherineThomas-NOAA:
I ran some low res tests on Gaea and compared the gaussian sfcanl files between develop and this PR.  The differences between the files were isolated to the soil states and a look with ncview confirmed that the differences matched what is in the sfc increment files.  The following forecasts were identical, as expected.  It seems to be working as expected.

PR #4205, opened 2025-10-31
Update stage YAMLs to point to EE2 filename links

DavidHuber-NOAA:
# Description
This updates the staging YAMLs to point to the EE2-standardized filenames in the ICSDIR.

Resolves #4173 
# Type of change
- [ ] Bug fix (fixes something broken)
- [x] New feature (adds functionality)
- [x] Maintenance (code refactor, clean-up, new CI test, etc.)

# Change characteristics
- Is this a breaking change (a change in existing functionality)? YES: Old experiments will not be able to restart from old filenames.
- Does this change require a documentation update? NO
- Does this change require an update to any of the following submodules? NO

# How has this been tested?
Will test staging and first cycle forecast jobs on all platforms.

# Checklist
- [ ] Any dependent changes have been merged and published
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have documented my code, including function, input, and output descriptions
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] This change is covered by an existing CI test or a new one has been added
- [ ] Any new scripts have been added to the .github/CODEOWNERS file with owners
- [ ] I have made corresponding changes to the system documentation if necessary

PR #4247, opened 2025-11-13
Restart capability from pre-EE2-compliant filenames

### LINKED ISSUE #4242: Add support for restarting GFS experiments with old filenames

Users should have the ability to restart experiments that are using old (pre-EE2-compliant) filenames. A script to generate EE2-compliant links pointing to old filenames would suffice for this purpose. This would not guarantee that the old files would work with the new system (e.g. restarts with older surface parameters, etc).

### Acceptance Criteria

- A user is able to run the script on a COM directory to create the links.
- The script is compatible with GFS ice, ocean, snow, waves, and atmosphere filenames.

_No response_

DavidHuber-NOAA:
# Description
This adds a script that creates links from the pre-EE2-compliant filenames to EE2-equivalent names.  This will allow experiments to run warm restarts locally or from HPSS.  This is a companion PR to #4205.

Resolves #4242 
Prerequisite for https://github.com/NOAA-EMC/GFS/issues/3
# Type of change

# Change characteristics

# How has this been tested?
Ran on the staged ICs to verify the same links were created as in #4205 

# Checklist

### DIFF
--- a/dev/ush/make_ee2_links.sh
+++ b/dev/ush/make_ee2_links.sh
@@ -47,7 +47,7 @@ link_file() {
         echo "Error: Targeted link ${link} exists and is not a symbolic link. Not overwriting data files."
         return 1
     fi
-    ln -sf "$target" "$link"
+    ln -sf "${target}" "${link}"
     return 0
 }
 
--- a/dev/ush/make_ee2_links.sh
+++ b/dev/ush/make_ee2_links.sh
@@ -295,15 +295,29 @@ for dir in "${enkfgdas_list[@]}" "${enkfgfs_list[@]}"; do
             cd "${cwd}/${dir}/${cyc}/ensstat/analysis/atmos"
             if [[ -f "${system_prefix}.t${cyc}z.abias.ensmean" ]]; then
                 link_file "${system_prefix}.t${cyc}z.abias.ensmean" "${system_prefix}.t${cyc}z.abias.ensmean.txt"
+            fi
+            if [[ -f "${system_prefix}.t${cyc}z.abias_air.ensmean" ]]; then
                 link_file "${system_prefix}.t${cyc}z.abias_air.ensmean" "${system_prefix}.t${cyc}z.abias_air.ensmean.txt"
+            fi
+            if [[ -f "${system_prefix}.t${cyc}z.abias_int.ensmean" ]]; then
                 link_file "${system_prefix}.t${cyc}z.abias_int.ensmean" "${system_prefix}.t${cyc}z.abias_int.ensmean.txt"
+            fi
+            if [[ -f "${system_prefix}.t${cyc}z.abias_pc.ensmean" ]]; then
                 link_file "${system_prefix}.t${cyc}z.abias_pc.ensmean" "${system_prefix}.t${cyc}z.abias_pc.ensmean.txt"
             fi
             if [[ -f "${system_prefix}.t${cyc}z.cnvstat.ensmean" ]]; then
                 link_file "${system_prefix}.t${cyc}z.cnvstat.ensmean" "${system_prefix}.t${cyc}z.cnvstat.ensmean.tar"
+            fi
+            if [[ -f "${system_prefix}.t${cyc}z.enkfstat" ]]; then
                 link_file "${system_prefix}.t${cyc}z.enkfstat" "${system_prefix}.t${cyc}z.enkfstat.txt"
+            fi
+            if [[ -f "${system_prefix}.t${cyc}z.gsistat.ensmean" ]]; then
                 link_file "${system_prefix}.t${cyc}z.gsistat.ensmean" "${system_prefix}.t${cyc}z.gsistat.ensmean.tar"
+            fi
+            if [[ -f "${system_prefix}.t${cyc}z.oznstat.ensmean" ]]; then
                 link_file "${system_prefix}.t${cyc}z.oznstat.ensmean" "${system_prefix}.t${cyc}z.oznstat.ensmean.tar"
+            fi
+            if [[ -f "${system_prefix}.t${cyc}z.radstat.ensmean" ]]; then
                 link_file "${system_prefix}.t${cyc}z.radstat.ensmean" "${system_prefix}.t${cyc}z.radstat.ensmean.tar"
             fi
             # Surface increments
--- a/dev/ush/make_ee2_links.sh
+++ b/dev/ush/make_ee2_links.sh
@@ -23,14 +23,6 @@ fi
 target_dir=$1
 cd "${target_dir}" || exit 1
 
-# Check for existence of at least one of gdas.YYYYMMDD, gfs.YYYYMMDD, or enkfgdas.YYYYMMDD directories
-dir_list=($(ls -d gdas.* gfs.* enkfgdas.* || true ))
-
-if [[ ${#dir_list[@]} -eq 0 ]]; then
-    echo "No gdas.*, gfs.*, or enkfgdas.* directories found in ${target_dir}."
-    exit 1
-fi
-
 # A helper function to create symbolic links with error checking
 link_file() {
     if [[ $# -ne 2 ]]; then
@@ -57,6 +49,12 @@ gcdas_list=($(ls -d gcdas.* || true ))
 gcafs_list=($(ls -d gcafs.* || true ))
 enkfgdas_list=($(ls -d enkfgdas.* || true ))
 enkfgfs_list=($(ls -d enkfgfs.* || true ))
+
+# If the length of all of the arrays is zero, exit with a message
+if [[ ${#gdas_list[@]} -eq 0 && ${#gfs_list[@]} -eq 0 && ${#gcdas_list[@]} -eq 0 && ${#gcafs_list[@]} -eq 0 && ${#enkfgdas_list[@]} -eq 0 && ${#enkfgfs_list[@]} -eq 0 ]]; then
+    echo "No gdas, gfs, gcdas, gcafs, enkfgdas, or enkfgfs directories found in ${target_dir}. Exiting."
+    exit 0
+fi
 
 cwd=${PWD}
 # Loop through the gdas, gfs, gcdas, and gcafs directories
@@ -190,7 +188,7 @@ for dir in "${enkfgdas_list[@]}" "${enkfgfs_list[@]}"; do
     cd "${dir}"
     cycle_list=($(ls -d ./?? || true ))
     for cyc in "${cycle_list[@]}"; do
-            cd "${cwd}/${dir}/${cyc}"
+        cd "${cwd}/${dir}/${cyc}"
         mem_list=($(ls -d mem* || true ))
         for mem in "${mem_list[@]}"; do
             # atmos
--- a/dev/ush/make_ee2_links.sh
+++ b/dev/ush/make_ee2_links.sh
@@ -70,7 +70,7 @@ for dir in "${gdas_list[@]}" "${gfs_list[@]}" "${gcdas_list[@]}" "${gcafs_list[@
         *) echo "Unknown directory prefix: ${dir}"; exit 1 ;;
     esac
 
-    cycle_list=($(ls -d ./?? || true ))
+    cycle_list=($(ls -d -- ?? || true ))
     for cyc in "${cycle_list[@]}"; do
         if [[ -d "${cwd}/${dir}/${cyc}/analysis/atmos" ]]; then
             cd "${cwd}/${dir}/${cyc}/analysis/atmos"
@@ -186,7 +186,7 @@ done
 
 for dir in "${enkfgdas_list[@]}" "${enkfgfs_list[@]}"; do
     cd "${dir}"
-    cycle_list=($(ls -d ./?? || true ))
+    cycle_list=($(ls -d -- ?? || true ))
     for cyc in "${cycle_list[@]}"; do
         cd "${cwd}/${dir}/${cyc}"
         mem_list=($(ls -d mem* || true ))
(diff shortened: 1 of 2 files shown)


dev/ush/make_ee2_links.sh
```diff
@@ -0,0 +1,307 @@
...
+    fi
+  done
+  cd "${cwd}"
+done
+
+for dir in "${enkfgdas_list[@]}" "${enkfgfs_list[@]}"; do
+  cd "${dir}"
+  cycle_list=($(ls -d ./?? || true ))
+  for cyc in "${cycle_list[@]}"; do
+    mem_list=($(ls -d mem* || true ))
```
JessicaMeixner-NOAA: Before this line, we need to cd into the $cyc directory.  I added: 

+    cd "${cwd}/${dir}/${cyc}"

dev/ush/make_ee2_links.sh
```diff
@@ -0,0 +1,307 @@
+#!/bin/bash
+#shellcheck disable=SC2207
+PS4='+ $LINENO: '
+set -eux
```
JessicaMeixner-NOAA, line 4: the e here means that if you have to run this multiple times then you exit out if the link already exists... not sure if that's a good or bad thing?

DavidHuber-NOAA: I will keep it for now for debugging, but can remove it later. I have an idea on how to add some robustness to this by creating a link function following this pseudocode:
```
link_file():
{
  target=$1
  link=$2
  if ! -f $target; then error(target doesn't exist); fi
  if -f $link and ! is_link($link); then error(do not overwrite data files); return; fi
  ln -sf $target $link
  return 0
}

Comment (DavidHuber-NOAA) [reply dev/ush/make_ee2_links.sh:4]: OK, that should allow you to rerun and keep `-e`.

File: dev/ush/make_ee2_links.sh (hunk 1 above)
Comment (DavidHuber-NOAA) [reply dev/ush/make_ee2_links.sh]: Fixed.

**Comment by DavidHuber-NOAA** [comment]
@CoryMartin-NOAA Could you try running this script on a GCAFS directory structure?

**Comment by CoryMartin-NOAA** [comment]
> @CoryMartin-NOAA Could you try running this script on a GCAFS directory structure?

@lipan-NOAA would it be easy for you to test this since you are maintaining the real time? If not, let me know and I will set up a test and try it.

File: dev/ush/make_ee2_links.sh (hunk 2 above)
Comment (JessicaMeixner-NOAA) [reply dev/ush/make_ee2_links.sh:4]: That's nice! Thanks!

File: dev/ush/make_ee2_links.sh [hunk 3]
```diff
@@ -0,0 +1,327 @@
...
+    case "${dir}" in
+        gdas.*) system_prefix="gdas" ;;
+        gfs.*) system_prefix="gfs" ;;
+        gcdas.*) system_prefix="gcdas" ;;
+        gcafs.*) system_prefix="gcafs" ;;
+        *) echo "Unknown directory prefix: ${dir}"; exit 1 ;;
+    esac
+
+    cycle_list=($(ls -d ./?? || true ))
+    for cyc in "${cycle_list[@]}"; do
```
JessicaMeixner-NOAA, line 77: cyc ends up getting defined as 
./00 ./06 and so when you use it below as .t${cyc} it doesn't work correctly.

JessicaMeixner-NOAA: output if it's needed: /gpfs/f6/gfs-cpu/world-shared/Jessica.Meixner/gwpr4247/global-workflow/dev/ush/savefordave.txt 

I'll see if I can't come up with a solution

JessicaMeixner-NOAA: Copilot says: 
```
I updated the two places where cycle_list was populated so that ls -d returns bare directory names (00, 06, etc.) instead of ./00 ./06. Specifically, I changed the occurrences of: cycle_list=($(ls -d ./?? || true )) to: cycle_list=($(ls -d ?? || true ))

This keeps the rest of the script logic unchanged (paths constructed with 
d
i
r
/
{cyc} will now use "00" rather than "./00").

If you want a more robust solution that handles directory names with spaces or avoids depending on ls parsing, I can replace those lines with a find-based approach (e.g., using find -maxdepth 1 -type d -name '??' -printf '%f\n') and update the array population accordingly.
```


I used cycle_list=($(ls -d ?? || true )) and this worked.

DavidHuber-NOAA: I see.  The easy solution is to revert to what I had before, but it will raise a shellcheck warning:

```bash
    cycle_list=($(ls -d ?? || true ))
```

I have an idea.

DavidHuber-NOAA: Great!  I just pushed the same solution (with `--` to keep shellcheck happy).

dev/ush/make_ee2_links.sh
```diff
@@ -0,0 +1,327 @@
...
+                fi
+            done
+        fi
+    done
+    cd "${cwd}"
+done
+
+for dir in "${enkfgdas_list[@]}" "${enkfgfs_list[@]}"; do
+    cd "${dir}"
+    cycle_list=($(ls -d -- ?? || true ))
```
JessicaMeixner-NOAA, line 192: Need to add: 

```
   # Determine the system prefix
    system_prefix=""
    case "${dir}" in
        enkfgdas.*) system_prefix="enkfgdas" ;;
        enkfgfs.*) system_prefix="enkfgfs" ;;
        *) echo "Unknown directory prefix: ${dir}"; exit 1 ;;
    esac
```

JessicaMeixner-NOAA: with this update, i can successfully run gdas and enkf staging jobs using the staged IC for the retros.

DavidHuber-NOAA: I just added it in along with the `enkfgcdas` RUN.

DavidHuber-NOAA:
Marking this PR as ready for review.

dev/ush/make_ee2_links.sh
JessicaMeixner-NOAA: I'm working to copy a fresh set of ICs over to then do a clean run of the script.    But I don't expect things to be different, just triple checking.

dev/ush/make_ee2_links.sh
```diff
@@ -0,0 +1,338 @@
...
+                  # This gets the first two fields of the filename separated by dots
+                  prefix=$(echo "${ice_file}" | cut -d. -f1-2)
+                  link_file "${ice_file}" "${prefix}.analysis.cice_model.res.nc"
+              fi
+          done
+      fi
+      cd "${cwd}"
+        if [[ -d "${cwd}/${dir}/${cyc}/analysis/snow" ]]; then
+            cd "${cwd}/${dir}/${cyc}/analysis/snow"
+            for snow_file in *.sfc_data.tile1.nc; do
```
JessicaMeixner-NOAA: Should this instead be looking for snowinc*sfc_data.tile1.n ?

dev/ush/make_ee2_links.sh
```diff
@@ -0,0 +1,338 @@
...
+                link_file "${system_prefix}.t${cyc}z.sfci009.nc" "${system_prefix}.t${cyc}z.increment.sfc.i009.nc"
+            fi
+            if [[ -f "${system_prefix}.t${cyc}z.loginc.txt" ]]; then
+                link_file "${system_prefix}.t${cyc}z.loginc.txt" "${system_prefix}.t${cyc}z.increment.done.txt"
+            fi
+        fi
+        # snow ensstat
+        if [[ -d "${cwd}/${dir}/${cyc}/ensstat/analysis/snow" ]]; then
+            cd "${cwd}/${dir}/${cyc}/ensstat/analysis/snow"
+            for snow_file in *.sfc_data.tile1.nc; do
```
JessicaMeixner-NOAA: Should this be looking for snowinc.*.sfc_data.tile1.nc ?

dev/ush/make_ee2_links.sh
```diff
@@ -0,0 +1,338 @@
...
+                if [[ -f "${snow_file}" ]]; then
+                    prefix=$(echo "${snow_file}" | cut -d. -f1-3)
+                    prefix_new=$(echo "${prefix}" | cut -d. -f2-)
+                    for tile in {1..6}; do
+                        link_file "${prefix}.tile${tile}.nc" "${prefix_new}.snow_increment.sfc_data.tile${tile}.nc"
```
JessicaMeixner-NOAA: Does the prefix contain the sfc_data from the file name snowinc.YYYYMMDD.HH0000.sfc_data.tile?.nc

dev/ush/make_ee2_links.sh
```diff
@@ -0,0 +1,338 @@
...
+                    if [[ -f "${ice_file}" ]]; then
+                        prefix=$(echo "${ice_file}" | cut -d. -f1-2)
+                        link_file "${ice_file}" "${prefix}.analysis.cice_model.res.nc"
+                    fi
+                done
+            fi
+            # snow
+            if [[ -d "${cwd}/${dir}/${cyc}/${mem}/analysis/snow" ]]; then
+                cd "${cwd}/${dir}/${cyc}/${mem}/analysis/snow"
+                for snow_file in *.sfc_data.tile1.nc; do
```
JessicaMeixner-NOAA: This will grab both snowinc. and YYYMMDD. *sfc_data.tile1.nc files.  And we just want the ones without snowinc i think here.

dev/ush/make_ee2_links.sh
DavidHuber-NOAA: That is safer, yes.

JessicaMeixner-NOAA (dismissed):
I think there are some issues with the snow files.  I did not catch it earlier because we do not stage snow increments or analysis with the staged ICs.   


Unfortunately I think i've over-written these files at this point so I'll have to grab from hpss or create more data to continue testing. 

I think the increment section is easy to fix, not sure the best way to get the analysis files.  Maybe be more specific in the listing?

dev/ush/make_ee2_links.sh
DavidHuber-NOAA: Yes, fixed.

dev/ush/make_ee2_links.sh
DavidHuber-NOAA: No, this was a bug.  Fixed.

dev/ush/make_ee2_links.sh
DavidHuber-NOAA: Yes, you're correct.  Now fixed.

dev/ush/make_ee2_links.sh
```diff
@@ -0,0 +1,338 @@
...
+                    if [[ -f "${ice_file}" ]]; then
+                        prefix=$(echo "${ice_file}" | cut -d. -f1-2)
+                        link_file "${ice_file}" "${prefix}.analysis.cice_model.res.nc"
+                    fi
+                done
+            fi
+            # snow
+            if [[ -d "${cwd}/${dir}/${cyc}/${mem}/analysis/snow" ]]; then
+                cd "${cwd}/${dir}/${cyc}/${mem}/analysis/snow"
+                for snow_file in snowinc.*.sfc_data.tile1.nc; do
```
JessicaMeixner-NOAA: We are naming the same files as snow_analsysis and snow_increment.  I think this should be 

for snow_file in ????????.??0000.sfc_data.tile1.nc  and explicitly exclude the snowinc files. 

@CoryMartin-NOAA @tsga can you help confirm?

CoryMartin-NOAA: Yes, this is not correct as-is, we should have both analysis (no prefix) and increment (snowinc) files in the "old" structure

DavidHuber-NOAA: @JessicaMeixner-NOAA I believe you are correct. For reference, this [spreadsheet](https://docs.google.com/spreadsheets/d/1pb1V3K7JyZi20GV_npUU3Lf-OYDSV5FqSqgvPyCOeLQ/edit?gid=0#gid=0) was what we used to rename the snow files. I think this script should follow the same procedure, but I will wait for confirmation.

DavidHuber-NOAA: I believe it is fixed now.  I am going to run this through a copy of the ICSDIR.  I may need to recreate the links in the real ICSDIR as well.

JessicaMeixner-NOAA: I'd highly recommend doing that on a copy first.  I know I've basically deleted files b/c things weren't working as expected, and now have to start over.

DavidHuber-NOAA:
I made a pseudo copy of the ICSDIR with 0-sized files and ran this script. It was determined that a few `ensmean` files were not being created, which I fixed (script updated). After that, I ran the script again and compared all contents with the following findings:

1. No files were overwritten by links
2. All links point to the same targets with the exception of the following
    1. The official ICSDIR contains 4 dead links: `C48C48mx500/20250327/enkfgdas.20210324/12/mem00{1-4}/analysis/ice/20210324.150000.analysis.cice_model.res.nc`.  There is no ice analysis in this directory with that timestamp.
    2. The snow analysis and increment files in the ICSDIR copy point to the correct targets while the current ICSDIR created links to the same `snowinc` files for both analyses and increments.
3. Some links were not created in the ICSDIR copy. This is because the naming convention of directories changed in 2024 from `<component>/analysis` to `analysis/<component>`.  This only applies to older ICs, and I think this finding can be ignored.

When this PR is approved/merged, I will remove `snow_analysis` and `snow_increment` links in the ICSDIR and run this script on the affected directories. Note that this has no baring on #4205 as snow increments and snow analyses are not used when restarting an experiment.

dev/ush/make_ee2_links.sh
DavidHuber-NOAA: Fixed.

dev/ush/make_ee2_links.sh
DavidHuber-NOAA: Fixed.

JessicaMeixner-NOAA:
Waiting to make sure all DA jobs start as expected and then will give this the thumbs up.  Should be in the next few hours I hope.

JessicaMeixner-NOAA (approved):
@DavidHuber-NOAA  Thank you for making the really nice script to do the EE2 file name changes and working with us to get the updates in.  This will be essential for the retros and it's super appreciated!! THANK YOU!

CoryMartin-NOAA (approved):

PR #4278, opened 2025-11-26
Update PR template for expected changes

DavidHuber-NOAA:
# Description
This updates the PR template to include checklist items for expected changes.

# Type of change

# Change characteristics

# How has this been tested?
Rendered template


DavidHuber-NOAA:
FYI @JessicaMeixner-NOAA @CatherineThomas-NOAA @RuiyuSun

DavidHuber-NOAA:
@JessicaMeixner-NOAA @CatherineThomas-NOAA @RuiyuSun could you provide guidance on what constitutes a 'change to baselines'? Would this include just changes to model/analysis answers or product contents? Or would it also include things like filename changes?

RuiyuSun:
> @JessicaMeixner-NOAA @CatherineThomas-NOAA @RuiyuSun could you provide guidance on what constitutes a 'change to baselines'? Would this include just changes to model/analysis answers or product contents? Or would it also include things like filename changes?

@DavidHuber-NOAA We need to watch for any changes from inputs to outputs and archives. So, I would say what you mentioned above should all be included.

RussTreadon-NOAA:
Question:  Will _baseline_ output be stored somewhere for developers to compare against? Otherwise, developers will need to run g-w CI twice (once for `develop` and a second time for the PR) and then compare the two sets of output to document if/how the PR alters the _baseline_?

aerorahul:
We should not update the PR template to enforce a check against a baseline. 
The global-workflow does not do baseline testing at this moment, and that feature is far from being 
executed.  Baseline testing is extremely expensive and intensive to perform with a workflow that supports these many applications, configurations, and options.  This feature needs to be carefully planned and executed, otherwise we risk falling in the same trap as some of the other components that do regression testing.

In addition, considerably more work is required to operationally harden the workflow for GFSv17 and that will include updates to filenames.  We also have not concluded the archiving work as we don't yet have a full list of what files need to be archived to partition the tarballs into its appropriate groups.

FWIW, thats my feedback.

RuiyuSun:
@DavidHuber-NOAA Thank you for starting the PR to fulfill the request we made to watch for the incoming changes. @aerorahul @DavidHuber-NOAA  You are the experts. What do you think is the best way to monitor the changes that could affect the GFSv17 at current stage?

aerorahul (approved):

PR #4297, opened 2025-12-04
Fix METplus dependency matrix

### LINKED ISSUE #4281: gfs_metp won't start when 18 Z cycle doesn't exist
### What is wrong?

In the retrov17_01_stream3, the following tasks won't start because we don't run the 18Z cycle in this experiment and  dependency does not exist. In this experiment, we only run 00Z and 12 Z cycles. 

202412011800            gfs_metpg2g1                           -                   -                   -         -             -
202412011800            gfs_metpg2o1                           -                   -                   -         -             -
202412011800            gfs_metppcp1 

The tasks should run 

WCOSS2

ac57169c398e6fb5c4497076477384eb9891c51

### Steps to reproduce

retrov17_01_stream3

### Additional information

_No response_

_No response_

DavidHuber-NOAA:
# Description
This fixes the METplus dependencies by checking backwards for up to (`GFS_INTERVAL / assim_freq`) cycles for completed `gfs_arch_vrfy` jobs.  It also does a separate check for the last cycle.

Resolves #4281

# Type of change

# Change characteristics

# How has this been tested?
Generated XMLs for several different cases.

# Checklist


dev/workflow/rocoto/gfs_tasks.py
```diff
@@ -2001,24 +2001,48 @@ def metp(self):
...
         dep_dict = {'type': 'task', 'name': f'{self.run}_arch_vrfy'}
         deps.append(rocoto.add_dependency(dep_dict))
-        if self._base["interval_gfs"] < to_timedelta("24H"):
-            n_lookback = self._base["interval_gfs"] // to_timedelta("6H")
-            for lookback in range(1, n_lookback + 1):
-                deps2 = []
+        interval_gfs = self._base.get('interval_gfs')
+        assim_freq = self._base['assim_freq']
+
+        if interval_gfs < to_timedelta("24H"):
```
aerorahul, line 2007: Should this include a check where `interval_gfs > 0`?

aerorahul (dismissed):
Changes look good to me.  Just one question. Approve pending CI

dev/workflow/rocoto/gfs_tasks.py
DavidHuber-NOAA: I think I will add a check for this elsewhere. If `interval_gfs <= 0` and `RUN==gfs`, then we have problems, since `interval_gfs` should only be 0 if we aren't running a forecast, so we shouldn't be trying to run verification.

JessicaMeixner-NOAA:
@DavidHuber-NOAA  - 

Our xml that we had no changes on for 12 hour gfs interval looks like: 

```
                <dependency>
                        <or>
                                <taskdep task="gfs_arch_vrfy"/>
                                <and>
                                        <not><taskvalid task="gfs_arch_vrfy"/></not>
                                        <taskdep task="gfs_arch_vrfy" cycle_offset="00:00:00"/>
                                </and>
                                <and>
                                        <not><taskvalid task="gfs_arch_vrfy"/></not>
                                        <not><cycleexistdep cycle_offset="-6:00:00"/></not>
                                        <taskdep task="gfs_arch_vrfy" cycle_offset="00:00:00"/>
                                </and>
                        </or>
                </dependency>
```

We know that the first `<taskdep task="gfs_arch_vrfy" cycle_offset="00:00:00"/>` should be `<taskdep task="gfs_arch_vrfy" cycle_offset="-6:00:00"/>` 

Using this branch, I made a xml file for the same stream and the metp dependency is now: 

```
                <dependency>
                        <or>
                                <taskdep task="gfs_arch_vrfy"/>
                                <and>
                                        <not><taskvalid task="gfs_arch_vrfy"/></not>
                                        <or>
                                                <taskdep task="gfs_arch_vrfy" cycle_offset="-6:00:00"/>
                                        </or>
                                </and>
                                <and>
                                        <not><taskvalid task="gfs_arch_vrfy"/></not>
                                        <taskdep task="gfs_arch_vrfy" cycle_offset="-18:00:00"/>
                                        <not><cycleexistdep cycle_offset="-6:00:00"/></not>
                                        <not><cycleexistdep cycle_offset="-12:00:00"/></not>
                                </and>
                        </or>
                </dependency>
```

I know there's an extra `<or>` in the first `<and>` and the second set of `<and>`s are different.

DavidHuber-NOAA:
@JessicaMeixner-NOAA Breaking this down:

> We know that the first `<taskdep task="gfs_arch_vrfy" cycle_offset="00:00:00"/>` should be `<taskdep task="gfs_arch_vrfy" cycle_offset="-6:00:00"/>`
> 
> Using this branch, I made a xml file for the same stream and the metp dependency is now:
> 
> ```
>                 <dependency>
>                         <or>
>                                 <taskdep task="gfs_arch_vrfy"/>
>                                 <and>
>                                         <not><taskvalid task="gfs_arch_vrfy"/></not>
>                                         <or>
>                                                 <taskdep task="gfs_arch_vrfy" cycle_offset="-6:00:00"/>
>                                         </or>
>                                 </and>
>```
^^Since the METplus jobs run on the 18z cycle and the forecasts run on 00z and 12z, this checks that the last forecast of the cycle completed (i.e. 12z). The inner `<and> <or>` block would expand further if ever someone were to run with `INTERVAL_GFS=18` (which seems highly unlikely) to
```xml
                                     <and>
                                         <not><taskvalid task="gfs_arch_vrfy"/></not>
                                         <or>
                                                 <taskdep task="gfs_arch_vrfy" cycle_offset="-6:00:00"/>
                                                 <taskdep task="gfs_arch_vrfy" cycle_offset="-12:00:00"/>
                                         </or>
                                     </and>
```
>```
>                                 <and>
>                                         <not><taskvalid task="gfs_arch_vrfy"/></not>
>                                         <taskdep task="gfs_arch_vrfy" cycle_offset="-18:00:00"/>
>                                         <not><cycleexistdep cycle_offset="-6:00:00"/></not>
>                                         <not><cycleexistdep cycle_offset="-12:00:00"/></not>
>                                 </and>
>                         </or>
>                 </dependency>
> ```
^^ This will only be triggered by the very last cycle.  Since this retro case ends on a 00z cycle and the METplus job only runs on 18z, it is looking back 18 hours AND verifying the previous 2 cycles do not exist, which would only happen on the last cycle.

Do you agree with this logic?

JessicaMeixner-NOAA:
Thank you for the explanations @DavidHuber-NOAA  - I did not fully understand the logic before and thanks for making these updates!!!

JessicaMeixner-NOAA (dismissed):

dev/workflow/rocoto/gfs_tasks.py
DavidHuber-NOAA: Actually, this cannot happen. The code to determine if `gfs` should be considered a valid run requires `interval_gfs>0`:
https://github.com/NOAA-EMC/global-workflow/blob/a9123819ef7c9f56b332de9b9a3c9b92e4d9bfdc/dev/workflow/applications/gfs_cycled.py#L49-L53

So no need to create a check for it.

DavidHuber-NOAA:
Launching CI on WCOSS2.

DavidHuber-NOAA:
All tests passed on WCOSS2.

DavidHuber-NOAA:
Will merge after final approvals.

aerorahul (approved):
thanks for the fix.

PR #4382, opened 2025-12-24
Fix bash code analysis issues

WalterKolczynski-NOAA:
# Description
Fixes some issues from the recent expansion of bash code analysis and switch to the reviewdog framework.

First and foremost, the permission issue that prevented errors from being posted to the PR are fixed. This was only an issue for pull requests from forks (which is all of them for us). The trigger is changed to `pull_request_target`, which changes the context to the target repo. This is safe as long as we are not attempting to build or run the code. Because of the context change, an additional argument is needed for checkout to checkout the head of the PR.

Second, some parameters are turned into conditionals to handle the fact that the action is run on both pull requests and merges to develop. This allows both the correct hash in the previous change, and also the correct GH reporter to be used. This uses a new GH action that will need to be added to the allowed list.

Third, the `shfmt` options are fixed so that it actually fails on error. Reviewdog has deprecated `fail_on_error`, but the `reviewdog/shfmt` action does not seem to support the replacement `fail_level`. So, `fail_level` is passed to reviewdog directly as a flag.

Finally, the README status badges are updated for the new actions. Additionally, clicking on the code analysis badges now brings up the GH action logs.

Note: the `bash_code_analysis` actions may not run for this PR because of the trigger change. It will run on develop after merge, and on PRs from then on.

Resolves #4369

# Type of change
- [x] Bug fix (fixes something broken)

# Change characteristics
- Is this change expected to change outputs (e.g. value changes to existing outputs, new files stored in COM, files removed from COM, filename changes, additions/subtractions to archives)? NO
- Is this a breaking change (a change in existing functionality)? NO
- Does this change require a documentation update? NO
- Does this change require an update to any of the following submodules? NO

# How has this been tested?
- Tested by making pull requests into my own fork from a third fork

# Checklist
- [x] Any dependent changes have been merged and published
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have documented my code, including function, input, and output descriptions
- [x] My changes generate no new warnings
- [x] New and existing tests pass with my changes
- [x] This change is covered by an existing CI test or a new one has been added
- [x] Any new scripts have been added to the .github/CODEOWNERS file with owners
- [x] I have made corresponding changes to the system documentation if necessary

PR #4390, opened 2025-12-29
Revert "Fix bash code analysis issues"

DavidHuber-NOAA:
Reverts NOAA-EMC/global-workflow#4382


DavidHuber-NOAA:
Reverting the bash code analysis updates.

PR #4392, opened 2025-12-30
Replace Links with `cpfs/cpreq` in Workflow Scripts

AntonMFernando-NOAA:
# Description
Some jobs in the current global-workflow create links from their working directories to files in the COM directory, allowing output to be written or accessed directly in COM.
- Use `cpreq` when files are copied from COM directory.
- Use `cpfs` when files are copied to COM directory

  Refs #712 

# Type of change

# Change characteristics

- Is this change expected to change outputs (e.g. value changes to existing outputs, new files stored in COM, files removed from COM, filename changes, additions/subtractions to archives)? YES/NO (If YES, please indicate to which system(s))
  - [ ] GFS
  - [ ] GEFS
  - [ ] SFS
  - [ ] GCAFS
  - [ ] EMC verif-global 
  - [ ] GDAS 
  - [ ] GFS-utils 
  - [ ] GSI 
  - [ ] GSI-monitor 
  - [ ] GSI-utils 
  - [ ] UFS-utils 
  - [ ] UFS-weather-model 
  - [ ] wxflow 

# How has this been tested?
- CI tests on hera and extended tests on wcoss2

# Checklist


dev/scripts/exgdas_enkf_post.sh
```diff
@@ -68,8 +68,8 @@ for imem in $(seq 1 "${NMEM_ENS}"); do
 
     for fhr in $(seq "${FHMIN}" "${FHOUT}" "${FHMAX}"); do
         fhrchar=$(printf %03i "${fhr}")
-        ${NLN} "${COMIN_ATMOS_HISTORY}/${PREFIX}sfc.f${fhrchar}.nc" "sfcf${fhrchar}_${memchar}"
-        ${NLN} "${COMIN_ATMOS_HISTORY}/${PREFIX}atm.f${fhrchar}.nc" "atmf${fhrchar}_${memchar}"
+        cpreq "${COMIN_ATMOS_HISTORY}/${PREFIX}sfc.f${fhrchar}.nc" "sfcf${fhrchar}_${memchar}"
+        cpreq "${COMIN_ATMOS_HISTORY}/${PREFIX}atm.f${fhrchar}.nc" "atmf${fhrchar}_${memchar}"
```
DavidHuber-NOAA: Since these are member forecasts, they can stay as links for the post job.

gempak/ush/gfs_meta_comp.sh
```diff
@@ -29,7 +29,7 @@ for cycle in $(seq -f "%02g" -s ' ' 0 "${INTERVAL_GFS}" "${cyc}"); do
     for file_in in "${gempak_dir}/gfs_1p00_${PDY}${cycle}f"*; do
         file_out="${COMIN}/$(basename "${file_in}")"
```
DavidHuber-NOAA, line 30: This should be updated to COMOUT.

DavidHuber-NOAA (commented):
Looks good.  There's just one member loop that should stay as links and I spotted a COMIN that should be a COMOUT.

ush/gfs_bufr.sh
```diff
@@ -92,9 +92,9 @@ if [[ -z ${filename} ]]; then
 fi
 
 #------------------------------------------------------------------
-${NLN} "${COMIN_ATMOS_HISTORY}/${RUN}.${cycle}.atm.f${fhr}.${atmfm}" "sigf${fhr}"
-${NLN} "${COMIN_ATMOS_HISTORY}/${RUN}.${cycle}.sfc.f${fhr}.${atmfm}" "flxf${fhr}"
-${NLN} "${COMIN_ATMOS_HISTORY}/${RUN}.${cycle}.sfc.f${fhr_p}.${atmfm}" "flxf${fhr_p}"
+cpreq "${COMIN_ATMOS_HISTORY}/${RUN}.${cycle}.atm.f${fhr}.${atmfm}" "sigf${fhr}"
```
BoCui-NOAA, line 95: The changes look good for ush/gfs_bufr.sh.

dev/scripts/exglobal_enkf_ecen.sh
```diff
@@ -168,7 +169,7 @@ for FHR in $(seq "${FHMIN}" "${FHOUT}" "${FHMAX}"); do
 
         # If available, link to ensemble mean guess.  Otherwise, compute ensemble mean guess
```
DavidHuber-NOAA:
```suggestion
        # If available, copy the ensemble mean guess.  Otherwise, compute ensemble mean guess
```

dev/scripts/exglobal_enkf_ecen.sh
```diff
@@ -214,10 +215,10 @@ for FHR in $(seq "${FHMIN}" "${FHOUT}" "${FHMAX}"); do
 
         # if we already have a ensemble resolution GSI analysis then just link to it
```
DavidHuber-NOAA:
```suggestion
        # if we already have a ensemble resolution GSI analysis then just copy it
```

dev/scripts/exglobal_enkf_update.sh
```diff
@@ -140,14 +140,14 @@ for imem in $(seq 1 "${NMEM_ENS}"); do
...
             "sfg_${PDY}${cyc}_fhr0${FHR}_${memchar}"
         if [[ "${hofx_2m_sfcfile}" == ".true." ]]; then
-            ${NLN} "${COMIN_ATMOS_HISTORY_MEM_PREV}/${GPREFIX}sfc.f00${FHR}${ENKF_SUFFIX}.nc" \
+            cpreq "${COMIN_ATMOS_HISTORY_MEM_PREV}/${GPREFIX}sfc.f00${FHR}${ENKF_SUFFIX}.nc" \
                 "bfg_${PDY}${cyc}_fhr0${FHR}_${memchar}"
         fi
         if [[ "${cnvw_option}" == ".true." ]]; then
-            ${NLN} "${COMIN_ATMOS_HISTORY_MEM_PREV}/${GPREFIX}sfc.f00${FHR}.nc" \
+            cpreq "${COMIN_ATMOS_HISTORY_MEM_PREV}/${GPREFIX}sfc.f00${FHR}.nc" \
                 "sfgsfc_${PDY}${cyc}_fhr0${FHR}_${memchar}"
```
DavidHuber-NOAA, line 151: These are all of the member forecasts and can stay as links.

gempak/ush/gdas_ecmwf_meta_ver.sh
DavidHuber-NOAA, line 1: Just noting that this file was removed from v16 production as ECMWF gempak products are no longer being disseminated. I'm fine with this change, but the TODO probably won't need to happen.

gempak/ush/gdas_meta_loop.sh
```diff
@@ -13,7 +13,7 @@ device="nc | gdasloop.meta"
 #
 export COMIN="${RUN}.${PDY}${cyc}"
```
DavidHuber-NOAA, line 14: Noting that `COMIN` here (and in most of the gempak ush scripts) seems like an EE2 violation. Should we open an issue for this @aerorahul to rename this variable since it is inside a `DATA` directory?

gempak/ush/gfs_meta_comp.sh
```diff
@@ -22,12 +22,12 @@ metatype="comp"
...
 
-export COMIN="gfs.multi"
-mkdir "${COMIN}"
+export COMOUT="gfs.multi"
+mkdir "${COMOUT}"
 for cycle in $(seq -f "%02g" -s ' ' 0 "${INTERVAL_GFS}" "${cyc}"); do
     YMD=${PDY} HH=${cycle} GRID="1p00" declare_from_tmpl gempak_dir:COM_ATMOS_GEMPAK_TMPL
     for file_in in "${gempak_dir}/gfs_1p00_${PDY}${cycle}f"*; do
-        file_out="${COMIN}/$(basename "${file_in}")"
+        file_out="${COMOUT}/$(basename "${file_in}")"
```
DavidHuber-NOAA: Please leave these as `COMIN` for now, please.

gempak/ush/gfs_meta_comp.sh
```diff
@@ -22,12 +22,12 @@ metatype="comp"
...
         if [[ ! -L "${file_out}" ]]; then
             ${NLN} "${file_in}" "${file_out}"
         fi
```
DavidHuber-NOAA:
```suggestion
        cpreq "${file_in}" "${file_out}"
```

gempak/ush/gfs_meta_bwx.sh
```diff
@@ -21,7 +21,7 @@ device="nc | ${metaname}"
 #
 export COMIN="${RUN}.${PDY}${cyc}"
 if [[ ! -L ${COMIN} ]]; then
-    ${NLN} "${COMIN_ATMOS_GEMPAK_1p00}" "${COMIN}"
+    cpreq -R "${COMIN_ATMOS_GEMPAK_1p00}" "${COMIN}"
 fi
```
DavidHuber-NOAA: This `-L` tests if a link already exists. Since we are now copying, there is no need to check for links.

```suggestion
cpreq -R "${COMIN_ATMOS_GEMPAK_1p00}" "${COMIN}"
```

gempak/ush/gfs_meta_comp.sh
```diff
@@ -101,7 +102,7 @@ for gareas in US NP; do
         HPCGFS="${RUN}.${init_time}"
         if [[ ! -L ${HPCGFS} ]]; then
```
DavidHuber-NOAA, line 103: The `-L` check can be removed.

gempak/ush/gfs_meta_crb.sh
```diff
@@ -23,7 +23,7 @@ device="nc | ${metaname}"
 #
 export COMIN="${RUN}.${PDY}${cyc}"
 if [[ ! -L ${COMIN} ]]; then
```
DavidHuber-NOAA: `-L` can be removed here and elsewhere.

DavidHuber-NOAA (changes requested):

DavidHuber-NOAA:
@AntonMFernando-NOAA The changes look good overall.  I just have a few cleanup items and a [question](https://github.com/NOAA-EMC/global-workflow/pull/4392#discussion_r2672911435) for @aerorahul.

dev/scripts/exgfs_wave_prdgen_gridded.sh
JessicaMeixner-NOAA: any reason why this isn't changed?

dev/scripts/exgfs_wave_post_pnt.sh
JessicaMeixner-NOAA: Shouldn't this now be a cpreq?

gempak/ush/gdas_meta_loop.sh
aerorahul: Ideally it would be defined in the j-job, since everything here is the same

aerorahul: I take this back.  
This is not `COMIN` as in, this is a directory name (not a directory path).
It is rather poor choice of name for the variable.  `COMIN` should have been an absolute path to `COM/` `+`  stuff.

gempak/ush/gdas_ecmwf_meta_ver.sh
```diff
@@ -19,9 +19,7 @@ device="nc | ecmwfver.meta"
 cpreq "${HOMEgfs}/gempak/fix/datatype.tbl" datatype.tbl
 
 export COMIN="gdas.${PDY}${cyc}"
-if [[ ! -L ${COMIN} ]]; then
-    ${NLN} "${COMIN_ATMOS_GEMPAK_1p00}" "${COMIN}"
-fi
+cpreq -R "${COMIN_ATMOS_GEMPAK_1p00}" "${COMIN}"
```
aerorahul: I don't think this is the intended use of `cpreq` with `-R` to copy a directory.
A required directory may exist, but may not contain all the required files.  `cpreq` operates on a file and if it does not exist or is corrupted, reports an error.

DavidHuber-NOAA: I realize now that this will require more engineering. These gempak scripts are called by the meta task which has a restart capability, thus the `COMIN` directory could already exist. Additionally, these scripts are called in parallel via MPMD, so multiple jobs will be trying to write to `COMIN` at the same time. This `cpreq` task should probably be moved up to the [exgfs_atmos_gempak_meta.sh](https://github.com/NOAA-EMC/global-workflow/blob/develop/dev/scripts/exgfs_atmos_gempak_meta.sh) script. Let's revert these and make them a `TODO`.

DavidHuber-NOAA: Scratch that. I see now each of these `ush` scripts run in a separate directory. I'll think on the solution a bit more...

gempak/ush/gdas_ecmwf_meta_ver.sh
DavidHuber-NOAA, line 1: Based on the update for GFS v16.3.28 (https://github.com/NOAA-EMC/global-workflow/commit/8cb46ad3b1db26faadaabf74bc8c1277b2df27cf), this file can be deleted. Let's do that instead of fixing anything here.

gempak/ush/gdas_ukmet_meta_ver.sh
DavidHuber-NOAA, line 1: This script can also be deleted.

gempak/ush/gdas_meta_loop.sh
```diff
@@ -12,9 +12,7 @@ device="nc | gdasloop.meta"
 # TODO: Replace this
 #
 export COMIN="${RUN}.${PDY}${cyc}"
-if [[ ! -L "${COMIN}" ]]; then
-    ${NLN} "${COMIN_ATMOS_GEMPAK_1p00}" "${COMIN}"
-fi
+cpreq -R "${COMIN_ATMOS_GEMPAK_1p00}" "${COMIN}"
```
DavidHuber-NOAA: Alright, sorry for the whiplash. This will take some more thought. Please revert these `cpreq` calls to `NLN` and add a `TODO`. I will create a follow-up issue.

gempak/ush/gdas_ecmwf_meta_ver.sh
DavidHuber-NOAA: @aerorahul I'm not sure if this is the intended purpose, but the `cpreq` script does take `$*` as the argument to `cp` ([cpreq#L3](https://github.com/NCO-HPC/prod_util/blob/27247dd76afd2b81825f433c50bee8cf9bd98a87/ush/cpreq#L3)):
```bash
cp $*
```
Thus, the `-R` does get passed as an argument to `cp` and if anything fails in the copy, it will call `err_exit`. That said, the standards only have [one example](https://github.com/NCO-HPC/nws-hpc-standards/blob/d0e8f079b66891d39fe7494a1c68bd7c77639425/docs/standards.rst?plain=1#L953) with a single file operation:
```bash
   cpreq $COMIN/inputfile inputfile       # copy essential input files into working directory
```

All of that said, I think these linking/copying operations need to be reworked anyway and I'll open a follow-up issue for that.

DavidHuber-NOAA: Oh, I misread your comment. You're correct about the potentially-missing files. Still, I'll open another issue.

aerorahul: Yes. `cpreq` is essentially this:  `cp $*; if [[ $? -ne 0 ]]; then echo "we have a problem"; exit 1; fi`

I think that's quite silly.  Instead, it should work on file-by-file basis, because a directory may exist with none of the required files and yet `cpreq` will succeed.

Instead, we should propose to NCO that `cpreq` should be more thorough.

gempak/ush/gfs_meta_opc_na_ver.sh
```diff
@@ -16,9 +16,7 @@ cpreq "${HOMEgfs}/gempak/fix/datatype.tbl" datatype.tbl
 # TODO: Replace this
 #
 export COMIN="${RUN}.${PDY}${cyc}"
-if [[ ! -L ${COMIN} ]]; then
-    ${NLN} "${COMIN_ATMOS_GEMPAK_1p00}" "${COMIN}"
-fi
+cpreq -R "${COMIN_ATMOS_GEMPAK_1p00}" "${COMIN}"
```
DavidHuber-NOAA: This should be reverted to linking.

DavidHuber-NOAA (approved):
LGTM

DavidHuber-NOAA:
Launching tests on Ursa.

DavidHuber-NOAA:
@AntonMFernando-NOAA has verified b4b identical results between this branch and develop. Merging.

PR #4402, opened 2026-01-05
Remove extraneous module use from build_fv3.sh (GFS v16.3.29->16.3.30)

DavidHuber-NOAA:
# Description
This GFS v16 production update removes an extraneous `module use` from the `build_fv3.sh` build script. I only tested that the build was successful after removing this `module use`. The model was not rebuilt in production following this update.

# Type of change
- [x] Production


aerorahul (approved):
lgtm.
Thanks @DavidHuber-NOAA

PR #4414, opened 2026-01-08
Bugfix for archiving netcdf files 

JessicaMeixner-NOAA:
# Description

The if statement in dev/parm/config/gfs/config.arch_tars was not working as intended.  This only calculates a nonzero ARCH_GAUSSIAN_FHMAX if FHMAX_GFS has a value. 

# Type of change
- [x] Bug fix (fixes something broken)
- [ ] New feature (adds functionality)
- [ ] Maintenance (code refactor, clean-up, new CI test, etc.)

# Change characteristics

- Is this change expected to change outputs (e.g. value changes to existing outputs, new files stored in COM, files removed from COM, filename changes, additions/subtractions to archives)? YES/NO (If YES, please indicate to which system(s))
- No changes to products, just additional things actually archived. 
- Is this a breaking change (a change in existing functionality)? NO
- Does this change require a documentation update? NO
- Does this change require an update to any of the following submodules? NO 

# How has this been tested?
This was tested by manually changing a gfsv17 test's expdir and confirmed archive existed as expected: https://github.com/NOAA-EMC/global-workflow/issues/4403#issuecomment-3723977553

# Checklist
- [ ] Any dependent changes have been merged and published
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have documented my code, including function, input, and output descriptions
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] This change is covered by an existing CI test or a new one has been added
- [ ] Any new scripts have been added to the .github/CODEOWNERS file with owners
- [ ] I have made corresponding changes to the system documentation if necessary

PR #4415, opened 2026-01-08
fix archiving issue that netcdf files were being skipped

### LINKED ISSUE #4403: ARCH_GAUSSIAN_FHMAX is not being set to 24 as expected
### What is wrong?

In https://github.com/NOAA-EMC/global-workflow/blob/develop/dev/parm/config/gfs/config.arch_tars#L20-L23 

we have: 

```
export ARCH_GAUSSIAN_FHMAX=0
if [[ -z "${FHMAX_GFS+x}" ]]; then
    export ARCH_GAUSSIAN_FHMAX=$(( FHMAX_GFS < 24 ? FHMAX_GFS : 24 ))
fi
```

My original understanding of this, was that ARCH_GAUSSIAN_FHMAX=24 unless FHMAX_GFS was less than 24, in which it would be 24.  

However, it was reported by @KaiWang-NOAA that we only have 000 in the tar file.  @CatherineThomas-NOAA  discovered the line `export ARCH_GAUSSIAN_FHMAX=$(( FHMAX_GFS < 24 ? FHMAX_GFS : 24 ))` is not being executed, because `if [[ -z "${FHMAX_GFS+x}" ]]; then` is only true if FHMAX_GFS is not set. 

We should execute ARCH_GAUSSIAN_FHMAX=$(( FHMAX_GFS < 24 ? FHMAX_GFS : 24 )). 

All or N/A

dev/gfs.v17

### Steps to reproduce

This should be reproducible by CI tests, but this was discovered in the GFS retros.  This issue is also in the develop branch.   A fix is needed for both develop and the dev/gfs.v17 branch. 

### Additional information

_No response_

Change the `if [[ -z "${FHMAX_GFS+x}" ]] `line to be true if FHMAX_GFS exists, not if it does not exist.

JessicaMeixner-NOAA:
# Description

The if statement in dev/parm/config/gfs/config.arch_tars was not working as intended.  This only calculates a nonzero ARCH_GAUSSIAN_FHMAX if FHMAX_GFS has a value. 

Fixes https://github.com/NOAA-EMC/global-workflow/issues/4403 
This is the PR for develop, PR to gfsv17 is https://github.com/NOAA-EMC/global-workflow/pull/4414

# Type of change

# Change characteristics

- Is this change expected to change outputs (e.g. value changes to existing outputs, new files stored in COM, files removed from COM, filename changes, additions/subtractions to archives)? YES/NO (If YES, please indicate to which system(s))
- No changes to products, just additional things actually archived. 

# How has this been tested?
This was tested by manually changing a gfsv17 test's expdir and confirmed archive existed as expected: https://github.com/NOAA-EMC/global-workflow/issues/4403#issuecomment-3723977553

# Checklist

### DIFF
--- a/dev/parm/config/gfs/config.arch_tars
+++ b/dev/parm/config/gfs/config.arch_tars
@@ -18,7 +18,7 @@ export ARCH_DIFFS='NO'       # Archive the output of 'git diff' for the GW; requ
 
 export ARCH_GAUSSIAN="YES"
 export ARCH_GAUSSIAN_FHMAX=0
-if [[ -n "${FHMAX_GFS}" ]]; then
+if [[ -n "${FHMAX_GFS:-}" ]]; then
     export ARCH_GAUSSIAN_FHMAX=$(( FHMAX_GFS < 24 ? FHMAX_GFS : 24 ))
 fi
 export ARCH_GAUSSIAN_FHINC=${FHOUT_GFS:-}
--- a/dev/parm/config/gfs/config.arch_tars
+++ b/dev/parm/config/gfs/config.arch_tars
@@ -18,7 +18,7 @@ export ARCH_DIFFS='NO'       # Archive the output of 'git diff' for the GW; requ
 
 export ARCH_GAUSSIAN="YES"
 export ARCH_GAUSSIAN_FHMAX=0
-if [[ -z "${FHMAX_GFS+x}" ]]; then
+if [[ -n "${FHMAX_GFS}" ]]; then
     export ARCH_GAUSSIAN_FHMAX=$(( FHMAX_GFS < 24 ? FHMAX_GFS : 24 ))
 fi
 export ARCH_GAUSSIAN_FHINC=${FHOUT_GFS:-}


dev/parm/config/gfs/config.arch_tars
```diff
@@ -18,7 +18,7 @@ export ARCH_DIFFS='NO'       # Archive the output of 'git diff' for the GW; requ
 
 export ARCH_GAUSSIAN="YES"
 export ARCH_GAUSSIAN_FHMAX=0
-if [[ -z "${FHMAX_GFS+x}" ]]; then
+if [[ -n "${FHMAX_GFS}" ]]; then
```
DavidHuber-NOAA: One suggestion I just thought of (sorry I didn't think of it before) to declare an empty string if `FHMAX_GFS` isn't defined:

```suggestion
if [[ -n "${FHMAX_GFS:-}" ]]; then
```

JessicaMeixner-NOAA: My understanding is that the -n checks if it's defined.  If it's not, then it doesn't go into that string at all.   But we should test copilot's info to confirm that is the case.   I don't think we want it to go into the next stream if it is empty.

aerorahul: `-n` checks if the string contained in the variable is empty or not.  If it is not empty, the condition is true.
If the variable `FHMAX_GFS` is undefined, `set -u` will be triggered, and the `set -e` will terminate the process.

JessicaMeixner-NOAA: Okay, I will test adding `if [[ -n "${FHMAX_GFS:-}" ]];` then to see if we still go through this loop and archive things as expected.

aerorahul: I don't think you need to retest.  I am 100% confident this is correct.  Plus, I think we define `FHMAX_GFS` in `config.base` which is sourced before this config file, so technically `:-` is not required as that variable will not be undefined, and the suggestion is programatically correct.

DavidHuber-NOAA (approved):

DavidHuber-NOAA:
Launching CI on Ursa. @JessicaMeixner-NOAA when the CI test completes, I will point you to the COMROOT directories if you would like to verify the outputs.

aerorahul (approved):
lgtm

DavidHuber-NOAA:
FYI @NeilBarton-NOAA

Checking the failure, the reported error was for the missing history file at 48 hours:

```
FileNotFoundError: Source file '/scratch3/NCEPDEV/global/role.glopara/GFS_CI_CD/URSA/BUILDS/GITLAB/pr_cases_4415_de48f20e_6911/RUNTESTS/COMROOT/C96mx025_S2S_de48f20e-6911/sfs.19940501/00/mem001/model/ocean/history/sfs.t00z.24hr_avg.f048.nc' does not exist
```

Checking that directory, I see

```bash
-rw-r--r-- 1 role.glopara global 877393444 Jan  8 20:31 sfs.t00z.24hr_avg.f024.nc                                                                                                         
-rw-r--r-- 1 role.glopara global 681368548 Jan  8 21:18 sfs.t00z.24hr_avg.f072.nc
```

It appears that the 48-hour forecast failed to write. This is clearly unrelated to the PR being tested, so a follow-up issue will be opened.

DavidHuber-NOAA:
All other tests passed.  Merging.

NeilBarton-NOAA:
@DavidHuber-NOAA thanks, we'll double check this in dev/sfs @YangxingZheng-NOAA

PR #4598, opened 2026-02-27
Add Jinja templates for forecast and surface analysis jobs

CoryMartin-NOAA:
# Description
Add Jinja2 templates for the forecast and surface analysis jobs. When running the deployment script, these templates will be rendered to only contain code/variables needed for their specific NET and RUN.


CoryMartin-NOAA:
@DavidHuber-NOAA any idea on how to get this shellcheck to ignore the j2 files? I'll keep digging and trying but the exclude line isn't working as I expected

dev/jobs/JGLOBAL_ATMOS_SFCANL.j2
```diff
@@ -0,0 +1,50 @@
...
+# Ignore possible spelling error (nothing is misspelled)
+# shellcheck disable=SC2153
+GDATE=$(date --utc -d "${PDY} ${cyc} - ${assim_freq} hours" +%Y%m%d%H)
+export GDATE
+
+{% if RUN == "gcdas" %}
+BKG_RUN="gcdas"
+{% else %}
+BKG_RUN="gdas"
+{% endif %}
```
DavidHuber-NOAA: @CoryMartin-NOAA something like this should work

```suggestion
#{% if RUN == "gcdas" %}
BKG_RUN="gcdas"
#{% else %}
BKG_RUN="gdas"
#{% endif %}
```

Jinja parsing ignores bash comments.

If that doesn't work, you can add

```bash
# shellcheck disable=all
```

to the top of the .j2 files (after the shebang).

CoryMartin-NOAA:
thanks @DavidHuber-NOAA I finally realized that the CI does not use changes in the PR and still uses what is in develop, probably as a security measure, hence why my changes to the exclude list were not working

dev/jobs/JGLOBAL_FORECAST.j2
```diff
@@ -0,0 +1,156 @@
...
+# Remove the Temporary working directory
+##########################################
+cd "${DATAROOT}" || true
+# do not remove DATAjob. It contains DATAoutput
+if [[ "${KEEPDATA}" == "NO" ]]; then
+    rm -rf "${DATA}"
+
+    # Determine if this is the last segment
+    commas="${FCST_SEGMENTS//[^,]/}"
+#    {% raw %}n_segs=${#commas}{% endraw %}
```
DavidHuber-NOAA: This may not work. Shellcheck will ignore this line and Jinja will render this, but it will then be commented and skipped by the Bash interpreter. You could perhaps instead do


```suggestion
#{% raw %}
    n_segs=${#commas}
#{% endraw %}
```

CoryMartin-NOAA: I'm thinking we need to just have it skip all .j2 files instead rather than trying to put in these hacks. Thoughts? Or perhaps just the # shellcheck disable=all line for now?

DavidHuber-NOAA: If my suggestion above doesn't work, then I'm fine with adding `# shellcheck disable=all` to the top of the templates.

dev/jobs/JGLOBAL_FORECAST.j2
```diff
@@ -0,0 +1,156 @@
...
+YMD="${PDY}" HH="${cyc}" RUN="${rCDUMP}" declare_from_tmpl -rx \
+    COMIN_TRACER_RESTART:COM_ATMOS_RESTART_TMPL
+YMD="${PDY}" HH="${cyc}" declare_from_tmpl -rx \
+    COMIN_CHEM_INPUT:COM_CHEM_INPUT_TMPL
+#{% endif %}
+
+###############################################################
+# Run relevant exglobal script
+###############################################################
+"${FORECASTSH:-${SCRgfs}/ex{{ RUN | lower }}_forecast.sh}" && true # The && true prevents the shell from exiting when set -e
```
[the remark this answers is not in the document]
CoryMartin-NOAA: @DavidHuber-NOAA any tips on this one?

dev/jobs/JGLOBAL_FORECAST.j2
```diff
@@ -0,0 +1,158 @@
...
```
DavidHuber-NOAA, line 108: This probably just needs a disable.
```suggestion
# shellcheck disable=SC2288
"${FORECASTSH:-${SCRgfs}/ex{{ RUN | lower }}_forecast.sh}" && true # The && true prevents the shell from exiting when set -e
```

DavidHuber-NOAA: Maybe you could get fancy with something like

```bash
FORECASTSH=${FORECASTSH:-"${SCRgfs}/ex"'{{ RUN | lower }}_forecast.sh'}
"${FORECASTSH}" && true # The && true prevents the shell from exiting when set -e
```

DavidHuber-NOAA (approved):

lipan-NOAA (approved):

PR #4626, opened 2026-03-06
ecflow definition on cycle 00z under testing to be used for NCO SPA review

lgannoaa:
This version of the 00z definition file is for NCO SPA to review on family/task arrangement only.
The job level testing is still work in progress.


DavidHuber-NOAA:
@lgannoaa since the structure was approved, can we go ahead and merge this?

PR #4645, opened 2026-03-11
Make compute build more resilient

### LINKED ISSUE #4644: Make compute build more resilient

Some job schedulers are volatile and can return unknown status of jobs even though the job is still queued. It would be good if build_all was more lenient and allowed job statuses to return unknown a couple times before failing when building on compute.

### Acceptance Criteria

- [ ] When building on compute with `build_all.sh`, the overall build suite does not fail until multiple consecutive unknown job statuses are reported
- [ ] Build system otherwise works as before

_No response_

WalterKolczynski-NOAA:
# Description
Some job schedulers are volatile and can return unknown status of jobs even though the job is still queued. This causes `build_all.sh` to fail immediately, rather than waiting for the issue to resolve.

This commit removes the immediate failure on receiving an unknown job state and instead only fails after three consecutive polls return at least one unknown state.

Most scheduler instability is transient and impacts all jobs at once. This means we can track just whether there was an unknown status without the added complexity of tracking individual job status across loops.

Resolves #4644

# Type of change

# Change characteristics

# How has this been tested?
- build_all on Ursa
- build_all using compute on Derecho (using other yet-uncommitted changes)

# Checklist


DavidHuber-NOAA (approved):
Looks good. Thanks for adding this resilience.

Issue #4670, opened 2026-03-19
Remove references to RUN_ENVIR


CarlosMDiaz-NCO:
### What is wrong?

References to RUN_ENVIR should be removed in production from production scripts

This is an item in row 42 in the[ EE2 Checklist](https://docs.google.com/spreadsheets/d/1a7Wayq7DAswW32wlNfsRPwY4aPJ2eyTXKq_4xq8IH-w/edit?gid=643873428#gid=643873428)

Link to line: https://github.com/NOAA-EMC/global-workflow/blob/bbd81294080e6a32d8d4e00a045df47c77fd86b6/dev/jobs/JGLOBAL_OFFLINE_ATMOS_ANALYSIS#L33 

### What should have happened?

Standards violation in JGLOBAL_OFFLINE_ATMOS_ANALYSIS

### Steps to reproduce

Standards violation in JGLOBAL_OFFLINE_ATMOS_ANALYSIS

### Additional information

Found when setting up GCAFS. After the setup process RUN_ENVIR is found in jobs/JGCDAS_ATMOS_INITIALIZE

### Do you have a proposed solution?

Adjust GCAFS setup process to remove RUN_ENVIR references

CoryMartin-NOAA:
I'll work to make it so that when the j-jobs that use RUN_ENVIR are moved/copied/templated from dev/jobs to jobs/ that this if-block is removed

PR #4672, opened 2026-03-19
[GCAFSv1] Create a jinja2 template for the Atmos Init J-job

CoryMartin-NOAA:
# Description
This PR makes a Jinja2 template for the Atmos Init J-Job so that when `setup_gcafs_for_nco` is executed, only the if-block where `$RUN_ENVIR` is `nco` is included. This works for v1 T2O and lays groundwork for future changes for development where the Jinja templates are more fully rendered.

Resolves https://github.com/NOAA-EMC/global-workflow/issues/4670

### DIFF
--- a/dev/jobs/JGLOBAL_OFFLINE_ATMOS_ANALYSIS.j2
+++ b/dev/jobs/JGLOBAL_OFFLINE_ATMOS_ANALYSIS.j2
@@ -46,7 +46,11 @@ mkdir -p "${COMOUT_ATMOS_ANALYSIS}"
 mkdir -p "${COMOUT_OBS}"
 
 # Execute generation of increments from offline analysis
-EXSCRIPT=${OFFLINEANLPY:-${SCRglobal}/exglobal_offline_atmos_analysis.py}
+{% if RUN == "gcdas" %}
+EXSCRIPT=${SCRglobal}/exgcdas_atmos_initialize.py && true
+{% else %}
+EXSCRIPT=${SCRglobal}/exglobal_offline_atmos_analysis.py && true
+{% endif %}
 ${EXSCRIPT} && true
 
 export err=$?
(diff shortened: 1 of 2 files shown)


dev/jobs/JGLOBAL_OFFLINE_ATMOS_ANALYSIS.j2
```diff
@@ -0,0 +1,68 @@
...
+declare -rx COMOUT_OBS="${ROTDIR}/${RUN}.${PDY}/${cyc}/obs"
+declare -rx COMOUT_ATMOS_ANALYSIS="${ROTDIR}/${RUN}.${PDY}/${cyc}/analysis/atmos"
+
+declare -rx COMIN_ATMOS_HISTORY_PREV="${ROTDIR}/${RUN}.${gPDY}/${gcyc}/model/atmos/history"
+
+mkdir -p "${COMOUT_ATMOS_ANALYSIS}"
+mkdir -p "${COMOUT_OBS}"
+
+# Execute generation of increments from offline analysis
+EXSCRIPT=${OFFLINEANLPY:-${SCRglobal}/exglobal_offline_atmos_analysis.py}
```
TravisElless-NOAA: I don't know if this is getting changed correctly. When I rendered this it stays as this line, but when I used the non-templated one, it changed the python file to be exgcdas_atmos_initialize.py. Did this change as intended for you @CoryMartin-NOAA? I very easily might have made a misstep in my setup of things, but wanted to raise here just in case.

CoryMartin-NOAA: @TravisElless-NOAA you have a better eye than me, I didn't look at this, let me check, I bet I missed something

CoryMartin-NOAA: @TravisElless-NOAA check now

CarlosMDiaz-NCO (approved):

TravisElless-NOAA (approved):
Looks good to me. Was able to run the `jgcdas_atmos_initialize` job successfully with these updated changes. Approved

PR #4679, opened 2026-03-20
Move EXPDIR definition to the top of each j-job

### LINKED ISSUE #4674: Remove EXPDIR from scripts
### What is wrong?

EXPDIR is needed to find the config files to source, but EXPDIR will not be defined in production. This needs to be rectified by using $PACKAGEHOME/parm/config

Anything to add @CarlosMDiaz-NCO ?

Files should be sourced properly but cannot

### Steps to reproduce

Run GCAFS in a production-like ecflow environment

### Bugzilla issue

_No response_

### Additional information

GCAFSv1 dev/gcafs.v1 branch

_No response_

CoryMartin-NOAA:
# Description
Move definition of $EXPDIR from inside jjob_header (which will eventually be deprecated anyways) to the top of each j-job. This also includes modifying the default path to be `$HOMEglobal/parm/config/$NET`

  Resolves #4674


CoryMartin-NOAA:
@DavidHuber-NOAA @TravisElless-NOAA if everyone likes this PR, would it make more sense to put most of it in `develop`?

CarlosMDiaz-NCO (approved):

bbakernoaa (approved):

bbakernoaa:
@CoryMartin-NOAA clean up the shellcheck and shfmt and it looks good.

CoryMartin-NOAA:
@bbakernoaa those are actually issues outside of my PR...

bbakernoaa:
> @bbakernoaa those are actually issues outside of my PR...

Understood. Thanks for quickly attacking them though.  Helps us in the future.

DavidHuber-NOAA:
> @CoryMartin-NOAA clean up the shellcheck and shfmt and it looks good.

@bbakernoaa the shfmt issue is a bug we've fixed in develop. It will continue flagging this whether it's changed or not. I'll open a PR to GCAFS today to fix it there, too.

CoryMartin-NOAA:
@DavidHuber-NOAA did I fix it in this PR in the latest commit? I also opened https://github.com/NOAA-EMC/global-workflow/issues/4680

DavidHuber-NOAA:
@CoryMartin-NOAA No, it's intermittent and shfmt doesn't know what it wants from one commit to the next. Sometimes it wants
```bash
            nip=${arrpar[${iparam}-1]}
```

other times it wants

```bash
            nip=${arrpar[${iparam} - 1]}
```

The real fix is neither of these:

```bash
            nip=${arrpar[iparam - 1]}
```

Bash treats variables in `[ ]` like math, so they don't need a `$`.

dev/jobs/JGCDAS_PREPARE_OBS
```diff
@@ -2,6 +2,7 @@
 set -x
 
 export DATA=${DATA:-${DATAROOT}/${RUN}aeroanl_${cyc}}
+export EXPDIR="${EXPDIR:-${HOMEglobal}/parm/config/${NET}}"
```
DavidHuber-NOAA, line 5: I'm fine with this approach. For develop, though, could we make this part of the `jjob_standard_vars.sh` and `jjob_var_setup.j2`? Does it matter if it goes in before or after sourcing the configs?

TravisElless-NOAA (commented):
I think this is fine here for now with GCAFS. For longer term develop, I think it might be better to handle EXPDIR naming with other standard variables being moved from config.base and properly use other variables currently in the `jjob_standard_vars.sh` area. That's at least my current 2 cents on the develop question; always happy to hear if other have a different opinions on develop

dev/jobs/JGCDAS_PREPARE_OBS
CoryMartin-NOAA: It has to go before, because the configs live in EXPDIR

CoryMartin-NOAA:
Sounds like everyone is fine with this for now, but wants to explore a better option for the future. Let's go with that approach then and I will merge this into the dev branch.

dev/jobs/JGCDAS_PREPARE_OBS
DavidHuber-NOAA: Ah, right, of course. OK, then it's fine here in develop as well.

TravisElless-NOAA:
FYI I think this change was missed for the `dev/jobs/JGLOBAL_OFFLINE_ATMOS_ANALYSIS.j2`. While running tests for #4689 the `jobs/JGCDAS_ATMOS_INITIALIZE` that gets created doesn't populate the `EXPDIR`

PR #4701, opened 2026-03-24
Use compath.py to define ROTDIR for GCAFS

CoryMartin-NOAA:
# Description
This PR allows for the use of compath.py to define ROTDIR for GCAFS cycling. Please note that this is not intended as a _**final**_ solution but rather a temporary one moving in the right direction to allow for testing and evaluation to proceed.

For example, eventually the export of COMPATH will be removed from the run.ver file. Also, we should probably define ROTDIR/use compath.py in each j-job rather than in config.base

### DIFF
--- a/parm/config/gcafs/config.base
+++ b/parm/config/gcafs/config.base
@@ -90,7 +90,7 @@ export assim_freq="6" # Assimilation frequency in hours
 export PSLOT=""
 #export EXPDIR="/lfs/h2/emc/da/noscrub/cory.r.martin/gcafs/runtests2/EXPDIR/${PSLOT}"
 #export ROTDIR="/lfs/h2/emc/da/noscrub/cory.r.martin/gcafs/runtests2/COMROOT/${PSLOT}"
-export ROTDIR=${COMROOT}
+export ROTDIR=$(compath.py "${envir}/gcafs/${gcafs_ver}")
 export DUMP_SUFFIX=""
 if [[ "${PDY}${cyc}" -ge "2019092100" && "${PDY}${cyc}" -le "2019110700" ]]; then
     export DUMP_SUFFIX="p"              # Use dumps from NCO GFS v15.3 parallel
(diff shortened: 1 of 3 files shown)


parm/config/gcafs/config.base
```diff
@@ -90,7 +90,7 @@ export assim_freq="6" # Assimilation frequency in hours
 export PSLOT=""
 #export EXPDIR="/lfs/h2/emc/da/noscrub/cory.r.martin/gcafs/runtests2/EXPDIR/${PSLOT}"
 #export ROTDIR="/lfs/h2/emc/da/noscrub/cory.r.martin/gcafs/runtests2/COMROOT/${PSLOT}"
-export ROTDIR=${COMROOT}
+export ROTDIR=$(compath.py "${envir}/gcafs/${gcafs_ver}")
```
TravisElless-NOAA:
```suggestion
export ROTDIR=$(compath.py "${envir}/${NET}/${gcafs_ver}")
```

TravisElless-NOAA (commented):
Just a minor suggestion from me. I was able to successfully run a full 00z cycle with this PR so should be good for testing purposes. Launching the following 06z just to double check that cycling continues work as intended

versions/run.wcoss2.ver
```diff
@@ -54,3 +55,5 @@ export prepobs_run_ver=1.1.3
 
 export ens_tracker_ver=v1.3.0
 export fit2obs_ver=1.1.10
+
+export COMPATH="/lfs/h2/emc/ptmp/cory.r.martin/gcafs/ops/para/com/gcafs"
```
TravisElless-NOAA: I guess a second one. I haven't really tested this out but it seemed logical to me. 
```suggestion
export COMPATH="/lfs/h2/emc/ptmp/${USER}/gcafs/ops/para/com/gcafs"
```

versions/run.wcoss2.ver
```diff
@@ -1,5 +1,6 @@
 #! /usr/bin/env bash
-export version=v16.3.7
+export version=v1.0.0
+export gcafs_ver=v1.0.0
```
CarlosMDiaz-NCO: Was gcafs_ver added for some reason? In operations it won't be allowed. ${model}_ver is something set for us at the ecflow level. If left in run.ver it can create a mismatch if we do a minor upgrade and don't update it

CarlosMDiaz-NCO (approved):
the updates to use compath.py look good

versions/run.wcoss2.ver
CoryMartin-NOAA: added out of ignorance :-) I'll remove it

CoryMartin-NOAA:
```suggestion
```

TravisElless-NOAA (approved):

PR #4704, opened 2026-03-24
Get rid of set_trace and replace with set -x

CoryMartin-NOAA:
# Description
This removes the set_trace bash function and replaces all calls to that function with `set -x`

  Resolves #4683

# Type of change
- [ ] Bug fix (fixes something broken)
- [ ] New feature (adds functionality)
- [x] Maintenance (code refactor, clean-up, new CI test, etc.)

# Change characteristics

- Is this change expected to change outputs (e.g. value changes to existing outputs, new files stored in COM, files removed from COM, filename changes, additions/subtractions to archives)? NO
- Is this a breaking change (a change in existing functionality)? NO
- Does this change require a documentation update? NO
- Does this change require an update to any of the following submodules? NO
  - [ ] EMC verif-global 
  - [ ] GDAS 
  - [ ] GFS-utils 
  - [ ] GSI 
  - [ ] GSI-monitor 
  - [ ] GSI-utils 
  - [ ] UFS-utils 
  - [ ] UFS-weather-model 
  - [ ] wxflow 

# How has this been tested?

# Checklist
- [ ] Any dependent changes have been merged and published
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have documented my code, including function, input, and output descriptions
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] This change is covered by an existing CI test or a new one has been added
- [ ] Any new scripts have been added to the .github/CODEOWNERS file with owners
- [ ] I have made corresponding changes to the system documentation if necessary

PR #4712, opened 2026-03-26
Add in some cherry picked commits from develop for EE2 compliance

CoryMartin-NOAA:
# Description

### Get rid of set_trace and replace with set -x https://github.com/NOAA-EMC/global-workflow/pull/4704

This removes the set_trace bash function and replaces all calls to that
function with `set -x`

### Changes so that setup_data_dir is now sourced instead of a bash function https://github.com/NOAA-EMC/global-workflow/pull/4706

This PR changes the setup_data_dir ush script so that it is intended to
be sourced when executed rather than sourced once and then called as a
bash function.


TravisElless-NOAA (approved):
Looks good to me

Issue #4725, opened 2026-03-30
Address CRTM related issues for GCAFSv1


CoryMartin-NOAA:
### What task needs to be completed?

- Need to link coefficient files as part of the build process
- Need to modify where coefficient files come from at runtime
- Need to figure out if CRTM-JEDI needs to be a library or not on WCOSS

CoryMartin-NOAA:
Noting here that CRTM-JEDI is not needed for GFSv17, so it is a one-off build as part of the GCAFS package. Some work will need to be done to make CRTM optional in the GFSv17 package build.

PR #4727, opened 2026-03-30
Copy CRTM coefficients at build time to FIXglobal and use from there at runtime

CoryMartin-NOAA:
# Description
This PR copies CRTM coefficients needed for the special JEDI-GCAFS version of CRTM to FIXglobal at build time and copies from there, not externally, at runtime.

Resolves https://github.com/NOAA-EMC/global-workflow/issues/4725

### DIFF
--- a/ecf/scripts/gcdas/chem/analysis/jgcdas_aerosol_analysis_initialize.ecf
+++ b/ecf/scripts/gcdas/chem/analysis/jgcdas_aerosol_analysis_initialize.ecf
@@ -95,10 +95,6 @@ export cyc=%CYC%
 export cycle=t%CYC%z
 export HOMEglobal=${HOMEgcafs}
 
-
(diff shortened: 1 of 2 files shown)


TravisElless-NOAA:
The inclusion of the fix files works and is included in the run. I think the ecf scripts still need to be updated to point at the new location of these files? (e.g., https://github.com/NOAA-EMC/global-workflow/blob/dev/gcafs.v1/ecf/scripts/gcdas/chem/analysis/jgcdas_aerosol_analysis_initialize.ecf#L100)

CoryMartin-NOAA:
Good catch @TravisElless-NOAA I actually think we don't even need those lines here now. I'll change

TravisElless-NOAA (approved):
Got a full 00z gcdas cycle completed. Things look alright to me

PR #4734, opened 2026-03-31
Revert "Re-enable and upgrade verif-global to use spack-stack 2.0.0, metplus 6, met 12"

DavidHuber-NOAA:
The verif-global package needs to be reverted to MET/METplus v9/v3 while issues are worked out.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -46,6 +46,7 @@ body:
         - Gaea-C6
         - Orion
         - Hercules
+        - Jet
         - Cloud
     validations:
       required: true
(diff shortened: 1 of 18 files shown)


RussTreadon-NOAA:
Thank you @DavidHuber-NOAA for letting me know about this development.  I've been struggling for a few days with plotting and web visualization using the new package.  Reverting to the old package requires regeneration of GFS v18 stats since old and new formats are incompatible.  On the upside, we should be able to plot and access via the web interface again.  Should we add verification team member(s) as reviewers for this PR?

RussTreadon-NOAA (approved):
Haven't tested but EMC_verif-global stat generation, plotting, and web interface worked with the previous version.  Assume (hope) this remains true following the revert.

JessicaMeixner-NOAA:
I'm a code owner of something in here so I got ping'd for review.  I did a quick glance and just as an FYI - looks like there are some JET removals mixed in with the other updates.  My guess is a clean revert is easier, but pointing out all the same.  I have no objections to this PR, but not sure I'm a great reviewer for it either.

DavidHuber-NOAA:
@JessicaMeixner-NOAA I will go ahead and merge this as-is then open another quick-fix PR to clean up a few things.

PR #4774, opened 2026-04-13
Add support for NSF/NCAR Derecho

### LINKED ISSUE #4562: Port to Derecho

As part of the public release of global-workflow, EPIC plans to add support for NSF/NCAR's Derecho on-prem machine.

### Acceptance Criteria

- [ ] All components ported to Derecho
- [ ] All necessary data staged on Derecho
- [ ] Global-workflow can be built and run on Derecho
- [ ] All CI/CD tests pass on Derecho (when run manually)

EPIC will be completing this work. Much of it is already done as of the creation of this issue.

WalterKolczynski-NOAA:
# Description
Adds full support for NSF/NCAR's Derecho in preparation for EPIC's release of global workflow.

Derecho uses a different queue based on whether the job is shared of exclusive, so new queue, partition, and cluster variables are added to allow different values when running exclusive jobs. They all default to the regular value when not set, so prior behavior should be maintained.

There are two places where machine-specific code had to spill out into scripts.

The first is in `build_all.sh`. This change is just to reduce the number of jobs when running on the login node. Derecho has stricter controls on login node usage, and running with more than four jobs seems to get builds killed by the system. Because of this, it is highly recommend users build on compute nodes.

The second is in the `run_mpmd.sh` script. Derecho uses PBS Pro as a scheduler and `mpiexec` to launch MPI jobs, just like WCOSS. However, the implementation of MPMD is different in how it parses files passed via `--configfile`. Unfortunately, no solution was found to allow the same format on both machines, so a machine-specific section had to be added to construct a different mpmd command file.

Because Derecho is not a NOAA machine, there is no rstprod capability and only non-restricted data is available. Running without rstprod seems to have not been tested in a while, as a few scripts had to have parts updated to only try to set rstprod or require rstprod files to be present, when `CHGRP_RSTPROD` is on.

The section in `config.base` that turns Met-plus off for certain machines is consolidated with the first occurance of the setting.

The GCAFS fetch yaml setting is moved into the host files. The default is set inline in the `config.fetch`, as placing it in `defaults.yaml` would override the value in the host file.

The expected tests are disabled on Derecho: the "extended" tests that use gempack, as well as C48_gsienkf_atmDA and C48_ufsenkf_atmDA.

Resolves #4562

# Type of change

# Change characteristics

- Does this change require an update to any of the following submodules? YES (All changes already merged)
  - [ ] EMC verif-global 
  - [x] GDAS 
  - [ ] GFS-utils 
  - [x] GSI 
  - [x] GSI-monitor 
  - [ ] GSI-utils 
  - [x] Nexus
  - [x] UFS-utils 
  - [ ] UFS-weather-model 
  - [ ] wxflow 

# How has this been tested?
- All "PR" cases on Derecho ([gist](https://gist.github.com/WalterKolczynski-NOAA/482f823561e19aba6bcd61e110433d41))
- All "PR" cases on Ursa ([gist](https://gist.github.com/WalterKolczynski-NOAA/4e7cb32f02bd7c329cea861a014879d0))
- All "PR" cases on Hercules (awaiting rstprod)

# Checklist


WalterKolczynski-NOAA:
Will take off draft once I confirm test pass on Ursa and Hercules, but I don't expect any big changes.

dev/workflow/hosts.py
weihuang-jedi, line 1: /glade/u disk is probably more "derecho" specific.

weihuang-jedi: We see [ush/detect_machine.sh](https://github.com/NOAA-EMC/global-workflow/pull/4774/changes#diff-d719a3382f6d6036a32e1bf1fe192a6c92ebb3bb58262f944077f4c42983fbb5) use "/glade/u" for derecho.

weihuang-jedi (dismissed):

WalterKolczynski-NOAA:
I got removed from rstprod on MSU at some point, so that test is currently blocked (I put in a ticket to have it restored). But, it got past the fetch job, which is the only one I substantively changed the code for on MSU, so I'm marking this as ready for review.

dev/job_cards/rocoto/prep.sh
```diff
@@ -45,6 +45,9 @@ declare -rx COMOUT_OBS_PREV="${ROTDIR}/${GDUMP}.${gPDY}/${gcyc}/obs"
 declare -rx COMINobsproc_PREV="${DMPDIR}/${GDUMP}.${gPDY}/${gcyc}/atmos"
 declare -rx COMOUT_ATMOS_ANALYSIS_PREV="${ROTDIR}/${GDUMP}.${gPDY}/${gcyc}/analysis/atmos"
 
+# Prepobs still uses old HOME name
+declare -rx HOMEgfs="${HOMglobal}"
```
DavidHuber-NOAA:
```suggestion
declare -rx HOMEgfs="${HOMEglobal}"
```

dev/workflow/hosts/derecho.yaml
```diff
@@ -0,0 +1,31 @@
...
+SCHEDULER: pbspro
+QUEUE: develop
+QUEUE_EXCLUSIVE: main
+QUEUE_SERVICE: develop
+# HPSS properties
+HPSS_PROJECT: null
+ARCHCOM_TO: null
+ATARDIR: null
+CHGRP_RSTPROD: 'NO'
+CHGRP_CMD: 'chgrp rstprod'
```
DavidHuber-NOAA: Should this be one of the following?
`CHGRP_CMD: 'ls'`
or
`CHGRP_CMD: ''`

dev/workflow/hosts/hercules.yaml
```diff
@@ -30,3 +30,4 @@ CLIENT_GLOBUS_UUID: '869912fe-f6de-46c0-af10-b22efd84a022'
 # Features
 SUPPORTED_RESOLUTIONS: ['C1152', 'C768', 'C384', 'C192', 'C96', 'C48']
 DO_ARCHCOM: 'NO'
+FETCH_GCDAS_YAML_TMPL_LIST: '${PARMglobal}/fetch/${NET}_${APP}_gdas-anl_msu.yaml.j2,'
```
DavidHuber-NOAA, line 33: @CoryMartin-NOAA what do you think of this approach?

sorc/build_all.sh
```diff
@@ -79,6 +79,10 @@ export HOMEglobal
 
 echo "Sourcing global-workflow modules ..."
 source "${HOMEglobal}/dev/ush/gw_setup.sh"
+if [[ ${MACHINE_ID} == "derecho" ]]; then
+    # Derecho has stricter limits on head node usage
+    max_cores=4
```
DavidHuber-NOAA, line 92: Might be worth it to warn the users as this will take a very long time to build and suggest using `build_compute.sh` instead.


```suggestion
    echo "WARNING Derecho can only build with 4 cores at a time. Consider using build_compute.sh instead"
    max_cores=4
```

DavidHuber-NOAA (commented):
Changes look good overall. I just have a few comments.

dev/workflow/hosts/derecho.yaml
WalterKolczynski-NOAA: Would it break if we just remove it (and the other `null` lines related to HPSS) entirely, since they aren't applicable on Derecho?

sorc/build_all.sh
WalterKolczynski-NOAA: Yeah, I had had the same thought. My only question is whether to just have it as a warning and continue automatically, or ask for confirmation before starting.

DavidHuber-NOAA: I think just a warning, but I'm fine with either approach. I know for myself at least I sometimes run `generate_workflows.sh -u -b ...` and walk away. I'd be pretty annoyed to come back to see a prompt waiting for me.

dev/workflow/hosts/derecho.yaml
DavidHuber-NOAA: Ah, yes, it would break with a null string. Better to go with `'ls'`. Another alternative would be `true`, which can take any number of arguments and won't return anything to the prompt/log.

sorc/build_all.sh
WalterKolczynski-NOAA: Warning added.

dev/workflow/hosts/derecho.yaml
WalterKolczynski-NOAA: I meant removing the settings from `derecho.yaml` entirely. It shouldn't ever be used with `CHGRP_RSTPROD: 'NO'` anyway.

After rethinking the ATAR settings though, I'm going to provide settings for those since turning on `DO_ARCHCOM` is conceivable to local. `HPSS_PROJECT` would also be removed though.

DavidHuber-NOAA: OK, that makes sense. Yeah, I think with the script changes you made, you should be good to go if you remove `CHGRP_CMD`.

If anyone is interested in the globus-to-Mercury-to-HPSS capability, we can try dusting that off as well. I think SFS is successfully using it on Hercules.

DavidHuber-NOAA (dismissed):
Changes look good to me. Thanks for the port @WalterKolczynski-NOAA.

DavidHuber-NOAA:
Running CI on WCOSS2 for sanity.

DavidHuber-NOAA:
All tests passed on WCOSS2 except C96_atm3DVar_extended, which is still running, but has completed two full cycles. Merging.

DavidHuber-NOAA (approved):

WalterKolczynski-NOAA:
Thanks Dave!

Issue #4802, opened 2026-04-20
Develop a forecast manager for the global workflow


DavidHuber-NOAA:
### What new functionality do you need?

A forecast manager job is needed for the global workflow. The manager will handle copying products to COM from the running directory and will run alongside the forecast jobs.

### Acceptance Criteria

- [ ] A manager job is written such that it can detect and copy products to COM as they are written
- [ ] Dead links are removed from the forecast running directories (at least in production mode)
- [ ] Products are identical
- [ ] No significant increase in overall runtime
- [ ] The manager job adapts automatically to the model components (atm, ocean, ice, waves, aerosols, etc)

DavidHuber-NOAA:
In a meeting with @CarlosMDiaz-NCO today, we discussed the requirements and proposed a design for the forecast manager job. There were also discussions on how the umbrella directory is currently designed and how it should be designed going forward. That will be covered in a separate issue.

Here are the notes on the requirements and design of the forecast manager:

* Forecast/forecast manager requirements
   * (Umbrella working directory) Data storage for two separate jobs (forecast and manager)
      * Common space currently in $DATAROOT/$RUN.$PDY/${RUN}fcst.${PDY}${cyc}
         * Carlos will ask NCO team on this naming convention
      * The model writes products and product logs to ${DATA}/${component}_OUTPUT/
      * Currently, these directories are linked to subdirectories under the common DATAROOT space
   * Dead links from the common DATAROOT space to COM should be removed
   * How many tasks may the manager run with?
      * Since 2-5 components (Atm, waves, ocean, ice, aerosols) are forecasted, it may make sense to have one core/component
      * MPMD may be an option if 1 core is insufficient
   * Use cpfs for copying back to COM
   * Specifically needed for the GFS forecast, not the GDAS or EnKFGDAS
      * With the GDAS and EnKFGDAS, we will just copy the products at the end of the job
* Forecast manager design
   * Proposal
      * At the start of the forecast job, create a set of ascii tables (one for each component) for all products
         * Target file, target log (indicating writing is complete), destination file in COM, destination log file in COM
      * Copy this file to COM to trigger the manager job
      * Forecast manager
         * Ingest the tables
         * Loop and wait for log files associated with targets
            * Establish a maximum wait time and error if that time is exceeded
               * Carlos will ask NCO team on preference of how long the timeout should be OR if there should be one at all
         * When the log files are present, copy the target file to COM first, then the associated target logs to COM
[...]

PR #4860, opened 2026-04-29
[develop] Forecast Manager: Real-Time Product Copy to COM with Adaptive, Component-Aware Segment Management

### LINKED ISSUE #4821: (develop) Develop a forecast manager for the global workflow

AntonMFernando-NOAA:
# Description
- This PR introduces a robust forecast manager job for the global workflow, designed to handle real-time copying of model products from the running directory to COM as they are written. The manager runs in parallel with forecast jobs and adapts automatically to the active model components (atmosphere, ocean, ice, waves, aerosols, etc).

  Resolves #4821
  Refs #4802 

# Type of change

# Change characteristics

- Is this change expected to change outputs (e.g. value changes to existing outputs, new files stored in COM, files removed from COM, filename changes, additions/subtractions to archives)? NO (If YES, please indicate to which system(s))
  - [ ] GFS
  - [ ] GEFS
  - [ ] SFS
  - [ ] GCAFS
  - [ ] EMC verif-global 
  - [ ] GDAS 
  - [ ] GFS-utils 
  - [ ] GSI 
  - [ ] GSI-monitor 
  - [ ] GSI-utils 
  - [ ] UFS-utils 
  - [ ] UFS-weather-model 
  - [ ] wxflow 

# How has this been tested?
- tests on all platforms

# Checklist
- [ ] A manager job is implemented to detect and copy products to COM as soon as they are available.
- [ ] Dead links are removed from the forecast running directories (at least in production mode).
- [ ] All products in COM are bitwise identical to those produced by the legacy workflow.
- [ ] No significant increase in overall runtime is observed.
- [ ] The manager job automatically adapts to the set of model components in use for each workflow configuration.


AntonMFernando-NOAA:
@CarlosMDiaz-NCO @lgannoaa @DavidHuber-NOAA @CoryMartin-NOAA @TravisElless-NOAA This is the initial implementation of the forecast manager for the GFS forecast. I ran tests on WCOSS2, and all passed except for C48_S2SW_extended and C96C48mx500_S2SW_cyc_gfs. I identified the cause of these failures, applied a fix, and am rerunning the tests. I am opening this draft PR to gather feedback before requesting a full review. WCOSS2 tests are available at `/lfs/h2/emc/stmp/anton.fernando/RUNTESTS/EXPDIR`.

DavidHuber-NOAA:
@AntonMFernando-NOAA I think NCO's preference is that the manager runs with one core if possible. I'd like to test this out to see what the timing works out as with one core vs four cores at production resolution.

Could you add an option to `exglobal_forecast_manager.sh` to concatenate the `ATM`, `OCN`, `ICE`, and `WW3` tables and then run `ush/forecast_manager.sh` on the whole table instead of running them in MPMD?

AntonMFernando-NOAA:
> @AntonMFernando-NOAA I think NCO's preference is that the manager runs with one core if possible. I'd like to test this out to see what the timing works out as with one core vs four cores at production resolution.
> 
> Could you add an option to `exglobal_forecast_manager.sh` to concatenate the `ATM`, `OCN`, `ICE`, and `WW3` tables and then run `ush/forecast_manager.sh` on the whole table instead of running them in MPMD?

@DavidHuber-NOAA Yes. Let me add that option.

AntonMFernando-NOAA:
@DavidHuber-NOAA Edit L64 of `dev/scripts/exglobal_forecast_manager.sh` to test:
`FCST_MGR_MPMD="${FCST_MGR_MPMD:-YES}`

ush/forecast_manager.sh
```diff
@@ -0,0 +1,139 @@
...
+            if [[ ${copy_err} -ne 0 ]]; then
+                echo "FATAL ERROR [${component}]: cpfs '${local_data[j]}' -> '${com_data[j]}' failed (err=${copy_err})" >&2
+                exit "${copy_err}"
+            fi
+        done
+
+        # Copy sentinel log last.
+        # FV3/WW3 pattern: local_log is a separate file written by the model after each
+        # output period; copy it to COM as the completion signal.
+        # MOM6/CICE pattern: local_log == local_data (no separate model-written sentinel);
```
JessicaMeixner-NOAA, line 113: MOM6 files now have a log file, we just need to enable that in the develop branch as it is in the dev/gfsv17 branch so that you can have the actual log.

ush/forecast_postdet_cice.sh
```diff
@@ -0,0 +1,166 @@
+#! /usr/bin/env bash
```
JessicaMeixner-NOAA: I get splitting this up, however it's now hard to see what's been changed since everything is now in it's own file instead of a changed file.  I wonder if that could be done independently of this PR to help with the review process?

ush/forecast_postdet_ww3.sh
```diff
@@ -0,0 +1,216 @@
...
+    while [[ ${fhr} -le ${FHMAX_WAV} ]]; do
+        fhr3=$(printf '%03d' "${fhr}")
+        vdate=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} + ${fhr} hours" +%Y%m%d.%H0000)
+        local ww3_pnt_local="${DATAoutput}/WW3_OUTPUT/${vdate}.out_pnt.ww3.nc"
+        local ww3_pnt_local_log="${DATAoutput}/WW3_OUTPUT/log.${vdate}.out_pnt.ww3.txt"
+        local ww3_pnt_com="${COMOUT_WAVE_HISTORY}/${RUN}.t${cyc}z.points.f${fhr3}.nc"
+        local ww3_pnt_com_log="${COMOUT_WAVE_HISTORY}/${RUN}.t${cyc}z.points.f${fhr3}.log"
+        if [[ "${use_mgr_ww3}" == "YES" ]]; then
+            # Each WW3 point file has its own per-file sentinel log
+            echo "${ww3_pnt_local} ${ww3_pnt_local_log} ${ww3_pnt_com} ${ww3_pnt_com_log}" >> "${ww3_table}"
```
JessicaMeixner-NOAA: Will everything be looked for in the order in which it's in ww3_table?    For example, will all of the above gridded files be looked for before looking at anything from the points?

ush/forecast_manager.sh
AntonMFernando-NOAA: @JessicaMeixner-NOAA That's great!! Should I leave this as is and remove the dummy log creation once the actual logging is enabled? I assume we can turn that on after this PR.

ush/forecast_postdet_cice.sh
AntonMFernando-NOAA: > I get splitting this up, however it's now hard to see what's been changed since everything is now in it's own file instead of a changed file. I wonder if that could be done independently of this PR to help with the review process?

@JessicaMeixner-NOAA Understood! Let me put everything back in `forecast_det.sh`.

ush/forecast_manager.sh
JessicaMeixner-NOAA: It's overdue to get the MOM6 logging turned on in develop... maybe i can just try to pull that in now instead.

ush/forecast_postdet_cice.sh
JessicaMeixner-NOAA: Others may disagree with me.  And I think splitting it up is likely the way to go, but maybe that's a PR by itself versus combined w/lots of other changes?   I might be alone in this thought - so before you do more work maybe some others can chime in too.   @DavidHuber-NOAA @TravisElless-NOAA

ush/forecast_postdet_ww3.sh
AntonMFernando-NOAA: > Will everything be looked for in the order in which it's in ww3_table? For example, will all of the above gridded files be looked for before looking at anything from the points?

@JessicaMeixner-NOAA No. The loop doesn't process entries strictly top-to-bottom in one pass. On each iteration of the outer while, it scans all `i` from `0` to `count-1`, but it only acts on an entry when its sentinel log is present. So the actual copy order is driven by when the model writes each sentinel, not by table row order.

If the gridded sentinel appears at hour 6 and the points sentinel also appears at hour 6 (or earlier), the manager could process points before some gridded files depending on which sentinel file lands first on disk. The for-loop scans index 0 upward, so among entries whose sentinel is already present in the same pass, table order does apply. But the outer while loop resets to i=0 after each sleep cycle, not after each individual copy.

JessicaMeixner-NOAA: Thank you!! Just wanted to double check.

ush/forecast_manager.sh
AntonMFernando-NOAA: @JessicaMeixner-NOAA Thank you!! It will make things clean.

JessicaMeixner-NOAA: https://github.com/NOAA-EMC/global-workflow/pull/4864 -- just need to run CI on this.

dev/jobs/JGLOBAL_FORECAST_MANAGER
```diff
@@ -0,0 +1,107 @@
...
+#           the DATAjob umbrella directory with the forecast job and reads
+#           product tables from COMOUT_CONF that were written by the forecast
+#           job pre-run setup phase.
+#
+#           Currently manages real-time copy for components that emit per-file
+#           sentinel log files (FV3ATM and WW3). Ocean and ice history files
+#           are copied at the end of the forecast job by JGLOBAL_FORECAST.
+#############################################################################
+
+export DATAjob="${DATAROOT}/${RUN}fcst.${PDY:-}${cyc}"
```
CarlosMDiaz-NCO: Is there a reason the manager's working directory $DATA needs to be under the $DATAjob shared space? If not, export DATA=${DATA:-${DATAROOT}/${jobid}} should work for this job. and $DATAjob would just be used to reference where the needed files will be to copy out

Also, this goes to the forecast job as well, but the shared working directory name should be closer to the job name, so in this case - ${RUN}_forecast_${PDY:-}${cyc}. ($PDY will not be available at this point in production, but it looks like this syntax handles that case so we'd see gcafs_forecast_00)

dev/jobs/JGLOBAL_FORECAST_MANAGER
```diff
@@ -0,0 +1,107 @@
...
+    declare -rx COMOUT_WAVE_HISTORY="${COMOUT}/model/wave/history"
+fi
+
+if [[ "${DO_OCN:-NO}" == "YES" ]]; then
+    declare -rx COMOUT_OCEAN_HISTORY="${COMOUT}/model/ocean/history"
+fi
+
+if [[ "${DO_ICE:-NO}" == "YES" ]]; then
+    declare -rx COMOUT_ICE_HISTORY="${COMOUT}/model/ice/history"
+fi
```
DavidHuber-NOAA, line 76: We'll also need a block for aerosols to support GCAFS. The GFS `ATMA` case hasn't been tested in a while, so I'm not sure it will work for testing this capability.

Once you have the GFS working, could you also add this capability to GCAFS so it can be tested? I'm OK with holding off on bringing this to SFS and GEFS to a later date.

ush/forecast_postdet_cice.sh
DavidHuber-NOAA: @JessicaMeixner-NOAA @AntonMFernando-NOAA Good points. I think splitting it up is the way to go, but I agree, it is harder to see what's changed. Let's target the split up for a follow-up PR.

ush/wait_for_table.sh
DavidHuber-NOAA, line 1: Can we use `wait_for_file.sh` for this instead? I don't really want to add another script if we don't need to. The `INFO` `echo`s could be added to the ex-script instead.

dev/jobs/JGLOBAL_FORECAST_MANAGER
AntonMFernando-NOAA: @CarlosMDiaz-NCO This was due to a misunderstanding on my part. I’ll update it as suggested.. Thank you!!

dev/job_cards/rocoto/fcst_manager.sh
```diff
@@ -0,0 +1,25 @@
...
+set -x
+
+###############################################################
+# Source FV3GFS workflow modules
+source "${HOMEglobal}/ush/detect_machine.sh"
+if [[ "${MACHINE_ID}" == "wcoss2" ]]; then
+    source "${HOMEglobal}/dev/ush/load_modules.sh" ufswm
+else
+    source "${HOMEglobal}/dev/ush/load_modules.sh" run
+fi
```
DavidHuber-NOAA: I think `run` modules will work for all machines.
```suggestion
source "${HOMEglobal}/dev/ush/load_modules.sh" run
```

dev/jobs/JGLOBAL_FORECAST_MANAGER
```diff
@@ -0,0 +1,112 @@
...
+export COMOUT_CONF COMOUT_ATMOS_HISTORY COMOUT_ATMOS_MASTER
+[[ "${DO_WAVE}" == "YES" ]] && export COMOUT_WAVE_HISTORY
+[[ "${DO_OCN:-NO}" == "YES" ]] && export COMOUT_OCEAN_HISTORY
+[[ "${DO_ICE:-NO}" == "YES" ]] && export COMOUT_ICE_HISTORY
+[[ "${DO_AERO_FCST:-NO}" == "YES" ]] && export COMOUT_CHEM_HISTORY
+##############################################
+# Begin JOB SPECIFIC work
+##############################################
+
+: "${FORECASTMGRSH:=${SCRglobal}/exglobal_forecast_manager.sh}"
```
DavidHuber-NOAA: Very nitpicky on my part, but I would prefer to have a standard way to override defaults. As such, can we go with what we use elsewhere?

```suggestion
FORECASTMGRSH="${FORECASTMGRSH:-${SCRglobal}/exglobal_forecast_manager.sh}"
```

dev/parm/config/gcafs/config.resources
```diff
@@ -1354,6 +1354,20 @@ case ${step} in
...
+    # Polls DATAoutput for sentinel log files and copies products to COM in real time.
+    # Walltime must be long enough to cover the full forecast run.
+    # is_exclusive must be reset here because config.fcst (loaded before this file)
+    # exports is_exclusive=True; the manager does not need exclusive node access.
+    export is_exclusive=False
+    ntasks=2
+    tasks_per_node=2
+    threads_per_task=1
+    memory="8GB"
+    walltime="08:00:00"
```
DavidHuber-NOAA, line 1359: Noting that this may need to be 6 on some machines. This can go in `config.resources.<HOST>` if needed.

Have you tested the wall time limits on the RDHPCS machines?

dev/parm/config/gfs/config.resources
```diff
@@ -985,6 +985,20 @@ case ${step} in
...
     ;;
 
+  "fcst_manager")
+    # Forecast manager: one MPMD rank per active component (ATM, WW3, OCN, ICE).
+    # Polls DATAoutput for sentinel log files and copies products to COM in real time.
+    # Walltime must be long enough to cover the full forecast run.
+    # is_exclusive must be reset here because config.fcst (loaded before this file)
+    # exports is_exclusive=True; the manager does not need exclusive node access.
+    export is_exclusive=False
+    ntasks=4
```
DavidHuber-NOAA: This should probably scale with the number of components. This may not be straightforward to calculate, though.

dev/jobs/JGLOBAL_FORECAST_MANAGER
```diff
@@ -0,0 +1,112 @@
...
+if [[ "${DO_AERO_FCST:-NO}" == "YES" ]]; then
+    declare -rx COMOUT_CHEM_HISTORY="${COMOUT}/model/chem/history"
+fi
+
+# Export COM variables needed by each forecast_manager.sh rank
+export COMOUT_CONF COMOUT_ATMOS_HISTORY COMOUT_ATMOS_MASTER
+[[ "${DO_WAVE}" == "YES" ]] && export COMOUT_WAVE_HISTORY
+[[ "${DO_OCN:-NO}" == "YES" ]] && export COMOUT_OCEAN_HISTORY
+[[ "${DO_ICE:-NO}" == "YES" ]] && export COMOUT_ICE_HISTORY
+[[ "${DO_AERO_FCST:-NO}" == "YES" ]] && export COMOUT_CHEM_HISTORY
```
DavidHuber-NOAA: Prefer these to be `if` blocks.

dev/scripts/exglobal_cleanup.sh
```diff
@@ -5,13 +5,24 @@ echo "Begin Cleanup ${DATAROOT}!"
...
 rm -rf "${DATAROOT}/${RUN}efcs"*"${PDY:-}${cyc}"
 ###############################################################
+# Remove MOM6/CICE completion sentinel .log files from the current cycle.
+# These marker files are written by the forecast manager to confirm that each
+# output .nc was fully copied to COM.  They are not needed after the cycle
+# completes and the cleanup job runs.
+for _sentinel_log in \
+    "${ROTDIR}/${RUN}.${PDY}/${cyc}/model/ocean/history"/*.nc.log \
+    "${ROTDIR}/${RUN}.${PDY}/${cyc}/model/ice/history"/*.nc.log; do
+    [[ -f "${_sentinel_log}" ]] && rm -f "${_sentinel_log}"
```
DavidHuber-NOAA: No need for the logic.
```suggestion
    rm -f "${_sentinel_log}"
```

dev/jobs/JGLOBAL_FORECAST_MANAGER
```diff
@@ -0,0 +1,112 @@
...
+#
+#           Currently manages real-time copy for FV3ATM, MOM6 ocean history,
+#           WW3, CICE, and GOCART/aerosol (all gfs only). MOM6 restarts and
+#           config files are still copied at the end of the forecast job by
+#           JGLOBAL_FORECAST.
+#############################################################################
+
+export DATAjob="${DATAROOT}/${RUN}_forecast.${PDY:-}${cyc}"
+export DATA="${DATA:-${DATAROOT}/${jobid}}"
+source "${HOMEglobal}/ush/jjob_header.sh" -e "fcst_manager" -c "base fcst"
```
DavidHuber-NOAA: Should this be using the `fcst_manager` config? Not the `fcst` config? Or maybe it needs both?

```suggestion
source "${HOMEglobal}/ush/jjob_header.sh" -e "fcst_manager" -c "base fcst_manager"
```

dev/workflow/applications/applications.py
```diff
@@ -291,6 +291,8 @@ def _source_configs(self, conf: Configuration, run: str = "gfs", log: bool = Tru
                 files += ['config.anal', 'config.eupd']
             elif config in ['efcs']:
                 files += ['config.fcst', 'config.efcs']
+            elif config in ['fcst_manager']:
+                files += ['config.fcst', 'config.fcst_manager']
```
DavidHuber-NOAA: Does the `fcst_manager` need `config.fcst`?

ush/forecast_manager.sh
```diff
@@ -0,0 +1,164 @@
...
+if [[ ! -f "${table_file}" ]]; then
+    echo "FATAL ERROR: Product table '${table_file}' not found for component '${component}'" >&2
+    exit 1
+fi
+
+# Load table into indexed arrays
+declare -a local_data local_log com_data com_log done_flag
+count=0
+while read -r ld ll cd cl; do
+    [[ -z "${ld}" || "${ld:0:1}" == "#" ]] && continue
```
DavidHuber-NOAA:
```suggestion
    if [[ -z "${ld}" || "${ld:0:1}" == "#" ]]; then
        continue
    
```

ush/forecast_manager.sh
```diff
@@ -0,0 +1,164 @@
...
+    local_data[count]="${ld}"
+    local_log[count]="${ll}"
+    com_data[count]="${cd}"
+    com_log[count]="${cl}"
+    done_flag[count]="NO"
+    ((count++)) || true
```
DavidHuber-NOAA, line 51: Why the `|| true`?

Oh, I think I know. This is a quirk of shell scripting. If `count++` evaluates to `1`, this will return a non-zero status. Is that right?

ush/forecast_manager.sh
```diff
@@ -0,0 +1,164 @@
...
+
+remaining=${count}
+start_time=$(date +%s)
+fcst_done_idle=0
+
+while [[ ${remaining} -gt 0 ]]; do
+    remaining_before=${remaining}
+    for ((i = 0; i < count; i++)); do
+        [[ "${done_flag[i]}" == "YES" ]] && continue
+        [[ ! -f "${local_log[i]}" ]] && continue
```
DavidHuber-NOAA:
```suggestion
        if [[ "${done_flag[i]}" == "YES" ]]; then
            continue
        fi
        
        if [[ ! -f "${local_log[i]}" ]]; then
            continue
        fi
```

ush/forecast_manager.sh
```diff
@@ -0,0 +1,164 @@
...
+                    ((remaining--)) || true
+                fi
+            done
+            continue
+        fi
+
+        # Copy all data files that share this sentinel (data first, log last)
+        for ((j = 0; j < count; j++)); do
+            [[ "${done_flag[j]}" == "YES" ]] && continue
+            [[ "${local_log[j]}" != "${this_ll}" ]] && continue
```
DavidHuber-NOAA:
```suggestion
            if [[ "${done_flag[j]}" == "YES" ]]; then
                continue
            fi
            if [[ "${local_log[j]}" != "${this_ll}" ]]; then
                continue
            fi
```

You could also combine these checks with `||` if desired.

ush/forecast_manager.sh
```diff
@@ -0,0 +1,164 @@
...
+            com_dir=$(dirname "${com_data[j]}")
+            if [[ ! -d "${com_dir}" ]]; then mkdir -p "${com_dir}"; fi
```
DavidHuber-NOAA:
```suggestion
            if [[ ! -d "${com_dir}" ]]; then
                mkdir -p "${com_dir}"
            fi
```

ush/forecast_manager.sh
```diff
@@ -0,0 +1,164 @@
...
+
+        # Copy sentinel log last.
+        # FV3/WW3 pattern: local_log is a separate file written by the model after each
+        # output period; copy it to COM as the completion signal.
+        # MOM6/CICE pattern: local_log == local_data (no separate model-written sentinel);
+        # the manager writes a small text marker to com_log after the data copy to confirm
+        # the copy completed successfully and to provide a clean downstream dependency.
+        if [[ ! -f "${this_cl}" ]]; then
+            cl_dir=$(dirname "${this_cl}")
+            if [[ ! -d "${cl_dir}" ]]; then mkdir -p "${cl_dir}"; fi
```
DavidHuber-NOAA:
```suggestion
            if [[ ! -d "${cl_dir}" ]]; then
                mkdir -p "${cl_dir}"
            fi
```

ush/forecast_manager.sh
```diff
@@ -0,0 +1,164 @@
...
+        # Mark all rows for this sentinel as done
+        for ((j = 0; j < count; j++)); do
+            if [[ "${done_flag[j]}" == "NO" && "${local_log[j]}" == "${this_ll}" ]]; then
+                done_flag[j]="YES"
+                ((remaining--)) || true
+            fi
+        done
+    done
+
+    [[ ${remaining} -eq 0 ]] && break
```
DavidHuber-NOAA:
```suggestion
    if [[ ${remaining} -eq 0 ]]; then
        break
    fi
```

ush/forecast_postdet.sh
```diff
@@ -324,26 +325,71 @@ EOF
...
+        case "${RUN}" in
+            gfs | gcafs) use_mgr="YES" ;;
+                # TODO: enable forecast manager for gefs, sfs once tested
+                # gefs | sfs) use_mgr="YES" ;;
+            *) ;;
+        esac
+
+        local atm_table="${DATAjob}/atm_products_seg${FCST_SEGMENT:-0}.txt"
+        if [[ "${use_mgr}" == "YES" ]]; then
+            rm -f "${atm_table}"
```
DavidHuber-NOAA, line 341: If the forecast for a particular segment needs to restart, does this table end up getting all of the products listed, or just those from the restart time onward? I think it's the former (which is what we want), but want to be sure.

ush/forecast_postdet.sh
```diff
@@ -324,26 +325,71 @@ EOF
...
+        # For GDAS/enkfGDAS: keep NLN symlinks so analysis jobs can read outputs during the run.
+        local use_mgr="NO"
+        case "${RUN}" in
+            gfs | gcafs) use_mgr="YES" ;;
+                # TODO: enable forecast manager for gefs, sfs once tested
+                # gefs | sfs) use_mgr="YES" ;;
+            *) ;;
+        esac
+
+        local atm_table="${DATAjob}/atm_products_seg${FCST_SEGMENT:-0}.txt"
```
DavidHuber-NOAA, line 339: Should `FCST_SEGMENT` be defaulted to 0 once in `exglobal_forecast.sh` rather than having `FCST_SEGMENT:-0` everywhere?

DavidHuber-NOAA (changes requested):
Overall, this looks great. I have a lot of nitpicky comments, but the overall structure is good.

dev/parm/config/gcafs/config.resources
AntonMFernando-NOAA: @DavidHuber-NOAA I think the limit is 12:00:00 for RDHPCS machines. We don’t need to keep walltime="08:00:00" for the forecast job since this is only for testing. Someone more familiar with the setup can suggest an appropriate walltime.

ush/forecast_manager.sh
AntonMFernando-NOAA: @DavidHuber-NOAA Yes. That's exactly why.

dev/jobs/JGLOBAL_FORECAST_MANAGER
AntonMFernando-NOAA: @DavidHuber-NOAA These are actually redundant. Got rid of them. `declare -rx` already exports variables at declaration time.

dev/workflow/applications/applications.py
AntonMFernando-NOAA: @DavidHuber-NOAA Seems you are right. Every variable of the manager comes from either `config.base` or `config.fcst_manager` it seems. Making the change.

ush/forecast_postdet.sh
AntonMFernando-NOAA: @DavidHuber-NOAA All products from `FHMIN` to `FHMAX`. The table is rebuilt unconditionally (`rm -f "${atm_table}`" then the full for `fhr` in ${`FV3_OUTPUT_FH`} loop) and `FV3_OUTPUT_FH` is always built from `FHMIN`. It is never trimmed to the restart hour. `RERUN_DATE` only affects which restart files are staged into DATA/INPUT. Iit does not change the output hour list.

ush/forecast_postdet.sh
AntonMFernando-NOAA: @DavidHuber-NOAA corrected.

DavidHuber-NOAA (approved):
Changes look good. I'm running a high resolution case now.

DavidHuber-NOAA:
Closing for now. The logging options in dev/gfs.v17 are not available in the main branch of the weather model and will be different. This will need to be revisited at a later time.

PR #4856, opened 2026-04-28
(develop) Reduce debug output from logit decorator

DavidHuber-NOAA:
# Description
This reduces the amount of debug output written by `Task` methods and Python ex-scripts by
1. Modifying the wxflow `logit` method to avoid printing `self`
2. No longer decorating `__init__` methods with `logit`

Refs #4721 

# Type of change
- [ ] Bug fix (fixes something broken)
- [ ] New feature (adds functionality)
- [x] Maintenance (code refactor, clean-up, new CI test, etc.)

# Change characteristics
- Is this change expected to change outputs NO
- Is this a breaking change (a change in existing functionality)? NO
- Does this change require a documentation update? NO
- Does this change require an update to any of the following submodules? YES
  - [x] wxflow https://github.com/NOAA-EMC/wxflow/pull/73
  - [x] GDASApp https://github.com/NOAA-EMC/GDASApp/pull/2129

# How has this been tested?
Testing is in progress on C6 and WCOSS2.

# Checklist
- [x] Any dependent changes have been merged and published
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing tests pass with my changes
- [x] This change is covered by an existing CI test or a new one has been added

PR #4865, opened 2026-04-29
Hotfix pipeline clone pr

TerrenceMcGuinness-NOAA:
## Discription

Fixes intermittent GitLab CI build failures on Gaea where `git submodule update --init --recursive` fails with:

```
fatal: remote error: upload-pack: not our ref e454a2324fc0bd8e7561725912138cafa7a1a741
```

…even though the pinned `sorc/gdas.cd` SHA is valid and reachable.

## Root Cause

A git client-side bug in versions **≤ 2.45** (Gaea's Spack stack currently ships **git 2.43.x**), interacting with `gh pr checkout`:

1. `gh pr checkout <PR>` sets `branch.<name>.remote` to the **PR author's fork** \
(e.g. `https://github.com/<author>/global-workflow`).
2. When a submodule is pinned to a SHA that isn't on any advertised branch (common for cross-repo PRs like gdas.cd against `refs/pull/*/head`), `git submodule--helper` falls back to fetching that bare SHA.
3. To build the fallback fetch URL, older git uses the parent repo's **branch upstream URL** as the base — i.e. the author's fork — instead of the submodule's own `.gitmodules` URL.
4. The author's fork doesn't contain the gdas.cd SHA, so the server replies `not our ref` and the build fails.

git **≥ 2.47** computes the fallback URL correctly and is unaffected.

## Fix

Three config operations added to `.build_template` in `.gitlab-ci.yml`, immediately after `gh pr checkout`:

```bash
git remote set-url origin "${GW_REPO_URL}"
_br=$(git branch --show-current)
git config "branch.${_br}.remote"     origin
git config "branch.${_br}.pushRemote" origin
```

This re-pins the checked-out PR branch's upstream to `NOAA-EMC/global-workflow`, so the submodule helper's fallback URL is correct.

## Scope / Risk

- Touches only `.gitlab-ci.yml` `.build_template`. No build/runtime code affected.
- Behavior unchanged on git ≥ 2.47 (the re-pin has no effect on submodule fetching there).
- No impact on `gh pr` operations — `origin` already pointed at NOAA-EMC; this just re-asserts it after `gh` mutated it.

## Validation

- Test scripts using (git 2.43.6) **without** patch → fails identically to Gaea (`not our ref e454a2324f…`).
- Test scripts **with** patch → `[OK] all submodules resolved`, `==> DONE`.
- Tested against PR #4856 (`feature/no_self`) which exposes the bug via gdas.cd pinned to `refs/pull/2129/head`.

## Follow-up

Remove this workaround once Gaea's Spack stack ships git ≥ 2.47.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -157,6 +157,17 @@ include:
         git remote add github "${GW_REPO_URL}"
         git fetch github
         ${GH} pr checkout "${PR_NUMBER}"
+        # Workaround for git <= 2.45 (e.g. Gaea Spack git 2.43.x):
+        # `gh pr checkout` repoints branch.<name>.remote to the PR author's fork.
+        # git submodule--helper then builds an unadvertised-SHA fallback fetch URL
+        # from that upstream, producing "<author-fork>.git <gdas-sha>" -> "not our ref".
+        # Re-pin the current branch upstream to NOAA-EMC/global-workflow so the
+        # fallback URL is correct. Remove once Gaea ships git >= 2.47.
+        git remote set-url origin "${GW_REPO_URL}"
+        _br=$(git branch --show-current)
+        git config "branch.${_br}.remote"     origin
+        git config "branch.${_br}.pushRemote" origin
+        unset _br
         git submodule update --init --recursive -j 8
       fi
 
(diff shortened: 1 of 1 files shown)


DavidHuber-NOAA (approved):
Thanks for sniffing this out @TerrenceMcGuinness-NOAA!!

Issue #4850, opened 2026-04-24
Output from the global_cycle executable does not write to job log file


TravisElless-NOAA:
### What is wrong?

Output from the global_cycle executable is currently being written to a file in the DATA directory. This output never makes it to the job log files that get written to COM output, which is problematic for debugging purposes. 

### What should have happened?

The standard output from the global cycle executable should be included in the log file included in the COM output

### What machines are impacted?

All or N/A

### What global-workflow hash are you using?

develop

### Steps to reproduce

1. Run a job that runs the ush/global_cycle.sh script
2. Look for output for the global_cycle executable

### Do you have a proposed solution?

Change `${APRUNCY} "${CYCLEXEC}" 1>> "${PGMOUT}" 2>> "${PGMERR}"` to `${APRUNCY} "${CYCLEXEC}" 1>&1 2>&2` in [ush/global_cycle.sh](https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/global_cycle.sh#L341C1-L341C57)

DavidHuber-NOAA:
Completed via #4904 and #4925

PR #4293, opened 2025-12-04
Rename a bunch of jgdas/exgdas to jglobal/exglobal

aerorahul:
# Description
This PR:
- Resolves #4241
- Renames the job and scripts as described in the issue.

# Type of change
- [X] Maintenance (code refactor, clean-up, new CI test, etc.)

# Change characteristics
- Is this a breaking change (a change in existing functionality)? NO
- Does this change require a documentation update? NO
- Does this change require an update to any of the following submodules? NO

# How has this been tested?
Tested on GaeaC6 at: 
`/gpfs/f6/drsa-precip4/world-shared/Rahul.Mahajan/RUNTESTS/gw-4241`

# Checklist
- [ ] Any dependent changes have been merged and published
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have documented my code, including function, input, and output descriptions
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] This change is covered by an existing CI test or a new one has been added
- [ ] Any new scripts have been added to the .github/CODEOWNERS file with owners
- [ ] I have made corresponding changes to the system documentation if necessary

PR #4904, opened 2026-05-11
[develop] Improve eupd/global cycle log output

TravisElless-NOAA:
# Description
It was discovered during GCAFS testing that standard output from the global cycle executable was not found in the job output log file. During this evaluation, it was also noted that the enkf update job also was having standard output written only to the `enkfstat.txt` file and not to the job log file. This PR updates the global cycle script to send the executable output to the job logfile instead to the `DATA` directory. For the enkf update script, changes allow output to be captured in the job log file, while maintaining this information in the `enkfstat.txt` file. Note the order of information within the `enkfstat.txt` file does change from the existing file, but all information from the existing file is captured in the new one. 

This PR also updates the CODEOWNERS file to reflect that the enkf jobs/scripts were renamed from gdas to global in #4293.

Refs #4850 
# Type of change

# Change characteristics

- Is this change expected to change outputs (e.g. value changes to existing outputs, new files stored in COM, files removed from COM, filename changes, additions/subtractions to archives)? YES, the order of information in the `enkfstat.txt` files changes
  - [x] GFS
  
# How has this been tested?
C96C48_hybatmDA ci test along with 2 cycles of an 80-member enkf test were run on WCOSS2

# Checklist


dev/scripts/exglobal_enkf_update.sh
```diff
@@ -323,14 +323,12 @@ export pgm=${ENKFEXEC}
 source prep_step
 
 cpreq "${ENKFEXEC}" "${DATA}"
-${APRUN_ENKF} "${DATA}/$(basename "${ENKFEXEC}")" 1> stdout 2> stderr && true
+${APRUN_ENKF} "${DATA}/$(basename "${ENKFEXEC}")" 2>&1 | tee enkfstat.txt && true
```
DavidHuber-NOAA, line 326: Noting that this kind of solution may be beneficial elsewhere now that `set -o pipefail` is enabled.

DavidHuber-NOAA (dismissed):
Looks good!

DavidHuber-NOAA:
@TravisElless-NOAA which machine did you run the `C96C48_hybatmDA` case on (asking so I can add the appropriate label)?

TravisElless-NOAA:
> @TravisElless-NOAA which machine did you run the `C96C48_hybatmDA` case on (asking so I can add the appropriate label)?

@DavidHuber-NOAA  sorry for not adding that info. I did my testing on WCOSS2

dev/scripts/exglobal_enkf_update.sh
RussTreadon-NOAA: Gemini and Copilot indicate that `&& true` is not needed if `set -o pipefail` is enabled.  Is this true?  I'm asking because I don't know.

RussTreadon-NOAA (commented):
Looks good to me.  Question about the need for `&& true` when `set -o pipefail` is enabled.

dev/scripts/exglobal_enkf_update.sh
DavidHuber-NOAA: Indeed, `&& true` is not needed for just `set -o pipefail`. It is needed, however, when running with `set -e`. In that case the `&& true` allows us to capture the error code in the following line on a failure.

ush/global_cycle.sh
```diff
@@ -336,7 +336,7 @@ cat << EOF > fort.37
  /
 EOF
 
-${APRUNCY} "${CYCLEXEC}" 1>> "${PGMOUT}" 2>> "${PGMERR}"
+${APRUNCY} "${CYCLEXEC}" 1>&1 2>&2
```
BrianCurtis-NOAA: Line 261-262 PGMOUT/PGMERR are defined. Did we want to remove the definition in the file there as well, or define at those lines where you want it to go?

dev/scripts/exglobal_enkf_update.sh
RussTreadon-NOAA: Thanks @DavidHuber-NOAA for the explanation.

RussTreadon-NOAA (dismissed):
Approve with regards to `exglobal_enkf_update.sh`

ush/global_cycle.sh
TravisElless-NOAA: @BrianCurtis-NOAA I was thinking going with a blend and adopted the approach I was using for the eupd script for the global cycle script as well. I deleted the `PGMERR` definition at 262, but left the `PGMOUT` in place. All of the output from the global cycle executable would write to both the job log file, and would keep the functionality of having a user defined out file in the `RUNDIR` if they desired. Does that approach seem reasonable for you?

BrianCurtis-NOAA: Yes that seems reasonable, thanks!

BrianCurtis-NOAA (approved):
Approve global_cycle changes.

DavidHuber-NOAA (approved):

DavidHuber-NOAA:
Merging based on passed CI tests on C6.

Issue #4822, opened 2026-04-20
(dev/gfs.v17) Develop a forecast manager for the global workflow


DavidHuber-NOAA:

DavidHuber-NOAA:
Completed via #4984.

Issue #4854, opened 2026-04-28
Do not remove cmd script or logs in DATA location


LinGan-NOAA:
### What is wrong?

Certain jobs are designed to automatically delete the CMD script from the DATA directory. This practice hinders the ability to effectively troubleshoot job failures or execution issues.
For example, atmos product job used $HOMEgfs/ush/run_mpmd.sh file to run CFP commend. However, this script removed the CMD files. This action disable the ability to troubleshoot even with KEEPDATA=YES.

### What should have happened?

The CMD script used in DATA should not be removed. If the same name for the script will be used again, the original CMD script should be protected by save as a trackable backup file.

### What machines are impacted?

All or N/A

### What global-workflow hash are you using?

abcdb8375

### Steps to reproduce

cd to DATA while job is running. Observe the cmd script been removed.

DavidHuber-NOAA:
@CarlosMDiaz-NCO @lgannoaa what do you think of this approach:

1. Before an MPMD runs, echo the command file to the parent log
2. On success
     1. Echo the MPMD jobs' logs to the parent log file
     2. Remove the MPMD jobs' logs
     3. Keep the MPMD command file(s)
3. On failure
     1. Exit immediately without echoing or removing the jobs' logs
4. (assuming success) if another MPMD needs to run, then remove all MPMD command files before generating new ones.

This way, the command file(s) are archived to the parent log. Would that satisfy the requirements/desires here?

LinGan-NOAA:
My recommendations are:
There is no need to echo the CFP serial job CMD in log file before execute. Because the production utility error_exit has build in capability to display list of the files in DATA. Developer can track the CMD file in DATA as needed. The key of this issue is to preserve the CMD file in a trackable way.

Given the high volume of CFP jobs, merging every individual log into the main job log can significantly degrade I/O performance and clutter the file. Instead, it is recommended to only output the CFP command summary (displaying the status of each rank) to the main log. This provides developers with immediate visibility into job progress without the overhead. To support detailed debugging, all individual CFP output files should be retained in the DATA directory.

If the design of the original job does need to share DATA to other jobs within the same cycle. Use Umbrella Shared Data location as required from the implementation standard. 
Example of the Umbrella Shared Data:
umbrella_JobName_data=${DATAROOT}/${model}_${JobName}_${cyc}_${version}_${optional_identifier}/atmos
[...]

PR #4907, opened 2026-05-11
[dev/gfs.v17] Forecast Manager: Real-Time Product Copy to COM with Adaptive, Component-Aware Segment Management

AntonMFernando-NOAA:
# Description

- This PR introduces a robust forecast manager job for the global workflow, designed to handle real-time copying of model products from the running directory to COM as they are written. The manager runs in parallel with forecast jobs and adapts automatically to the active model components (atmosphere, ocean, ice, waves, aerosols, etc).

  Refs #4822 
  Refs #4802 

# Type of change

# Change characteristics

- Is this change expected to change outputs (e.g. value changes to existing outputs, new files stored in COM, files removed from COM, filename changes, additions/subtractions to archives)? NO (If YES, please indicate to which system(s))
  - [ ] GFS
  - [ ] GEFS
  - [ ] SFS
  - [ ] GCAFS
  - [ ] EMC verif-global 
  - [ ] GDAS 
  - [ ] GFS-utils 
  - [ ] GSI 
  - [ ] GSI-monitor 
  - [ ] GSI-utils 
  - [ ] UFS-utils 
  - [ ] UFS-weather-model 
  - [ ] wxflow 

# How has this been tested?
- CI tests on wcoss2 and gaea

# Checklist


dev/jobs/JGLOBAL_FORECAST
```diff
@@ -7,7 +7,7 @@ if [[ 10#${ENSMEM:--1} -ge 0 ]]; then
     export DATA="${DATAjob}/${jobid}"
     source "${HOMEgfs}/ush/jjob_header.sh" -e "efcs" -c "base fcst efcs"
 else
-    export DATAjob="${DATAROOT}/${RUN}fcst.${PDY:-}${cyc}"
+    export DATAjob="${DATAROOT}/${RUN}_forecast.${PDY:-}${cyc}"
```
LinGan-NOAA, line 10: Recommend not to make unassociated change. This PR is not about changing forecast job name.

AntonMFernando-NOAA: @lgannoaa: This was a recommendation from @CarlosMDiaz-NCO .
see [here](https://github.com/NOAA-EMC/global-workflow/pull/4860#discussion_r3164146189).

comment: 
Is there a reason the manager's working directory $DATA needs to be under the 
DATAjob shared space? If not, `exportDATA={DATA:-${DATAROOT}/${jobid}}` should work for this job. and $DATAjob would just be used to reference where the needed files will be to copy out
Also, this goes to the forecast job as well, but the shared working directory name should be closer to the job name, so in this case - ${RUN}forecast${PDY:-}${cyc}. ($PDY will not be available at this point in production, but it looks like this syntax handles that case so we'd see gcafs_forecast_00)

dev/scripts/exglobal_cleanup.sh
```diff
@@ -5,13 +5,24 @@ echo "Begin Cleanup ${DATAROOT}!"
 
 # Remove DATAoutput from the forecast model run
 # TODO: Handle this better
-DATAfcst="${DATAROOT}/${RUN}fcst.${PDY:-}${cyc}"
+DATAfcst="${DATAROOT}/${RUN}_forecast.${PDY:-}${cyc}"
```
LinGan-NOAA, line 8: Recommend to not making name change here. Because the PR is not to change the forecast job name.

dev/scripts/exglobal_forecast.sh
```diff
@@ -176,11 +179,17 @@ else
 fi
 
 cpreq "${EXECgfs}/${FCSTEXEC}" "${DATA}/"
+# Signal to the forecast manager that this segment's model run is about to start.
+# This sentinel is cleaned during postdet so a rewound segment does not trigger the
+# manager prematurely from a table written by the previous run.
+echo "${RUN}_fcst_seg${FCST_SEGMENT} started" > "${DATAjob}/fcst_started_seg${FCST_SEGMENT}"
```
LinGan-NOAA: Unrelated to this PR. Recommend to remove.

dev/scripts/exglobal_forecast.sh
```diff
@@ -176,11 +179,17 @@ else
...
 ${APRUN_UFS} "${DATA}/${FCSTEXEC}" 1>&1 2>&2 && true
 export err=$?
 if [[ ${err} -ne 0 ]]; then
     err_exit "The forecast failed to run to completion!"
 fi
+# Signal to the forecast manager that the model run has completed successfully for this segment.
+echo "${RUN}_fcst_seg${FCST_SEGMENT} done" > "${DATAjob}/fcst_done_seg${FCST_SEGMENT}"
```
LinGan-NOAA, line 200: Unrelated to this PR. Recommend to remove.

dev/scripts/exglobal_forecast_manager.sh
```diff
@@ -0,0 +1,95 @@
...
+num_ranks=$(wc -l < "${FCST_MANAGER_CMDFILE}")
+echo "INFO: Launching ${num_ranks} MPMD component manager rank(s)"
+
+# Tell forecast_manager.sh where to find the model-completion sentinel so it can
+# exit gracefully when the model is done but some product files were not produced.
+export FCST_DONE_SENTINEL="${DATAjob}/fcst_done_seg${FCST_SEGMENT}"
+
+# Launch all component managers concurrently via run_mpmd.sh
+export USE_CFP=YES
+"${USHgfs}/run_mpmd.sh" "${FCST_MANAGER_CMDFILE}"
```
LinGan-NOAA: Additional application logic maybe needed to know which file has already been copied. To avoid overwrite COM in case of rerun.

ecf/defs/gfs_prod.def
```diff
@@ -842,6 +842,9 @@ suite gfs_v17
             event 1682 release_gfs_atmos_goesupp_f384
           task jgfs_fcst
             trigger ../analysis/atmos/jgfs_atmos_analupp==complete and ../init==complete and ../analysis/marine==complete
+            event release_fcst_manager
```
LinGan-NOAA: Missing components will not work as is.

ecf/scripts/gfs/forecast/jgfs_fcst_manager.ecf
```diff
@@ -0,0 +1,38 @@
...
+
+model=gfs
+export cyc="%CYC%"
+%include <head.h>
+%include <envir-p1.h>
+
+############################################################
+# Load modules
+############################################################
+module load intel/${intel_ver}
```
LinGan-NOAA: Missing components will not work as is

ush/forecast_manager.sh
```diff
@@ -0,0 +1,183 @@
...
+# Usage:    forecast_manager.sh <component> <table_file>
+#             component  - short name used in log output, e.g. "atm", "ww3",
+#                          "ocn", or "ice"
+#             table_file - absolute path to the 4-column product table file
+################################################################################
+
+# Suppress xtrace that may be inherited from the parent J-job via SHELLOPTS.
+# The inner copy loops iterate hundreds of entries; xtrace would generate
+# millions of log lines and significantly slow filesystem I/O.
+set +x
```
LinGan-NOAA: Do not use set +x (violated implementation standard)

ush/forecast_manager.sh
```diff
@@ -0,0 +1,183 @@
...
+                mkdir -p "${cl_dir}"
+            fi
+            if [[ "${this_ll}" == "${local_data[i]}" ]]; then
+                # Data-as-sentinel fallback: MOM6 did not write the period-start
+                # sentinel (e.g. FHMAX last window or sentinel dir was cleaned).
+                # Write a synthetic marker instead of copying the NC binary.
+                echo "$(basename "${com_data[i]}") completed $(date --utc +%Y%m%d%H%M%S)" > "${this_cl}"
+                log_err=0
+            else
+                cpfs "${this_ll}" "${this_cl}"
```
LinGan-NOAA, line 259: Application logic needed to ensure rerun case will not overwrite the COM file from previous running job.

ush/python/pygfs/task/oceanice_products.py
```diff
@@ -282,8 +282,9 @@ def subset(config: Dict) -> None:
         logger.info(f"Subsetting {varlist} from {input_file} to {output_file}")
 
         try:
-            # open the netcdf file
-            ds = xr.open_dataset(input_file)
+            # open the netcdf file; use_cftime=True handles non-standard calendars
+            # (e.g. julian) that pandas cannot decode
+            ds = xr.open_dataset(input_file, use_cftime=True)
```
LinGan-NOAA: Please do not make unrelated change.

ush/forecast_manager.sh
```diff
@@ -0,0 +1,183 @@
...
+                continue
+            fi
+            if [[ "${local_log[j]}" != "${this_ll}" ]]; then
+                continue
+            fi
+            com_dir=$(dirname "${com_data[j]}")
+            if [[ ! -d "${com_dir}" ]]; then
+                mkdir -p "${com_dir}"
+            fi
+            cpfs "${local_data[j]}" "${com_data[j]}"
```
LinGan-NOAA, line 178: Application logic needed to ensure rerun case will not overwrite the COM file from previous running job.

ush/forecast_postdet.sh
```diff
@@ -304,25 +304,82 @@ EOF
...
+        case "${RUN}" in
+            gfs) use_mgr="YES" ;;
+                # TODO: enable forecast manager for gefs, sfs, gcafs once tested
+                # gefs | sfs | gcafs) use_mgr="YES" ;;
+            *) ;;
+        esac
+
+        local atm_table="${DATAjob}/atm_products_seg${FCST_SEGMENT:-0}.txt"
+        if [[ "${use_mgr}" == "YES" ]]; then
+            rm -f "${atm_table}"
```
LinGan-NOAA: Please do not implement remove file from DATA location in ush level script.

ush/forecast_postdet.sh
```diff
@@ -304,25 +304,82 @@ EOF
...
+            # Remove the started sentinel so the forecast manager does not trigger from a
+            # previous run when this segment is rewound and re-queued.
+            rm -f "${DATAjob}/fcst_started_seg${FCST_SEGMENT:-0}"
```
LinGan-NOAA: Please do not implement remove file from DATA location in ush level script.

ush/forecast_postdet.sh
```diff
@@ -304,25 +304,82 @@ EOF
...
+                # Product table entries: local_data  local_log  com_data  com_log
+                for ((i = 0; i < ${#local_files[@]}; i++)); do
+                    echo "${local_files[i]} ${local_log} ${com_files[i]} ${com_log}" >> "${atm_table}"
+                done
+            else
+                # GDAS/enkfGDAS: NLN symlinks to COM so analysis jobs can read outputs during run
+                for ((i = 0; i < ${#local_files[@]}; i++)); do
+                    ${NLN} "${com_files[i]}" "${local_files[i]}"
+                done
+                ${NLN} "${com_log}" "${local_log}"
```
LinGan-NOAA, line 416: No link from $DATA to $COM

ush/forecast_postdet.sh
```diff
@@ -665,28 +801,36 @@ MOM6_postdet() {
                 ((midpoint = last_fhr + interval / 2))
 
                 vdate=$(date --utc -d "${current_cycle:0:8} ${current_cycle:8:2} + ${fhr} hours" +%Y%m%d%H)
```
LinGan-NOAA, line 848: Recommend using production utility for CDATE related calculation

DavidHuber-NOAA:
@AntonMFernando-NOAA the forecast manager is taking a very long time to copy atmospheric files to COM at full resolution. Currently, the forecast is at fhour 48, and wave, ocean, and ice products are up to date, but the atmospheric files are only at fhour 14. This is due to both the size difference and the cadence of the ATM products. The ATM (and wave) products come in every hour (to 120 fhours, then every 3 thereafter), while the ocean and ice come in every 6.

We may need to consider adding another core (or more) for the ATM products. One core handles the even hours and the other handles the odd hours, or something like that. (If the math works out, then we probably need 4 cores just for ATM products, then 1 core for each of the other).

DavidHuber-NOAA:
@AntonMFernando-NOAA Disregard the above comment. I forgot that I had launched the forecast manager after the forecast was already underway. I think it will work out OK, but will run another test where the manager runs alongside the forecast from the beginning.

dev/scripts/exglobal_forecast.sh
DavidHuber-NOAA: We have the capability to run the forecast in segments. This is a supported feature in the GFS (though we don't currently plan on exercising it in v17). This will also be very important for SFS, so I think it should stay.

dev/scripts/exglobal_forecast.sh
DavidHuber-NOAA: All forecasts run in segments in the GFS. It's just that for operations, we only run the 0th segment, so this needs to stay.

ush/forecast_manager.sh
DavidHuber-NOAA: This is handled at lines 85-93 and 97-99.

dev/scripts/exglobal_forecast_manager.sh
DavidHuber-NOAA: This is handled in the ush/forecast_manager.sh script.

DavidHuber-NOAA: @CarlosMDiaz-NCO a question on overwriting files already in COM. If the forecast needs to be rerun, and the new products differ from those already in COM, should those in COM be overwritten, or remain?

ush/forecast_manager.sh
DavidHuber-NOAA: Agreed.
```suggestion
```

ush/python/pygfs/task/oceanice_products.py
DavidHuber-NOAA: Fixed.

ush/forecast_postdet.sh
DavidHuber-NOAA: @lgannoaa This isn't a `DATA` location, but `DATAjob`. This is a shared space for the forecast and the forecast manager and on rerun, this table needs to be refreshed. Does this comment still apply?

ush/forecast_postdet.sh
DavidHuber-NOAA: This is only for `NET`s that do not (yet) support the forecast manager (SFS, GEFS, and GCAFS), so this will not affect the GFS (at least for `RUN=gfs`).

dev/parm/config/gfs/config.resources
```diff
@@ -983,6 +983,30 @@ case ${step} in
...
+    fi
+    if [[ "${DO_ICE:-NO}" == "YES" ]]; then
+        (( ntasks += 1 ))
+    fi
+    if [[ "${DO_AERO_FCST:-NO}" == "YES" ]]; then
+        (( ntasks += 1 ))
+    fi
+    tasks_per_node=${ntasks}
+    threads_per_task=1
+    memory="8GB"
```
DavidHuber-NOAA: On a high resolution test case, I noted the following:

```
 Cgroup mem limit exceeded: t
```

This memory request should thus be increased.

```suggestion
    memory="12GB"
```

dev/parm/config/gfs/config.resources
```diff
@@ -983,6 +983,30 @@ case ${step} in
     unset NTASKS_TOT
     ;;
 
+  "fcst_manager")
+    # Forecast manager: one MPMD rank per active model component.
+    # Polls DATAoutput for sentinel log files and copies products to COM in real time.
+    # Walltime must be long enough to cover the full forecast run.
+    export is_exclusive=False
```
DavidHuber-NOAA: I'd suggest making this a prepost node job. Also, these variables don't need to be exported:


```suggestion
    is_exclusive=False
    prepost=True
```

dev/parm/config/gfs/config.resources
```diff
@@ -983,6 +983,30 @@ case ${step} in
...
+    if [[ "${DO_ICE:-NO}" == "YES" ]]; then
+        (( ntasks += 1 ))
+    fi
+    if [[ "${DO_AERO_FCST:-NO}" == "YES" ]]; then
+        (( ntasks += 1 ))
+    fi
+    tasks_per_node=${ntasks}
+    threads_per_task=1
+    memory="8GB"
+    walltime="08:00:00"
```
DavidHuber-NOAA: Forecast manager at high res completed in ~3.5 hours (forecast completed in ~3.5 hours, fancy that). Thus, I think this can be decreased to the forecast walltime:

```suggestion
    walltime="06:00:00"
```

ecf/defs/gfs_prod.def
DavidHuber-NOAA: Can you elaborate? I think there is supposed to be a number with this, and it likely needs to be sequential (i.e. `1683`). Is there anything else?

DavidHuber-NOAA: @RussellManser-NCO Is there anything else missing with this event besides the `1683` that you see? (asking you as I think Lin is OOO)

ecf/scripts/gfs/forecast/jgfs_fcst_manager.ecf
```diff
@@ -0,0 +1,38 @@
...
+module load craype/${craype_ver}
+
+module list
```
DavidHuber-NOAA:
```suggestion
#TODO: load only the modules needed for each job
source ${HOMEgfs}/dev/ush/load_modules.sh run

module list
```

ush/forecast_postdet.sh
DavidHuber-NOAA: @AntonMFernando-NOAA we likely need to turn these into copies at the end of the job for `RUN=(gdas | enkfgdas | enkfgfs)`. I think we can tackle that task in another issue.

ush/forecast_postdet.sh
DavidHuber-NOAA: I think this use is fine. It is also not touched by this PR. Resolving.

ush/python/pygfs/task/oceanice_products.py
DavidHuber-NOAA: Actually, it looks like this change is needed. I am getting this error sequence without it:

```
2026-05-15 12:10:09,462 - ERROR    - oceanice_products: FATAL ERROR: Error occurred during netCDF subset: ocean.nc
...
pandas._libs.tslibs.np_datetime.OutOfBoundsDatetime: Cannot decode times from a non-standard calendar, 'julian', using pandas.
...
ValueError: attempt to get argmin of an empty sequence
...
ValueError: unable to decode time units 'hours since 2026-05-14 00:00:00' with "calendar 'julian'". Try opening your dataset with decode_times=False or installing cftime if it is not installed.
...more of the same...
UnboundLocalError: cannot access local variable 'ds' where it is not associated with a value
```

It doesn't really make sense to me that copying the files instead of linking them could cause a change to the internal time structures. I _**think**_ that what is going on here is that the file is getting modified after the model says it is done writing it (i.e. it closes). I think this is a MOM6 bug.

Since I am running a warm restart case, I will investigate the differences between the ocean product netCDF in this case and the real time to see if my hunch is correct.

DavidHuber-NOAA: Indeed, the files are different:

```console
> ls -l /lfs/h2/emc/ptmp/david.huber/realtime_manager/COM/fm_17/gfs.20260514/06/model/ocean/history/gfs.t06z.6hr_avg.f054.nc
-rw-r--r-- 1 david.huber emc 650M May 14 21:06 /lfs/h2/emc/ptmp/david.huber/realtime_manager/COM/fm_17/gfs.20260514/06/model/ocean/history/gfs.t06z.6hr_avg.f054.nc
> ls -l /lfs/h2/emc/gfstemp/emc.global/comroot/retrov17_01_realtime/gfs.20260514/06/model/ocean/history/gfs.t06z.6hr_avg.f054.nc
-rw-r--r-- 1 emc.global global 914M May 14 10:57 /lfs/h2/emc/gfstemp/emc.global/comroot/retrov17_01_realtime/gfs.20260514/06/model/ocean/history/gfs.t06z.6hr_avg.f054.nc
```

And if we check the `DATAjob` space for the originating file for forecast hour 54:
```console
> /lfs/h2/emc/stmp/david.huber/RUNDIRS/fm_17/gfs.2026051406/gfs_forecast.2026051406/output/MOM6_OUTPUT/ocn_2026_05_16_09_00.nc
-rw-r--r-- 1 david.huber emc 914M May 14 21:08 /lfs/h2/emc/stmp/david.huber/RUNDIRS/fm_17/gfs.2026051406/gfs_forecast.2026051406/output/MOM6_OUTPUT/ocn_2026_05_16_09_00.nc
```

We see that its size matches the real-time retro. So the forecast manager is copying the file too soon.

ecf/scripts/gfs/forecast/jgfs_fcst_manager.ecf
```diff
@@ -0,0 +1,38 @@
+#PBS -S /bin/bash
+#PBS -N gfs_fcst_manager_%CYC%
+#PBS -j oe
+#PBS -q %QUEUE%
+#PBS -A %PROJ%-%PROJENVIR%
+#PBS -l walltime=08:00:00
+#PBS -l select=1:ncpus=4:ompthreads=1:mem=32GB
```
DavidHuber-NOAA:
```suggestion
#PBS -l walltime=06:00:00
#PBS -l select=1:ncpus=4:ompthreads=1:mem=12GB
```

ush/python/pygfs/task/oceanice_products.py
```diff
@@ -282,7 +282,8 @@ def subset(config: Dict) -> None:
         logger.info(f"Subsetting {varlist} from {input_file} to {output_file}")
 
         try:
-            # open the netcdf file
+            # open the netcdf file; use_cftime=True handles non-standard calendars
+            # (e.g. julian) that pandas cannot decode
```
DavidHuber-NOAA:
```suggestion
            # open the netcdf file
```

ecf/defs/gfs_prod.def
RussellManser-NCO: I believe what @lgannoaa is referring to are missing events. I think the idea here was to replace `gfs/forecast/jgfs_fcst_fsm`. If that's correct, then two changes may be needed:

1. Move all events from `gfs/forecast/jgfs_fcst_fsm` to `gfs/forecast/jgfs_fcst_manager`
2. In `ush/forecast_manager.sh` use `ecflow_client --event release_<job_name>`

It isn't clear to me how `jgfs_fcst_manager` is releasing or triggering the corresponding jobs otherwise. Is there a script or step in the workflow that I overlooked?

DavidHuber-NOAA: At the moment, the forecast manager does not release/trigger any tasks. The FSM job is still doing that. We could probably combine the two at a later time. This PR is only introducing the forecast manager, which currently only copies files to COM.

So, at the moment, I'm just looking for guidance on how to set up this copying job.

RussellManser-NCO: Sounds like I misunderstood the context for this PR, so thank you for the clarification. I don't see any changes that need to be made other than adding the event number.

DavidHuber-NOAA: Thanks @RussellManser-NCO!

DavidHuber-NOAA:
```suggestion
            event 1683 release_fcst_manager
```

ush/forecast_postdet.sh
DavidHuber-NOAA: After discussion with the leads, we will need this job enabled for `RUN=(gdas | enkfgdas)`. This will be handled in a separate PR.

ecf/scripts/gfs/forecast/jgfs_fcst_manager.ecf
```diff
@@ -0,0 +1,38 @@
+#PBS -S /bin/bash
+#PBS -N gfs_fcst_manager_%CYC%
+#PBS -j oe
+#PBS -q %QUEUE%
+#PBS -A %PROJ%-%PROJENVIR%
+#PBS -l walltime=08:00:00
+#PBS -l walltime=06:00:00
+#PBS -l select=1:ncpus=4:ompthreads=1:mem=12GB
```
DavidHuber-NOAA:
```suggestion
#PBS -l walltime=06:00:00
### Note, the exact number of CPUs will need to be fine tuned.
#PBS -l select=1:ncpus=7:ompthreads=1:mem=12GB
```

DavidHuber-NOAA:
Thorough testing of this capability at low and high res has been completed. The capability works in all configurations available for `RUN=gfs`. In particular, I tested the C1152 forecast and verified that
1. Products copied in a timely manner.
     1. The maximum delay between completing the ATM component GRIB2 and netCDF products was 3 minutes
           1. GRIB2 is written to DATA first and the forecast_manager waits for the netCDF to also finish before writing.
           2. This accounted for 2 minutes of the 3-minute delay.
     2. The delay for other products was less.
4. All products (except CICE) copied successfully.
5. All product jobs (except CICE) executed successfully.

There is an issue that causes the 00Z CICE logs not to be written correctly. However, as a proof of concept, this solution will work.

DavidHuber-NOAA:
>    1. The maximum delay between completing the ATM component GRIB2 and netCDF products was 3 minutes

An important clarification here: this 3 minute delay is 1 minute longer than the current design. The model writes ATM products in batches: gridded GRIB2, surface flux GRIB2, surface netCDF, and atmospheric netCDF products. Only once all of these products are written will the log be written to indicate the products are done for the forecast hour. The delay between when this log is written and the forecast manager finishes writing the products to COM is at most 1 minute.

DavidHuber-NOAA:
@AntonMFernando-NOAA duplicate forecast hours are being created in the ATM tables. This is due to a longstanding duplicate hour bug in `forecast_predet.sh`. Can you apply this patch to fix it?

```diff
diff --git a/ush/forecast_predet.sh b/ush/forecast_predet.sh
index 79ad38798..8a44fb61c 100755
--- a/ush/forecast_predet.sh
+++ b/ush/forecast_predet.sh
@@ -177,7 +177,7 @@ FV3_predet() {
     local fhr=${FHMIN}
     if [[ "${FHOUT_HF}" -gt 0 && "${FHMAX_HF}" -gt 0 ]]; then
         FV3_OUTPUT_FH="${FV3_OUTPUT_FH} $(seq -s ' ' "${FHMIN}" "${FHOUT_HF}" "${FHMAX_HF}")"
-        fhr=${FHMAX_HF}
+        fhr=$(( FHMAX_HF + FHOUT ))
     fi
     FV3_OUTPUT_FH="${FV3_OUTPUT_FH} $(seq -s ' ' "${fhr}" "${FHOUT}" "${FHMAX}")"
```

DavidHuber-NOAA:
Testing of these changes at C1152 shows that the maximum latency between when the products are generated and when they are successfully copied in full to COM is now 36s. However, further testing shows that this latency can be further reduced to 11s by changing this [one line](https://github.com/AntonMFernando-NOAA/global-workflow/blob/6c21b9f7834358a3382ba14c2347dbc2e2fcf4f3/dev/parm/config/gfs/config.fcst_manager#L29)

from

```bash
export FCST_MGR_SLEEP=30
```

to

```bash
export FCST_MGR_SLEEP=5
```

I have tested this solution at low and high resolution and for both ATM-only and S2SW cases. I believe it is now ready for review.

dev/jobs/JGLOBAL_FORECAST_MANAGER
```diff
@@ -0,0 +1,95 @@
...
+PYTHONPATH="${HOMEgfs}/ush/python${PYTHONPATH:+:${PYTHONPATH}}"
+export PYTHONPATH
+
+##############################################
+# Construct COM variables (mirrors JGLOBAL_FORECAST)
+##############################################
+
+rCDUMP="${RUN/gfs/gdas}"
+export rCDUMP="${rCDUMP/gcafs/gcdas}"
+
```
DavidHuber-NOAA: I don't think these are needed.

```suggestion

File: dev/jobs/JGLOBAL_FORECAST_MANAGER [hunk 24]
```diff
@@ -0,0 +1,95 @@
...
+    declare -rx COMOUT_WAVE_HISTORY="${COMOUT}/model/wave/history"
+fi
+
+if [[ "${DO_OCN:-NO}" == "YES" ]]; then
+    declare -rx COMOUT_OCEAN_HISTORY="${COMOUT}/model/ocean/history"
+fi
+
+if [[ "${DO_ICE:-NO}" == "YES" ]]; then
+    declare -rx COMOUT_ICE_HISTORY="${COMOUT}/model/ice/history"
+fi
```
DavidHuber-NOAA, line 65: We should probably also have one for aerosols in develop. I'm fine with it not being here in dev/gfs.v17. Just making a note.

DavidHuber-NOAA:
> Testing of these changes at C1152 shows that the maximum latency between when the products are generated and when they are successfully copied in full to COM is now 36s.

Noting that the actual copy time for the largest file (`*atmf*.nc`) generally took a little less than 0.1s per file. There were a couple of instances where this was close to 1s. This latency is likely to decrease when `DATA` is located on `/f1` or `/f2`. The bulk of the latency comes from the aforementioned sleep parameter and a sanity check sleep, [here](https://github.com/AntonMFernando-NOAA/global-workflow/blob/6c21b9f7834358a3382ba14c2347dbc2e2fcf4f3/ush/forecast_manager.sh#L166), currently set to 5s. This parameter could likely be reduced to 0 once the CICE logging issue is resolved.

dev/parm/config/sfs/config.fcst_manager
```diff
@@ -0,0 +1,22 @@
...
+# Maximum time (seconds) to wait for product tables to appear in COMOUT_CONF.
+# Must be less than the batch walltime of the fcst_manager job.
+export FCST_MGR_INIT_TIMEOUT=7200
+
+# Maximum time (seconds) to wait for the model to finish before the manager
+# declares a timeout and exits with an error (0 = no timeout).
+export FCST_MGR_TIMEOUT=0
+
+# Polling interval (seconds) between sentinel checks.
+export FCST_MGR_SLEEP=30
```
DavidHuber-NOAA: Consider turning this down to 10 (maybe after addressing #4854 so the logs aren't gargantuan).

dev/parm/config/gfs/config.resources
```diff
@@ -983,6 +983,31 @@ case ${step} in
...
+    fi
+    if [[ "${DO_ICE:-NO}" == "YES" ]]; then
+        (( ntasks += ${MGR_NTASKS_ICE:-1} ))
+    fi
+    if [[ "${DO_AERO_FCST:-NO}" == "YES" ]]; then
+        (( ntasks += 1 ))
+    fi
+    tasks_per_node=${ntasks}
+    threads_per_task=1
+    memory="12GB"
```
DavidHuber-NOAA: I noted during testing that the `Cgroup mem limit exceeded: t` message was present, so this should be increased.


```suggestion
    memory="20GB"
```

dev/scripts/exglobal_forecast.sh
```diff
@@ -144,6 +147,13 @@ if [[ "${cplchm}" == ".true." ]]; then
 fi
 echo "MAIN: Post-determination set up of run type finished"
 
+# Signal to the forecast manager that all product tables are ready.
+# This is created AFTER all component postdet functions complete to ensure
+# the manager only starts when all tables (ATM, WW3, OCN, ICE) are fully written.
+if [[ -n "${DATAjob:-}" ]]; then
+    echo "${RUN}_fcst_seg${FCST_SEGMENT} table ready" > "${DATAjob}/fcst_table_ready_seg${FCST_SEGMENT}"
```
DavidHuber-NOAA: Let's put a date in this file as well

```suggestion
    echo "${RUN}_fcst_seg${FCST_SEGMENT} table ready at $(date)" > "${DATAjob}/fcst_table_ready_seg${FCST_SEGMENT}"
```

dev/scripts/exglobal_forecast.sh
```diff
@@ -144,6 +147,13 @@ if [[ "${cplchm}" == ".true." ]]; then
...
+fi
```
DavidHuber-NOAA, line 160: We also need an ecflow trigger here (I will have another suggestion in the `gfs_prod.def` file to accompany this):

```suggestion
    if [[ "${SENDECF}" == "YES" ]]; then
        if [[ "${RUN}" == "gfs" ]]; then
            ecflow_client --event release_gfs_fcst_manager
        fi
    fi
fi
```

ecf/defs/gfs_prod.def
```diff
@@ -837,6 +837,9 @@ suite gfs_v17_nco
             event 1682 release_gfs_atmos_goesupp_f384
           task jgfs_fcst
             trigger ../analysis/atmos/jgfs_atmos_analupp==complete and ../init==complete and ../analysis/marine==complete
+            event 1683 release_fcst_manager
+          task jgfs_fcst_manager
+            trigger jgfs_fcst:release_fcst_manager
```
DavidHuber-NOAA:
```suggestion
            event 1683 release_gfs_fcst_manager
          task jgfs_fcst_manager
            trigger jgfs_fcst:release_gfs_fcst_manager
```

ecf/defs/gfs_prod.def
```diff
@@ -6723,6 +6726,9 @@ suite gfs_v17_nco
             event 1682 release_gfs_atmos_goesupp_f384
           task jgfs_fcst
             trigger ../analysis/atmos/jgfs_atmos_analupp==complete and ../init==complete and ../analysis/marine==complete
+            event release_fcst_manager
+          task jgfs_fcst_manager
+            trigger jgfs_fcst:release_fcst_manager
```
DavidHuber-NOAA:
```suggestion
            event release_gfs_fcst_manager
          task jgfs_fcst_manager
            trigger jgfs_fcst:release_gfs_fcst_manager
```

ecf/defs/gfs_prod.def
```diff
@@ -12609,6 +12615,9 @@ suite gfs_v17_nco
...
```
DavidHuber-NOAA:
```suggestion
            event release_gfs_fcst_manager
          task jgfs_fcst_manager
            trigger jgfs_fcst:release_gfs_fcst_manager
```

ecf/defs/gfs_prod.def
```diff
@@ -18495,6 +18504,9 @@ suite gfs_v17_nco
...
```
DavidHuber-NOAA:
```suggestion
            event release_gfs_fcst_manager
          task jgfs_fcst_manager
            trigger jgfs_fcst:release_gfs_fcst_manager
```

ecf/scripts/gfs/forecast/jgfs_fcst_manager.ecf
```diff
@@ -0,0 +1,38 @@
+#PBS -S /bin/bash
+#PBS -N gfs_fcst_manager_%CYC%
+#PBS -j oe
+#PBS -q %QUEUE%
+#PBS -A %PROJ%-%PROJENVIR%
+#PBS -l walltime=06:00:00
+### Note, the exact number of CPUs will need to be fine tuned.
+#PBS -l select=1:ncpus=7:ompthreads=1:mem=12GB
```
DavidHuber-NOAA:
```suggestion
#PBS -l select=1:ncpus=14:ompthreads=1:mem=20GB
```

dev/scripts/exglobal_forecast_manager.sh
DavidHuber-NOAA: This was answered offline by Carlos. In general, files in COM should not be overwritten.

ush/forecast_manager.sh
```diff
@@ -0,0 +1,313 @@
...
+
+if [[ ! -f "${table_file}" ]]; then
+    echo "FATAL ERROR: Product table '${table_file}' not found for component '${component}'" >&2
+    exit 1
+fi
+
+# Load table into indexed arrays
+declare -a local_data local_log com_data com_log done_flag
+count=0
+while read -r ld ll cd cl; do
```
DavidHuber-NOAA: Can `cd` be changed here? Maybe to `comd`?

ush/forecast_manager.sh
```diff
@@ -0,0 +1,313 @@
...
+                    fi
+                done
+                if [[ -n "${_ref_size}" ]]; then
+                    _sz_diff=$((_sz_new - _ref_size))
+                    if [[ ${_sz_diff} -lt 0 ]]; then
+                        _sz_diff=$((-_sz_diff))
+                    fi
+                    if [[ ${_sz_diff} -gt 1048576 ]]; then
+                        echo "WARN [${component}]: size mismatch for '$(basename "${com_data[j]}")': ${_sz_new}B vs reference '${_ref_name}': ${_ref_size}B (diff=${_sz_diff}B > 1 MB) -- possible partial/corrupt output"
+                        _size_check_msgs+="WARN: size mismatch for '$(basename "${com_data[j]}")': ${_sz_new}B vs ref '${_ref_name}': ${_ref_size}B diff=${_sz_diff}B (>1 MB)"$'\n'
```
DavidHuber-NOAA: Warnings should be `WARNING`, per standards.

ush/forecast_manager.sh
```diff
@@ -0,0 +1,313 @@
...
+    # FCST_MGR_DONE_IDLE_MAX idle cycles (default 3) exit with a warning for any
+    # entries the model never produced (e.g. optional GOCART output types).
+    if [[ -n "${FCST_DONE_SENTINEL:-}" && -f "${FCST_DONE_SENTINEL}" ]]; then
+        if [[ ${remaining} -lt ${remaining_before} ]]; then
+            fcst_done_idle=0
+        else
+            ((fcst_done_idle++)) || true
+            idle_max=${FCST_MGR_DONE_IDLE_MAX:-3}
+            if [[ ${fcst_done_idle} -ge ${idle_max} ]]; then
+                echo "WARN [${component}]: Model run complete; no new files for ${fcst_done_idle} consecutive poll cycle(s). ${remaining} of ${count} table entry(s) were not produced by the model — skipping."
```
DavidHuber-NOAA: `WARNING`

ush/forecast_atm_barrier.sh
```diff
@@ -0,0 +1,167 @@
...
+barrier_table="${1:?Usage: forecast_atm_barrier.sh <barrier_table>}"
+
+echo "INFO [atm_barrier]: Starting barrier; reading '${barrier_table}'"
+
+if [[ ! -f "${barrier_table}" ]]; then
+    echo "FATAL ERROR [atm_barrier]: Barrier table '${barrier_table}' not found" >&2
+    exit 1
+fi
+
+FCST_POLL_INTERVAL="${FCST_MGR_POLL_INTERVAL:-30}"
```
DavidHuber-NOAA: An entry for this should be added to `config.fcst_manager`. Consider reducing this to 5.

DavidHuber-NOAA (changes requested):
Looks good, just some minor adjustments needed.

ecf/scripts/gfs/forecast/jgfs_fcst_manager.ecf
```diff
@@ -0,0 +1,38 @@
...
+############################################################
+# Load modules
+#TODO: load only the modules needed for each job
+source ${HOMEgfs}/dev/ush/load_modules.sh run
+
+module list
+############################################################
+# CALL executable job script here
+############################################################
+export FCST_SEGMENT='0'
```
DavidHuber-NOAA, line 26:
```suggestion
export HOMEglobal=${HOMEgfs}
export FCST_SEGMENT='0'
```

dev/scripts/exglobal_forecast_manager.sh
```diff
@@ -0,0 +1,156 @@
...
+
+    while read -r ld ll cd cl; do
+        [[ -z "${ld}" || "${ld:0:1}" == "#" ]] && continue
+        if [[ -z "${sentinel_rank[${ll}]+_}" ]]; then
+            sentinel_rank["${ll}"]=${rank}
+            ((rank = (rank + 1) % n_ranks)) || true
+        fi
+        r="${sentinel_rank[${ll}]}"
+        echo "${ld} ${ll} ${cd} ${cl}" >> "${output_prefix}${r}.txt"
+    done < "${table_file}"
```
DavidHuber-NOAA: Recommend expanding these variable names a bit as `ld`, `ll`, and `cd` are linux commands (or common aliases).
```suggestion
    local r logd logl comd coml
    local rank=0
    declare -A sentinel_rank

    for ((r = 0; r < n_ranks; r++)); do
        : > "${output_prefix}${r}.txt"
    done

    while read -r logd logl comd coml; do
        [[ -z "${logd}" || "${logd:0:1}" == "#" ]] && continue
        if [[ -z "${sentinel_rank[${logl}]+_}" ]]; then
            sentinel_rank["${logl}"]=${rank}
            ((rank = (rank + 1) % n_ranks)) || true
        fi
        r="${sentinel_rank[${logl}]}"
        echo "${logd} ${logl} ${comd} ${coml}" >> "${output_prefix}${r}.txt"
    done < "${table_file}"
```

DavidHuber-NOAA (approved):

DavidHuber-NOAA:
Launching tests on WCOSS2 and C6.

dev/jobs/JGLOBAL_FORECAST_MANAGER
```diff
@@ -0,0 +1,91 @@
...
+# When running the workflow in development, source jjob_standard_vars.sh #}
+source "${HOMEglobal}/ush/jjob_standard_vars.sh"
+#{% else %}
+#{# When rendered for operations, place the contents of jjob_var_setup.j2 here #}
+#{# Note that the first line of jjob_var_setup.j2 will be commented #}
+#{% include jjob_var_setup.j2 %}
+#{% endif %}
+
+source "${USHgfs}/jjob_shell_setup.sh"
+source "${USHgfs}/set_strict.sh"
```
DavidHuber-NOAA:
```suggestion
source "${USHglobal}/jjob_shell_setup.sh"
source "${USHglobal}/set_strict.sh"
```

dev/jobs/JGLOBAL_FORECAST_MANAGER
```diff
@@ -0,0 +1,91 @@
...
+
+if [[ "${DO_ICE:-NO}" == "YES" ]]; then
+    declare -rx COMOUT_ICE_HISTORY="${COMOUT}/model/ice/history"
+fi
+
+##############################################
+# Begin JOB SPECIFIC work
+##############################################
+
+FORECASTMGRSH="${FORECASTMGRSH:-${SCRgfs}/exglobal_forecast_manager.sh}"
```
DavidHuber-NOAA:
```suggestion
FORECASTMGRSH="${FORECASTMGRSH:-${SCRglobal}/exglobal_forecast_manager.sh}"
```

dev/scripts/exglobal_forecast_manager.sh
```diff
@@ -0,0 +1,156 @@
...
+            echo "FATAL ERROR: Timed out after ${MGR_INIT_TIMEOUT}s waiting for ${_atm_tbl}" >&2
+            exit 1
+        fi
+    done
+    {
+        echo "${USHgfs}/forecast_manager.sh atm_atmf ${atm_atmf_tbl}"
+        echo "${USHgfs}/forecast_manager.sh atm_sfcf ${atm_sfcf_tbl}"
+        echo "${USHgfs}/forecast_manager.sh atm_grib ${atm_grib_tbl}"
+        echo "${USHgfs}/forecast_manager.sh atm_flux ${atm_flux_tbl}"
+        echo "${USHgfs}/forecast_atm_barrier.sh ${atm_barrier_tbl}"
```
DavidHuber-NOAA:
```suggestion
        echo "${USHglobal}/forecast_manager.sh atm_atmf ${atm_atmf_tbl}"
        echo "${USHglobal}/forecast_manager.sh atm_sfcf ${atm_sfcf_tbl}"
        echo "${USHglobal}/forecast_manager.sh atm_grib ${atm_grib_tbl}"
        echo "${USHglobal}/forecast_manager.sh atm_flux ${atm_flux_tbl}"
        echo "${USHglobal}/forecast_atm_barrier.sh ${atm_barrier_tbl}"
```

dev/scripts/exglobal_forecast_manager.sh
```diff
@@ -0,0 +1,156 @@
...
+    WW3_TABLE="${DATAjob}/ww3_products_seg${FCST_SEGMENT}.txt"
+    echo "INFO: Waiting for WW3 product table at ${WW3_TABLE}"
+    if ! wait_for_file "${WW3_TABLE}" "${mgr_sleep_interval}" "${mgr_max_tries}"; then
+        echo "FATAL ERROR: Timed out after ${MGR_INIT_TIMEOUT}s waiting for ${WW3_TABLE}" >&2
+        exit 1
+    fi
+    echo "INFO: WW3 product table found (${MGR_NTASKS_WW3} rank(s))"
+    split_table_by_sentinel "${WW3_TABLE}" "${MGR_NTASKS_WW3}" "${DATA}/ww3_mgr_rank"
+    for ((r = 0; r < MGR_NTASKS_WW3; r++)); do
+        echo "${USHgfs}/forecast_manager.sh ww3 ${DATA}/ww3_mgr_rank${r}.txt" >> "${FCST_MANAGER_CMDFILE}"
```
DavidHuber-NOAA:
```suggestion
        echo "${USHglobal}/forecast_manager.sh ww3 ${DATA}/ww3_mgr_rank${r}.txt" >> "${FCST_MANAGER_CMDFILE}"
```

dev/scripts/exglobal_forecast_manager.sh
```diff
@@ -0,0 +1,156 @@
...
+    OCN_TABLE="${DATAjob}/ocn_products_seg${FCST_SEGMENT}.txt"
+    echo "INFO: Waiting for OCN product table at ${OCN_TABLE}"
+    if ! wait_for_file "${OCN_TABLE}" "${mgr_sleep_interval}" "${mgr_max_tries}"; then
+        echo "FATAL ERROR: Timed out after ${MGR_INIT_TIMEOUT}s waiting for ${OCN_TABLE}" >&2
+        exit 1
+    fi
+    echo "INFO: OCN product table found (${MGR_NTASKS_OCN} rank(s))"
+    split_table_by_sentinel "${OCN_TABLE}" "${MGR_NTASKS_OCN}" "${DATA}/ocn_mgr_rank"
+    for ((r = 0; r < MGR_NTASKS_OCN; r++)); do
+        echo "${USHgfs}/forecast_manager.sh ocn ${DATA}/ocn_mgr_rank${r}.txt" >> "${FCST_MANAGER_CMDFILE}"
```
DavidHuber-NOAA:
```suggestion
        echo "${USHglobal}/forecast_manager.sh ocn ${DATA}/ocn_mgr_rank${r}.txt" >> "${FCST_MANAGER_CMDFILE}"
```

dev/scripts/exglobal_forecast_manager.sh
```diff
@@ -0,0 +1,156 @@
...
+    ICE_TABLE="${DATAjob}/ice_products_seg${FCST_SEGMENT}.txt"
+    echo "INFO: Waiting for ICE product table at ${ICE_TABLE}"
+    if ! wait_for_file "${ICE_TABLE}" "${mgr_sleep_interval}" "${mgr_max_tries}"; then
+        echo "FATAL ERROR: Timed out after ${MGR_INIT_TIMEOUT}s waiting for ${ICE_TABLE}" >&2
+        exit 1
+    fi
+    echo "INFO: ICE product table found (${MGR_NTASKS_ICE} rank(s))"
+    split_table_by_sentinel "${ICE_TABLE}" "${MGR_NTASKS_ICE}" "${DATA}/ice_mgr_rank"
+    for ((r = 0; r < MGR_NTASKS_ICE; r++)); do
+        echo "${USHgfs}/forecast_manager.sh ice ${DATA}/ice_mgr_rank${r}.txt" >> "${FCST_MANAGER_CMDFILE}"
```
DavidHuber-NOAA:
```suggestion
        echo "${USHglobal}/forecast_manager.sh ice ${DATA}/ice_mgr_rank${r}.txt" >> "${FCST_MANAGER_CMDFILE}"
```

dev/scripts/exglobal_forecast_manager.sh
```diff
@@ -0,0 +1,156 @@
...
+echo "INFO: Launching ${num_ranks} MPMD component manager rank(s)"
+
+# Tell forecast_manager.sh where to find the model-completion sentinel so it can
+# exit gracefully when the model is done but some product files were not produced.
+export FCST_TABLE_READY_SENTINEL="${DATAjob}/fcst_table_ready_seg${FCST_SEGMENT}"
+export FCST_DONE_SENTINEL="${DATAjob}/fcst_done_seg${FCST_SEGMENT}"
+
+# Launch all component managers concurrently via run_mpmd.sh
+export USE_CFP=YES
+"${USHgfs}/run_mpmd.sh" "${FCST_MANAGER_CMDFILE}"
```
DavidHuber-NOAA:
```suggestion
"${USHglobal}/run_mpmd.sh" "${FCST_MANAGER_CMDFILE}"
```

DavidHuber-NOAA (changes requested):
A few `VARgfs` variables need to be renamed to `VARglobal`.

DavidHuber-NOAA:
All tests passed on WCOSS2:

```
Tue Jun  2 12:19:35 UTC 2026
******** C48_ATM_4907 ********
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202103231200        Done    Jun 01 2026 15:00:18    Jun 01 2026 16:15:33

******** C48mx500_3DVarAOWCDA_4907 ********
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202103241800        Done    Jun 01 2026 15:00:22    Jun 01 2026 15:25:42
202103250000        Done    Jun 01 2026 15:00:22    Jun 01 2026 16:50:27

******** C48mx500_hybAOWCDA_4907 ********
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202103241800        Done    Jun 01 2026 15:00:25    Jun 01 2026 15:24:49
202103250000        Done    Jun 01 2026 15:00:25    Jun 01 2026 16:30:39

******** C48_S2SW_extended_4907 ********
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202103231200        Done    Jun 01 2026 15:00:53    Jun 01 2026 17:55:36
202103231800        Done    Jun 01 2026 15:00:53    Jun 01 2026 17:30:26

******** C96_atm3DVar_extended_4907 ********
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202112201800        Done    Jun 01 2026 15:00:40    Jun 01 2026 15:24:56
202112210000        Done    Jun 01 2026 15:00:40    Jun 01 2026 19:50:48
202112210600        Done    Jun 01 2026 15:00:40    Jun 01 2026 20:40:23
202112211200        Done    Jun 01 2026 15:25:22    Jun 01 2026 22:05:27
202112211800        Done    Jun 01 2026 19:55:20    Jun 02 2026 00:35:12

******** C96C48_hybatmDA_4907 ********
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202112201800        Done    Jun 01 2026 15:00:39    Jun 01 2026 15:25:00
202112210000        Done    Jun 01 2026 15:00:39    Jun 01 2026 17:35:28
202112210600        Done    Jun 01 2026 15:00:39    Jun 01 2026 17:20:40

******** C96C48_hybatmsnowDA_4907 ********
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202112201200        Done    Jun 01 2026 15:00:10    Jun 01 2026 15:25:04
202112201800        Done    Jun 01 2026 15:00:10    Jun 01 2026 17:25:22
202112210000        Done    Jun 01 2026 15:00:10    Jun 01 2026 17:25:23

******** C96C48_hybatmsoilDA_4907 ********
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202205150600        Done    Jun 01 2026 15:00:24    Jun 01 2026 15:25:09
202205151200        Done    Jun 01 2026 15:00:24    Jun 01 2026 17:30:52
202205151800        Done    Jun 01 2026 15:00:24    Jun 01 2026 17:30:52

******** C96C48mx500_S2SW_cyc_gfs_4907 ********
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202112201200        Done    Jun 01 2026 15:00:42    Jun 01 2026 15:25:14
202112201800        Done    Jun 01 2026 15:00:42    Jun 01 2026 17:55:37
202112210000        Done    Jun 01 2026 15:00:42    Jun 01 2026 18:20:39
202112211800        Done    Jun 01 2026 15:25:36    Jun 01 2026 18:31:04
```

DavidHuber-NOAA:
All tests passed on C6:

```
############################# C48_ATM_4907 #############################
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202103231200        Done    Jun 01 2026 14:35:12    Jun 01 2026 20:10:07
###################### C48mx500_3DVarAOWCDA_4907 #######################
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202103241800        Done    Jun 01 2026 14:35:16    Jun 01 2026 14:50:15
202103250000        Done    Jun 01 2026 14:35:16    Jun 01 2026 20:15:12
####################### C48mx500_hybAOWCDA_4907 ########################
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202103241800        Done    Jun 01 2026 14:35:20    Jun 01 2026 14:55:35
202103250000        Done    Jun 01 2026 14:35:20    Jun 01 2026 16:40:38
############################ C48_S2SW_4907 #############################
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202103231200        Done    Jun 01 2026 14:35:24    Jun 01 2026 20:15:19
########################## C96_atm3DVar_4907 ###########################
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202112201800        Done    Jun 01 2026 14:35:28    Jun 01 2026 14:55:42
202112210000        Done    Jun 01 2026 14:35:28    Jun 01 2026 20:10:21
202112210600        Done    Jun 01 2026 14:35:28    Jun 01 2026 18:25:44
######################### C96C48_hybatmDA_4907 #########################
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202112201800        Done    Jun 01 2026 14:35:32    Jun 01 2026 14:55:46
202112210000        Done    Jun 01 2026 14:35:32    Jun 01 2026 20:10:25
202112210600        Done    Jun 01 2026 14:35:32    Jun 01 2026 18:20:47
####################### C96C48_hybatmsnowDA_4907 #######################
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202112201200        Done    Jun 01 2026 14:35:35    Jun 01 2026 14:55:51
202112201800        Done    Jun 01 2026 14:35:35    Jun 01 2026 17:50:51
202112210000        Done    Jun 01 2026 14:35:35    Jun 01 2026 18:20:50
####################### C96C48_hybatmsoilDA_4907 #######################
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202205150600        Done    Jun 01 2026 14:35:39    Jun 01 2026 14:55:54
202205151200        Done    Jun 01 2026 14:35:39    Jun 01 2026 17:50:55
202205151800        Done    Jun 01 2026 14:35:39    Jun 01 2026 18:25:56
#################### C96C48mx500_S2SW_cyc_gfs_4907 #####################
   CYCLE         STATE           ACTIVATED              DEACTIVATED     
202112201200        Done    Jun 01 2026 14:35:43    Jun 01 2026 14:55:59
202112201800        Done    Jun 01 2026 14:35:43    Jun 01 2026 19:30:42
202112210000        Done    Jun 01 2026 14:35:43    Jun 01 2026 21:15:41
```

DavidHuber-NOAA (approved):

ecf/defs/gfs_prod.def
```diff
@@ -6723,6 +6726,9 @@ suite gfs_v17_nco
             event 1682 release_gfs_atmos_goesupp_f384
           task jgfs_fcst
             trigger ../analysis/atmos/jgfs_atmos_analupp==complete and ../init==complete and ../analysis/marine==complete
+            event release_gfs_fcst_manager
```
TravisElless-NOAA:
```suggestion
            event 1683 release_gfs_fcst_manager
```

ecf/defs/gfs_prod.def
```diff
@@ -12609,6 +12615,9 @@ suite gfs_v17_nco
...
```
TravisElless-NOAA:
```suggestion
            event 1683 release_gfs_fcst_manager
```

ecf/defs/gfs_prod.def
```diff
@@ -18495,6 +18504,9 @@ suite gfs_v17_nco
...
```
TravisElless-NOAA:
```suggestion
            event 1683 release_gfs_fcst_manager
```

TravisElless-NOAA (commented):
I think a event release number is needed for these triggers as well

PR #4932, opened 2026-05-20
HOTFIX Create RESTART directory to prevent FMS from doing it

DavidHuber-NOAA:
# Description
This fixes a crash seen at high resolution where FMS fails to create the RESTART directory in parallel.

# Type of change

# Change characteristics

# How has this been tested?
Ran the marine recentering job showing that both the directory and it's permissions were correct.

# Checklist

### DIFF
--- a/dev/jobs/JGLOBAL_MARINE_ANALYSIS_ECEN
+++ b/dev/jobs/JGLOBAL_MARINE_ANALYSIS_ECEN
@@ -26,7 +26,7 @@ PYTHONPATH="${HOMEgfs}/ush/python${PYTHONPATH:+:${PYTHONPATH}}"
 export PYTHONPATH
 
 ## HOTFIX
-# TODO Move this to the GDASApp config. Better yet, fix this in the model so RESTART doesn't get created.
+# TODO Fix this in the input.nml template so RESTART doesn't get created in the first place.
 mkdir -p "${DATA}/RESTART"
 
 # Ignore possible spelling error (nothing is misspelled)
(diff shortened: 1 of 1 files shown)


guillaumevernieres (approved):
Thanks for the quick patch, @DavidHuber-NOAA.

One clarification: this is FMS, called from the JEDI system, failing to create the directory for some reason, not MOM6.

I’m OK with this fix, but the more appropriate long-term fix is probably in the Jinja template that generates input.nml, since RESTART is no longer used.

DavidHuber-NOAA:
Thanks for the clarification @guillaumevernieres. I've updated the title and description accordingly and will also update my `TODO` comment.

dev/jobs/JGLOBAL_MARINE_ANALYSIS_ECEN
```diff
@@ -25,6 +25,10 @@ source "${USHgfs}/set_strict.sh"
 PYTHONPATH="${HOMEgfs}/ush/python${PYTHONPATH:+:${PYTHONPATH}}"
 export PYTHONPATH
 
+## HOTFIX
+# TODO Move this to the GDASApp config. Better yet, fix this in the model so RESTART doesn't get created.
```
DavidHuber-NOAA:
```suggestion
# TODO Fix this in the input.nml template so RESTART doesn't get created in the first place.
```

AndrewEichmann-NOAA (approved):
What @guillaumevernieres  said

PR #5057, opened 2026-06-24
[dev/gfs.v17] Hotfix for analysis ecf triggers

TravisElless-NOAA:
# Description
The ecflow analysis jobs are launching too early and are missing observations that are needed for downstream analysis jobs. This PR updates when the emc sfc prep job runs which subsequently kicks off the analysis job. Note this is a development/para testing issue. Operations will trigger these jobs off of production obsproc, however we can't do this for para testing and rely on the time.  

# Type of change
- [ ] Bug fix (fixes something broken)

# Change characteristics

- Is this change expected to change outputs (e.g. value changes to existing outputs, new files stored in COM, files removed from COM, filename changes, additions/subtractions to archives)? NO 
- Is this a breaking change (a change in existing functionality)? NO
- Does this change require a documentation update? NO
- Does this change require an update to any of the following submodules? NO 
# How has this been tested?

# Checklist
- [ ] Any dependent changes have been merged and published
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have documented my code, including function, input, and output descriptions
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] This change is covered by an existing CI test or a new one has been added
- [ ] Any new scripts have been added to the .github/CODEOWNERS file with owners
- [ ] I have made corresponding changes to the system documentation if necessary

PR #5067, opened 2026-06-26
(dev/gfs.v17) Adjust analysis triggers to fix occasionally early launched jobs

### LINKED ISSUE #4853: Improve fbwind job design with exception handling
### What is wrong?

The fbwind jobs need some input files. Such as *pres_a and *pres_b grib2 files. The product jobs does not running sequentially to generate the grib2 file. Job failed when unconditional cpreq statement used for the input grib2 files.
The design should be enhanced to include exception handling that triggers a sleep/wait cycle if the input file is missing.

The job should wait for a short amount of time on the input file if file not exist before running the cpreq commend.
The job should FATAL ERROR and show which file not found in the log file. 

All or N/A

abcdb83

### Steps to reproduce

Remove one of the input file to test the exception handling capability. Such as "gfs.t${cyc}z.pres_*.0p25.f006.grib2"
Run the job and see how the job failed.

### Additional information

_No response_

_No response_

### LINKED ISSUE #5064: (dev/gfsv17) ecflow trigger for jgfs_atmos_fbwind needs updated
### What is wrong?

We had a failed jgfs_atmos_fbwind job because data was not available.  On further investigation this job depends on f_32 product is released, however, that is indicative that the master file is available but this job uses the pgrb2 files.  Therefore this trigger should be that the f032 product job has been completed. 

job should not fail for data not available. 

All or N/A

dev/gfsv17

### Steps to reproduce

run ecflow and happen to catch failure. 

### Additional information

Note - This does still assume that all previous forecast hours have finished which technically f032 could be finished and the one before not. 

Change the trigger.

DavidHuber-NOAA:
# Description
This adjusts the triggers to the analysis, sfcprep, and tropcy_qc_reloc jobs to give more freedom and flexibility to the workflow. The analysis in particular has triggered too early and the tropcy_qc_reloc job has failed to launch at least once.

Resolves #4853 
Resolves #5064 
# Type of change

# Change characteristics

# How has this been tested?
To be tested in the ecflow real-time parallel.

# Checklist

### DIFF
--- a/ecf/defs/gfs_prod.def
+++ b/ecf/defs/gfs_prod.def
@@ -22,11 +22,11 @@ suite gfs_v17_nco
             edit WGF 'atmos'
             task jgfs_atmos_tropcy_qc_reloc
               #event 1 jtwc_bull_email
-              time 02:41
+              trigger :TIME >= 0241 and :TIME < 0245
             task jgfs_atmos_emcsfc_sfc_prep
               #NCO/Ops should use obsproc trigger.  Time trigger is a workaround for development
               #trigger /gfs_v17_0/primary/00/obsproc/v1.3/gfs/atmos/dump/jobsproc_gfs_atmos_dump:release_sfcprep
-              time 03:07
+              trigger :TIME >= 0307 and :TIME < 0312
           endfamily		//atmos
           family marine
             edit WGF 'marine'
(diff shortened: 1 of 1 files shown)


JessicaMeixner-NOAA (approved):
Just one question on the timing of the prep jobs -- did we need to push these back further than what they were now? Or just give the range of time?

DavidHuber-NOAA:
I think just give the range. I haven't seen a situation where the prep jobs failed because they launched too early, only that they didn't launch at all.

JessicaMeixner-NOAA:
I thought @TravisElless-NOAA had seen a case where they launched too early, but maybe we already made that adjustment.

TravisElless-NOAA:
From what I've seen the prep jobs were always running successfully when they should because the specific prep jobs didn't require all the obsproc data to be there. When we were just triggering the analysis job off the completion of the prep job, that would launch before all the obsproc data was there. #5057 pushed back the prep job until the obsproc data was there so the analysis job would not trigger too early.

TravisElless-NOAA (approved):

CatherineThomas-NOAA (approved):

PR #5100, opened 2026-07-07
[dev/gfs.v17] Remove marine prep fsm from ecflow def

TravisElless-NOAA:
# Description
This PR removes the dependency of the marine prep fsm job from the ecflow definition file. Triggers are updated to use time triggers for marine prep job as a substitute from the obsproc completion trigger needed during operations. The bmat_init job trigger is also updated to remove the dependency on the fsm job.  

# Type of change

# Change characteristics

# How has this been tested?
def file successfully loads into ecflow server

# Checklist

### DIFF
--- a/dev/scripts/exglobal_fsm.sh
+++ b/dev/scripts/exglobal_fsm.sh
@@ -109,12 +109,12 @@ while [[ "${proceed_trigger_scan}" == "YES" ]]; do
         # TODO remove/change this and look at obsproc for the bufr files
         for ty_md in adt icec sst; do
             # Check for the existence of files for each type of marine observation; if any type is missing, skip the rest and wait for the next scan
-	    if [[ ${ty_md} == "adt" && ${cyc} != "00" ]]; then
-		echo "adt files are only produced at 00z...skipping check for ${cyc}z"
-		continue
-	    else
+            if [[ ${ty_md} == "adt" && ${cyc} != "00" ]]; then
+                echo "adt files are only produced at 00z...skipping check for ${cyc}z"
+                continue
+            else
                 tty_files=("${COMIN_OCEAN_OBS_gfs}/"*"${ty_md}"*)
-	    fi
+            fi
             count_tty=${#tty_files[@]}
             if [[ ${count_tty} -eq 0 ]]; then
                 skip_this_scan="NO"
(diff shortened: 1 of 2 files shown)


JessicaMeixner-NOAA (approved):

CatherineThomas-NOAA (approved):

dev/scripts/exglobal_fsm.sh
```diff
@@ -105,7 +107,7 @@ while [[ "${proceed_trigger_scan}" == "YES" ]]; do
         skip_this_scan="YES"
         echo "Proceeding with scan_release_gfs_marine_prepoceanobs"
         # TODO remove/change this and look at obsproc for the bufr files
-        for ty_md in adt icec sst insitu; do
+        for ty_md in adt icec sst; do
```
DavidHuber-NOAA, line 110: The `gfs_marine_obs_dump` did not generate any `adt` files either. Should this also be removed? Are these generated by the `jgfs_marine_obs_bufr_dump` job?

JessicaMeixner-NOAA: Are those files being generated by atm? Or can they not exist? Is that the issue here? I thought both adt and insitu obs were being used --- do we need to re-check the marine obs jobs?

TravisElless-NOAA: The marine obs dump job log file /lfs/h2/emc/gfstemp/emc.global/ecflow/output/today/gfs_marine_obs_dump_12.o127852525 looks to have adt obs in there while the obs_bufr_dump job does not. So adt obs should come from the marine_obs_dump job is my assumption

FYI @CoryMartin-NOAA

CoryMartin-NOAA: my understanding is that only the 'insitu_*' come from ObsProc/BUFR and everything else comes from DCOMROOT.

There's also one ob type that is only around for one cycle per day, which I do not recall. @apchoiCMD may be able to confirm which one

apchoiCMD: @DavidHuber-NOAA `adt` obs are only generated in `00z` , I assume you are looking at the cycles other than `00`z if my understanding is correct-

DavidHuber-NOAA: @apchoiCMD yes, that is correct. Thanks!

We should rework this file check to only look for `adt` files when `cyc=00`.

I suspect that without `nullglob`, this was returning a false positive result before. Here's a hotfix option:

```suggestion
        for ty_md in adt icec sst; do
            if [[ ${ty_md} == "adt" && ${cyc} != "00" ]]; then
                # adt observations are only available on the 00z cycle
                continue
            fi
        done
```

TravisElless-NOAA: I think that fix option should work, minus the done statement at the end. This for loop will need to extend several lines down to do the proper checks of the other files. Will apply to the gfs and gdas marine here shortly

Issue #5116, opened 2026-07-10
(dev/gfs.v17) Allow ecflow cases to run in near real time


DavidHuber-NOAA:
### What new functionality do you need?

The GFS's ecflow suite currently uses the PDY set by prod_util in the production suite. This means that if the ecflow workflow falls too far behind, it will have the wrong PDY set and will not be able to run any further.

There should be a way for the ecflow suite to know that it is running in a development environment and set the PDY according to the actual date of interest.

Note additional issues are time triggered jobs and if we are running too far behind and miss those time windows then the workflow will fall behind as well.

### Acceptance Criteria

- [ ] The ecflow suite is able to run in near real time, lagging up to 5 days behind (any longer and the data in prod/com will start to age off).

### Suggest a solution (optional)

One possible solution is to set the PDY in the cycle_end ecflow script. This gets tricky because cycle_end will be called before the previous cycle is complete. We could wait until the entirety of the previous cycle is complete, but this could cause significant delays.

JessicaMeixner-NOAA:
Note additional issues are time triggered jobs and if we are running too far behind and miss those time windows.

DavidHuber-NOAA:
Resolved by #5119

PR #5119, opened 2026-07-10
(dev/gfs.v17) Manage PDY in ecflow suite for near-real-time GFS runs

DavidHuber-NOAA:
This AI-generated solution was reviewed by @DavidHuber-NOAA before opening this PR.

# Description
The GFS ecflow suite read PDY from the operational prod date file (/lfs/h1/ops/prod/com/date/tCYCz) in head.h. When a development suite lags behind real time, that file reflects the current operational date, so PDY was wrong and the workflow could not proceed (issue #5116).

head.h now sources PDY from a suite-managed ecFlow variable when MACHINE_SITE=development (falling back to the operational date file when unset, e.g. the first cycle). cycle_end.ecf advances the PDY by one day per cycling loop and publishes it via 'ecflow_client --alter add variable PDY' on the cycle family so head.h picks it up through variable inheritance. Production/NCO behavior is unchanged.

# Type of change

# Change characteristics

# How has this been tested?
The new python script was tested with PDYs from yesterday and today and shown to successfully trigger jobs that had passed their time windows.
The full solution will need to be tested in ecflow.

# Checklist


ecf/scripts/cycle_end.ecf
```diff
@@ -23,24 +23,49 @@ case ${cyc} in
...
+    setpdy.sh
+    . ./PDY
+    cur_pdy=${PDY}
+  fi
+
+  # Advance one day to obtain the PDY of the next occurrence of this cycle.
+  # This is what allows the suite to march forward through dates and catch up
+  # when running behind real time (near-real-time), independent of the
+  # operational date file.
+  next_pdy=$(date --utc --date="${cur_pdy} + 1 day" +%Y%m%d)
```
TravisElless-NOAA: Perhaps I'm reading this wrong, but I think this will run on every cycle right? So 12z would add a day to PDY and then 18z would too. So you would get 20260709 for 12z and then 20260710 for 18z and 20260711 for 00z?

DavidHuber-NOAA: Yes, you're correct. That won't work. I'm realizing now there are other holes in this solution, like not accounting for time-dependent triggers. I'll close this and try again.

DavidHuber-NOAA: I've fixed this issue and added a method for detecting/triggering timed jobs that have passed their window.

ecf/scripts/cycle_begin.ecf
```diff
@@ -0,0 +1,109 @@
...
+  # Update the output directory for the current cycle
+  ecflow_client --alter add variable OUTPUTDIR $outputdir_ecf_today /${suite_name}/primary/${cyc}/gfs
+  ecflow_client --alter add variable OUTPUTDIR $outputdir_ecf_today /${suite_name}/primary/${cyc}/gdas
+  ecflow_client --alter add variable OUTPUTDIR $outputdir_ecf_today /${suite_name}/primary/${cyc}/enkfgdas
+
+  # Requeue all jobs for the current cycle (except cycle_begin of course)
+  ecflow_client --requeue=force /${suite_name}/primary/${cyc}/gfs
+  ecflow_client --requeue=force /${suite_name}/primary/${cyc}/gdas
+  ecflow_client --requeue=force /${suite_name}/primary/${cyc}/enkfgdas
+  ecflow_client --requeue=force /${suite_name}/primary/${cyc}/cycle_end
```
DavidHuber-NOAA:
```suggestion
```

ecf/scripts/cycle_begin.ecf
```diff
@@ -0,0 +1,108 @@
...
+     next_cyc="18"
+     ;;
+  18)
+     next_cyc="00"
+     ;;
+esac
+# retro=YES enables the developer near-real-time / retrospective cycling
+# driver, which manages its own PDY.  Set retro=NO for an NCO code delivery,
+# where PDY is provided by the operational date file (see ecf/include/head.h).
+retro=YES
```
JessicaMeixner-NOAA: Should another existing variable be used to manage this more cleanly? Or should this variable be added to the gfs_prod.def file, which would be less flexible in some sense, but also wouldn't have something that needs to be changed hard-coded in a scirpt?

DavidHuber-NOAA: Yeah, I think we could use `MACHINE_SITE == development`, which is a standard variable and defined in `gfs_prod.def`. I'll update this.

DavidHuber-NOAA: Updated.

DavidHuber-NOAA:
I tested this approach in an ecflow workflow and it successfully triggered jobs with a 24 hour lag time. This approach should work for near-real-time runs, though I have one caveat. If we are running in near-real-time, the `qc_tropcy` jobs will be working with out-of-date inputs. I can think of four possible solutions:
1. Add logic to the `qc_tropcy` jobs to check the $PDY$cyc to check if we are running retrospectively and copy the `syndat` files from an archive.
2. If we are running retrospectively, run have `cycle_begin.ecf` run the `dev/job_cards/rocoto/prep.sh` and then use `ecflow_client` to mark all prep jobs as complete.
3. Mark the `qc_tropcy` jobs as complete using automated ecflow_client commands in the `cycle_begin.ecf` script rather than running them. The GSI will then not have inputs to assimilate for tropical cyclones.
4. Allow the `qc_tropcy` jobs to run as normal. I _**think**_ the GSI has the logic to conclude that the inputs from `qc_tropcy` are not appropriately dated and will exclude them.

Noting that options 3 and 4 will cause solutions to diverge from a real-time run. Which of these do you think would be the best option here?

TravisElless-NOAA:
My vote would be option 1. Option 2 can get messy since we are not regularly utilizing those scripts in the ecflow workflow; could be a more of development-focused approach but I don't think a strategy for operational style configurations. I don't like option 3&4 because if we have the obs we should be utilizing them instead of intentionally discarding them

CatherineThomas-NOAA:
Thanks @DavidHuber-NOAA for outlining potential solutions.  While all of these options should allow cycling to continue and give scientifically reasonable results, I'm also voting for Option 1.  These are files that should be saved anyway for reproducibility, so we should try to make use of them and attempt to reproduce a real time run as closely as we can.

DavidHuber-NOAA:
@CatherineThomas-NOAA @JessicaMeixner-NOAA @TravisElless-NOAA I think I have a method now to copy in the syndata observation file. I'm going to test this out and will report when it is complete.

DavidHuber-NOAA:
My test ecflow workflow now appears to be running correctly. The syndata files are being copied in from the archive and jobs are being launched correctly by `cycle_begin`. I have suspended the 12z cycle and will start it back up tomorrow morning to simulate a workflow lagging behind real time.

The workflow can be monitored by connecting to my server on `ddecflow02` on port 34566.

DavidHuber-NOAA:
After restarting the workflow today, jobs failed because `PDY` was not set correctly. Something about my changes to copy in the `syndata` files seems to have undermined the capability of the workflow to manage its own `PDY`. I'm converting this to draft for the time being while I work that out.

DavidHuber-NOAA:
Reopening for review. After a more thorough review of the ecflow test, I see now that the PDY was updated correctly and as intended for the 2nd cycle, but not the first. I have fixed this and am retesting now. For a more thorough explanation:

The workflow was initialized on 2026080712. The `cycle_begin` job ran and successfully instantiated the PDY to 20260807 for cycle **18** (not 12). I then suspended the workflow before cycle 12 was complete. It was then restarted the following day. At that time, PDY was not set for cycle 12, so the workflow read PDY from the operational file (at that time 20260808). This caused the failures I saw.

My fix was to set the PDY variable in ecflow for the first cycle and for future cycles.

dev/ush/ecflow_time_triggers.py
```diff
@@ -0,0 +1,378 @@
...
+    # Split PDYcyc into PDY (YYYYMMDD) and cyc (HH)
+    PDY = PDYcyc[:8]
+    cyc = PDYcyc[8:]
+
+    # Get COMROOT to know where we are copying to
+    comroot = get_comroot()
+
+    copy_commands = []
+    # Extract the task path from the command
+    match = re.search(r"ecflow_client --complete (.+)", complete_command)
```
TravisElless-NOAA: The syndat cp command was not working for me. I traced things back to this line returning `None` and it was therefore skipping copy command section. Gemini suggested changing the line to this and it now seems to work for me.
```suggestion
    match = re.search(r"ecflow_client --force=complete .*/([^/]+)/prep/", complete_command)
```

dev/ush/ecflow_time_triggers.py
```diff
@@ -0,0 +1,378 @@
...
+def build_syndata_copy_commands(complete_command, PDYcyc):
+    """
+    Builds a list of commands to copy syndata for skipped tasks instead of triggering them.
+    """
+    # Split PDYcyc into PDY (YYYYMMDD) and cyc (HH)
+    PDY = PDYcyc[:8]
+    cyc = PDYcyc[8:]
+
+    # Get COMROOT to know where we are copying to
+    comroot = get_comroot()
```
TravisElless-NOAA, line 197: Also noting that get_comroot (I believe it was in here) did not process the `${USER}` correctly to my username and instead was trying to copy to `ptmp/${USER} `itself and said I did not have the permissions. Changing the COMPATH in run.ver to use the explicit username instead of `${USER} `allowed things to copy correctly.

ecf/include/head.h
```diff
@@ -22,8 +22,32 @@ if [ -d /apps/ops/prod ]; then # On WCOSS2
...
+#   situation would yield the wrong (current operational) PDY and stall the
+#   workflow.  See NOAA-EMC/global-workflow#5116.
+if [ "%MACHINE_SITE%" = "development" ]; then
+  # Prefer the suite-provided PDY (the ecFlow PDY variable set by cycle_begin).
+  # Fall back to the operational date file when it is not yet set, e.g. on the
+  # very first cycle before cycle_begin has run.
+  export ecf_pdy_set="YES"
+  export PDY="%PDY:%"
+  if [ -z "${PDY}" ]; then
+    export PDY=$(cut -c7-14 /lfs/h1/ops/prod/com/date/t%CYC%z)
```
TravisElless-NOAA: This solution currently here works when starting from the same cycle as ops PDY, however, if starting from further behind this had to be changed by the user to a hardcoded value. A potential cleaner solution would be to have a user include a `PDY_start` variable in their def file that was read automatically here and would not require coming and making a local edit here. 
```suggestion
    export PDY=%PDY_start%
```

TravisElless-NOAA (commented):
A couple of things I've come across when initially trying to run this for a case that was a couple of days behind ops.

dev/ush/ecflow_time_triggers.py
DavidHuber-NOAA: Yes, that is a gotcha. There's probably a better way to handle this.

ecf/include/head.h
DavidHuber-NOAA: This is solved in two places, first here:

https://github.com/DavidHuber-NOAA/global-workflow/blob/f6632d0dd5e8357a04185dae2bb45f0b8c10ed32/ecf/include/head.h#L38-L47

and then here

https://github.com/DavidHuber-NOAA/global-workflow/blob/f6632d0dd5e8357a04185dae2bb45f0b8c10ed32/ecf/scripts/cycle_begin.ecf#L30-L34

It's only if an unfortunate soul (thank you for your efforts, Travis! and sorry I didn't give you warning about this) tries to run this without the ecflow `PDY` variable being set in the first place, that this comes up.

dev/ush/ecflow_time_triggers.py
DavidHuber-NOAA: Oof, good catch. Thank you!

ecf/include/head.h
DavidHuber-NOAA: Yeah, we could do something like this. Maybe issue a warning if PDY_start isn't set and then try to use the ops PDY.

TravisElless-NOAA: Yeah I was coming to the same conclusion on the drive home last night. Make this a nested if block that if there isn't a PDY or a PDY start that it would default to the current day.

ecf/defs/gfs_prod.def
```diff
@@ -15,6 +15,8 @@ suite gfs_v17_nco
     family 00
       edit CYC '00'
       edit ECF_FILES '%PACKAGEHOME%/ecf/scripts'
+      task cycle_begin
+        trigger ../18/gdas/forecast == active or ../18/gdas/forecast == complete
```
TravisElless-NOAA: Would adding a trigger to make sure there aren't too many gfs cycles running at once be a good idea? Depending on machine availability I fear a worse case scenario where we can cycle through a day of gdas and it would force restart a gfs cycle that hasn't completed yet. 
```suggestion
        trigger ../12/gfs/forecast == complete and ../18/gdas/forecast == complete
```

DavidHuber-NOAA: Yeah, this could work, but I think it would need to be a `== complete or == inactive` so cycling doesn't stall on the first day.

TravisElless-NOAA:
Just noting that I did a "package update" simulated test where I forced the cycle begin job to run after the gfs tropcy job time window, but before the gdas time window. The job saw it was past the gfs time window and ran those jobs via catchup (copying the syndat file for gfs instead of actually running the tropcy job), and correctly did not touch the gdas cycle.

dev/ush/ecflow_time_triggers.py
DavidHuber-NOAA: I added a patch to `get_comroot` so it should now render `$USER` or `${USER}`.

ecf/defs/gfs_prod.def
DavidHuber-NOAA: I think the most recent commit addresses this.

TravisElless-NOAA:
Another testing note: I restarted a case using a PDY that is now 3 days old (20260818) for an 18z cycle. This PDY managed solution PR (at hash `0999bc3`) successfully completed the 2026081818, 2026081900, and 2026081906 cycles.

ecf/include/head.h
DavidHuber-NOAA: I believe I my most recent commit adds this capability.

DavidHuber-NOAA:
After testing, I made a slight modification to the documented procedure for setting the ecflow `PDYstart` variable and adjusted how it is used in head.h so that the ecflow variable will be referenced first, not the one declared in `head.h`.

Using this procedure, I was able to start cycle 2026082312.

I believe I have addressed all comments. Ecflow testing of cycle 2026082312 is now about half way complete. I will post again when the cycle finishes and 2026082318 starts.

DavidHuber-NOAA:
@TravisElless-NOAA pointed out that the syndat data generated by the `tropcy_qc` jobs will not be copied back to `${ROTDIR}/syndat` in cases where we are running behind. This means that we could end up in a non-reproducible situation. To maintain reproducibility, the operational `syndat` files will need to be copied in so that `tropcy_qc` has all of the up-to-date data to run on. I have added this capability and am currently testing. My ecflow test case is currently running cycle 2026082312 (with a 96-hour GFS forecast and no product generation). Once it catches up to real time, I will check the outputs of the `tropcy_qc` jobs to verify they generated correctly and are identical to operations.

dev/ush/ecflow_time_triggers.py
```diff
@@ -0,0 +1,391 @@
...
+        print(f"Creating destination directory: {os.path.dirname(dest_path)}")
+        os.makedirs(os.path.dirname(dest_path), exist_ok=True)
+
+        copy_command = f"cp {source_path} {dest_path}"
+        copy_commands.append(copy_command)
+
+        # We also need to update the syndat files used by future cycles. Copy these from operations.
+        syndat_source_path = f"{SYNDAT_LOCATION}/*"
+        syndat_dest_path = f"{comroot}/syndat/"
+        copy_commands.append(copy_command)
```
TravisElless-NOAA, line 239: These` syndat_*_path` variables would need to be changed to remove the `syndat `part to use the existing `copy_command` or would need to create a `syndat_copy_command `and update the `append` statement?

DavidHuber-NOAA: Yes, that would help 🤦‍♂️

DavidHuber-NOAA: Done.

TravisElless-NOAA (approved):
Thanks @DavidHuber-NOAA!!

DavidHuber-NOAA:
The ecflow test case successfully caught up to real time overnight and successfully launched the gdas_atmos_tropcy_qc job.

PR #5292, opened 2026-09-21
Restore the middle-of-window entry in soilinc_fhrs

arpitjain099:
`exglobal_atmos_anal_sfc_gcycle.sh` builds `soilinc_fhrs` with `+=` only:

```sh
gcycle_dates=("${PDY}${cyc}")        # Always update surface restarts at middle of window
if [[ "${DOIAU:-}" == "YES" ]]; then # Update surface restarts at beginning of window
    half_window=$((assim_freq / 2))
    soilinc_fhrs+=("${half_window}")
```

Element 0 is never set, and the loop below reads `soilinc_fhrs[hr]` using the `gcycle_dates` index, whose element 0 is the middle of the window. Running that block with the values a cycled gdas experiment supplies:

```
DOIAU=YES  2022051512 -> increment.sfc.i003.tile1.nc
           (beginning) -> increment.sfc.i00.tile1.nc
DOIAU=NO   2022051512 -> increment.sfc.i00.tile1.nc
```

So the middle of the window gets the beginning-of-window increment, and the other pass builds a path with an empty hour. `STRICT="NO"` is the shipped default, so `set -u` is off and this is silent.

`regrid_gsi_sfc_incr_to_tile.sh` and `exglobal_atmos_anal_sfc_regrid.sh` both start the array with `soilinc_fhrs=("${assim_freq}")` under the same comment, and `gaussian_sfcanl.sh` hardcodes `i006` for that cycle time, so restoring that line here makes the three agree. The same run afterwards stages i006 then i003.

The guard needs `DO_LAND_IAU=".false."` with `DO_GSISOILDA="YES"`, which the CI yamls never produce since they all set `DO_LAND_IAU: ".true."`, but `config.base.j2` forces it false on the first cycle of a cold start and whenever `DOIAU="NO"`.


ClaraDraper-NOAA:
Hi @arpitjain099 Thanks for catching this. I don't quite follow the last sentence about the guards needed - which combinations of settings (DOIAU, DO_LAND_IAU, and DO_GSISOILDA) would this result in incorrect/unintended results for?

tsga (approved):

